PR #18, opened 2024-11-29
Closest point to geometric median instead of centroid

giorgioangel:
Instead of the centroid, first computes the geometric median and then select the point in the point cloud closest to it

### DIFF
--- a/thaumato-anakalyptor/ThaumatoAnakalyptor/instances_to_graph.py
+++ b/thaumato-anakalyptor/ThaumatoAnakalyptor/instances_to_graph.py
@@ -1,6 +1,7 @@
 ### Julian Schilliger - ThaumatoAnakalyptor - Vesuvius Challenge 2024
 
 import numpy as np
+from numba import njit
 from tqdm import tqdm
 import pickle
 import glob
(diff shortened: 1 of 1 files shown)


thaumato-anakalyptor/ThaumatoAnakalyptor/instances_to_graph.py
```diff
@@ -32,6 +33,72 @@
...
+        # Calculate the norm along rows (for 2D array)
+        return np.sqrt(np.sum(arr ** 2, axis=1))
+    elif axis == 0:
+        # Calculate the norm along columns (for 2D array)
+        return np.sqrt(np.sum(arr ** 2, axis=0))
+    else:
+        raise ValueError("Unsupported axis. Use 0 or 1.")
+    
+@njit
+def geometric_median_(X, eps=1e-5):
```
schillij95, line 68: could this be more concise + not needing a new library?
np.exp(np.mean(np.log(arr), axis=axis))

schillij95 (commented):
Related to #25

schillij95:
Geometric Median from this PR has been added with PR #26

PR #24, opened 2024-12-01
Add .zarr file loading for faster upfront training

StewartSethA:
Adds a new segment_to_ome.py utility based on existing scroll_to_ome.py for converting segments into .zarr format for faster loading, and an implementation of .zarr array loading in one of the training scripts

### DIFF
--- a/foundation/scroll2zarr/README.md
+++ b/foundation/scroll2zarr/README.md
@@ -561,3 +561,8 @@ in `volumes` are not compressed, so this discussion does not apply
 to them).
 
 A script `decompress_tiffs.py` is provided for this purpose.
+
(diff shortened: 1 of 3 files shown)


schillij95:
Thank you Seth for your PR

After looking over the scroll2zarr adjustments, i have some questions:
- what is the difference between the scroll_to_ome.py and the segment_to_ome.py?
- scroll_to_ome.py can also take the "input_tiff_dir" argument as path_so_segment/layers as tif/jpg input stack and then output the zarr as the argument "output", ie segment_id.zarr

giorgioangel:
This is very outdated now, have to close it

PR #25, opened 2024-12-02
geometric mean to calculate umbilicus distance

schillij95:
Using geometric mean to calculate the umbilicus distance
- Improves the same block edge accuracy from 2/3 to 3/4
Adding option to only recompute edges
Fixing the ground truth assignment when generating a subgraph (for faster testing and development)


schillij95:
need to switch to geometric mean as well during the centroid calculation for patches_centroids

PR #27, opened 2024-12-03
thaumato readme

giorgioangel:
Without this line in the README the Docker won't start properly


schillij95:
Good point, the docker runs without it, but if you changed code and wanted to run the updated code you would need -v $(pwd)/:/workspace

PR #28, opened 2024-12-03
thaumato - compute centroid (or equivalent) only once

giorgioangel:
Computing the centroid (or equivalent) only once and storing the value in a dictionary can spare $O(10^8)$ redundant computations if I am not mistaken.


schillij95:
Great addition! There seems to be some unresolved conflicts as the branch is 29 commits behind

PR #34, opened 2024-12-05
thaumato - centroid computation optimization

giorgioangel:
Computing the centroid (or equivalent) only once and storing the value in a dictionary can spare $O(10^8)$ redundant computations if I am not mistaken.


schillij95:
Thank you for this nice speedup!

PR #66, opened 2025-01-08
Datasets page GP dataset edits

stephenrparsons:
@giorgioangel could you take a look at these edits (minor things to make this one stylistically resemble the other one)?

This breaks the URLs - if you let me know that it looks good, I'll rename `/datasets/GrandPrizeBannerRegion` to `/datasets/grand-prize-banner-region` on the data server when merging so they're still valid. Wanted to check that I wasn't missing or breaking anything else here in the process.


giorgioangel:
> @giorgioangel could you take a look at these edits (minor things to make this one stylistically resemble the other one)?
> 
> This breaks the URLs - if you let me know that it looks good, I'll rename `/datasets/GrandPrizeBannerRegion` to `/datasets/grand-prize-banner-region` on the data server when merging so they're still valid. Wanted to check that I wasn't missing or breaking anything else here in the process.

I think this should be fine, I am merging it and changing the directory name on the dataserver.

stephenrparsons:
@giorgioangel thank you!

PR #68, opened 2025-01-13
remove padding from pred before writing

bruniss:
added to remove the padding from the prediction before writing it so it is easier used in other tasks


stephenrparsons:
lgtm - @younader look good to you?

younader:
@bruniss Thanks for this PR! I think this is a needed change, but I believe we need to sync this with the training so that either both expect padded labels or unpadded labels. I think adding paddings to the labels in the training should do the trick.

giorgioangel:
@younader @repushko up!

PR #84, opened 2025-01-28
[website] Add virtual unrolling page

giorgioangel:
This PR draft adds the Virtual Unrolling page to the website.

I did some changes to Docsaurus to be able to write big Latex formulas.
Before deploying locally, I ran once:
`yarn add remark-math rehype-katex`
@stephenrparsons how is the webserver deployed, is it containerized?

Fyi, I also changed `$` to `\$` in several pages to avoid the dollar symbol being misrecognized as the latex math mode command. To keep in mind when editing or creating new pages.

Unwanted behaviour: 
While navigating, I notice a strange behaviour - the added `Virtual Unrolling` page sometimes loads without the background color. It gets it back when the page is refreshed. Investigating...


stephenrparsons:
@giorgioangel it's deployed on Vercel and will pick up the new dependencies via the `package.json` changes you have here - I just added you to the Vercel org so you can log in and look at those test deployments. It creates a build each time you push to a branch. If the branch is `main`, it also deploys the build to scrollprize.org. But in either case the build can be previewed there.

PR #121, opened 2025-02-18
Create chop_and_stack.py

skyward7187:
Script to chop up horizontally long images and stack the pieces into one image with a more easily viewable aspect ratio.


stephenrparsons:
Thanks @skyward7187! I'm going to go ahead and merge. We could probably make `num_pieces` have a default that is computed based on the input image width/height and that would make the output most resemble a square

PR #172, opened 2025-03-31
Model optimization framework

erdpx:
This PR adds a model optimization framework for nnUNet and two separate warmup schedulers (linear and exponential).

The framework makes it easier to run hyperparameter tuning via config files, generates trainer variants automatically, supports parallel training/inference, and uses the existing evaluation setup. This should help standardize experimentation and make it easier to add new losses, augmentations, and other components later.

Recommend creating a new branch called `model-optimization-framework`.


giorgioangel:
Thanks, Elian! Is this compatible with the wandb branch? https://github.com/ScrollPrize/villa/tree/nnunet-wandb/segmentation/models/arch/nnunet/nnunetv2

PR #177, opened 2025-04-06
Inference only

bruniss:
### volume 

move from models/volume_old.py to data/volume.py with 

added more normalization options

changed norm schemes to not use simple minmax 
    - 'none': No normalization
    - 'instance_zscore': Z-score normalization per slice/volume
    - 'global_zscore': Z-score with pre-computed mean/std
    - 'instance_minmax': Min-max scaling to [0, 1]
    - parameters for global_mean and global_std when using global normalization (would have to precompute)

extended return types

  - Added return_as_tensor option to directly return PyTorch tensors
  - Ensures tensors are properly contiguous for PyTorch compatibility
  - return_as_type parameter to control output dtype

### inference pipeline 

more in readme, but short explanation: 

1. `inference.py` - Performs model inference on input data in multiple parts
2. `blending.py` - Merges model prediction parts with Gaussian weighting
3. `finalize_outputs.py` - Processes blended logits into final outputs
4. `vesuvius_pipeline.py` - Orchestrates all steps with multi-GPU support

take model inputs , pass thru model, save patches as non-overlapping chunks by creating a zarr of [patch z * num_patches , patch y , patch z] and appending patches along the single dimension , this is necessary so gpus dont get backed up by blocking writes from long running partial chunk rights (given that each patch is overlapped 50%). also saves an array of all positions so that we can later recover this info in blending. if using num_parts > 1 and more than one part id, you can run this on an arbitrary number of nodes and blend data at a different time 

```python
vesuvius.predict \
  --model_path /path/to/model \
  --input_dir /path/to/input/volume.zarr \
  --output_dir /path/to/output/directory \
  --part_id 0 \
  --num_parts 1 \
  --overlap 0.5 \
  --batch_size 2 \
  --device cuda:0 \
  --skip_empty_patches
 ```
 
 compute a gaussian importance map which assigns higher values to logits closer to the center, and blend each patch into this weight array, then normalize by the number of overlaps (this needs some improvements to speed but its difficult to figure out exactly how to do this better). 
 
 ```python
vesuvius.blend_logits \
  /path/to/parts/directory \
  /path/to/output/blended.zarr \
  --cache_gb 8 \
  --sigma_scale 8.0
```

for 2 channel data, apply softmax over the bg/fg channels, and save only the fg. if threshold is provided, apply argmax over the fg channel. for multiclass, do the same but over all the channels

```python
vesuvius.finalize_outputs \
  /path/to/blended.zarr \
  /path/to/final_output.zarr \
  --mode binary \
  --threshold \
  --cache-gb 8
```

runs all of this in one go, with an option for multi-gpu, which will just spawn a different process for it. does not use ddp because i want to use this in k8s and i hate ddp 

```python
vesuvius.inference_pipeline \
  --input /path/to/input/volume.zarr \
  --output /path/to/final_output.zarr \
  --model /path/to/model \
  --gpus 0,1 \
  --mode binary \
  --threshold \
  --batch-size 2
```


giorgioangel:
@bruniss are these chatbot placeholders intended or there were some messages that were supposed to be printed?
https://github.com/ScrollPrize/villa/blob/98b59335f519cb1bbda20000c67ebd05ba882087/vesuvius/data/volume.py#L471
this message appears multiple time in this file

PR #218, opened 2025-05-13
Small corrections for Segmentation page (Docker file name and Docker image name)

spacegaier:
- Fix Docker file name (guide mentions Ubuntu 24.04 but uses 22.04)
- Fix incorrect Docker image name


bruniss (approved):
Nice catch, thank you!

PR #258, opened 2025-06-02
invert_surface_volume.py

giorgioangel:
Still to be optimized, but seems working. To be determined.

So far the script is taking as image input layers only `*.png`

Some multiprocessing in writing out to zarr could be beneficial.

Example command:
`python invert_surface_volume.py --obj /home/giorgio/Desktop/rendering_test/20231221180251.obj --images /home/giorgio/Desktop/rendering_test/layers --output /home/giorgio/Desktop/rendering_test`

### DIFF
--- a/vesuvius/rendering/invert_surface_volume_aggregator.py
+++ b/vesuvius/rendering/invert_surface_volume_aggregator.py
@@ -1,0 +1,393 @@
+#!/usr/bin/env python3
+"""
+invert_surface_volume_aggregator.py
+
+Aggregate per‐segment Zarr volumes into per‐scroll Zarr volumes by taking the voxelwise maximum
+across all segments of a given scroll—and optionally merging in a “predictions” Zarr—using Dask
+for parallel, out‐of‐core computation.
+
+Folder structure assumed:
+    segments_root/
+        <scroll_id>_<segment_id>/
+            labels/    ← a uint8 3D Zarr (DirectoryStore) with shape=(Z,Y,X)
+            mask/      ← a uint8 3D Zarr (DirectoryStore) with the same shape & chunks
+
+If --predictions-zarr is provided, it should point to a directory containing
+<scroll_id_predictions.zarr> stores. Each predictions Zarr must have the same overall shape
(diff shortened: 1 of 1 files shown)


giorgioangel:
I modified `invert_surface_volume.py` and added `invert_surface_volume_wrapper.py`

Having a folder full of images in this format `<scrollid>_<segment_id>_<layer_id>.png` , one should run the wrapper as follows with `--invppm_script invert_surface_volume.py` and with a number of workers such that your system memory can handle it.

giorgioangel:
After the segments renders have been inverted, one can merge them in a single per scroll zarr with the following:

invert_surface_volume_aggregator.py

Aggregate per‐segment Zarr volumes into per‐scroll Zarr volumes by taking the voxelwise maximum
across all segments of a given scroll—and optionally merging in a “predictions” Zarr—using Dask
for parallel, out‐of‐core computation.

Folder structure assumed:
    segments_root/
        <scroll_id>_<segment_id>/
            labels/    ← a uint8 3D Zarr (DirectoryStore) with shape=(Z,Y,X)
            mask/      ← a uint8 3D Zarr (DirectoryStore) with the same shape & chunks

If --predictions-zarr is provided, it should point to a directory containing
<scroll_id_predictions.zarr> stores. Each predictions Zarr must have the same overall shape
as the segment Zarrs, but may use a different chunk layout. For each scroll_id, this script
will produce:
    output_root/
        <scroll_id>_labels.zarr
        <scroll_id>_mask.zarr

Each output Zarr has the same shape, chunk‐layout, compressor, filters, etc., as the
segment inputs, and each voxel in “labels” is the maximum over:
    • all segment‐labels at that position, and
    • the predictions volume (if provided, over the same voxel‐range).

Usage:
    python invert_surface_volume_aggregator.py \
        --segments-root /path/to/segments \
        --output-root   /path/to/aggregated_zarrs \
        [--predictions-zarr /path/to/predictions] \
        [--use-distributed] \
        [--n-workers N] \
        [--threads-per-worker T] \
        [--log-level LEVEL]

PR #260, opened 2025-06-04
Add redirect to license page and remove redundant license

stephenrparsons:


stephenrparsons:
FYI @giorgioangel keeps scrollprize.org/license functional but makes us have just a single source of truth on that so I don't lose track of updating both

PR #261, opened 2025-06-04
Add tests to vesuvius

stephenrparsons:


vesuvius/setup.py
```diff
@@ -21,89 +22,89 @@ def run(self):
...
-        'psutil',
-        'pytorch-optimizer',
-        'tensorstore',
-        'blosc2'
+        "numpy",
+        "requests",
+        "aiohttp",
+        "fsspec",
+        "huggingface_hub",
+        "zarr",
```
jrudolph, line 43: zarr needs to be pinned to 2.x

PR #278, opened 2025-07-07
add anisotropic zarr with full z resolution

erdpx:
Usage example: `zarr_to_ome_fullres_z.py segments/20231005123336.zarr segments/20231005123336_anisotropic.zarr --algorithm mean --nlevels 7 --enable_smoothing --smooth_start_level 4 --keep_continuous`


giorgioangel:
@erdpx the code looks ok, thanks, just
https://github.com/erdpx/villa/blob/9d39cd1b6f5299876c4d9e8819b05069e260311d/foundation/scroll2zarr/zarr_to_ome_fullres_z.py#L625 this flag is not doing anything right now, it's always True

can we implement this as a PR to chuck's repository to the original file? https://github.com/KhartesViewer/scroll2zarr/blob/main/zarr_to_ome.py

@stephenrparsons can give a hand if needed

PR #282, opened 2025-07-08
Inital Grand Prize 3D import

SuperOptimizer:
This is the 3d version of the 2023 resnet ink detection code. Changes include

- The model now takes in a 3D ink label.
- It now works on full 64x64x64 chunks
- only zarr is supported in the training, but there are scripts for converting image stacks to zarr
- more 3D augmentations
- a script for downloading from webknossos
- improved surface volume samling

here is a sample inference of 20231005123336
20231005123336_ink_prediction


jrudolph (commented):
Great to have that here!

A few questions / remarks:
 * What is a "fragment"? Can we use consistent terminology in this repo and call things the same everywhere? Should it be "segment", "chunk", or "tile", instead?
 * For efficient deployment in the cluster we need:
   * containerization (not necessarily in this PR, I can also work on this at some point)
   * Don't expect multi-GPU nodes (for inference) but instead we need to be able to distribute the inference work load over multiple nodes by partitioning the work. Is that already prepared? What is needed to do that? E.g. the nnunet inference code can be passed a `--num_parts` and `--part_id` parameter to split up work across nodes. That's the most important bit to run inference code on scale.
 * Adding all the inklabels to the monorepo adds quite a tax on everyone cloning the repo for various purposes not related to this code. I don't have an immediate solution, just noting that adding binary file sets will slow down common operations and can add significant latency esp. to automation runs.
 * Can we put the project into the right subfolder, so that the root folder doesn't become a grab bag? @stephenrparsons where should things go?
 * A more high-level question is what the benefit is of using this model vs. "just using nnunet"? Every new pipeline we add brings additional integration and maintenance overhead and we need to be sure it will be worth it. Not saying that this hasn't been evaluated, I just don't know. In any case, it would be good if we could at least converge to single input/output (command line parameter) data format / library for all the inference jobs, so that we don't have to rewrite all pipelines for all new models from scratch (which with testing takes at least a week of work per pipeline)

bruniss:
@jrudolph i talked w/ youssef and elian and we are going to integrate the timesformer and i3d models into the vesuvius inference and training pipeine. this addition by @SuperOptimizer will still be great for running iterative runs and such through webknossos and the models from this could also easily work with the vesuvius inferencer. i think all of our large scale inference should run through vesuvius.inference_pipeline and its parts. 

there is no technical reason the inferencer we have cannot work with timesformer or i3d, it is model agnostic. it is just referred to in many cases as "nnunet" because it was initially built to use the same model architecture

SuperOptimizer:
I've not done much work with the code under the vesuvius folder or the inference pipeline so I wasn't aware of it, but I'll spend some time adapting this so that the integration is tighter. I didn't look hard, but I didn't see any ink detection stuff under there yet? 

At least for the onsite labeling, annotation, and model iteration work that's going on right now, keeping this code fairly separate is fine, but we don't necessarily need to merge this PR as is. I mostly want the papyrologists to have an easy, fairly standalone tool to work with.

gp3d/README.md
```diff
@@ -0,0 +1,119 @@
+# GP3D: Vesuvius 2023 Grand Prize Ink Detection 3D Version
```
giorgioangel, line 1: Why this title?

The Grand Prize model should be only one of the models usable by the pipeline, whose main purpose is automatizing ink iterative labeling from and to webknossos annotations.

gp3d/model.py
giorgioangel, line 1: Why are we defining a new model rather than using what we already have?

gp3d/train.py
giorgioangel, line 1: why are we defining a new training script and not using the trainers we already have?

SuperOptimizer:
I used the resnet 3d model from the grandprize ink detection, modified to work with 3d labels. The model definition had to change to use volumetric labels but I tried to keep it as close as possible to what was in the 2023 GP model. I spent some time trying to adapt the timesformer to use 3d labels but couldn't get it to work well. @younader should have the 3d version of timesformer but IIRC he hasn't uploaded it yet. The alternative would be to collapse the 3d ink labels into 2d to use the timesformer and/or resnet models currently in the ink-detection folder, though we'd lose the benefit of the 3d ink labels.

It seems I misunderstood the intent. I'll work on trying to more tightly integrate this into the existing vesuvius inference and training pipeline and see what I can come up with.

PR #300, opened 2025-07-22
Update 27_master_plan.md

giorgioangel:


bruniss:
just to extend my thought on this , would you be comfortable with the statement "ink on other scrolls is not detectable by either intensity or morphological structure" ? i would not

giorgioangel:
Merging it because the current version is too old. We can also edit later to modify, refine and add more details.

PR #303, opened 2025-07-23
Move vesuvius into module

bruniss:
this pr moves vesuvius into a module within the vesuvius directory,  which should allow easier imports ie : from vesuvius.models or import vesuvius.models.training , etc . additionally should (hopefully) fix installation on uv


stephenrparsons:
I made a handful of additions:
* Moved the library source into `src/`
* Simplified .gitignore
* Fixed a path for `vesuvius.accept_terms`
* Remove empty `data/io` directory

There are some others I'd like to make (e.g. moving `docs/` outside of `src/`) but we can make those separately.

This **does not** fix `uv` installation yet.

Basic commands seem to work per some tests with @bruniss's input, but let's add some tests ASAP so we don't have to do this manually every time.

Looks good to me - as mentioned above, can't guarantee full functionality since we don't have tests, but I think this is a step in the right direction and is not a regression.

stephenrparsons (approved):

PR #307, opened 2025-07-24
move aux targets / aux processing into module , remove binarize label…

bruniss:
…s and threshold labels

### DIFF
--- a/vesuvius/src/vesuvius/models/configuration/config_manager.py
+++ b/vesuvius/src/vesuvius/models/configuration/config_manager.py
@@ -8,6 +8,7 @@ from pathlib import Path
 import torch
 import torch.nn as nn
 from vesuvius.utils.utils import determine_dimensionality
+from vesuvius.models.training.auxiliary_tasks import create_auxiliary_task
 
 
 Image.MAX_IMAGE_PIXELS = None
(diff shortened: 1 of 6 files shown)


vesuvius/src/vesuvius/models/configuration/config_manager.py
```diff
@@ -331,56 +319,22 @@ def _apply_auxiliary_tasks(self):
         """
         if not self.auxiliary_tasks:
             return
-            
+
+        from models.training.auxiliary_tasks import create_auxiliary_task
```
stephenrparsons: move to top level import?

vesuvius/src/vesuvius/models/training/auxiliary_tasks/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+    create_distance_transform_config,
+    compute_distance_transform
+)
+
+from .aux_surface_normals import (
+    create_surface_normals_config,
+    compute_surface_normals_from_sdt
+)
+
+__all__ = [
```
stephenrparsons: `__all__` is probably not necessary here since this is straightforward imports that are quite clear in the remainder of `__init__.py` and we don't intend to encourage wildcard imports. could remove?

bruniss: makes sense ,  will remove this after you finsh the review

PR #309, opened 2025-07-24
fix ndim in spatial aug, and rework compose to respect --no-spatial flag

bruniss:


stephenrparsons (commented):
Looks reasonable, please test these locally with actual data since these are substantial logic changes and we don't have tests yet

PR #310, opened 2025-07-24
refactor-img-dataset

bruniss:
simplify initialization of volumes, remove extraneous comments, add support for unlabeled data


stephenrparsons (changes requested):
Please break up into component parts (at least commits if not PRs)

PR #311, opened 2025-07-24
Refactor img dataset

bruniss:
- remove mask support ( should be a label index) 
- remove useless comments
- rework volume initialization to support unlabeled patches, and track them


vesuvius/src/vesuvius/models/datasets/image_dataset.py
```diff
@@ -214,124 +185,118 @@ def _initialize_volumes(self):
...
-        if not labels_dir.exists():
-            raise ValueError(f"Labels directory does not exist: {labels_dir}")
 
-        images_group, labels_group = self._get_or_create_zarr_groups()
+        has_labels = labels_dir.exists()
+        if not has_labels and not self.allow_unlabeled_data:
+            raise ValueError(f"Labels directory does not exist: {labels_dir} and allow_unlabeled_data=False")
+
         configured_targets = set(self.mgr.targets.keys())
+        configured_targets = {t for t in configured_targets
```
stephenrparsons, line 166: Comment could help here. This filters for only those that do not have auxiliary_task in their targets dict?

bruniss: yeah i'll throw a comment in , if an aux task exists the target handling is passed off to the aux task handling (which just takes the original targets and appends them) -- this is a little janky but previously i had to have aux task handling in every single dataset, which was not ideal

PR #350, opened 2025-08-15
Fix reference links in vesuvius/src/README.md

ragul-kachiappan-dev:
Per the moves to files made in #303 , several reference links in [readme](https://github.com/ScrollPrize/villa/blob/main/vesuvius/README.md) for Python package have become dead links including links to other reference docs and sample images. This PR updates those links to the correct path to restore README to its original state


stephenrparsons:
@ragul-kachiappan-dev thank you!

PR #360, opened 2025-09-02
Implementation of an optimized inference container

repushko:
This PR introduces a production-focused, containerized inference workflow for ink detection using a TimeSformer model. It downloads model weights from Hugging Face, fetches input layers from S3, performs tiled GPU inference, uploads a PNG prediction to S3, and exposes the result S3 URI via a well-defined file output for workflow orchestration.

### DIFF
--- a/ink-detection/optimized_inference/Dockerfile
+++ b/ink-detection/optimized_inference/Dockerfile
@@ -30,7 +30,6 @@ ENV PYTHONPATH=/app
 ENV CUDA_VISIBLE_DEVICES=0
 
 COPY --from=build /opt/venv /opt/venv
-COPY config.py /app
 COPY entrypoint.py /app
 COPY inference_timesformer.py /app
 
--- a/ink-detection/optimized_inference/Dockerfile
+++ b/ink-detection/optimized_inference/Dockerfile
@@ -1,0 +1,40 @@
+ARG BASE_IMAGE=pytorch/pytorch:2.1.0-cuda12.1-cudnn8-runtime
+
+FROM ${BASE_IMAGE} as base
+ENV DEBIAN_FRONTEND=noninteractive
+WORKDIR /app
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    libgl1-mesa-glx \
+    libglib2.0-0 \
+    libsm6 \
+    libxext6 \
+    libxrender1 \
+    libgomp1 \
+    ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
+
+# Build venv layer
+FROM base as build
+COPY requirements.txt /tmp/requirements.txt
+RUN python -m venv /opt/venv \
+ && . /opt/venv/bin/activate \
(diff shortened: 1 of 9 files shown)


ink-detection/optimized_inference/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+ENV PYTHONPATH=/app
+ENV CUDA_VISIBLE_DEVICES=0
+
+COPY --from=build /opt/venv /opt/venv
+COPY config.py /app
+COPY entrypoint.py /app
+COPY inference_timesformer.py /app
+
+# Ensure minimal size
+RUN rm -rf /root/.cache /app/__pycache__
```
jrudolph, line 37: This does not work, it will not make the image smaller as the files are already part of previous layers.

ink-detection/optimized_inference/entrypoint.py
```diff
@@ -0,0 +1,339 @@
...
+@dataclass
+class Inputs:
+    model_key: str
+    s3_path: str
+    start_layer: int
+    end_layer: int
+    force_reverse: bool = False
+
+
+MODEL_TO_HF_REPO: Dict[str, str] = {
```
jrudolph: Can you remove hard-coded references and replace by both hf:// and s3:// addresses? We need to host models ourselves for own usage because hf has very tight limits and we will quickly run out of quota running into errors when pulling from hf.

ink-detection/optimized_inference/Dockerfile
jrudolph: Maybe it's a left-over because where would the files even come from?

jrudolph (commented):
I added a few initial comments

ink-detection/optimized_inference/entrypoint.py
repushko: Yep, properly fixed

giorgioangel:
@jrudolph can this be merged?

ink-detection/optimized_inference/entrypoint.py
```diff
@@ -0,0 +1,372 @@
...
+    # Stack layers using the same method as local script
+    # This creates (H, W, C) format like the working version
+    stacked_layers = np.stack(images, axis=2)
+    
+    # Ensure proper dtype - match the local script behavior
+    # Don't convert to float32 here, let the inference function handle it
+    return stacked_layers
+
+
+def download_model_weights(model_name: str, dest_dir: str, s3_client) -> str:
```
jrudolph, line 170: Can we use an url instead of the model name here? The logic should be:

 * if an `hf` url is given translate to well-known `s3` caching location. If the weights can be found there, use that s3 location. Otherwise, download the model weight to that location first.
 * If an `s3` url is given download to local fs and use it.

repushko: We'll use a model name, but we'll use s3 as a first place to look at, and if it's not found, we'll use HF. Model name is more convenient than s3 path

PR #389, opened 2025-09-08
Support additional formats of layers (png+jpeg)

repushko:
As one of the first tests, we've figured out that default layers format can be not only `.tiff`, but also `.png` and `.jpeg`. In this PR, this support is introduced.


jrudolph:
Does the inference pipeline just support it like that?

repushko:
> Does the inference pipeline just support it like that?

@jrudolph  yep, I've tested that with `.jpeg` -- all good. Here is the `.jpeg` results from Scroll5
prediction_timesformer_scroll5_27112024_00_26

PR #393, opened 2025-09-09
Ordered candidate set

giorgioangel:
1. Ordering the candidate set instead of unordered
2. Fixing shared_unlocked() in some points
3. Stabilizing the deterministic jitter with the hash of the surface name rather than the pointer

Please have a look because I am not entirely sure point 1 is well implemented


hendrikschilling:
Looks right to me. Just note that for deterministic behaviour it must be run without omp (or threads = 1), otherwise timing influence the choice of candidates available.

PR #464, opened 2025-09-28
CWindow watchdir + Autoreload surfaces

SuperOptimizer:
Hendrik wanted a way to reload surfaces that are changed externally by subprocess calls to vc_grow_seg_from_seed and related. He suggested inotify but after some experimenting I wanted to do something a bit easier. It seems like inotify requires one watch per directory, and with some 10s of thousands of directories that seemed a bit much. Instead I did up a simple watch directory: cwindow watches a directory in /tmp for files. the names of the files are the segment(s) to add, remove, or reload. We just need to pass this watch directory to subprocesses, and when the subprocess writes out the changes, needs to just do a `touch /tmp/watchdir/auto_grown_segment_i_just_modified` and CWindow will reload it. CWindow checks every 1 second right now which seems like a good frequency.

### DIFF
--- a/volume-cartographer/apps/VC3D/CWindow.cpp
+++ b/volume-cartographer/apps/VC3D/CWindow.cpp
@@ -68,7 +68,9 @@ CWindow::CWindow() :
     _cmdRunner(nullptr),
     _seedingWidget(nullptr),
     _drawingWidget(nullptr),
-    _point_collection_widget(nullptr)
+    _point_collection_widget(nullptr),
+    _inotifyFd(-1),
+    _inotifyNotifier(nullptr)
 {
     _point_collection = new VCCollection(this);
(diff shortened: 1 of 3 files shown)


hendrikschilling:
I think a single inotify watch on the paths directory should be enough to detect the events we are interested in?

hendrikschilling:
Thats why I made the write to a temp dir and then a rename to actual target dir (also to make the rewrite atomic). I think inotify should catch that.

SuperOptimizer:
> Thats why I made the write to a temp dir and then a rename to actual target dir (also to make the rewrite atomic). I think inotify should catch that.

ahh gotcha. ok, Ive got the old inotify code in a commit. I'll resurrect it to check and see if that works

giorgioangel:
@SuperOptimizer is this ready?

PR #470, opened 2025-09-30
fix(CodeQL): replace std::gmtime with thread-safe version

giorgioangel:
- Avoids static storage from std::gmtime (CodeQL rule cpp/potentially-dangerous-function)
- Uses gmtime_s on Windows and gmtime_r on POSIX
- Stops shadowing the time_t type; adds <ctime>


jrudolph:
Can we put code like this into a proper method, so that we don't get platform specific code everywhere? How significant is that warning for us in the first place?

jrudolph (approved):
:+1:

PR #488, opened 2025-10-05
Add Symmetric‑Dirichlet (SD) regularizer to the surface growth solver

giorgioangel:
**TL;DR**
Introduce a **Symmetric-Dirichlet (SD)** distortion term to `GrowPatch.cpp` (opt-in via `sdir_weight`, default **0.0**). SD adds a **collapse barrier**, is **rotation-invariant** (cell level), and treats **shrink/stretch symmetrically**. It complements our existing **edge-length** terms (axis **and** diagonals) and **straightness** smoothing, improving robustness and final surface quality—especially in low-signal regions or under strong external alignment/snapping.

---

## Motivation

We currently optimize with:

* **Edge-length preservation** (`DistLoss`) on **axis neighbors** and **diagonals**.
* **Straightness** (`StraightLoss`) along rows/cols and diagonals.
* Optional **direction/normal field** alignment and **snapping** to data.

This works well, but in challenging zones we still see:

* **Near-degenerate quads** (usually acute angles) when other terms dominate.
* **Excessive anisotropy** (poor local conditioning) in low-signal areas.
* **Occasional solver instability** around ill-conditioned patches.

Edge terms—including both diagonals—do provide a *quadratic* shear penalty near the optimum, but they still:

1. Have **no true collapse barrier** (cost remains bounded as area → 0).
2. Are **not rotation-invariant** at the cell level (they sample a finite set of directions).
3. Respond **asymmetrically** to shrink vs. stretch.

**Symmetric-Dirichlet** directly addresses these gaps.

---

### 1) Adds SD as a per-cell residual

* New loss: `SymmetricDirichletLoss` in `CostFunctions.hpp`.
  It minimizes (dimensionless)
  $$E_{\text{SD}}(J) = \mathrm{tr}(G) + \mathrm{tr}(G^{-1}) - 4 \approx \mathrm{tr}(G) + \dfrac{\mathrm{tr}(G)}{\mathrm{det}(G) + \varepsilon} - 4 ,$$ $$\qquad G = J^\top J$$ and $$J=[e_u;e_v]$$
  with a small $\varepsilon$ in the denominator implementation to stabilize near-degeneracy.

* Integrated via:

  * `gen_sdirichlet_loss(...)` and `conditional_sdirichlet_loss(...)`
  * Added in `add_continuous_losses(...)` (local solves) and `add_missing_losses(...)` (global/patch-wide completion)

* **Normalization**: Each edge vector is divided by `unit = step*scale`, so SD is zero at the desired scale (an isometry).

### 2) Configuration & defaults (backwards compatible)

* `sdir_weight` (default **0.0**, **off** by default → preserves current behavior)
* $$\varepsilon$$ (HARDCODED to **1e-8**) to mollify the barrier near $\det(G)\to 0$

Example:

```json
{
  "sdir_weight": 0.03,
}
```

---

## Why SD with our current edge/diagonal setup?

**What we already have:**

* **Axis edges**: $|e_u|\approx 1$, $|e_v|\approx 1$ (in unit-less form)
* **Diagonal edges**: $|e_u\pm e_v|\approx \sqrt{2}$, which does penalize **shear** to second order near the optimum
* **Straightness**: smooths curvature; no metric control

**What SD adds:**

* A **collapse barrier** via the $1/\det(G)$ behavior (no bounded “cheap” path to nearly zero area).
* **Symmetric** control of shrink/stretch in **all** tangent directions (singular values and their inverses).
* **Rotation-invariant** distortion control (cell level): acts on the eigenvalues of $G$, not on a finite sampling of directions.

**Net effect:** The **edge+diagonal** terms continue to **anchor scale** cleanly and provide simple gradients (great for early growth), while **SD** prevents pathological shapes and improves local conditioning in hard regions.

---

## Code overview (key additions)

* `CostFunctions.hpp`

  * `struct SymmetricDirichletLoss`
* `GrowPatch.cpp`

  * `LossType::SDIR` and JSON plumbing (`sdir_weight`)
  * `gen_sdirichlet_loss(...)` / `conditional_sdirichlet_loss(...)`
  * Calls inside `add_continuous_losses(...)` and `add_missing_losses(...)`

---

## Risks & mitigations

* **Over-constraint early in growth** → keep `sdir_weight` modest or ramp it during growth.
* **Sensitivity near singularities** → $$\varepsilon$$ and modest `sdir_weight` prevent blow-ups while preserving a strong barrier.

---

## Conclusion

Adding **Symmetric-Dirichlet** provides the missing **collapse barrier**, **symmetric distortion control**, and **rotation-invariant** regularization that edge+diagonal and straightness terms cannot offer alone. It is **opt-in**, **inexpensive**, and materially improves robustness and surface quality where it matters most.

### DIFF
--- a/volume-cartographer/apps/VC3D/CWindowContextMenu.cpp
+++ b/volume-cartographer/apps/VC3D/CWindowContextMenu.cpp
@@ -2,6 +2,10 @@
 #include "CSurfaceCollection.hpp"
 #include "SurfacePanelController.hpp"
 
+#include <functional>
+#include <algorithm>
+#include <iostream>
+
 #include <QSettings>
 #include <QMessageBox>
 #include <QProcess>
@@ -18,18 +22,17 @@
 #include <QTextStream>
 #include <QtGlobal>
 #include <QProcessEnvironment>
+#include <QProgressDialog>
+#include <QPointer>
+#include <QTimer>
 #if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
-#include <QStandardPaths>   // for QStandardPaths::findExecutable
+#include <QStandardPaths>
 #endif
 
 #include "CommandLineToolRunner.hpp"
 #include "vc/core/types/VolumePkg.hpp"
 #include "vc/core/util/Surface.hpp"
 #include "ToolDialogs.hpp"
-
-
-
-
 
 // --------- local helpers for running external tools -------------------------
 static bool runProcessBlocking(const QString& program,
@@ -42,11 +45,8 @@ static bool runProcessBlocking(const QString& program,
     if (!workDir.isEmpty()) p.setWorkingDirectory(workDir);
     p.setProcessChannelMode(QProcess::SeparateChannels);
 
-    // Print the entire program invocation
     std::cout << "Running: " << program.toStdString();
-    for (const QString& arg : args) {
-        std::cout << " " << arg.toStdString();
-    }
+    for (const QString& arg : args) std::cout << " " << arg.toStdString();
     std::cout << std::endl;
 
     p.start(program, args);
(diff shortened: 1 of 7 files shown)


hendrikschilling:
Could you observe any improvements or effects from the loss?
I see no issue adding additional losses that are not enabled by default, as long as they don't complicate the code, but these changes do cleanup, remove sensible comments and reorder some code, if this is desired its better if done separately (and I would like to keep most of the comments and some of the removed code). If the reason is LLMs - prompt it to do changes as small diffs instead of file re-writes.

giorgioangel:
@hendrikschilling Agreed. Yes I usually ask to LLM to an audit of the code, but it rewrote it deleting all the comments to preserve tokens. I am slapping them back.

Regarding the effect of the loss, I do notice (visually at least) that there is less warping in some regions when I put the weight between `0.01` and `0.03`. I am personally liking `0.03`. Already going `> 0.05` up to  `0.1` seems to be useful in some cases, but it blows the patch up in many others.
 `> 0.1` is already too much.
 
 I am looking forward to observe the plots with vis_losses true, because my visual interpretation could be wrong.

hendrikschilling:
What does blow up mean in this case? Creating misformed quads or just lots of sheet switches / not following the other losses.

giorgioangel:
@hendrik image

this for instance is with `0.32`

giorgioangel:
This can blow up when the det(G) is small... which means when during the optimization process the area of the proposed quad is going to zero. We do want a strong barrier for small areas, but we do not want it to be so strong to dominate over the other losses (data terms).

PR #527, opened 2025-10-22
inotify watcher

SuperOptimizer:
this adds an inotify watcher to cwindow to update the ui and detect added, removed, and renamed directories in paths/ or traces/ . It also includes some cahnges to Segmentation and Volume Pkg to move around surface ownership.

The inotify events are collected in a queue and processed every 10 seconds. I didn't want to update every time an event fires immediately because this can trigger some gui redraws and intersection redraws which can get slow if you get more than >1 or so events / second

### DIFF
--- a/volume-cartographer/apps/VC3D/CWindow.cpp
+++ b/volume-cartographer/apps/VC3D/CWindow.cpp
@@ -3376,7 +3376,6 @@ void CWindow::processInotifySegmentRemoval(const std::string& dirName, const std
     }
 }
 
-
 void CWindow::processPendingInotifyEvents()
 {
     if (_pendingInotifyEvents.empty() && _pendingSegmentUpdates.empty()) {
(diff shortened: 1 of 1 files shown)


volume-cartographer/cmake/VCFindDependencies.cmake
```diff
@@ -4,6 +4,11 @@ include(FetchContent)
 option(VC_BUILD_JSON "Build in-source JSON library" OFF)
 option(VC_BUILD_Z5   "Build (vendor) z5 header-only library" ON)
 
+find_package(CURL REQUIRED)
```
SuperOptimizer: I was having some weird linker errors with cmake picking up stuff from miniconda and this seemed to have fixed it? along with the core/cmakelsits.txt change

hendrikschilling:
Ok, just got to testing it:
- it seems that the patch grower creates move events and those are not actually processed
- Also I understand the desire to throttle it, but its also nice to not wait 10s. Could we check wether the last update is less than 10s ago and if yes create the timer as a one-shot with the remaining time to the 10s and of not then just process immediately.
- lastly - the timer needs to be initialied before the openVolpkgAt from auto-open as that will call close and also access the timer.

SuperOptimizer:
> it seems that the patch grower creates move events and those are not actually processed

fixed. running `(base) forrest@forlinlap ~/CLionProjects/villa/volume-cartographer/cmake-build-releasegcc/bin (main*) $ for i in {1..1000}; do /home/forrest/CLionProjects/villa/volume-cartographer/cmake-build-releasegcc/bin/vc_grow_seg_from_seed "/home/forrest/scroll3.volpkg/volumes/ensemble_ome.zarr" "/home/forrest/scroll3.volpkg/paths" "/home/forrest/scroll3.volpkg/expand.json"; done` from the commandline shows the latest segment getting added and the ui auto updating

> Also I understand the desire to throttle it, but its also nice to not wait 10s. Could we check wether the last update is less than 10s ago and if yes create the timer as a one-shot with the remaining time to the 10s and of not then just process immediately.

fixed by adding the timer

> lastly - the timer needs to be initialied before the openVolpkgAt from auto-open as that will call close and also access the timer.

fixed

PR #543, opened 2025-11-02
more robust ignore label support

bruniss:


giorgioangel:
@bruniss what's this?

PR #546, opened 2025-11-02
vc sync support for backups, download only upload only, and enable segment

SuperOptimizer:


SuperOptimizer:
David wanted some extra features for vc_sync but I'm not sure if he's had the time to actually look. I'll ping him about it

Issue #295, opened 2025-07-19
in vesuvius.train if the user names one of their targets 'mask' the dataloader/trainer will mix up these and the ignore masks


bruniss:
solution is to assert user named targets cannot be called 'mask'

PR #555, opened 2025-11-09
vesuvius.train - Validate target names

Regan-Koopmans:
Resolves https://github.com/ScrollPrize/villa/issues/295

Currently we allow users to pass in target names that conflict with internal names. Now, if a user attempts to use a reserved name they will see something like:
```
Target name 'mask' is reserved and cannot be used. Reserved names: is_unlabeled, mask, plane_mask. Please choose a different name for your target.
```


bruniss:
thank you for the PR ! good catch on this

PR #557, opened 2025-11-11
better ish intersection rendering

SuperOptimizer:
Our colored intersection line rendering has a few deficiencies: non deterministic rendering, pop in and out, discontinuities, not tracing the entire length of the viewport, and being slow. This fixes all the issues except the slowness, but that is coming along with the spatial grid I am using, but more improvement is needed there.


SuperOptimizer:
this is actually fucked

PR #563, opened 2025-11-13
stretch values from readinterpolated3d to 0 - 255

SuperOptimizer:


SuperOptimizer:
gonna tackle this in a different way

PR #756, opened 2026-03-02
[wip] inference: attempt to integrate kaggle post processing

jrudolph:
Used verbatim it created gaps between chunks (which was removed in 5529086bc8e4718974cff97c0bd86a72da34ddc1).


giorgioangel:
@codex review

jrudolph:
Let's close this for now, if we think we need better results in surface inference we can recycle.

PR #799, opened 2026-03-28
[codex] Add multiscale surface training and BCE+Dice ignore-label support

giorgioangel:
## What changed

This PR adds multiscale surface-training support for `vesuvius` and introduces a BCE+Dice training path that works with the existing `ignore_label: 2` surface labels.

Concretely it:
- loads `dataset_config.ome_zarr_resolution` into the training config manager and validates it against `valid_patch_find_resolution`
- makes patch-cache generation and lookup scale-aware by including the training resolution in the cache key
- converts training-resolution patch sizes to full-resolution patch sizes for patch finding
- converts cached full-resolution patch coordinates back to the selected OME-Zarr training level when reading cached patches
- adds `BinaryBCEAndDiceLoss` for single-channel surface training with `ignore_label: 2`
- adds a GPU fit-probe helper for patch-size and batch-size experiments
- adds 8 surface training configs covering scale 0/2, `256^3 @ bs=3` and `128^3 @ bs=28`, with both MedialSurfaceRecall and BCE+Dice variants
- adds focused regression tests for multiscale config loading, cache separation, coordinate scaling, and the new BCE+Dice loss path

## Why

Two feature gaps were blocking the intended surface experiments:

1. Multiscale training was not wired correctly end-to-end.
   Patch finding cached full-resolution coordinates, but the dataset could open lower OME-Zarr levels without converting those coordinates back to the training level. Cache keys also did not distinguish between scale 0 and scale 2.

2. The existing nnUNet BCE+Dice path was not a good fit for the current scalar `0/1/2` surface labels.
   It expects region-based targets, while these runs need single-channel binary logits with `ignore_label: 2` preserved.

## Impact

- Surface experiments can now target both OME-Zarr scale 0 and scale 2 correctly.
- Cache files are isolated per training scale, avoiding cross-scale cache reuse bugs.
- Surface BCE+Dice runs can use the same ignore-label semantics as the MedialSurfaceRecall runs.
- The added configs make the 8-run comparison matrix reproducible.

## Validation

Executed and verified on the remote H100 node:
- `python3 -m py_compile` for the modified Python files
- `uv run --all-extras pytest vesuvius/tests/models/test_surface_multiscale_training.py -q`
- `uv run --all-extras pytest vesuvius/tests -k "ome_zarr_resolution or patch_cache or zarr_dataset or surface_multiscale_training" -q`
- fit probe results:
  - `256^3 @ bs=3` fits on a single H100
  - `128^3 @ bs=28` fits on a single H100
- `vesuvius.find_patches` completed for the 8 configs, producing 4 unique cache files and 4 expected cache hits across loss variants
- the corresponding remote training runs were launched successfully with W&B logging enabled

## Notes

This PR intentionally scopes to the multiscale training fix, the BCE+Dice compatibility path, the probe helper, tests, and the new experiment configs. It does not include unrelated local or remote worktree changes.

### DIFF
--- a/vesuvius/src/vesuvius/models/build/build_network_from_config.py
+++ b/vesuvius/src/vesuvius/models/build/build_network_from_config.py
@@ -849,9 +849,11 @@ class NetworkFromConfig(nn.Module):
             return False
         return True
 
-    def forward(self, x, return_mae_mask=False):
+    def forward(self, x, return_mae_mask=False, apply_activation=None):
         # Check input channels and warn if mismatch
         self.check_input_channels(x)
+        if apply_activation is None:
+            apply_activation = not self.training
 
         # Get features from encoder (works for both U-Net and Primus)
         # For MAE training with Primus, we need to get the mask
@@ -897,7 +899,7 @@ class NetworkFromConfig(nn.Module):
                 logits = logits[0]
             logits = self._apply_z_projection(task_name, logits)
             activation_fn = self.task_activations[task_name] if task_name in self.task_activations else None
-            if activation_fn is not None and not self.training:
+            if activation_fn is not None and apply_activation:
                 if isinstance(logits, (list, tuple)):
                     logits = type(logits)(activation_fn(l) for l in logits)
                 else:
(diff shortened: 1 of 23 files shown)


giorgioangel:
@bruniss can we merge this?

PR #808, opened 2026-04-07
scrollprize.org: fix vercel deployment

jrudolph:
It seems vercel started failing probably because of some transitive dependency resolution hiccups.

According to claude, vercel's yarn classic build might be not picking up our lock file which would prevent the issue and the fix is to pin a package manager version.

### DIFF
--- a/scrollprize.org/package.json
+++ b/scrollprize.org/package.json
@@ -2,6 +2,7 @@
   "name": "my-website",
   "version": "0.0.0",
   "private": true,
+  "packageManager": "yarn@4.6.0",
   "scripts": {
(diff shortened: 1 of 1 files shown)


jrudolph:
This didn't fix it. It seems something changed behind our backs that leads to a different dependency resolution than before which isn't compatible any more (if I rebuild old versions on vercel the same problem now shows up).

jrudolph:
It now worked after also adding an env var

```
ENABLE_EXPERIMENTAL_COREPACK=1
```

in the vercel config for the project. This allows vercel to pick up the latest yarn version when building and then actually uses the yarn lock file which was ignored before.

PR #817, opened 2026-04-16
Optimize surface patch index and intersection line rendering

SuperOptimizer:
Cuts foreground intersection cost roughly in half (~18% -> ~8% of CPU under "10 segments visible, panning XY/XZ/YZ" workload).

- forEachTriangleImpl is now a private template; queryTriangles and computePlaneIntersections instantiate it with raw lambdas instead of going through std::function. The public std::function-taking forEachTriangle overloads are unchanged for external callers.
- computePlaneIntersections now clips inside the rtree visitor and appends TriangleSegments straight to per-surface buckets. The intermediate TriangleCandidate vector and the bySurface grouping pass are gone, along with their per-triangle push_back of a ~150-byte struct. Optional triangleBuf/surfaceBuf parameters are removed (only caller didn't use them).


SuperOptimizer:
@codex review

PR #818, opened 2026-04-16
Decouple rtree-stride from triangulation-stride; enable stride 1

SuperOptimizer:
## Summary

- **Decouple rtree storage stride from triangulation stride** in `SurfacePatchIndex`. The rtree stores one entry per `tileStride×tileStride` source-mesh tile (`tileStride = max(samplingStride, 8)`); the visitor sub-iterates at the user-facing `samplingStride` within each tile to emit fine triangles. Lets us actually use stride-1 triangulation without OOM (was ~21 GB for 9 surfaces previously; now ~330 MB).
- **Refine all tiers to stride 1** in the `ViewerManager` tier table. Coarse default still chosen for fast first-paint based on surface count.
- **Fix tier-vs-refinement infinite loop**: the refinement step's re-prime was re-running the tier code each cycle, clobbering the refined stride back to the coarse default. Added a `_surfacePatchStrideTiered` flag so tier code only runs once per surface set.

## Perf

`10 segments visible, panning XY/XZ/YZ` workload, foreground intersection cost vs. previous PR (#817):

| | Stride 4 (PR #817) | Stride 1 + tiled (this PR) |
|---|---|---|
| R-tree visitor | 2.87% | **0.76%** |
| R-tree pack (bg) | 3.21% | 0.57% |
| Total foreground | ~3% | **~1%** |

Stride-1 triangulation is now actually cheaper than stride-4 was before — the bbox-vs-plane reject kills 64-cell tiles cheaply, and only surviving tiles do per-sub-cell work.

## Test plan

- [x] No OOM at stride 1 (was 21 GB → now ~330 MB)
- [x] Visual: stride-1-fine intersection lines render correctly
- [x] Manual stride switching (e.g. user-set stride 32 → back to 1) works
- [x] Tier defaults still apply on first-time load per surface set; manual override sticky

### DIFF
--- a/volume-cartographer/apps/VC3D/ViewerManager.cpp
+++ b/volume-cartographer/apps/VC3D/ViewerManager.cpp
@@ -481,37 +481,34 @@ void ViewerManager::primeSurfacePatchIndicesAsync()
         _surfacePatchIndex.clear();
         _indexedSurfaceIds.clear();
         _surfacePatchIndexNeedsRebuild = false;
+        _surfacePatchStrideTiered = false;
+        _targetRefinedStride = 0;
         return;
     }
 
-    // Apply tiered default stride based on surface count (if not user-set)
+    // Apply tiered default stride only on the first prime per surface set.
+    // Subsequent primes (refinement step, edits) preserve the current stride
+    // and the (possibly already-consumed) refinement target. Without this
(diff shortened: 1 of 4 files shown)


SuperOptimizer:
@codex review

PR #819, opened 2026-04-16
VC3D usability: flattened z-scroll fix + retry remote segments

SuperOptimizer:
## Summary

Two independent usability fixes for VC3D.

**Flattened-view shift+scroll rigid translation**
Previously passed `zOff` through to `QuadSurface::gen` as `offset[2]`, applied as a per-pixel local-normal offset. On a curved scroll that's not a rigid translation, so plain zoom at nonzero `zOff` exposed drift that looked like simultaneous z-scroll + zoom. Now `zOff` is applied as a single world-space translation whose direction is captured from the surface normal at view center on each shift+scroll. Plain zoom just magnifies the translated content, no drift.

**File → Attach Remote Segments…**
The prompt for a remote segments path only appeared once on remote-volume open. A wrong paste / missing creds / dismissed prompt left no way to retry. Adds a menu entry that reuses the existing fetch flow against the currently loaded remote volume. Uses `volume->path().parent_path()` as the cache root (since `NewFromUrl` stages at `cacheRoot/volumeId`), so previously-cached segments are reused.

## Test plan

- [x] Shift+scroll in flattened segmentation viewer translates rigidly; plain zoom at nonzero zOff is pure magnification
- [x] Shift+scroll in XY/XZ/YZ plane viewers unchanged
- [x] File → Attach Remote Segments succeeds after an initial bad path
- [x] Attaching reuses cached segment data (no re-download)

### DIFF
--- a/volume-cartographer/apps/VC3D/MenuActionController.cpp
+++ b/volume-cartographer/apps/VC3D/MenuActionController.cpp
@@ -145,6 +145,9 @@ void MenuActionController::populateMenus(QMenuBar* menuBar)
     _attachRemoteZarrAct = new QAction(QObject::tr("Attach Remote &Zarr..."), this);
     connect(_attachRemoteZarrAct, &QAction::triggered, this, &MenuActionController::attachRemoteZarr);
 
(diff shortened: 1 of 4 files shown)


SuperOptimizer:
@codex review

PR #821, opened 2026-04-18
Move a bunch of the rendering pipeline to a tick based update system

SuperOptimizer:
tldr: moved a bunch of the zarr encode decode disk read network read block caching etc to a tick based update system to try to cut down on useless work and instead just try to update everything on a 16ms based tick system.

## Summary

Accumulated perf and architecture work on the render + cache path over several sessions.

- **Lock-free BlockCache shard map + emptyChunks set.** Open-addressed, 32 shards × 256K atomic entries + 8-way L2. Replaces the previous per-shard `pthread_rwlock`/`shared_mutex`. `BlockCache::get` dropped from ~44% → ~2.7% of render CPU.
- **Tick-batched state publication (`TickCoordinator`).** Dedicated `std::jthread` at 16 ms, double-buffered `FrameState`, monotonic release-gen protocol. No `shared_ptr`, multi-reader safe. Main thread untouched.
- **MPSC producer events** (`ChunkLanded`, `EmptyChunkNoted`) drain into `FrameState::emptyChunkKeys` (sorted vec) and `FrameState::slice` ({packedBlockKey, pipeline*, Block*}).
- **Coalesced prefetch.** Viewer render paths enqueue fetch keys through `TickCoordinator`; tick thread groups by `(pipeline, targetLevel)`, dedups, dispatches one `fetchInteractive` per group per tick instead of 8–20 per render cycle across 4 viewers.
- **ViewportSnapshot seqlock** per viewer so the coordinator can scope the slice to active pyramid levels.
- **`BlockSampler` reader side** binary-searches `emptyChunkKeys` and the slice on slot miss — plain memory, skips atomic probes inside `BlockPipeline::blockAt` for hot data.
- **Per-thread `ChunkData` pool** on decode path to cut malloc/mmap churn (~30% of CPU was in kernel mm on heavy decode).
- **Fused sample loop** + skip of inactive TF LUTs.
- **Configurable TF + composite pipeline** (pre-TF, post-TF, DVR, PBR, volumetric, first-hit iso, deviation-from-mean).
- **Progressive render** + 8-way L2 + tile-parallel render + sharded block cache.
- **FPS readout** rewritten to report 1 / mean(render duration) instead of render count / wallclock, so the number reflects render kernel speed rather than user-input pacing.

Profile after all this: 55% raw sample/composite kernel (irreducible), 2.7% `BlockCache::get`, 0.35% `BlockPipeline::blockAt`. Everything else < 1%.

## Test plan

- [ ] Build `cmake-build-releaseunsafe` cleanly (all warnings pre-existing).
- [ ] Open a volume in VC3D, verify 4 viewers render.
- [ ] Pan/zoom each viewer; confirm smooth rendering.
- [ ] Switch composite modes (Max / Mean / Min / Alpha / BeerLambert / DVR / FirstHitIso / DevFromMean); confirm each renders without crash.
- [ ] Toggle pre-TF and post-TF with non-identity knots; confirm output changes as expected.
- [ ] Run `perf record -F 999 --call-graph dwarf` for ~30 s and compare top symbols against baseline (`perf.before*.data` in the repo during development).
- [ ] Leave idle for 60 s; verify TickCoordinator CPU stays < 1%.
- [ ] Close cleanly; no jthread join hangs, no leaks.

### DIFF
--- a/volume-cartographer/apps/CMakeLists.txt
+++ b/volume-cartographer/apps/CMakeLists.txt
@@ -96,6 +96,24 @@ target_link_libraries(vc_diffuse_winding
 add_executable(vc_render_bench src/vc_render_bench.cpp)
 target_link_libraries(vc_render_bench vc_core)
 
+add_executable(vc_io_bench src/vc_io_bench.cpp)
+target_link_libraries(vc_io_bench vc_core)
+
+add_executable(vc_blockcache_bench src/vc_blockcache_bench.cpp)
+target_link_libraries(vc_blockcache_bench vc_core)
+
+add_executable(vc_transpose_bench src/vc_transpose_bench.cpp)
+target_link_libraries(vc_transpose_bench vc_core)
+
+add_executable(vc_coord_bench src/vc_coord_bench.cpp)
+target_link_libraries(vc_coord_bench vc_core)
+
+add_executable(vc_tifxyz_bench src/vc_tifxyz_bench.cpp)
+target_link_libraries(vc_tifxyz_bench vc_core)
+
+add_executable(vc_coord_regression src/vc_coord_regression.cpp)
+target_link_libraries(vc_coord_regression vc_core)
+
 add_executable(vc_create_segment_mask src/vc_create_segment_mask.cpp)
 target_link_libraries(vc_create_segment_mask vc_core Boost::program_options)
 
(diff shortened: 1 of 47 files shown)


giorgioangel:
@codex review

PR #829, opened 2026-04-21
perf(io): VolumeSource byte-range fetches + codec output shrink_to_fit

SuperOptimizer:
## Summary

Two independent cleanups extracted from the compress3d branch for standalone review. Split 1/5 of the compress3d work:

- **`HttpSource` inner-chunk fetches switch to HTTP byte-range GETs** instead of whole-shard downloads. For a viewport that only touches a handful of inner chunks per shard this is ~100× less wire bytes and avoids the whole-shard latency spike. Whole-shard downloads stay available for the prefetch path.
- **`video_encode_into` shrink_to_fit's the output** before returning. x265's output pattern bursts to many MB then tapers, so without the shrink each encoded chunk carries a big over-allocated tail into downstream buffers. On long recompress runs the slack adds up to gigabytes of leaked capacity.

No public API changes.

## Test plan
- [x] Full-tree ThinLTO MinSizeRel build on this branch links clean.
- [ ] CI green
- [ ] VC3D smoke test against a remote volume

### DIFF
--- a/volume-cartographer/core/include/vc/core/cache/VolumeSource.hpp
+++ b/volume-cartographer/core/include/vc/core/cache/VolumeSource.hpp
@@ -113,6 +113,8 @@ private:
     int totalChunksPerShard() const noexcept;
     std::vector<uint8_t> fetchFromShard(const ChunkKey& key);
     std::vector<uint8_t> httpGet(const std::string& url);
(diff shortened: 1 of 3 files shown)


SuperOptimizer:
@codex review

PR #830, opened 2026-04-21
feat(c3d): vendor libc3d + register as zarr v3 codec

SuperOptimizer:
## Summary

Split 2/5 of the compress3d branch — the c3d library + codec wrapper + zarr plumbing. **Fully independent of #829**; can merge in either order. Nothing in `vc_core` actively calls the codec yet — that wiring lives in follow-up PRs:

- PR3: BlockPipeline + Volume integrate c3d as the canonical codec
- PR4: c3d-aware `vc_zarr_{prefetch, recompress, raw_frames}`
- PR5: remove the legacy H.265 path entirely

## Highlights

- **`libs/c3d/{c3d.c, c3d.h, CMakeLists.txt}`**: vendored single-TU libc3d (DCT-III 3D volumetric codec, ~4k lines, Apache-2.0). Inner chunk atom is fixed at 256³ u8; every encoded chunk starts with `C3DC` magic so no wrapping header is needed.
- **`utils/{c3d_codec.hpp, c3d_codec.cpp}` + `utils_c3d_codec` target**: thin C++ wrapper exposing encode / decode / is_c3d_compressed on top of libc3d with a target-ratio knob (default 50 ≈ 40 dB PSNR on scroll CT; 10 → 46 dB, 100 → 35 dB).
- **`utils/zarr.hpp`**:
  - `is_canonical_c3d()` structural check for zarr v3 + 256³ inner chunks.
  - Inner-chunk payloads written with 4k alignment inside a shard so NVMe block writes aren't split across pages, with back-compat detection on read.
  - Sharded-array read path now walks `shard_config->sub_codecs` in addition to the outer codec pipeline, so inner c3d bytes get decompressed (both `read_chunk` and `needs_decompression`).
- **`core/util/Pool3d.hpp`**: header-only 3D average-pool / max-pool kernels used by the c3d LOD downscaler.
- **`core/src/VcDataset.cpp`**: parse `c3d` sub_codec config (inc. `target_ratio`) so zarr v3 arrays that declare c3d are handled.
- **`core/CMakeLists.txt`**: link `vc_core` against `utils_c3d_codec`.
- **`volume-cartographer/CMakeLists.txt`**: `add_subdirectory(libs/c3d)`.

## Test plan
- [x] Full-tree ThinLTO MinSizeRel build on this branch (on current `main`, without #829 merged) links clean — VC3D and all other binaries.
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #831, opened 2026-04-21
feat(cache): wire c3d into BlockPipeline + Volume + VC3D LOD setting

SuperOptimizer:
## Summary

Split 3/5 of the compress3d branch. Builds on #830 (c3d foundation, merged); adds the runtime paths that actually use libc3d. PR4 wires the apps; PR5 removes the legacy H.265 path entirely.

## Highlights

- **`BlockPipeline::Config`** gains a `Codec` enum (`{H265, C3d}`) and a `c3dEncodeParams` field, and `canonicalChunkSide()` dispatches 128³ for H265 / 256³ for C3d. `decodeCanonicalChunk` / `encodeCanonicalChunk` / `bytesAreCanonical` branch on `cfg.codec` so both formats coexist in the same pipeline during the migration window. `canonicalSourceShard` detection handles both 1024³ (h265) and 4096³ (c3d) shard layouts.
- **`Volume::createTieredCache`** defaults remote disk caches to c3d but keeps legacy h265 caches working: when a per-level `zarr.json` already exists we detect `is_canonical_c3d` / `is_canonical_vc3d` and honour the on-disk codec. `VC3D_REMOTE_CODEC=h265|c3d` env override is the manual knob. The canonical-passthrough fast path now recognises both c3d 4096³ sources (default) and h265 1024³ sources (opt-in), skipping decode + re-encode when the source already matches our local format.
- **`apps/VC3D/VCSettings.hpp`** registers a `LOD_METHOD` setting key for the VC3D settings dialog: `downsample` / `codec_synthesis` / `average`. Default `codec_synthesis` calls `c3d_chunk_decode_lod` directly from the compressed shard for a much cheaper coarse-level path.

## Test plan
- [x] Full-tree ThinLTO MinSizeRel build on this branch (on current `main` with #829 + #830 merged) links clean.
- [ ] CI green
- [ ] VC3D smoke test against an existing c3d volume and an existing h265 volume

### DIFF
--- a/volume-cartographer/apps/VC3D/VCSettings.hpp
+++ b/volume-cartographer/apps/VC3D/VCSettings.hpp
@@ -191,6 +191,17 @@ namespace perf {
     constexpr bool ENABLE_FILE_WATCHING_DEFAULT = true;
     constexpr int RAM_CACHE_SIZE_GB_DEFAULT = 10;
 
+    // LOD synthesis method.  Selects how c3d chunks are decoded when a
+    // downscaled view is requested.  Value is one of:
+    //   "codec_synthesis"   — call c3d_chunk_decode_lod; codec-native filter.
+    //   "full_decode_box"   — full decode + box-average pool.
+    //   "full_decode_min"   — full decode + min pool.
+    //   "full_decode_max"   — full decode + max pool.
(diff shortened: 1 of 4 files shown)


SuperOptimizer:
@codex review

PR #832, opened 2026-04-21
feat(apps): c3d-aware zarr tools + raw-frames streamer

SuperOptimizer:
## Summary

Split 4/5 of the compress3d branch. Depends on #830 (c3d foundation) and #831 (BlockPipeline / Volume c3d wiring) — both merged. PR5 (remove H.265 entirely) follows.

## Highlights

- **`vc_zarr_prefetch`** — canonical-passthrough now detects both c3d (256³/4096³) and h265 (128³/1024³) sources. `openOrCreateDiskLevel` creates matching local layout; per-level codec is captured in the `PassthroughLevel` queue so the magic-bytes spot-check validates against the right magic (`C3DC` vs `VC3D`). A mismatch between source codec and an existing on-disk cache falls back to the BlockPipeline transcode path instead of corrupting bytes.
- **`vc_zarr_recompress`** — `--codec {h265, c3d}` switch selects output geometry + encoder. c3d mode uses 4096³ shards with 256³ inner chunks and `--target-ratio` for rate control; h265 mode keeps the legacy 1024³/128³/`--qp` path. Passthrough + coalescing reuse `is_c3d_compressed` so already-c3d input bytes skip decode+re-encode.
- **`vc_zarr_raw_frames` (new)** — streams raw u8 z-slice frames from a sharded zarr to stdout for piping into ffmpeg. Processes one full z-chunk-layer at a time (decode every (iy,ix) chunk in parallel, emit every stride-th z from the decoded buffers) so each chunk is decoded exactly once regardless of stride. Intended for ad-hoc video encoding from a c3d volume.
- **`vc_encode_test`** — grew a `--codec` flag so the synthetic round-trip runs against both backends for side-by-side throughput / PSNR.

## Test plan
- [x] Full-tree ThinLTO MinSizeRel build on this branch (on current `main` with #829/#830/#831 merged) links clean — VC3D and all 4 affected binaries.
- [ ] CI green
- [ ] Smoke test `vc_zarr_prefetch` on the paris4 c3d volume; confirm canonical-passthrough path still engages.

### DIFF
--- a/volume-cartographer/apps/CMakeLists.txt
+++ b/volume-cartographer/apps/CMakeLists.txt
@@ -180,6 +180,9 @@ target_link_libraries(vc_tifxyz vc_core Boost::program_options opencv_core)
 add_executable(vc_zarr_to_tiff src/vc_zarr_to_tiff.cpp)
 target_link_libraries(vc_zarr_to_tiff vc_core Boost::program_options TIFF::TIFF)
 
+add_executable(vc_zarr_raw_frames src/vc_zarr_raw_frames.cpp)
+target_link_libraries(vc_zarr_raw_frames vc_core)
+
 add_executable(vc_zarr_recompress src/vc_zarr_recompress.cpp)
 target_link_libraries(vc_zarr_recompress vc_core blosc_static)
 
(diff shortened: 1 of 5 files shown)


SuperOptimizer:
@codex review

PR #834, opened 2026-04-22
perf: debounce intersection render + c3d decode cleanups

SuperOptimizer:
## Summary

Two independent VC3D performance fixes found via `perf record` on a warm-cache session (paris4_c3d_r50, 256³ c3d chunks). Together they eliminate ~14% of total CPU (the denoise skip is env-gated so that part isn't commit-default yet — see below) and ~83% of the absolute memcpy cycles in the c3d decode wrapper.

### Commit 1 — `perf(vc3d): debounce intersection renders + cache flattened fingerprint`

`CAdaptiveVolumeViewer::renderIntersections()` was paying the full rtree + triangle-clip cost on the main thread on every UI setter fire — opacity slider, thickness slider, surface change, intersect-target slider drag. Each of those spams setters and each setter was doing real work.

- Split `renderIntersections()` into a public debounce shim that sets `_intersectionsDirty` and a synchronous `renderIntersectionsNow()` drained by the existing `_renderTimer` tick. No new timer — reuses the 16 ms tick that already kicks `submitRender()`.
- Added a fingerprint cache for the flattened-view path: hash of the 3 plane poses + active seg identity + generation + opacity / thickness / patch count. `renderFlattenedIntersections()` now short-circuits when nothing material changed (the plane-view path already had this cache).

**Measured impact:** `SurfacePatchIndex::forEachTriangleImpl` + `clipTriangleToPlane` + related intersection render helpers dropped from ~10% of main-thread CPU to ~0.03% (i.e. off the hot list entirely) in the warm-cache capture.

### Commit 2 — `perf(c3d_codec): drop dead memset, decode in-place when aligned`

`utils/src/c3d_codec.cpp` was doing two redundant 16 MiB memory operations per c3d chunk decode, showing up as `__memcpy_oryon1` 8.02% and `__memset_oryon1` 2.13% in the before-capture.

- **Dead memset removed:** upfront `memset(staging, 0, 16 MiB)` before every decode was unnecessary. The c3d decoder writes every output voxel unconditionally — `c3d.c:3659` for the `lod_end==0` fast path, `c3d.c:3778–3790` otherwise. The §T9 truncated-entropy case zero-fills in the coefficient buffer, not the output, and the final output loop still writes every voxel.
- **In-place decode when aligned:** when `out.data()` is already 32-byte aligned (which it is in practice — glibc routes 16 MiB allocations through mmap, giving 4 KiB page alignment), decode directly into the `std::vector<std::byte>` and skip both the staging buffer and the 16 MiB copy-out. Falls back to the staging path if alignment ever doesn't hold.

Applied to both `c3d_decode` and `c3d_decode_lod`. No API change.

**Measured impact** (warm-cache VC3D, ~190k samples): `__memcpy_oryon1` 8.02% → 2.76% (~83% absolute cycle reduction); `__memset_oryon1` absolute cycles −40%.

## Profiling note — denoise

Separately, `c3d_denoise_3d` accounts for ~14% of warm-cache CPU and can be disabled at runtime with `C3D_DENOISE=0` (already plumbed in `c3d.c:3763–3764`) for a near-free win at negligible quality cost (~0.03 dB PSNR at our default r=50 per the table at `c3d.c:3449–3458`). Not part of this PR — worth a follow-up that either defaults it off for the viewer or adds a `skip_denoise` flag to `utils::C3dCodecParams`.

## Test plan

- [ ] Build MinSizeRel: intersections still render correctly (opacity slider, thickness slider, surface change, intersect-target drag)
- [ ] Decode a c3d-compressed volume: visual tiles render identically to before (the dead memset really was dead)
- [ ] `vc_zarr_recompress` + `VcDataset::c3dDecompress` paths still round-trip correctly
- [ ] Perf capture on a warm-cache session: `__memcpy_oryon1` drops from ~8% to ~3%, intersection render helpers below 0.1%

### DIFF
--- a/volume-cartographer/apps/VC3D/adaptive/CAdaptiveVolumeViewer.cpp
+++ b/volume-cartographer/apps/VC3D/adaptive/CAdaptiveVolumeViewer.cpp
@@ -1726,6 +1726,28 @@ void CAdaptiveVolumeViewer::renderFlattenedIntersections(const std::shared_ptr<S
     fp.activeSegHash = std::hash<const void*>{}(activeSeg.get());
     fp.targetGenerationHash = std::hash<uint64_t>{}(
         patchIndex->generation(activeSeg));
+    // Fold the camera state into the fingerprint so pan/zoom invalidates
+    // the cache — surfaceToScene() below consumes all of this.
+    std::size_t cameraHash = 0;
+    auto hashInt = [&](std::size_t s, int v) {
+        return mix(s, std::hash<int>{}(v));
+    };
+    cameraHash = hashInt(cameraHash, int(std::lround(_camSurfX * 1000.0f)));
+    cameraHash = hashInt(cameraHash, int(std::lround(_camSurfY * 1000.0f)));
+    cameraHash = hashInt(cameraHash, int(std::lround(_camScale * 1000.0f)));
+    cameraHash = hashInt(cameraHash, _framebuffer.width());
+    cameraHash = hashInt(cameraHash, _framebuffer.height());
+    if (_view) {
+        const QTransform t = _view->transform();
+        auto q = [](qreal v) { return int(std::lround(v * 1000.0)); };
+        cameraHash = hashInt(cameraHash, q(t.m11()));
+        cameraHash = hashInt(cameraHash, q(t.m12()));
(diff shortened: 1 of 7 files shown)


SuperOptimizer:
@codex review

PR #835, opened 2026-04-22
perf(vc3d): async render pipeline + RAM backpressure + mmap shards

SuperOptimizer:
## Summary

Pile of RAM + UI-thread wins measured on heavy warm-cache sessions (paris4_c3d_r50, ~1.97M SPI patches, 9 resident surfaces).

**Baseline** (main, pre-PR): peak RSS 26.9 GB, peak mi_commit 15.0 GB, 5.7 GB swap, visible input jank during download bursts.

**After this PR**: peak RSS ~16 GB, peak mi_commit ~1.6 GB, 0 swap, main thread ~5% CPU (almost entirely Qt event dispatch + small kernel).

## Biggest single contributors

| Change | Effect |
|---|---|
| mimalloc `purge_decommits=1`/`purge_delay=0` + swap `malloc_trim` → `mi_collect` | mi_commit 15 GB → 1.6 GB (~90% cut). Freed pages go back to the OS immediately instead of marinating in mimalloc's decommit queue. |
| BlockPipeline backpressure (encode/decode staging + in-flight shard reads) | 5.7 GB swap → 0. Loader can no longer hold tens of GiB of in-flight shard bytes while encoder/decoder drain. |
| mmap zarr shard reads (new `utils::ShardBytes` wrapper) | Shard bytes live in the kernel page cache; drop under memory pressure automatically; eliminate per-read 256 MiB heap allocation. Shard cache budget 1 → 4 GiB (cheap now). |
| Async render pipeline (`submitRender` now dispatches → QThreadPool worker → main-thread finish slot) | All the heavy per-frame work (gen + tile sampling + CLAHE + raking + stretch scan) runs on a worker; main thread does Qt + scene commit only. Input events no longer stall behind the sample loop. |
| Flattened + plane-view intersection workers | `SurfacePatchIndex::forEachTriangleImpl` + `clipTriangleToPlane` + `computePlaneIntersections` all off the main thread. |
| QuadSurface `gen()` bespoke alloc-free warp + pooled scratch Mats | Replaces `cv::warpAffine` (which heap-allocates internal coord maps ~27 MiB × N calls). Hand-rolled scale+translate warp is auto-vectorised on NEON and reuses `mutable` Mat members via OpenCV's `Mat::create` no-op-on-match. |

## Observability

New thread names (`bpDl*/bpEnc*/bpLd*/bpDec*` for BlockPipeline I/O pools, `vcRender*` for render tile workers) let `perf` attribute CPU accurately — the remaining `VC3D` comm is now the true main thread alone. `RamStats` dumper wires up `mi_process_info` + all new staging/inflight fields for live monitoring; `mi_commit` underflow values at mimalloc teardown are clamped for display.

## Shutdown safety

`~CAdaptiveVolumeViewer` spins on `_backgroundWorkers` reaching zero before destroying members. Fixes a SIGSEGV on app exit when an async render or intersection worker was mid-execution.

## Minor

- Fixed z-offset status bar always showing 0 on plane viewers — shift+scroll there moves the focus POI, not `_camera.zOff`; now shows the plane's signed offset along its normal.
- `SurfaceLRU` max resident 8 → 4 (~680 MiB peak saved) + async warm on touch so evicted surfaces re-load on a worker.
- `NormalGridVolume` cache cap 4096 → 512 (prior theoretical ceiling was ~8 GiB).
- Main thread no longer participates as a render tile worker (pool grows by one to compensate).
- Short-circuited `SurfacePatchIndex` bbox cull (was ~1% of total CPU in `std::max/min<initializer_list>`).

## Test plan

- [ ] Build MinSizeRel: rendering works across the 4 viewers (xy/xz/yz planes + flattened segmentation)
- [ ] Pan / zoom / shift-scroll: no visible jank during heavy S3 downloads
- [ ] Maximise a viewer: framebuffer resizes correctly (no stuck-at-pre-resize size)
- [ ] Status bar z=X updates on plane viewers as you shift-scroll
- [ ] Clean exit — no SIGSEGV at shutdown after a warm-cache session with the pipeline busy
- [ ] `[RAM]` stats line ticks every second with sensible `mi_commit`, `enc_stage`, `dec_stage`, `shard_inflight`, `surfs` values

### DIFF
--- a/volume-cartographer/apps/VC3D/CState.hpp
+++ b/volume-cartographer/apps/VC3D/CState.hpp
@@ -109,5 +109,6 @@ private:
     std::unordered_map<std::string, std::shared_ptr<Surface>> _surfs;
     std::unordered_map<std::string, std::unique_ptr<POI>> _pois;
 
-    SurfaceLRU _surfaceLRU{8};
+    // See SurfaceLRU default — 4 resident × ~170 MiB/surface cap.
+    SurfaceLRU _surfaceLRU{4};
 };
--- a/volume-cartographer/apps/VC3D/CWindow.cpp
+++ b/volume-cartographer/apps/VC3D/CWindow.cpp
@@ -5,12 +5,10 @@
 #include <cstdlib>
 #include <functional>
 #if defined(__GLIBC__)
-#if defined(__GLIBC__)
 #include <malloc.h>
 #endif
 #if defined(VC_HAVE_MIMALLOC)
 #include <mimalloc.h>
-#endif
 #endif
 
 #include "vc/core/cache/HttpMetadataFetcher.hpp"
--- a/volume-cartographer/apps/VC3D/CWindow.cpp
+++ b/volume-cartographer/apps/VC3D/CWindow.cpp
@@ -5,7 +5,12 @@
 #include <cstdlib>
 #include <functional>
 #if defined(__GLIBC__)
+#if defined(__GLIBC__)
 #include <malloc.h>
+#endif
+#if defined(VC_HAVE_MIMALLOC)
+#include <mimalloc.h>
+#endif
 #endif
 
 #include "vc/core/cache/HttpMetadataFetcher.hpp"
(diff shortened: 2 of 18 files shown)


SuperOptimizer:
@codex review

PR #845, opened 2026-04-29
Fix OME-Zarr physical scale metadata

maxliebscher:
## Summary

Fixes OME-Zarr physical scale metadata written by `vc_render_tifxyz` so rendered output stores the actual physical spacing for the generated Z/Y/X axes.

## What changed

- `writeZarrAttrs` now accepts separate physical sizes for Z and Y/X instead of one isotropic base voxel size.
- `vc_render_tifxyz` computes output Y/X spacing from the selected source group and `--scale`.
- Rendered Z spacing uses `--slice-step` and remains unchanged across output pyramid levels.
- Pyramid metadata doubles only Y/X scale per downsampled output level.
- TIFF DPI uses the rendered Y/X pixel size.
- Added a focused doctest for anisotropic OME-Zarr scale metadata.

## Why

The previous metadata path used one base voxel size for all output axes and levels. That loses the distinction between:

- the source OME-Zarr group resolution,
- the render pixel scale,
- the rendered slice spacing along Z,
- and the Y/X-only downsampling applied to output pyramid levels.

This can make downstream OME-Zarr consumers interpret rendered volumes with the wrong physical scale.

## Validation

Rebased onto current `main` and ran in WSL2 Ubuntu 24.04:

```text
git diff --check origin/main...HEAD
cmake -S volume-cartographer -B volume-cartographer/build/wsl-vc-pr845 -GNinja -DCMAKE_BUILD_TYPE=Release -DVC_TESTING=ON -DLAPACKE_INCLUDE_DIRS=/usr/include -DLAPACKE_LIBRARIES=/usr/lib/x86_64-linux-gnu/liblapacke.so
cmake --build volume-cartographer/build/wsl-vc-pr845 --target vc_render_tifxyz test_zarr_attrs -j2
ctest --test-dir volume-cartographer/build/wsl-vc-pr845 -R '^test_zarr_attrs$' --output-on-failure
```

Result:

```text
100% tests passed, 0 tests failed out of 1
```

The generated build directory was removed after verification.

### DIFF
--- a/volume-cartographer/apps/src/vc_render_tifxyz.cpp
+++ b/volume-cartographer/apps/src/vc_render_tifxyz.cpp
@@ -1163,6 +1163,10 @@ int main(int argc, char *argv[])
     std::filesystem::path seg_path = parsed["segmentation"].as<std::string>();
 
     float tgt_scale = parsed["scale"].as<float>();
+    if (!std::isfinite(tgt_scale) || tgt_scale <= 0) {
+        logPrintf(stderr, "Error: --scale must be positive\n");
+        return EXIT_FAILURE;
+    }
     int group_idx = parsed["group-idx"].as<int>();
     int num_slices = parsed["num-slices"].as<int>();
     double slice_step = parsed["slice-step"].as<float>();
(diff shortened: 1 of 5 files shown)


maxliebscher:
Quick note on the Vercel check: this PR only touches volume-cartographer C++/CMake test code. I validated the affected path locally in WSL2 by building `vc_render_tifxyz` and `test_zarr_attrs`, then running `ctest -R test_zarr_attrs --output-on-failure`. The Vercel preview is blocked because this comes from a fork and needs team authorization; I do not think it is relevant for this patch unless you want preview deployments enabled for it.

giorgioangel:
@codex review

PR #846, opened 2026-04-29
Mirror overlap links in tracer growth

maxliebscher:
## Summary

Makes tracer overlap handling more tolerant of one-sided overlap metadata.

## What changed

- When `grow_surf_from_surfs` builds its in-memory overlap map, each valid overlap is now recorded in both directions.
- The source and target surfaces are both resolved through the non-defective surface map before adding the edge.

## Why

`overlapping.json` metadata can be one-sided: surface A may list B while B does not list A. Tracer growth treats overlaps as traversal/candidate relationships, so losing the reverse edge can make approved or relevant neighboring patches unavailable when growth starts from the other side.

This keeps the on-disk metadata untouched and only hardens the runtime graph.

## Validation

Ran in WSL2 Ubuntu 24.04:

```text
cmake -S volume-cartographer -B build/wsl-issue373 -GNinja -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DVC_BUILD_TESTS=OFF -DVC_WITH_CUDA_SPARSE=off
cmake --build build/wsl-issue373 --target vc_tracer -j8
cmake --build build/wsl-issue373 --target vc_grow_seg_from_segments -j8
./build/wsl-issue373/bin/vc_grow_seg_from_segments --help
```

Also ran:

```text
git diff --check
```

### DIFF
--- a/volume-cartographer/core/src/GrowSurface.cpp
+++ b/volume-cartographer/core/src/GrowSurface.cpp
@@ -1669,10 +1669,22 @@ static QuadSurface *grow_surf_from_surfs_impl(QuadSurface *seed,
     for(auto sm : approved_sm)
         std::cout << "approved: " << surface_name(sm) << std::endl;
 
-    for(auto &sm : surfs_v)
-        for(const auto& name : sm->overlappingIds())
-            if (surfs.contains(name))
-                overlaps[sm].insert(surfs[name]);
+    for(auto &sm : surfs_v) {
+        auto source_it = surfs.find(surface_name(sm));
+        if (source_it == surfs.end())
+            continue;
+
+        for(const auto& name : sm->overlappingIds()) {
+            auto target_it = surfs.find(name);
+            if (target_it == surfs.end())
+                continue;
+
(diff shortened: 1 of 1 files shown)


maxliebscher:
Quick note on the Vercel check: this PR only touches volume-cartographer tracer C++ code. I validated the affected target locally in WSL2 by building `vc_tracer`, linking `vc_grow_seg_from_segments`, and smoke-testing `vc_grow_seg_from_segments --help`. The Vercel preview is blocked because this comes from a fork and needs team authorization; I do not think it is relevant for this patch unless you want preview deployments enabled for it.

maxliebscher:
Closing this as superseded by the current `main` refactor. This PR targeted the old `GrowSurface.cpp` in-memory `overlapping_map`; that map no longer exists on `main`, and tracer growth now uses `SurfacePatchIndex` for candidate discovery instead of relying on one-sided overlap metadata there. The current overlap writers also update both sides when recording discovered overlaps, so keeping this patch open would mostly create conflict noise.

PR #849, opened 2026-04-29
Avoid mutating grow seed JSON params

maxliebscher:
## Summary

Stop `CommandLineToolRunner::setGrowParams` from rewriting the package `seed.json` / `expand.json` file when selecting the runtime grow mode.

## Why

The runner only needs to override `mode` for the command invocation. Previously it opened the source params file and wrote the modified object back to that same file, so running seed/expand actions from VC3D could mutate package configuration as a side effect.

This keeps the package JSON stable and passes a temporary runtime JSON file to `vc_grow_seg_from_seed` instead.

## What changed

- Parse and validate grow params as a JSON object before applying the runtime mode.
- Write the mode-adjusted params to a temporary file.
- Pass the temporary params path to the command.
- Clean up the temporary file when the process finishes, errors, the runner is destroyed, or another JSON-param command replaces it.

## Validation

- `git diff --check`
- `cmake --build build/wsl-issue373 --target VC3D -j2`

### DIFF
--- a/volume-cartographer/apps/VC3D/CommandLineToolRunner.cpp
+++ b/volume-cartographer/apps/VC3D/CommandLineToolRunner.cpp
@@ -9,9 +9,11 @@
 #include <QTextStream>
 #include <QJsonDocument>
 #include <QJsonObject>
+#include <QJsonParseError>
 #include <QMessageBox>
 #include <QClipboard>
 #include <QApplication>
+#include <QTemporaryFile>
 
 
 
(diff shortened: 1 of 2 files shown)


maxliebscher:
Note on the Vercel bot: this PR only changes VC3D command-runner behavior for seed/expand params. I am not authorizing a fork deployment; local validation was a VC3D CMake build.

maxliebscher:
Closing this as superseded by the current `main` UI refactor. The mutating `CommandLineToolRunner::setGrowParams` path this PR fixed no longer exists: seed/expand actions now invoke `vc_grow_seg_from_seed` directly with the package JSON path, and the newer neighbor-copy flow writes runtime params to `QTemporaryFile` instances before invoking the tool. Keeping this open would reintroduce obsolete runner code rather than fixing a current path.

PR #852, opened 2026-04-29
JSON based volpkgs

SuperOptimizer:
## Summary
- Replaces the legacy directory-based volpkg layer with a JSON-only project loader (`*.volpkg.json`). Volumes / segments / normal_grids are referenced by URI (local paths or `s3://` / `https://` URLs); the core no longer auto-creates `volumes/`, `paths/`, etc.
- Remote segment metadata download moves to a detached background thread (16 workers) with a `setSegmentsChangedCallback` that fires every ~16 segments, so the surface panel populates progressively and the GUI never freezes on remote opens.
- Legacy folder/URL volpkgs convert via the new `vc_volpkg_convert` CLI (also called transparently when File → Open Project picks a folder or remote URL); `UnifiedBrowserDialog` replaces `S3BrowserDialog` and now respects typed URLs without a navigate-first; new File → Detach action plus pre-attach `validateLocation()` checks.

## Test plan
- [ ] Open a local `.volpkg.json` directly via File → Open Project — volumes, segments, and normal grids show up.
- [ ] Open a legacy local volpkg folder via File → Open Project — prompted to save the converted `.volpkg.json`, then opens it.
- [ ] Paste an `s3://bucket/legacy.volpkg/` URL into File → Open Project — converts and opens; volume folder of zarrs is auto-scanned; segments list populates progressively in the surface panel.
- [ ] File → Attach Volume / Segments / Normal Grid: invalid path is rejected with a structural reason; valid attaches succeed.
- [ ] File → Detach: lists every attached entry; selected one is removed and the autosave updates.
- [ ] Swap volume in the volume picker on a remote project — old pipeline tears down cleanly, new one loads, segments stay populated.
- [ ] Shut down VC3D mid-download — no crash, in-flight worker threads exit cleanly.


SuperOptimizer:
@codex review

giorgioangel:
@codex review

giorgioangel (approved):

PR #855, opened 2026-04-30
Crash handler with rich diagnostics + move volpkg convert into vc_core

SuperOptimizer:
## Summary

- **Crash handler**: catch SIGSEGV/SIGBUS/SIGFPE/SIGILL/SIGABRT/SIGTRAP/SIGSYS plus unhandled C++ exceptions, write a self-contained diagnostic report to stderr so users can paste it into bug reports / hand it to AI for triage. Handler runs on a 16 MB sigaltstack with async-signal-safe I/O. Linked into VC3D via an explicit `vc::crash::install()` at the top of `main`.
- **Volpkg conversion** moved out of the standalone CLI and into `vc_core` (`vc::convertVolpkg` in `util/VolpkgConvert.hpp`). The CLI is now a thin wrapper. The GUI calls the library directly instead of shelling out to `vc_volpkg_convert` via `QProcess`, so a conversion failure now shows the real error in a properly-sized dialog instead of an unresizable `QMessageBox` whose title was truncated to `Con...`.
- **Build flags**: every config now compiles with `-g3 -fno-omit-frame-pointer -fasynchronous-unwind-tables` (+ `-fstandalone-debug` on clang) and binary stripping is off by default, so libbacktrace can resolve `function/file:line` for every frame in user-submitted crash dumps regardless of build type.

## Crash report contents

For each crash:
- version, commit hash, compiler, build date, arch, PID, TID + thread name
- signal name, decoded `si_code`, faulting `si_addr`, `si_pid`/`si_uid`/`si_errno`
- full register set (x86_64 and aarch64)
- libgcc `backtrace()` raw addresses, then per-frame `function` + `file:line` via libbacktrace's `backtrace_pcinfo` (with inlined-frame expansion); `dladdr()` fallback for frames with no DWARF (system libs print as `<symbol> + <off> in <path>`)
- "PROBABLE CRASH SITE" marker above the first non-handler frame
- manual frame-pointer walk fallback for crashes where PC is corrupt; ARMv8.3 PAC bits stripped on aarch64
- 64 bytes of machine code at PC; 256 bytes of stack from SP; 128 bytes around `si_addr` for SEGV/BUS (probed via `pread()` on `/proc/self/mem` so the handler never re-faults)
- `uname`, `/proc/version`, `/etc/os-release`, first `/proc/cpuinfo` block (ARM-friendly: doesn't require a "model name" line)
- per-thread summary (`tid`, `comm`, `state`, `wchan`) for every thread in `/proc/self/task/`
- open file-descriptor listing via `readlink` of every entry in `/proc/self/fd`
- environment variables with values redacted when the name contains `KEY`/`TOKEN`/`SECRET`/`PASSWORD`/`AUTH`/`CRED`/`PRIVATE`/`API_`/`SESSION`/`COOKIE`
- `getrusage`, `/proc/self/status`, `/proc/self/limits`, `/proc/meminfo`, `/proc/self/maps`
- recursive-crash guard `_exit`s if the handler itself faults; on completion the original signal is re-raised with the default handler so a coredump still drops if enabled

Linux-only. Adds `libbacktrace` as a system dependency (it ships with gcc; available on every distro that ships a working C++ toolchain).

## Test plan

- [x] Build VC3D in `Release` config with the new flags (no stripping, full DWARF)
- [x] Trigger SIGSEGV via a deliberate null-deref menu item (since removed) and verify:
  - signal info, registers, `si_addr` decoded
  - resolved frames include the user-code lambda with `MenuActionController.cpp:169` location and the inlined Qt slot-dispatch chain
  - "PROBABLE CRASH SITE" marker lands on the first user frame, skipping crash-handler internals and the `??` Qt trampoline
  - bytes at PC show the exact aarch64 store instruction (`str w9, [x8]`) confirming the fault is the deliberate null write
  - 127 threads enumerated with comm/state/wchan
  - 39 open FDs listed via readlink
  - environment dumped with `SESSION`/`AUTH`/`CRED` values redacted
- [ ] Verify on x86_64 (only tested on aarch64 in development)
- [ ] Verify GUI volpkg conversion failure dialog now shows an actionable message body (the original "Con..." truncated dialog)


SuperOptimizer:
@codex review

PR #863, opened 2026-05-04
Fix tracer growth / fix masks / extend Volume / switch (most) zarr readers to volume 

bruniss:


SuperOptimizer (approved):

giorgioangel:
@codex review

giorgioangel:
@bruniss Claude says:

The regression                                                                                                                                                                        
                                                                                                                                                                                        
  The PR replaces the slab-read inner loop with a much weaker per-thread cache. Concretely:                                                                                             
                                                                                                                                                                                        
  - Old path (Slicing.cpp:77-183, ChunkSampler<T>): every OMP worker holds a lock-free 16384-slot per-thread chunk table (HotSlot[kSlots]). The hot path is 3 int compares              
  (lastCz/lastCy/lastCx); on a chunk hop, it hashes into the per-thread slot table; only on a first-time hit per thread does it call cache.getChunkBlocking(), which takes the global
  state->mutex_.                                                                                                                                                                        
  - New path (Volume.cpp:220-285, the lambda inside readFromChunkedArrayZYX): each OMP worker holds a single cached struct — a per-thread cache of size 1. Every chunk hop calls
  array.getChunkBlocking(level, cz, cy, cx), which in ChunkCache::getChunkBlocking (ChunkCache.cpp:189) acquires state->mutex_ every single call — even if the chunk is already decoded.
   
  So whenever the inner-x loop crosses an X-chunk boundary, the OLD code did one lookup in a per-thread hashmap (lock-free), while the NEW code grabs a process-wide mutex.             
                  
  Why this hits vc_gen_normalgrids particularly hard                                                                                                                                    
                  
  For SliceDirection::XY, slab_shape = {chunkZ, fullY, fullX} and the inner loop walks x. With chunk shape 128³ and a typical 4096³ volume:                                             
                  
  - Crossings per row: ~32 (one per X-chunk boundary).                                                                                                                                  
  - Rows per slab: 128 × 4096 = 524k.
  - NEW: ~16M global-mutex acquisitions per slab (32 × 524k).                                                                                                                           
  - OLD: ~1024 (the number of unique chunks in the slab), and only the first time each one is touched per thread; the slot table caches the rest with no lock.                          
                                                                                                                                                                                        
  With 16+ OMP workers all contending on state->mutex_, that's the textbook cause of a 10× (or worse) wall-clock regression on a CPU-bound, decode-already-cached read. It will look    
  like CPUs are pegged but throughput is awful.                                                                                                                                         
                                                                                                                                                                                        
  Two corroborating issues, smaller magnitude                                                                                                                                           
   
  1. Cache budget set after lazy creation — vc_gen_normalgrids.cpp:484-487 calls input_volume.chunkedCache() for metadata before computing cache_budget_bytes and calling               
  setCacheBudget(...). setCacheBudget happens to do chunkedCache_.reset() (Volume.cpp:1380), so the cache is destroyed and re-created lazily on first read — correct, but it tosses any
  decoded chunks acquired for metadata, and the cycle costs an OpenedChunkedZarr re-open. Not a 10× factor, but unnecessary churn.                                                      
  2. Volume::chunkedCache() mutex per-call — once per readZYX, not per voxel, so negligible compared to (1).
                                                                                                                                                                                        
  How to confirm and fix                                                                                                                                                                
                                                                                                                                                                                        
  - Confirm: profile with perf top or perf record -g during a vc_gen_normalgrids run on the new commit; you should see massive time in pthread_mutex_lock / __lll_lock_wait under       
  ChunkCache::getChunkBlocking. Or just count: instrument getChunkBlocking with an atomic counter and compare pre/post commit.
  - Fix options (any one will recover most of the loss):                                                                                                                                
    a. Have readFromChunkedArrayZYX use ChunkSampler<T> (or an equivalent per-thread slot table) instead of the size-1 cached struct. Cleanest — reuses the existing optimized helper.  
    b. Add a lock-free per-thread or per-call slot table directly in the lambda (small N, e.g. enough to hold the X-chunk-row).                                                         
    c. Cheaper short-term: change the inner pattern so loadChunk only re-fetches when cx actually changes (it already does, but the re-fetch itself is the expensive bit — this won't   
  help unless you also cache more than one chunk).                                                                                                                                      
    d. Make ChunkCache::getChunkBlocking take a shared/read lock for already-resolved entries (drop to Resolved fast path with std::shared_mutex), so contention drops on the           
  steady-state hit case.                                                                                                                                                                
                  
  Option 1 is the smallest diff and matches what the OLD code had. The structural issue is that the PR moved the read into Volume but ported only the naive iteration, not the          
  per-thread chunk cache that made the old slab read fast.

volume-cartographer/apps/src/vc_gen_normalgrids.cpp
```diff
@@ -517,16 +522,14 @@ void run_generate(const po::variables_map& vm) {
...
             chunk_budget_mib,
-            ds->dtypeSize());
+            dtype_size);
         max_estimated_batch_bytes = std::max(max_estimated_batch_bytes, batch_plan.estimatedBatchBytes);
     }
     const size_t cache_budget_bytes = std::min<size_t>(
         256ull * 1024ull * 1024ull,
         std::max<size_t>(64ull * 1024ull * 1024ull, max_estimated_batch_bytes / 2));
 
-    auto cache = vc::render::createChunkCache(
```
SuperOptimizer, line 527: here

volume-cartographer/apps/src/vc_gen_normalgrids.cpp
```diff
@@ -684,7 +691,7 @@ void run_generate(const po::variables_map& vm) {
 
                     const auto read_start = std::chrono::steady_clock::now();
                     Array3D<uint8_t> chunk_data(slab_shape);
-                    readArea3D(chunk_data, slab_offset, cache.get(), 0);
```
SuperOptimizer, line 687: here

SuperOptimizer:
I shouldn't have approved this without looking harder and investigating the performance implications of whatever new stuff landed and testing them myself. That is my mistake.

This removed the multitier caching layer and the optimized slicing routines I had added and replaced them with the new thing.

PR #864, opened 2026-05-05
Fix approval mask drift and SegmentationEditManager segfault

SuperOptimizer:
## Summary

Two bugs traced to a single offending operation: `QuadSurface::trimToValidBbox()` running inside the save path.

- **Approval mask drift**: trim shrank `_points` but left `_channels` at the pre-trim size, so the on-disk mask was saved larger than points. The next growth's mask-preserve copy applied a (2,2) offset that assumed they matched, dropping 2 rows/cols of mask data and shifting the rest down-right each save+grow cycle. After enough cycles, masks visibly stopped displaying.
- **Segfault in `SegmentationEditManager::worldToGridIndex`** after a debounced approval save followed by alpha-push: trim did `_points = std::move(trimmed)`, destroying the `cv::Mat_` object that `_previewPoints` (raw pointer cached at session start) was pointing into. Push/pull keypress dereferenced the dangling pointer.

## Fix

**QuadSurface (`core/src/QuadSurface.cpp`):**
- `trimToValidBbox`: in-place crop (`*_points = (*_points)(rect).clone()`) so the `cv::Mat_` object address is stable, and crops matching `_channels` in lockstep so the size invariant survives save/load.
- `writeDataToDirectory`: dropped the `trimToValidBbox()` call. Saving no longer mutates in-memory state; trim happens on the next `ensureLoaded()` only.
- `ensureLoaded`: eager-loads channels before trim so trim sees them, plus a post-trim center-crop pass that re-aligns any oversized legacy channels saved before this fix.
- `channel()` lazy-load: same center-crop fallback for channels re-fetched after `unloadCaches()`.

**SegmentationEditManager (`apps/VC3D/segmentation/tools/SegmentationEditManager.{cpp,hpp}`):**
- Dropped the `_previewPoints` raw-pointer cache. Replaced with a `previewPointsPtr()` accessor that always returns `_baseSurface->rawPointsPtr()`. Every site that used the cached pointer now samples it fresh at function entry, eliminating the dangling-pointer class of bug by construction.

## Test plan

- [ ] Edit a segment, paint approvals, let the debounced save fire, then trigger alpha-push (the original crash repro). Should no longer SEGV.
- [ ] Run a full edit → grow → edit → grow cycle multiple times. Approval mask should stay aligned with the surface mesh; no (2,2) drift, no edge data loss.
- [ ] Reload a project that contains pre-fix legacy approval masks (saved at the un-trimmed size). Mask should display aligned via the center-crop heal on load.
- [ ] Verify no regressions in tracer growth, push/pull, drag-brush, manual add, surface-rotate, and resample.

### DIFF
--- a/volume-cartographer/CMakePresets.json
+++ b/volume-cartographer/CMakePresets.json
@@ -1,0 +1,30 @@
+{
+  "version": 6,
+  "configurePresets": [
+    {
+      "name": "macos-homebrew-llvm",
+      "displayName": "macOS Homebrew LLVM",
+      "description": "Native macOS build using Homebrew LLVM/Clang and Homebrew dependencies.",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build-macos",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "CMAKE_C_COMPILER": "$env{HOMEBREW_PREFIX}/opt/llvm/bin/clang",
+        "CMAKE_CXX_COMPILER": "$env{HOMEBREW_PREFIX}/opt/llvm/bin/clang++",
+        "CMAKE_OSX_DEPLOYMENT_TARGET": "13.3",
+        "VC_WITH_CUDA_SPARSE": "OFF",
+        "VC_WITH_PASTIX": "OFF",
+        "VC_USE_OPENMP": "OFF",
+        "VC_BUILD_TESTS": "OFF"
+      }
+    }
+  ],
+  "buildPresets": [
+    {
+      "name": "macos-homebrew-llvm",
+      "displayName": "Build macOS Homebrew LLVM",
+      "configurePreset": "macos-homebrew-llvm",
+      "jobs": 8
+    }
+  ]
+}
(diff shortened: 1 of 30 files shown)


giorgioangel:
@codex review

giorgioangel:
@SuperOptimizer @bruniss @skyward7187 Claude generated PRs have a "Test plan". Are you always doing these tests, or you just merge usually?

SuperOptimizer:
> @SuperOptimizer @bruniss @skyward7187 Claude generated PRs have a "Test plan". Are you always doing these tests, or you just merge usually?

It depends on what I'm doing whether or not the claude generated test plan is relevant. In this case I sent the branch to Kendra who reported it to test and she reported back that it's fine. I always manually test everything. Automated tests have been more varied; I've done some, like on the tiled renderer, but then since we don't have gating for tests they became stale quickly. Until we have real CI gating on code and test coverage spending tons of time on automated testing isn't worth it imo.

bruniss:
> @SuperOptimizer @bruniss @skyward7187 Claude generated PRs have a "Test plan". Are you always doing these tests, or you just merge usually?

i do not listen to any tests claude proposes because it just does whatever it needs to do to make them pass. i have never had one fail after claude had his way with it.

PR #866, opened 2026-05-05
ui: load attached remote volumes asynchronously

giorgioangel:
loadAttachedRemoteVolumesForCurrentPackage() ran every persisted Volume::NewFromUrl() call synchronously on the GUI thread. With N remote volumes registered in remote_volumes.json, opening a project froze the UI for the sum of N HTTP round-trips.

Lift the existing async pattern from attachRemoteZarrUrl() (which wraps a single NewFromUrl in QtConcurrent::run + QFutureWatcher): snapshot per-entry url+auth on the UI thread (auth resolution may prompt), hand the slow loop off to a worker, post the resolved volumes back to the UI thread to call vpkg->addVolume, refreshCurrentVolumePackageUi, and UpdateView. Status bar shows "Loading N attached remote volume(s)..." while the worker runs.

Use QPointer<MenuActionController> to guard the UI-thread reentry against window destruction during load.

### DIFF
--- a/volume-cartographer/apps/VC3D/MenuActionController.cpp
+++ b/volume-cartographer/apps/VC3D/MenuActionController.cpp
@@ -773,13 +773,24 @@ void MenuActionController::loadAttachedRemoteVolumesForCurrentPackage()
         std::string error;
     };
 
+    // Bind the async load to the package that's active right now. If the
+    // user opens or switches to a different volpkg while the loads are in
+    // flight, the completion callback must NOT mutate the new package
+    // with results meant for the old one. Pre-async this was implicit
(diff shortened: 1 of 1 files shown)


giorgioangel:
@codex review

bruniss:
deferring volume loading on async workers gives us an unstable state on app load. we have many things that run which _require_ a volume to be loaded, and not knowing if one is loaded unless this gives a callback saying its loaded or checking if one is loaded before running them is complex.

if the issue is the ui not loading for a second or two on app start, i dont believe this is an issue worth fixing. a small delay on a remote volume load is not an issue.

PR #867, opened 2026-05-05
ui: atomic remote_volumes.json write

giorgioangel:
Both writers (persistRemoteVolumeRegistryCacheRoot and persistAttachedRemoteVolume) opened the registry with ofstream::trunc, no error checks, no atomicity. A crash mid-write left an empty or torn JSON file, breaking the persisted remote volume list and the cache_root entry on next project open.

Factor a small atomicWriteJsonFile() helper in the anonymous namespace: write to <path>.tmp inside the same directory, flush, check the stream is good, std::filesystem::rename to the final path. The .tmp lives next to the target so the rename never crosses a filesystem boundary (EXDEV). All errors log via the existing Logger and the prior on-disk file is preserved on any failure.

Both registry-writer call sites now go through the helper.


bruniss:
remote_volumes.json is deprecated in favor of the volpkg json and should be removed not updated

PR #868, opened 2026-05-05
Pr/saveoverwrite backup atomic mask

giorgioangel:

PR #870, opened 2026-05-05
ui: run applyRotation save on a worker thread

giorgioangel:
## Summary                                                                                                                                                                            
  - applyRotation() ran rotate() + saveOverwrite() inline on the GUI
    thread, freezing the UI for several seconds on a large surface.                                                                                                                     
    Move both into a QtConcurrent::run worker (mirrors the                                                                                                                              
    performAutosave pattern in SegmentationModule) and post the UI                                                                                                                      
    work — refreshSurfacePatchIndex, setSurface, widget teardown,                                                                                                                       
    state reset — back via QFutureWatcher::finished.                                                                                                                                    
  - saveOverwrite() now (post-#868) snapshots the on-disk state before                                                                                                                  
    overwriting, so the backup ring captures the pre-rotation                                                                                                                           
    x/y/z.tif automatically. No explicit pre-rotate saveSnapshot is                                                                                                                     
    needed in the worker — rotate() only mutates _points in memory                                                                                                                      
    and the on-disk files are still pre-rotate when saveOverwrite runs.                                                                                                                 
  - Trivial rotations (|angle| < 0.01°) still resolve synchronously.                                                                                                                    
  - Failure handling stays equivalent: a thrown rotate or saveOverwrite                                                                                                                 
    surfaces from waitForFinished() and the same "Rotation Failed"                                                                                                                      
    warning fires from the finished slot.                                                                                                                                               
  - QPointer guards UI-thread reentry against controller destruction                                                                                                                    
    during the save.                                                                                                                                                                    
                                                                                                                                                                                        
  ## Test plan                                                                                                                                                                          
  - [ ] Apply a non-trivial rotation on a large surface — UI stays                                                                                                                      
    responsive while the save runs; segment view updates on completion.                                                                                                                 
  - [ ] Apply rotation with a read-only target path — the                                                                                                                               
    "Rotation Failed" warning fires.

### DIFF
--- a/volume-cartographer/apps/VC3D/overlays/SurfaceRotationOverlayController.cpp
+++ b/volume-cartographer/apps/VC3D/overlays/SurfaceRotationOverlayController.cpp
@@ -288,9 +288,14 @@ void SurfaceRotationOverlayController::cancelRotate()
     _previewSurface.reset();
     // Invalidate any in-flight save worker — the session it was
     // started for is over, so its completion callback will see a
-    // mismatched session id and drop its state changes.
+    // mismatched session id and drop its state changes. We do NOT
+    // clear _saveInFlight here: the worker is still running and a
+    // new Apply on a freshly begun session must wait for it to
+    // finish, otherwise two workers could race rotate()/saveOverwrite()
+    // on the same QuadSurface (the new session may target the very
+    // same segment). The finished slot clears _saveInFlight whether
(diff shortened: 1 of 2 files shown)


giorgioangel:
@codex review

giorgioangel:
@codex review

giorgioangel:
@codex review

PR #876, opened 2026-05-06
cmake simplification and ci improvmeents

SuperOptimizer:
### Build system
- Inlined `cmake/VC*.cmake` module logic into top-level `CMakeLists.txt`; deleted legacy modules.
- `find_package(Blosc2/zstd/lz4)` instead of `pkg_check_modules`.
- Portable defaults (`-march`/`-mcpu` floor); opt-in host tuning via `VC_MARCH_NATIVE`.
- `-ffunction-sections -fdata-sections` + `-Wl,--gc-sections` (skipped on Apple/ld64).
- LTO gated on no sanitizers + non-Debug builds.
- OpenMP enforced for gcc, stub target for clang; dropped `VC_USE_OPENMP`.
- Renamed `VC_BUILD_TESTS` → `VC_TESTING`; dropped `VC_USE_PCH`, `VC_USE_VALGRIND`, `VC_WITH_PASTIX`.
- `CMakePresets.json` with dev/CI presets, per-(preset, image) build dirs.
- `ProjectInfo::VersionString` now uses the git `<sha>-<date>` we already compute.

### flatboi wired into the in-tree build
- `libs/flatboi` is now a normal subdirectory like every other lib — `ninja flatboi` from the build dir builds it; the regular full build produces the binary in `bin/` and the Dockerfile picks it up.
- Pulls libigl + PaStiX 6.4 via `FetchContent`. PaStiX trimmed:
  - Single precision (`d` only, not `p;s;d;c;z` — 5× compile cut).
  - No tests, examples, wrappers, Fortran, Parsec, StarPU, CUDA, MPI.
  - Built as static lib inside the FetchContent scope so the parent's static archives don't need PIC and we don't ship `libpastix.so`.
- Dropped the stale `OMP_NUM_THREADS = hardware_concurrency()` block (giorgio's review): PaStiX doesn't read OMP_NUM_THREADS, hwloc auto-sizes to physical cores. `--tol` default reverted to 0 (off) so existing iteration counts stay reproducible.

### CI / sanitizer matrix
- Single matrix workflow (`.github/workflows/vc3d-ci.yml`): ubuntu-{24.04, 26.04} × {gcc, clang}, `ci-release` compile + `ci-tests` test, plus clang-only ASan+UBSan, TSan, TySan, NSan. Coverage and dead-code report run on ubuntu-26.04.
- `scripts/ci.sh` is the single driver (same script in GHA / dev / EC2).
- Pre-built builder images on GHCR (`vc-builder:ubuntu-24.04`, `vc-builder:ubuntu-26.04`); PR jobs `docker pull` instead of building.
- **PR gating is intentionally conservative for now**: only the four compile-only cells gate merges. Tests, sanitizers, coverage, and dead-code are `continue-on-error: true` while the new infra stabilizes — failures show up red on the PR but don't block. Flip per-cell `blocking: true` to gate in a follow-up.
- Per-job `timeout-minutes` caps everywhere so a slow runner doesn't burn 6h before timing out.

### Sanitizer tooling
- `core/test/sanitizer_canary.cpp`: one exe with subcommands that intentionally trip ASan, UBSan, TSan, TySan, NSan. ctest matches the sanitizer report via `PASS_REGULAR_EXPRESSION`, so the canary fails if instrumentation isn't actually active. Excluded from CodeQL via `.github/codeql/codeql-config.yml`.
- `misc/sanitizer_config.c` cleaned up: stripped comment-per-line bloat, dropped dangling `suppressions=tsan.supp` / `log_path` settings, added `__tysan_default_options` / `__nsan_default_options`. Linked into the canary so all sanitizer-instrumented binaries share the same baseline.
- New `test_post_process.cpp` exercises `buildWindowLevelLut` / `buildWindowLevelColormapLut` across boundaries and all three colormap kinds; passes under every sanitizer.

### CodeQL
- Fast PR scan + deep push scan workflows refreshed: deep-scan apt deps now match the builder image; dropped stale `VC_WITH_PASTIX` / `VC_BUILD_TESTS` cache flags.
- Path-ignore for the canary (intentional UB inside is the point of the test).


giorgioangel:
@codex review

giorgioangel:
CI fails?

SuperOptimizer:
> CI fails?

yeah I'm still working on the new CI workflows. I had to open a PR to get them to trigger on github

jrudolph:
We need to make sure no to use `-march=native` on the docker builds but use a sensible default because the binaries will be used on a large range of different CPU generations on AWS (and potentially on user machines as well).

I haven't looked into it recently, not sure what the current default settings are for x64_86, IIRC the last time I checked it was a bit too conservative for my liking, e.g. we might want to enable AVX2 by default.

SuperOptimizer:
> We need to make sure no to use `-march=native` on the docker builds but use a sensible default because the binaries will be used on a large range of different CPU generations on AWS (and potentially on user machines as well).
> 
> I haven't looked into it recently, not sure what the current default settings are for x64_86, IIRC the last time I checked it was a bit too conservative for my liking, e.g. we might want to enable AVX2 by default.

good point. I think gcc and clang support a generic `march=x86_64_v1` or similar that should work on all 64 bit processes. Or I can just not add that switch at all since that's the default if not specified.

I think `x86_64_v3` enables AVX2

jrudolph:
I think we should use x86_64_v3 (still supports down to Haswell from 2013, and adds AVX2, FMA and all lower) and something like `-march=armv8.4-a+sve -mtune=neoverse-512tvb` for arm (should support both Graviton V1 and V2).

volume-cartographer/ubuntu-24.04-noble.Dockerfile
```diff
@@ -1,127 +1,50 @@
...
-# Create prefixes we will actually keep (not deleted later)
-RUN make prefix=/usr/local/scotch install
-    
-WORKDIR /src/libs/pastix-install/src
-RUN cp /src/libs/config.in config.in
-RUN make SCOTCH_HOME=/usr/local/scotch
-RUN make SCOTCH_HOME=/usr/local/scotch install
-    
-    # --- build libigl-based target (after overlay) ---
-WORKDIR /src/libs/flatboi
```
jrudolph, line 98: Where is all that stuff gone?

SuperOptimizer: flatboi needs a sparse linear solver backend for which it uses libigl. Libigl can use Pastix, which relies on Scotch _or_ flatboi can use CHOLMOD which is provided via suitesparse which we already use. So this swaps the backend and deletes the old vendored libraries.

giorgioangel:
@SuperOptimizer if you right click on a segment and do SLIM Flatten, does it work?

SuperOptimizer:
> @SuperOptimizer if you right click on a segment and do SLIM Flatten, does it work?

I tested just via the CLI yesterday, I'll try it out through the VC3D GUI today

giorgioangel:
PaStiX auto mode is not using nproc; it initializes its own scheduler through hwloc and defaults to detected physical core capacity unless PASTIX_NUM_THREADS is set. On some machines nproc reports a number of online logical CPUs thifferent from the number of cores reported by lscpu, so PaStiX picked automatically picks #lscpu workers to avoid SMT/hyperthread scheduling by default, unless PASTIX_NUM_THREADS is set on.

SuperOptimizer:
@codex review

volume-cartographer/ubuntu-24.04-noble.Dockerfile
SuperOptimizer: update to this: pastix is way faster so I went back to using pastix but now we vendor it in the cmakelists.txt rather than building in the docker scripts

SuperOptimizer:
@codex review

SuperOptimizer:
@codex review

SuperOptimizer:
@codex review

SuperOptimizer:
@codex review

volume-cartographer/CMakeLists.txt
```diff
@@ -1,55 +1,317 @@
...
+    set(CMAKE_BUILD_TYPE Debug CACHE STRING "" FORCE)
+endif()
+
+# Portable shippable defaults; VC_MARCH_NATIVE=ON (set by dev-* presets) opts
+# into host tuning. armv8.2-a floor avoids SVE so Apple M1-M3 + Qualcomm Oryon
+# work alongside Graviton.
+if(VC_MARCH_NATIVE)
+    add_compile_options(-pipe -march=native)
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(x86_64|amd64|AMD64)$")
+    add_compile_options(-pipe -march=x86-64-v3)
```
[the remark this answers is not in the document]
SuperOptimizer: v3 is compatible with everything we care about IIRC so ignoring

volume-cartographer/CMakeLists.txt
```diff
@@ -1,55 +1,317 @@
...
+ProcessorCount(_nproc)
+if(_nproc EQUAL 0)
+    set(_nproc 1)
+endif()
+set(_libbt_install ${CMAKE_BINARY_DIR}/_deps/libbacktrace-install)
+set(_libbt_lib ${_libbt_install}/lib/libbacktrace.a)
+file(MAKE_DIRECTORY ${_libbt_install}/include)
+ExternalProject_Add(libbacktrace_build
+    GIT_REPOSITORY https://github.com/ianlancetaylor/libbacktrace.git
+    GIT_TAG master
```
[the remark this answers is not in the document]
SuperOptimizer: eh I don't really care. IIRC libbacktrace sees commits pretty seldom so dont think this will be an issue

giorgioangel (dismissed):

hendrikschilling (approved):

jrudolph:
After this was merged, vc3d-publish-ghcr.yml is now failing for every push. Why is it even triggered?

PR #880, opened 2026-05-07
vc: in VcDataset use blosc_decompress_ctx to avoid global lock

jrudolph:


giorgioangel:
@codex review

SuperOptimizer:
I don't think there's any reason we can't use multiple threads for the blosc context. 

It was previously set to 1 thread because the compression and decompression were done in multiple threads from a static thread pool on a chunk by chunk basis but that was deleted. I don't know what exactly happens now but I don't think it's particularly multithreaded.

That being said I have no idea if there will be any noticeable performance improvement from giving blosc multiple threads.

SuperOptimizer (approved):

jrudolph:
> That being said I have no idea if there will be any noticeable performance improvement from giving blosc multiple threads.

There are basically two things here: 1. without using the _ctx variant all caller threads block on a global lock 2. we also set nthreads=1 here because chunks are usually small enough that adding fine-grained parallelism just for decompression is not a likely win and also often not used depending on how many blocks blosc created in the first place (there's basically no good reason to create multiple blocks inside a blosc chunk).

The idea is to use more coarse-grained parallelism and allow loading whole chunks in parallel instead (which did not work before because of the lock...).

PR #881, opened 2026-05-07
vc: enable OPENMP in docker build

jrudolph:


giorgioangel:
@codex review

SuperOptimizer (approved):

jrudolph:
Somewhat funny / scary that it wasn't enabled before but maybe the workflows are the main docker image users right now...

SuperOptimizer:
> Somewhat funny / scary that it wasn't enabled before but maybe the workflows are the main docker image users right now...

IIRC It was enabled by default in the top level CMakeLists.txt but I think I accidentally swapped it to default OFF recently. I use LLVM pretty exclusively for development and their openmp implementation is bugged and spins the CPUs to 100% usage even when no work is being done.

I'm working on new CI and CMake enhancements and I swapped it to default ON for GCC and default OFF for LLVM which I think that's the right way to go.

PR #882, opened 2026-05-07
vc: setup partitioning for vc_gen_normalgrids

jrudolph:
This sets up the basic infrastructure / params to run normalgrids in a partitioned setting.

The current state of vc_gen_normalgrids is both functionality and performance-wise broken, so this just sets the stage for the bigger changes required.

### DIFF
--- a/volume-cartographer/apps/src/vc_gen_normalgrids.cpp
+++ b/volume-cartographer/apps/src/vc_gen_normalgrids.cpp
@@ -3,6 +3,7 @@
 #include <vector>
 #include <filesystem>
 #include <iomanip>
+#include <limits>
 
 #include <boost/program_options.hpp>
 #include <opencv2/opencv.hpp>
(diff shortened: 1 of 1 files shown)


SuperOptimizer:
I will actually review a little later, I just wanted to note what I observed by looking at the normal grids code a bit

IIRC the normal grid generation code creates slices the size of the entire volume, so we can create slices that are up to say 80000 x 30000 or so pixels for scroll 1. But I don't think that any of the consumers of the normal grids actually care about 99% of that data. We should probably run some tests with much, _much_ smaller windows, like 500x500 or so, and see if that has a negative impact on downstream usage

bruniss:
> I will actually review a little later, I just wanted to note what I observed by looking at the normal grids code a bit
> 
> IIRC the normal grid generation code creates slices the size of the entire volume, so we can create slices that are up to say 80000 x 30000 or so pixels for scroll 1. But I don't think that any of the consumers of the normal grids actually care about 99% of that data. We should probably run some tests with much, _much_ smaller windows, like 500x500 or so, and see if that has a negative impact on downstream usage



> I will actually review a little later, I just wanted to note what I observed by looking at the normal grids code a bit
> 
> IIRC the normal grid generation code creates slices the size of the entire volume, so we can create slices that are up to say 80000 x 30000 or so pixels for scroll 1. But I don't think that any of the consumers of the normal grids actually care about 99% of that data. We should probably run some tests with much, _much_ smaller windows, like 500x500 or so, and see if that has a negative impact on downstream usage

you can run it on smaller windows during processing, but there is no way to avoid changing the result w.r.t running on the entire slice at once. some parts are more amenable to being split into windows than others

jrudolph:
The interesting stuff is not yet in here ;)

PR #883, opened 2026-05-07
vc: add bring-your-own-buffer variants of chunk reading at various levels of APIs

jrudolph:
For the usual reasons, reusing buffers can help a lot of memory/allocation churn when you go through many chunks.

### DIFF
--- a/volume-cartographer/core/include/vc/core/types/Volume.hpp
+++ b/volume-cartographer/core/include/vc/core/types/Volume.hpp
@@ -212,6 +212,18 @@ public:
     [[nodiscard]] std::optional<std::vector<std::byte>> readChunk(
         int level,
         const std::array<size_t, 3>& chunkZYX) const;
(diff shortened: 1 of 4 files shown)


bruniss:
this is a cool idea to extend the Volume api with . is there any reason we need vcdataset  ? does it have any functionality that volume or  utils::zarr  can't do ? i was hoping to just get rid of it at some point so we dont have three volume reading classes

jrudolph:
> this is a cool idea to extend the Volume api with . is there any reason we need vcdataset ? does it have any functionality that volume or utils::zarr can't do ? i was hoping to just get rid of it at some point so we dont have three volume reading classes

I haven't checked but just added what was necessary without larger changes. These methods in VcDataset.cpp are all global and are used to set up the compression codec registry. I think we can and should at some point move them to their own file.

giorgioangel:
@codex review

PR #888, opened 2026-05-08
Add Spec 1.5 surface predictor to community projects

ciscoriordan:
Adds a one-line entry under **Segmentation > 📦 Materials > Scroll Surface Predictions**, linking to an open-source surface segmentation model that addresses #191 ("Surface and Fiber Predictions in Compressed or Highly Curved areas").

**Spec 1.5** = MedNeXt-L kernel5 + bruniss's SkeletonRecall loss + Dataset059. Beats `d058` by **+0.27 high_compressed IoU absolute (+66% relative)** on held-out Scroll-1 cubes (6/6 cubes won, biggest gap +0.32 on the truly held-out z=0 cube). Drop-in compatible with `d058`'s output contract; `scripts/inference/run_on_zarr.py` accepts OME-Zarr / plain Zarr volumes directly.

- Repo (code + writeup + figures): https://github.com/ciscoriordan/mednext-vs-umamba-scroll
- Weights: https://huggingface.co/ciscoriordan/mednext-l-scroll-surface


giorgioangel:
I don't think it addresses the issue. The validation is outdated because seven months ago we launched a Kaggle competition exactly for this. https://www.kaggle.com/competitions/vesuvius-challenge-surface-detection

PR #893, opened 2026-05-12
vc_gen_normalgrids: prune short thinning traces

jrudolph:
Drop thinning traces whose polyline length is below --prune-min-length-px. Short stubs from the seed-based thinning produce noisy artifacts downstream; pruning them removes the worst of those without measurably hurting useful traces. Pass --prune-min-length-px=0 to keep all raw traces (default).


giorgioangel:
@codex review

PR #912, opened 2026-05-15
fine cyl init, use model center for seed for any non-init opt, remove…

hendrikschilling:
… data choice logic from vc3d
This also fixes exploding non-init corr configs.


hendrikschilling:
@codex review

PR #914, opened 2026-05-16
fix(vesuvius-c): accept file:// URLs in vs_download

jonmarrs:
`vs_download` rejected any response where libcurl reported a non-200 HTTP
status code. For non-HTTP transports (notably `file://`), libcurl leaves
`CURLINFO_RESPONSE_CODE` at `0`, so local-file fetches were treated as
failures even when the buffer was populated successfully.

This change allows `http_code == 0` through the success path so
`vs_download` works for `file://` URLs (useful for local datasets, tests,
and offline mirroring). HTTP behavior is unchanged.

```diff
-    if (http_code != 200) {
+    // file:// urls will have http_code == 0
+    if (http_code != 200 && http_code != 0) {
         free(chunk.buffer);
         return -1;
     }
```


SuperOptimizer:
vesuvius-c is deprecated in favor of volume-cartographer

PR #917, opened 2026-05-16
Handle chunk fetch errors outside OpenMP workers

giorgioangel:
## Crash cause
Remote chunk fetch failures could throw from inside the OpenMP worker loop in `Volume::readFromChunkedArrayZYX`, which can terminate the process instead of returning a normal caller-thread error.

## Fix summary
- Capture the first chunk fetch error in thread-safe shared state.
- Stop using further chunks after the captured failure and let the OpenMP region exit cleanly.
- Throw one `std::runtime_error` on the caller thread with generic level/chunk coordinates only.
- Added a synthetic `IChunkedArray` regression test that returns `ChunkStatus::Error` for one chunk.

## Tests run
- `AGENTS_AGENT_MODE=1 cmake --preset ci-tests-gcc`
- `AGENTS_AGENT_MODE=1 cmake --build --preset ci-tests-gcc --parallel $(nproc)`
- `ctest --test-dir build/ci-tests-gcc -R "chunk|volume" --output-on-failure`
- `ctest --preset ci-tests-gcc`
- `AGENTS_AGENT_MODE=1 cmake --build --preset ci-release-gcc --parallel $(nproc)`

The new regression test uses synthetic data only: no live buckets, URLs, real volumes, tokens, or secrets.


jrudolph:
Wouldn't any exceptions bubbling out here be bugs anyway?

Wouldn't it be better to encapsulate any error handling in the downloader (and log noisily or fail) when such a thing happens instead of relying on exceptions to deal with bugs.

bruniss:
> Wouldn't any exceptions bubbling out here be bugs anyway?
> 
> Wouldn't it be better to encapsulate any error handling in the downloader (and log noisily or fail) when such a thing happens instead of relying on exceptions to deal with bugs.

sort of , but not all chunk errors are genuine bugs. with our zarrs we dont write empty chunks but they are still part of a fetch attempt (at least once, then they are marked as "empty"). if there are exceptions that are not coming out as ChunkStatus/ChunkResult::Error then yes though those should be addressed at the downloader

giorgioangel:
@codex review

PR #919, opened 2026-05-16
ec2_setup.sh: add CGAL/MPFR/GMP + scotch-metis apt deps

SuperOptimizer:
## Summary
- `scripts/ec2_setup.sh` was missing `libcgal-dev`, `libmpfr-dev`, `libgmp-dev`. The top-level `CMakeLists.txt` does `find_package(CGAL REQUIRED)` (CGAL transitively needs MPFR/GMP), so on a fresh EC2 host `cmake --preset ci-release-gcc` fails configure with *"Could not find a package configuration file provided by CGAL"*.
- Also add `libscotchmetis-dev` alongside the existing `libscotch-dev`.

## Test plan
- [x] Fresh aarch64 Ubuntu 26.04 EC2 host: run `ec2_setup.sh`, then `cmake --preset ci-release-gcc && cmake --build --preset ci-release-gcc --target vc_zarr_recompress` — configure + build succeed.

### DIFF
--- a/volume-cartographer/scripts/ec2_setup.sh
+++ b/volume-cartographer/scripts/ec2_setup.sh
@@ -25,12 +25,13 @@ apt-get install -y \
     build-essential git clang llvm ninja-build lld cmake pkg-config \
     qt6-base-dev libboost-system-dev libboost-program-options-dev \
     libceres-dev libsuitesparse-dev \
+    libcgal-dev libmpfr-dev libgmp-dev \
     libopencv-dev libopencv-contrib-dev \
     libblosc-dev libcurl4-openssl-dev \
     libavahi-client-dev nlohmann-json3-dev \
     liblz4-dev libtiff-dev \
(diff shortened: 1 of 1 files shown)


jrudolph (approved):
LGTM, ultimately it would be good if we could maintain all the ubuntu dependencies for local/ec2 and docker build in a single file but not sure how feasible that is and how quick it will break again.

SuperOptimizer:
> LGTM, ultimately it would be good if we could maintain all the ubuntu dependencies for local/ec2 and docker build in a single file but not sure how feasible that is and how quick it will break again.

yeah I want to do this but am a little unsure how. I think right now we have this sort of setup duplication in 4 places: 24.04 dockerfile, 26.04 dockerfile, ec2_Setup.sh, and somewhere in the github workflows files to do the codeQL I believe. I'd prefer exactly one script but I am not sure how well, e.g. docker files will cache the results from that and such.

I'll look into it but want this in particularly for some of the compress3d export stuff I am doing

PR #920, opened 2026-05-16
fix(aws): query IMDSv2 directly + cache instance-role creds

SuperOptimizer:
## Summary
- `AwsAuth::load()` resolved EC2 instance-role credentials by forking the `aws` CLI (`aws configure export-credentials`) on **every call**. `vc_zarr_recompress` calls `load()` per `S3Backend` construction across its 96 inner workers, so dozens of `aws` subprocesses hammered the link-local IMDS endpoint at once. IMDS rate-limits aggressively → some calls returned empty/throttled → `load()` returned **empty credentials** and the S3 op failed (observed 11+ times on a multi-TB recompress run).
- Now query **IMDSv2 directly** via libcurl (no subprocess) and **cache** the resolved creds in-process, refreshing ~5 min before the STS expiry. The metadata fetch happens at most once per rotation window regardless of caller concurrency.

## Behavior
Resolution order: explicit profile → **IMDSv2 (cached)** → SSO profiles → default `export-credentials` → INI files → env vars. Non-EC2 hosts fail the IMDSv2 probe in ~1s (tight link-local timeout) and fall through unchanged.

## Test plan
- [x] EC2 aarch64 host, recompress run that previously logged 11+ `FAILED`/empty-cred events: now logs exactly **1** `IMDSv2: got role ... creds` fetch, **0** `aws configure export-credentials` forks, **0** failures, run completes `errors: 0`.
- [x] Credentials served from cache on all subsequent `load()` calls within the run.

### DIFF
--- a/volume-cartographer/utils/include/utils/http_fetch.hpp
+++ b/volume-cartographer/utils/include/utils/http_fetch.hpp
@@ -193,10 +193,17 @@ struct AwsAuth {
     }
 
     /// Full credential resolution. Tries in order:
-    ///   1. `aws configure export-credentials` (resolves SSO, assume-role, etc.)
-    ///   2. ~/.aws/credentials + ~/.aws/config INI files
-    ///   3. Environment variables
-    /// Respects AWS_PROFILE for methods 1 & 2.
+    ///   1. Explicit profile via `aws configure export-credentials`
+    ///      (only when AWS_PROFILE is set or a non-default profile is passed)
+    ///   2. EC2 instance role via IMDSv2 — queried directly (no subprocess)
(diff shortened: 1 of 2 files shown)


jrudolph:
Would be really good to not do most of this here and require some external tools to gather a set of credentials into env vars (which is how everything else does it).

SuperOptimizer:
> Would be really good to not do most of this here and require some external tools to gather a set of credentials into env vars (which is how everything else does it).

yeah the s3 auth stuff is just getting ridiculous. I vibed this up to do the compress3d export. It's good enough for that but I'll close this and rethink aws auth in general

PR #924, opened 2026-05-17
Add DBbun executable companion examples as related resource

kartoun:
This pull request adds DBbun Companions as a related resource. DBbun Companions is a collection of executable companion examples that transform documents, images, papers, patents, articles, and other source materials into reproducible simulation packages with generated outputs, figures, structured data, and documentation.

This resource is not Vesuvius-specific, but may be relevant as a related workflow pattern for organizing complex visual/document-based analysis outputs into reproducible packages.


giorgioangel:
Not vesuvius specific

PR #931, opened 2026-05-19
build: fix concurrency in vc3d-publish-ghcr.yml

jrudolph:
Always only build the latest HEAD, no need to publish for old commits.

### DIFF
--- a/.github/workflows/vc3d-publish-ghcr.yml
+++ b/.github/workflows/vc3d-publish-ghcr.yml
@@ -23,15 +23,14 @@ permissions:
   contents: read
   packages: write
 
-concurrency:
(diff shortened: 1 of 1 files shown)


jrudolph:
This is the quick fix since this was broken since the CI/build changes.

@SuperOptimizer why do we need multiple docker builds again? The users basically don't care what user space distro we use, so it's our decision whether to use Ubuntu 24.04 or 26.04, there's no compatibility constraint on the user machines. If we think we should move to 26.04, we should just do it and remove another build dimension.

giorgioangel:
@jrudolph agreed, it's good to test that it builds also on a mac but we don't need multiple dockers.

PR #934, opened 2026-05-19
consolidate ci and docker deps into one script + drop ubuntu 24.04 docker + speedup CI

SuperOptimizer:
- drops the ubuntu 24.04 dockerfile in favor of just the 26.04 now that it's available on aws
- consolidates ec2_setup, dockerfile setup, and some apt install setup from the workflows into one place
- wraps all test executable stuff into vc_add_test so that the sanitizer suppressions correctly get linked in
- fix a flaky test error thing with how volpkg writes our its current project tmp file
- fix teh ci script to correctly call cmd_builder so we arent using a stale ghcr docker image when changes are made to docker or ci itself
- actually fix the threadsan suppressions
- dropped xpra from ec2 setup because its binary release doesnt exist in 26.04 and trying to work around that sucks
- fix the GHA running out of storage
- swap to mostly dynamic linking to save storage space and build times
- drop gcov in favor of clang coverage because gcov doesnt work as well with optimization passes like Os causing binary bloat
- try and get PCH working again on CI for faster CI times


jrudolph:
Would be nice if it works. 

Did you actually test all the paths that it works?
 - [ ] GHA CI
 - [ ] ec2_setup
 - [ ] local
 - [ ] docker local

jrudolph:
I reran the failing thread sanitizer test. Does that now fail regularly?

SuperOptimizer:
> I reran the failing thread sanitizer test. Does that now fail regularly?

the threadsan workflow is still failing, yes. AFAICT the failures are coming from threading issues in third party libraries so I've tried suppressing them but still haven't managed to actually suppress the right things

SuperOptimizer:
closing in favor of the like 5 PRs I'm going to open

PR #938, opened 2026-05-19
perf: cut slice-sampler per-pixel overhead

SuperOptimizer:
Profiling interactive VC3D rendering showed the slice sampler spending ~24% of render time on `LocalChunkCache` bookkeeping rather than actual sampling: `get()` returned `ChunkResult` by value, so trilinear sampling paid a `shared_ptr` refcount bump + `std::string` copy per voxel (8 per pixel), and re-hashed the same chunk key 8 times per pixel. The `~ChunkResult` destructor was the #4 hotspot at 7.7%.

Changes:
- `LocalChunkCache::get` returns `const ChunkResult&` + a last-key shortcut for the common adjacent-pixel case.
- Drop the redundant per-pixel `finiteCoord` check (`inLevelBounds` rejects non-finite coords identically — NaN compares false) and the libm `floor` (`inLevelBounds` guarantees coords ≥ 0, so truncation == floor).
- Remove the `WHOLE_ARCHIVE` wrapper on `utils_c3d_codec`: its stated ThinLTO/lld rationale is false (c3d symbols are referenced directly by name); verified VC3D links clean on gcc+LTO and clang+ThinLTO+lld.
- Add a best-of-N `ChunkedPlaneSampler` perf bench (behind `VC_RUN_BENCHES`).

Re-render profile after: cache bookkeeping 24% → 5%, `~ChunkResult` gone. Existing sampler correctness tests pass unchanged.

Note: built/measured on top of #937 (c3d re-vendor); independent at the code level but best landed with or after it.

### DIFF
--- a/volume-cartographer/core/CMakeLists.txt
+++ b/volume-cartographer/core/CMakeLists.txt
@@ -53,11 +53,9 @@ target_link_libraries(vc_core
 )
 
 # utils header-only library + vendored c3d codec.
-# WHOLE_ARCHIVE is needed because ThinLTO + lld can internalize symbols from
-# static archives before cross-archive references are resolved.
 target_link_libraries(vc_core PUBLIC
     $<BUILD_INTERFACE:utils>
-    $<BUILD_INTERFACE:$<LINK_LIBRARY:WHOLE_ARCHIVE,utils_c3d_codec>>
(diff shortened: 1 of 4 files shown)


SuperOptimizer:
@codex review

PR #960, opened 2026-05-22
Streamline trainer + new copy model inference / dataset / python Volume bindings 

bruniss:


SuperOptimizer:
I think you will need to merge the latest main to bring in the fix for all these failures

SuperOptimizer:
the coverage fail can be ignored; I'm still trying to get the coverage regression workflow to a good baseline. Granted if you want to add tests to any of the c++ side stuff feel free to; that might get that workflow green

hendrikschilling (approved):
lasagna stuff should be fine - I was just using cached neural tracer anyways

PR #970, opened 2026-05-23
VC3D: wire vc_merge_patch into the surface workflow

skyward7187:
## Summary

Adds VC3D GUI surface for `vc_merge_patch` (introduced in #969), mirroring the existing `vc_merge_tifxyz` integration so the two patch/merge entry points feel symmetric. Five files touched, ~730 lines of additive code.

**Depends on #969 — please review/merge that first.** This branch is stacked on top of #969, so the diff below currently includes both that PR's commits and this PR's GUI commit. Once #969 lands, GitHub will collapse this PR's diff to just the GUI commit.

Opening as a draft so it doesn't get merged before the binary lands.

## What's added

Five layers, modelled directly on `MergeTifxyzDialog`'s existing integration:

**1. `MergePatchDialog`** (`ToolDialogs.{hpp,cpp}`) — modal QDialog with:
- Two surface combo pickers (defaulted from the right-click selection or empty when launched from the top menu)
- A live **role hint** that auto-detects parent vs child by calling `QuadSurface::countValidPoints()` against the already-loaded VolumePkg cache (in-memory scan, no I/O)
- A **Swap parent/child** button that flips the auto-detect and pins the choice as explicit
- Border-cells / blend-cells / IDW-k spinners
- Checkable **Advanced** fold-out with the RANSAC set + anchor-cap + OMP threads
- A red **destructive-action banner** naming the parent dir + the backup-ring path (`<volpkg>/backups/<parent>/{0..7}/`)
- Per-session sticky defaults via statics so the user's last config persists across one VC3D run

**2. `CommandLineToolRunner`** (`CommandLineToolRunner.{hpp,cpp}`):
- New `Tool::MergePatch` enum entry
- `setMergePatchParams(...)` stores the twelve params into fresh `_patch*` private fields
- `buildArguments(Tool::MergePatch)` emits either two positional paths (auto-detect mode) or `--parent`/`--child` (when the user swapped), followed by `--border-cells`/`--blend-cells`/`--idw-k`/the RANSAC set/`--anchor-cap`
- `toolName(Tool::MergePatch)` returns `<appdir>/vc_merge_patch`
- `getOutputPath` returns the overwritten parent path so the post-run status message points at what changed

**3. `SegmentationCommandHandler::onMergePatch`**:
- Standard preconditions (vpkg loaded, runner idle, ≥2 segments)
- Resolves `volpkgDir` / `pathsDir` from `vpkg->outputSegmentsPath()`
- Constructs the dialog, on Accept calls `setMergePatchParams` + `execute(Tool::MergePatch)` with console streaming
- Installs a one-shot lambda on `toolFinished` filtered to `Tool::MergePatch`; on success calls `_surfacePanel->reloadSurfacesFromDisk()`. The mask.tif mtime delta in `detectSurfaceChanges` drives the per-surface reload, so the in-memory QuadSurface picks up the patched x/y/z without any per-surface invalidation here

**4. Menu + context-menu wiring**:
- `MenuActionController`: `_mergePatchAct` next to `_mergeTifxyzAct` in the Actions menu, `mergePatchFromMenuRequested()` signal
- `SurfacePanelController`: context-menu item gated on `selectedSegmentIds.size() == 2` (the patch tool is strictly 2-input, unlike merge's ≥2), `mergePatchRequested(QStringList)` signal

**5. `CWindow`** wires both signals into `_segmentationCommandHandler->onMergePatch(...)` — same lambda-bind pattern used for `mergeTifxyzRequested`.

## Smoke test

Walked through the workflow on the PHerc Paris 4 volpkg:
- Right-click 2 surfaces in the SurfacePanel → **Patch tifxyz...** opens the dialog with both combos populated
- Role hint correctly identifies the larger surface as parent (`20231210121321_v6`, valid=7,636,244) and the smaller as child (`dj_test_001_v003.tifxyz`, valid=2,896); tag reads `(auto-detect)`
- **Swap parent/child** flips them and changes the tag to `(swapped — explicit)`; flipping back restores auto-detect
- **Actions → Patch tifxyz...** from the top menu opens with empty combos
- Banner shows: \"Parent <name> will be overwritten in place... snapshot to /volpkg/backups/<name>/{0..7}/\"
- **OK** launches `vc_merge_patch`, the console widget streams stdout live (same as `vc_merge_tifxyz`)
- After successful exit, `reloadSurfacesFromDisk()` fires, the parent surface's mask.tif mtime delta triggers `detectSurfaceChanges`, and the panel re-loads the patched parent

## Open items I'd appreciate input on

- **Dialog sizing**: I have `resize(560, 720)` + `setSizeGripEnabled(true)` so the user can drag taller if the OMP threads row feels tight on their display. Happy to adjust the baked-in size if your displays render differently.
- **Confirmation UX**: chose banner-only (no \"I understand\" checkbox, no separate modal) on @skyward7187's preference — recoverable via the backup ring, and the banner makes the destination + recovery path obvious before OK.
- **Role hint** uses HTML (`<br/>`, colored `<span>`). Auto-detected as rich-text by Qt; switch to plain text if your theme has issues with that.

## Test plan

- [x] VC3D builds cleanly with the change (`cmake --build build --target VC3D`)
- [x] Dialog opens from both entry points (top menu + context menu)
- [x] Role auto-detect works on real data; Swap toggles to explicit
- [x] OK launches the binary, console streams, parent reloads on success
- [ ] Synthetic e2e test for the GUI (would need QTest fixtures; not added here)

### DIFF
--- a/volume-cartographer/apps/VC3D/CWindow.cpp
+++ b/volume-cartographer/apps/VC3D/CWindow.cpp
@@ -479,6 +479,10 @@ CWindow::CWindow(size_t cacheSizeGB) :
             this, [this]() {
                 _segmentationCommandHandler->onMergeTifxyz(QStringList{});
             });
+    connect(_menuController.get(), &MenuActionController::mergePatchFromMenuRequested,
+            this, [this]() {
+                _segmentationCommandHandler->onMergePatch(QStringList{});
+            });
 
     if (isDarkMode()) {
         applyDarkPalette();
@@ -1530,6 +1534,10 @@ void CWindow::CreateWidgets(void)
     connect(_surfacePanel.get(), &SurfacePanelController::mergeTifxyzRequested,
             this, [this](const QStringList& segmentIds) {
                 _segmentationCommandHandler->onMergeTifxyz(segmentIds);
+            });
+    connect(_surfacePanel.get(), &SurfacePanelController::mergePatchRequested,
+            this, [this](const QStringList& segmentIds) {
+                _segmentationCommandHandler->onMergePatch(segmentIds);
             });
     // Note: the Actions -> Merge tifxyz... menu wiring lives in the
     // constructor after _menuController is initialized -- _menuController
--- a/volume-cartographer/apps/VC3D/CommandLineToolRunner.cpp
+++ b/volume-cartographer/apps/VC3D/CommandLineToolRunner.cpp
@@ -230,8 +230,11 @@ bool CommandLineToolRunner::execute(Tool tool)
         return false;
     }
 
-    // vc_merge_tifxyz operates on segment paths only; no volume required.
-    const bool needsVolume = !isCustom && tool != Tool::MergeTifxyz;
+    // vc_merge_tifxyz and vc_merge_patch are path-only tools (segment dirs);
+    // no volume required.
+    const bool needsVolume = !isCustom
+                             && tool != Tool::MergeTifxyz
+                             && tool != Tool::MergePatch;
     if (needsVolume) {
         if (_explicitVolumePath) {
             if (_volumePath.isEmpty()) {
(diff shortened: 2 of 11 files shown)


hendrikschilling:
@codex review

hendrikschilling (dismissed):

SuperOptimizer (approved):

PR #988, opened 2026-05-27
flatten: heal degenerate cells so SLIM stops NaNing / crashing

SuperOptimizer:
## Problem

Full-res SLIM flattening of some segments failed hard: `-nan` energy at iter 1 (and, once that was worked around, a PaStiX heap corruption / SIGSEGV during factorization). It happened at every decimation level, so decimation never rescued it.

Root cause is degenerate input geometry the tracer emits: **grid-adjacent** points collapsed onto the same (or sub-voxel) 3D location. A healthy sheet keeps adjacent cells ~`1/scale` voxels apart; the repro segment (`20230702185753_v11`, 3.9M verts) had ~6k coincident adjacent points scattered in ~2,500 tiny clusters. These make zero-area triangles → infinite symmetric-Dirichlet energy → singular system → NaN.

## Fix (three layers, all geometry-side)

1. **`healDegenerateCells()`** (core, next to `inpaintSurfaceHoles`): detects collapsed grid cells, drops the connected cluster (dilated by a ring) so the inpaint gets a clean interior hole with healthy boundary, and refills via `inpaintSurfaceHoles`, iterating a few passes. Threshold is relative to `1/scale` so it adapts across segments. Optional `dropped_mask` out-param lets callers clear the same cells in other channels (approval, etc.). Repairs ~70k cells on the repro.

2. **Scale-relative emit guard** in `vc_tifxyz2obj`: a few fully-collapsed 2×2 blocks can't be separated locally, and even one zero-area triangle re-NaNs SLIM. The guard skips any triangle with a collapsed edge or sliver altitude → **0** zero-area triangles reach the OBJ (worst case: a sub-cell hole the flattener tolerates).

3. **Two-pass emit** in `vc_tifxyz2obj`: skipping triangles can leave a vertex referenced by no surviving face. Such an **orphan** is an all-zero row/col in `L = AᵀWA`, which crashes the PaStiX Cholesky (the "double free / SIGSEGV" that looked like a solver bug). Now we collect surviving triangles first, then emit only the vertices they use — no orphans.

Plus a one-line latent fix in the `Pastix6LLT` wrapper: `spmInit` leaves `spm.replicated = -1` ("uninitialized"); set it to `1` (single local matrix) so PaStiX doesn't take distributed-matrix branches.

## Verification

`20230702185753_v11`: was `-nan` at iter 1 at every decimation. Now flattens cleanly **via the VC3D SLIM panel** (GUI, decimated path, new segment created) and via CLI at `keep=2/6/25` — 0 orphans, 0 zero-area triangles, converged L2 ≈ 1.0, valid output OBJ.

## Notes

- The tracer is the upstream source of these collapsed cells (it initializes new points at a neighbor's position and can accept them without a min-spacing check). Fixing the tracer is a separate, larger change; this PR makes the flatten pipeline robust to the defects regardless.
- Full-res (3.9M verts) may still OOM on low-RAM hosts — that's a genuine memory limit, not a crash; decimate as needed.

### DIFF
--- a/volume-cartographer/apps/src/vc_tifxyz2obj.cpp
+++ b/volume-cartographer/apps/src/vc_tifxyz2obj.cpp
@@ -186,7 +186,20 @@ static cv::Mat_<cv::Vec3f> build_vertex_normals_from_faces(
 static void surf_write_obj(QuadSurface *surf, const std::filesystem::path &out_fn, bool normalize_uv, bool align_grid, float keep_percent, bool clean_surface, float clean_sigma_k, bool inpaint_holes)
 {
     cv::Mat_<cv::Vec3f> points = surf->rawPoints();
-    
+
+    // Heal degenerate (collapsed grid-adjacent) cells before anything else.
+    // These tracer defects make zero-area triangles that NaN SLIM at iter 1;
+    // healthy adjacent cells are ~1/scale voxels apart. No-op if none found.
+    {
+        const cv::Vec2f sc = surf->scale();
+        const double scale = (std::isfinite(sc[0]) && sc[0] > 0.f) ? sc[0] : 0.0;
+        const int healed = vc::core::util::healDegenerateCells(points, scale);
+        if (healed > 0) {
+            std::cerr << "note: healed " << healed
+                      << " degenerate (collapsed) grid cell(s) before flattening\n";
+        }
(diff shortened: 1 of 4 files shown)


giorgioangel:
@codex review

PR #989, opened 2026-05-27
add scale for in / out in lasagna panel

bruniss:


hendrikschilling:
should be adressed in:
https://github.com/ScrollPrize/villa/pull/990

PR #1013, opened 2026-06-04
Community Projects: add Plumbline (segmentation trace-quality checker)

GitHwwb:
Adds Plumbline to Segmentation → Tools.

Plumbline is a CPU-only tool that reads a segment's ink-detection prediction and flags where the trace likely went wrong (orientation breaks, row-spacing jumps, garble, and vertical sheet-jumps), output as a self-contained HTML trace-quality dashboard and JSON sidecar. No model inference, GPU, 3D, or server.

- Repo: https://github.com/GitHwwb/plumbline (Apache-2.0)
- Live demo: https://githwwb.github.io/plumbline/

This PR edits only `scrollprize.org/docs/20_community_projects.md` (one entry).

### DIFF
--- a/scrollprize.org/docs/20_community_projects.md
+++ b/scrollprize.org/docs/20_community_projects.md
@@ -164,7 +164,7 @@ For state-of-the-art updates join our [Discord server](https://discord.com/invit
   
 - [ScrollFiesta -- virtual meshing & unwrapping for the Herculaneum papyri](https://github.com/Hob3rMallow/scrollfiesta_public) by HariSeldon and friends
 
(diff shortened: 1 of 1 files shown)


GitHwwb:
Update since opening: Plumbline went through a hardening round, 76 → 115 tests, CI on 3.11/3.12.

- Rotated segments are detected and analyzed in the text's own frame (validated on GP-banner label sets stored at ~−74° and ~90°).
- Photos and diffuse predictions no longer over-flag legible text (rowness and ink density are now measured relative to each tile's background level).
- Reports are now interactive (per-tile measurements on hover, click-to-locate), and batch runs in parallel with chunked Zarr reads.

Live demo reflects the current build: https://githwwb.github.io/plumbline/

giorgioangel:
Hi! Can you explain a bit more how this is helpful? In this current state I don't think it is, but I could be wrong!

erdpx:
we’re closing this for now, but feel free to share any updates later!

PR #1021, opened 2026-06-06
Add StreakArtifactTransform: GPU CT streak-artifact augmentation (#201)

pscamillo:
# StreakArtifactTransform: GPU CT streak-artifact augmentation

Adds a scroll/CT-specific augmentation that simulates **streak artifacts** — the
straight bright/dark lines that cross the reconstructed axial plane from
sparse/limited-angle acquisition (aliasing streaks) or from highly attenuating
inclusions. Like rings they lie in the axial plane and are extruded along the
scan axis, but they are linear rather than concentric. They create spurious
straight structures a segmentation network can mistake for sheet boundaries.
Follows the "additional augmentations welcome" note in #201.

No existing transform produces line streaks: `BlankRectangleTransform` makes
filled boxes, `SmearTransform` makes a directional blur — neither generates thin
straight lines. So this fills a real gap.

**Grounding in the Vesuvius pipeline.** Per the project FAQ, reconstruction is
filtered backprojection from synchrotron parallel-beam scans. Streaks are a
textbook FBP artifact from angular undersampling / limited-angle and from highly
attenuating inclusions, so this models an artifact present in the reconstructed
volumes (augmentation for robustness, not reconstruction). As with rings,
amplitudes are in per-channel std units because the released intensities are
unitless / relative (raw floats ~[-0.1, 0.1]); the transform is additive and
non-destructive.

## Design

- `ImageOnlyTransform` (batchgeneratorsv2 style), fully on-device, in-place.
- Per call: pick an in-plane orientation θ, project plane coordinates onto it
  (t = a·cosθ + b·sinθ), build a 1-D profile of Gaussian-shaped streaks over t,
  and gather it onto the volume by the per-voxel projected coordinate (linear
  interpolation), broadcast along the scan axis (no per-voxel materialization).
- Amplitudes in per-channel std units by default (consistent under z-scored
  nnU-Net inputs). Sign drawn per streak (bright/dark).
- Handles `(C,X,Y)` and `(C,X,Y,Z)`; singleton spatial axes are never used to
  define the plane (no-op if < 2 non-singleton spatial axes).

## Speed (RTX 5070, Blackwell sm_120, single sample)

| patch (C,X,Y,Z)   | Streak ms/sample | Streak Mvox/s |
|-------------------|------------------|---------------|
| (1,192,192,1)     | 0.391            | 94            |
| (1,192,192,64)    | 0.405            | 5821          |
| (1,128,128,128)   | 0.333            | 6292          |
| (1,256,256,96)    | 0.469            | 13418         |

Sub-millisecond per sample (up to ~13.4 Gvox/s); cost is flat as the scan axis
grows because a single 2-D field is broadcast along it. The vendored 0.2.1
`GaussianNoiseTransform` is CPU-only, so it is shown as a CPU reference.

## Ablation (controlled synthetic, multi-seed, RTX 5070)

Two identical `SmallSeg3D` nets per seed, same init / data / steps (400, patch
64³). A = baseline (flips only); B = baseline + `StreakArtifactTransform`. Both
evaluated across a sweep of test-time streak severities. The test streaks are
produced by an **independent** operator (hard-edged top-hat lines, vs the
transform's Gaussian-profiled lines), so this is not testing on the training
augmentation. 5 seeds; mean ± std across seeds, with the paired B − A.

| test streak amp | A (flips)       | B (+streak)     | B − A (paired)    | seeds B>A |
|-----------------|-----------------|-----------------|-------------------|-----------|
| 0.00 (clean)    | 0.9808 ± 0.0074 | 0.9863 ± 0.0051 | +0.0055 ± 0.0056  | 4/5       |
| 0.20            | 0.9328 ± 0.0124 | 0.9486 ± 0.0079 | +0.0158 ± 0.0089  | 5/5       |
| 0.40            | 0.8296 ± 0.0265 | 0.8640 ± 0.0218 | +0.0344 ± 0.0177  | 5/5       |
| 0.60            | 0.7421 ± 0.0297 | 0.7936 ± 0.0331 | +0.0514 ± 0.0254  | 5/5       |

Pooled (amp>0): **B − A = +0.0339 ± 0.0236, 15/15 runs favour streak-aug**. The
gain grows monotonically with streak severity (the streak is a strong confounder —
the flips-only baseline falls to 0.74 Dice at high severity, and the augmentation
recovers +0.05), with no clean-data cost (clean is a slight gain). The real
magnitude is for the villa pipeline on scroll data; this synthetic ablation
matches the format of the merged Squeeze / Decohesion / Warp contributions.

## Figure
streak_triptych

*Illustrative effect of `StreakArtifactTransform`, rendered at 256² for visibility
(training patches are 64³). Left: clean synthetic papyrus. Centre: the additive
field (after − before) — straight bright/dark line streaks at an in-plane
orientation. Right: result, with streaks crossing the sheets. Parameters here are
stronger than the proposed training defaults (`intensity=(0.05,0.3)`) for
legibility.*

## Open questions for bruniss / giorgioangel

- Default streak orientation is uniform random in-plane; pin to a range if a
  specific acquisition geometry dominates.
- Scan axis: defaults to any non-singleton axis; set `streak_axes=[2]` to pin to Z.
- Wrapper probability in the training list (defaulting to ~0.15–0.25).

### DIFF
--- a/segmentation/models/batchgeneratorsv2/batchgeneratorsv2/transforms/noise/streak_artifact.py
+++ b/segmentation/models/batchgeneratorsv2/batchgeneratorsv2/transforms/noise/streak_artifact.py
@@ -44,6 +44,10 @@ class StreakArtifactTransform(ImageOnlyTransform):
         Scale amplitudes by per-channel spatial std.
     streak_axes : Optional[List[int]]
         Candidate scan/extrusion axes (0-based spatial). None -> all non-singleton.
+        For anisotropic data with a distinct acquisition axis (e.g. scroll CT),
+        set this to the scan/rotation axis so streaks lie in the true
+        reconstruction plane; None randomizes across all spatial axes and can
+        place streaks in non-physical orientations on anisotropic patches.
     p_per_channel : float
     synchronize_channels : bool
     benchmark : bool
--- a/segmentation/models/batchgeneratorsv2/batchgeneratorsv2/transforms/noise/streak_artifact.py
+++ b/segmentation/models/batchgeneratorsv2/batchgeneratorsv2/transforms/noise/streak_artifact.py
@@ -1,0 +1,198 @@
+from typing import List, Optional
+
+import torch
+
+from batchgeneratorsv2.helpers.scalar_type import RandomScalar, sample_scalar
+from batchgeneratorsv2.transforms.base.basic_transform import ImageOnlyTransform
+
+
+class StreakArtifactTransform(ImageOnlyTransform):
+    """
+    Simulates CT streak artifacts on the GPU (directional / undersampling streaks).
+
+    Physics
+    -------
+    Streak artifacts are straight bright/dark lines that cross the reconstructed
+    axial plane. They arise from sparse-angle / limited-angle acquisition
+    (aliasing streaks) and from highly attenuating inclusions. Like rings, they
+    live in the axial (reconstruction) plane and are extruded along the scan axis.
+    Unlike rings (concentric) they are linear: a family of parallel lines at some
+    in-plane orientation. They create spurious straight structures the network can
+    mistake for sheet boundaries; training with them as an augmentation builds
+    robustness.
+
+    Implementation
+    --------------
+    Per call: pick an in-plane orientation theta, project the plane coordinates
+    onto it (t = a·cosθ + b·sinθ), build a 1-D profile of Gaussian-shaped streaks
+    over t, and gather it onto the volume by the per-voxel projected coordinate
+    (linear interpolation), broadcast along the scan axis. Amplitudes are in
+    per-channel std units by default. Image-only; (C,X,Y) and (C,X,Y,Z); singleton
+    spatial axes are never used for the plane (no-op if < 2 non-singleton axes).
+
+    Parameters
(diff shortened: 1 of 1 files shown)


pscamillo:
**Validated on real Scroll-5 CT**

I ran this transform against a real raw CT crop from Scroll 5 (PHerc172, from the open zarr store) — before/after plus the per-voxel difference across depth.

With the scan axis set explicitly, the artifact reproduces cleanly: a family of straight, parallel directional streaks at a single in-plane orientation, additive over the real papyrus texture without erasing the fibers, consistent along the scan axis — matching the docstring physics. No edge or interpolation artifacts.

One usage note (now in the docstring): with `streak_axes=None` the scan axis is sampled across all spatial axes, so on anisotropic patches the streaks can land in a plane containing the through-axis — a non-physical orientation for scroll CT, where an arbitrary streak could also mimic a real sheet boundary. For the scroll trainer I'd set `streak_axes` to the acquisition scan axis.

GPU cost (RTX 5070, Blackwell sm_120): **0.401 ms/call** for a `(1, 192, 192, 64)` patch.

fidelity_StreakArtifactTransform

PR #1022, opened 2026-06-06
Community projects: add Ring + Streak CT-artifact augmentations (#201)

pscamillo:
Updated the description of scroll-specific augmentations to include additional features and validation details.

### DIFF
--- a/scrollprize.org/docs/20_community_projects.md
+++ b/scrollprize.org/docs/20_community_projects.md
@@ -79,7 +79,7 @@ For state-of-the-art updates join our [Discord server](https://discord.com/invit
 
 ### 🛠️ Tools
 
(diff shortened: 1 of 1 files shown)


pscamillo:
Both transforms referenced here are now validated on a real raw CT crop from Scroll 5 (PHerc172, open zarr store), not just the synthetic self-tests — before/after plus per-voxel difference across depth.

- **Ring** (#1020): concentric rings in the reconstruction plane, additive over papyrus; floored `radius_range` to `(0.1, 0.9)` so a ring never degenerates into a central blob.
- **Streak** (#1021): straight parallel directional streaks, additive over papyrus, no edge artifacts.

For both, when wiring into the scroll trainer I'd constrain `ring_axes` / `streak_axes` to the acquisition scan axis so the artifacts lie in the true reconstruction plane (the `None` default randomizes across all spatial axes). GPU cost on RTX 5070 (Blackwell sm_120): Ring 0.379 ms/call, Streak 0.401 ms/call for a `(1, 192, 192, 64)` patch. Per-transform detail and montages in the individual PRs.
ring_floored_1
fidelity_StreakArtifactTransform

PR #1036, opened 2026-06-10
VC3D: matter-compressor render cache (one .mca per volume, mc_cache resident) + fetch/render perf

SuperOptimizer:
## What

Replaces VC3D's render caching with matter-compressor: remote (and local) chunks are fetched once, re-encoded into ONE persistent `volume.mca` per volume (all LODs, crash-safe, appendable), and served as 16³ blocks through mc_cache (sharded S3FIFO resident cache inside the library). VC3D-side caching scaffolding is deleted; the ChunkCache keeps only per-256³-region fetch status plus the async orchestration (worker pool, listeners). Non-uint8 volumes keep the legacy byte-LRU path.

Vendored matter-compressor is at `1e53dbb` (format v6: per-axis dims, self-contained blocks, SIMD decode). Stale/older-format cache files are detected and rebuilt automatically.

## Fetch path

- per-region single-flight; worker-pool tasks interleave across regions so 48 I/O workers fetch distinct regions concurrently (workers are ~85% network-blocked)
- remote sharded zarr reads: shard index table cached after one GET — one S3 round trip per chunk instead of two (halves cold-fetch latency)
- 128³-chunk sources assemble their 256³ region with parallel sub-fetches
- download stats/status bar count only bytes actually pulled from the source; archive serves are not "downloading"

## Rendering / frontend

Profile-driven (perf + `--profile` logs, before/after in the commit messages):

- interactive preview system deleted (~420 lines): no more transformed-stale-frame previews, no 140ms settle timer, no 250/500ms chunk-ready coalescing windows (which starved repaints while chunks streamed). Interactions schedule normal full renders behind the single 16ms debounce; cache-side chunk-ready callbacks are throttled to ~30Hz with a guaranteed final notify
- speculative prefetch deleted: render-worker `tryGetChunk` misses queue exactly what requested frames need; the mca archive makes refetches cheap
- stale-render coalescing: a submit while the worker is busy only invalidates the in-flight frame if view parameters changed (fingerprint); data-only refreshes display it. Discarded renders: 19% → 5%
- pool submission moved outside the cache mutex; ChunkKey hash packs to 64 bits + one mix; shift/mask chunk indexing in the trilinear sampler hot path
- main-thread stalls (formerly up to 284ms in `submitRender`) are gone — the UI thread no longer registers in CPU profiles

## Also

- fixed a segfault on every exit: gnutls/libtasn1 DSO destructors free through mimalloc after its teardown in `_dl_fini`; VC3D now runs real teardown then `_exit()`

## Measured end state

Steady-state (warm archive): blocks decode from `volume.mca` via mc_cache, zero network, ~4% CPU in mc block decode. Cold exploration: c3d decode + mc encode dominate pool threads (once per region ever, amortizes out). Render frames are 94% sampler kernel — follow-up territory (pan reuse, kernel work).

Known: `test_volume_local` exact-roundtrip fails on this branch (reads go through the lossy q=8 mca; pre-existing design question, not a regression).


SuperOptimizer:
@codex review

SuperOptimizer:
Both codex P1s addressed in 8daf20ca6:

- **Non-cubic chunk shapes**: `applyMatterCache` now requires cubic source chunks whose edge divides 256 on every level (the region assembler's assumption) and falls back to the raw cache otherwise. No known producer emits non-cubic chunks, so this is a guard rather than full anisotropic support.
- **Stale local archives after writes**: `Volume::invalidateCache` — which is called from every write path (`writeZYX`, `removeChunk`, pyramid rebuild) and nowhere else — now also deletes the local sibling `.mcacache`, so reads after an edit rebuild from the updated zarr. Remote archives are unaffected (immutable sources).

SuperOptimizer:
@codex review

SuperOptimizer:
Round 2 addressed:

- **Sub-fetch failures** (`ef1fa63`→ fixed in latest): any HTTP/IO/decode error during region assembly now unclaims the region for retry instead of appending a zero-holed region to the archive. Only genuinely Missing sub-chunks are treated as air.
- **Local read exactness**: `Volume::readZYX` on local volumes reads the zarr directly — the data API returns exact bytes; the lossy mca cache remains the render/sampling path only.
- **CodeQL popen/path findings**: profile names are now validated against the AWS profile-name alphabet before reaching `popen` (upstream libs3 @93b2952, re-vendored). The env-derived `fopen` paths ($HOME/.aws etc.) are credential-file resolution by design, same trust domain as the process.

Also since the last review: remote `.mca` volume streaming — compressed chunk blobs are range-GET'd and appended verbatim to the local mirror archive (no decode/re-encode), verified bit-identical against the source archive over real S3.

SuperOptimizer:
@codex reivew

SuperOptimizer:
Round 3 addressed in 0dad89bc7:

- **Append failures**: a failed `appendChunkRaw` (disk full, write error) now unclaims the region for retry — same contract as the sub-fetch failure path — instead of publishing Present and serving zeros.
- **Priors after stale recreate**: `MatterArchive` tracks `createdFresh()` through its ctor (covering the stale-delete+recreate path), and the remote-mirror priors install keys off that instead of a pre-construction existence check.
- **CodeQL unbounded write**: sigv4/userpwd writes in libs3 are snprintf-bounded with sizes threaded through `apply_auth` (upstream a4396ff, re-vendored).

SuperOptimizer:
@codex review

PR #1039, opened 2026-06-13
Add vc_straighten + self-calibrating vc_merge_tifxyz gates + VC3D Straighten menu

skyward7187:
## Summary

Tooling for straightening flattened tifxyz surfaces, plus a fix for a winding-merge failure mode, developed on P.Herc. 1667. Three pieces:

1. **`vc_merge_tifxyz`: self-calibrating RANSAC gates for fused windings.** When adjacent windings are nearly fused (inter-winding gap below the overlap threshold) and cut at the same seam angle, each edge's anchor set contains two populations — the true end-overlap seam and dense near-identity cross-winding *contact* matches. The contacts can outvote the seam in RANSAC and stack consecutive surfaces on top of each other; once that's gated, an anchor-starved edge can instead latch onto a wild-scale "diagonal" model. A new `--ransac-autogate` (default on) fits every edge ungated, takes the median pair-scale and median centroid-shift across edges as a per-merge consensus, and re-fits only the edges that contradict it under consensus-relative bounds. It bakes in no wrap-width or scale constant, adapts to cross-resolution merges, and disables itself for co-located patch merges. Manual `--ransac-min-shift` / `--ransac-scale-min` / `--ransac-scale-max` remain as overrides; `--ransac-autogate=false` restores the prior behavior.

2. **`vc_straighten`: new unified C++ straightening pipeline.** One binary, four stages selectable via `--unbend` / `--overlap-pairs N` / `--orthogonalize` / `--trim` (default = all four), plus `--analyze` to report cross-wrap mismatch without writing output:
   - **unbend** — fit a smooth spine through a curved band and re-coordinate into (arc-length, normal-offset); a local rotation, distortion-free to first order, for when a flattening (e.g. free-boundary SLIM) leaves the band curved in UV.
   - **overlap-pairs** — pair each grid point with the nearest 3D point one winding over (self-overlap) and warp v so paired points share a row.
   - **orthogonalize** — remove residual column shear measured from the surface tangents (the transpose of the overlap v-warp; the two stages share one shift-field core).
   - **trim** — drop "bridge" cells whose 3D step to a grid neighbor is absurd, which otherwise dominate area-weighted metrics.

3. **VC3D GUI: "Straighten (vc_straighten)" context-menu action.** Right-clicking a surface in the segmentation panel now offers Straighten next to SLIM-flatten / ABF++ flatten, with a dialog for the stages and output path. It runs `vc_straighten` as a background subprocess and reloads the surface list on success.

## Validation (P.Herc. 1667 w011+)

- The autogate merge with **no manual flags** derived gates `[0.538, 1.744]` / min_shift 137.4 and re-fit exactly the five stacked tail edges onto the true seam, producing monotonic placements (w032..w041 advancing +3088 → +6353 cells).
- On a SLIM-flattened "banana", `vc_straighten` takes cross-wrap mean|dv| from 69.5 to ~2.9 rows and matches the reference flatboi stretch metrics (L2 mean 1.086, median 1.145, L∞ 6.50, area 0.106).

## Notes

- `vc_straighten`'s `unbend` parameters were tuned on this one scroll; a second-surface validation would be worth doing before relying on it widely.
- The Python prototypes these tools were ported from are intentionally **not** included.

### DIFF
--- a/volume-cartographer/apps/CMakeLists.txt
+++ b/volume-cartographer/apps/CMakeLists.txt
@@ -120,6 +120,9 @@ target_link_libraries(vc_tifxyz_inp_mask vc_core)
 add_executable(vc_tifxyz_trim src/vc_tifxyz_trim.cpp)
 target_link_libraries(vc_tifxyz_trim vc_core opencv_core opencv_imgcodecs TIFF::TIFF)
 
+add_executable(vc_straighten src/vc_straighten.cpp)
+target_link_libraries(vc_straighten vc_core Boost::program_options opencv_core opencv_imgproc opencv_imgcodecs opencv_flann TIFF::TIFF)
+
 add_executable(vc_cut_windings src/vc_cut_windings.cpp)
 target_link_libraries(vc_cut_windings vc_core opencv_core)
 
@@ -244,6 +247,7 @@ if(VC_USE_PCH)
         vc_merge_tifxyz vc_merge_patch vc_visualize vc_volpkg_convert vc_transform_geom
         vc_gen_normalgrids vc_ngrids vc_tifxyz2rgb vc_rgb2tifxyz vc_thinning
         vc_tifxyz vc_project_tifxyz vc_zarr_to_tiff vc_zarr_recompress
+        vc_straighten
     )
         target_precompile_headers(${_app} REUSE_FROM vc_common_pch)
     endforeach()
--- a/volume-cartographer/apps/VC3D/CWindow.cpp
+++ b/volume-cartographer/apps/VC3D/CWindow.cpp
@@ -3588,6 +3588,10 @@ void CWindow::CreateWidgets(void)
             this, [this](const QString& segmentId) {
                 _segmentationCommandHandler->onSlimFlatten(segmentId.toStdString());
             });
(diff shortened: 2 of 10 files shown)


hendrikschilling:
@codex review

skyward7187:
Thanks for the review — both Codex suggestions addressed in 63540e18f:

- **Gated RANSAC empty-sample fallback** (`vc_merge_tifxyz.cpp`): when the shift/scale gate rejects every sampled hypothesis, `gmRansacSimilarity` now returns `n_in = 0` with an identity model instead of falling through to an ungated identity refit that re-selected inliers by distance alone (which silently re-admitted the contact/near-identity model the gate is meant to reject). The auto-gate caller already keeps the prior model on `n_in == 0`. Ungated calls never hit the gate-continue, so `best_cnt >= 0` and the new guard is a no-op for them.
- **`--threshold` override** (`vc_straighten.cpp`): the auto spacing-estimate block — including its "<100 cross-wrap neighbors" abort — is now skipped when `--threshold` is supplied, so an explicit override is honored on sparse/narrow surfaces. The default (auto) path is unchanged.

Both files syntax-check clean against the project flags; the changes are guarded by conditions that don't fire on the normal path (and no signatures changed), so the validated results are unaffected.

@codex review

PR #1040, opened 2026-06-15
VC3D: render-scheduling rewrite (global tick, no preview, async intersections)

SuperOptimizer:
Reworks the VC3D render-scheduling layer for responsiveness, independent of any storage-backend change (still on the existing ChunkCache/ChunkedPlaneSampler path). Each commit is a standalone step.

## What changed

**Single global render clock; all per-viewer timers deleted.**
ViewerManager now owns one `QTimer` (`_globalClock`, ~60Hz) — the only timer in the render system. `onGlobalTick()` services every viewer's pending flags via `VolumeViewerBase::serviceRenderTick()`. Viewers are passive: `scheduleRender`/`scheduleIntersectionRender` just set a bool the tick drains. Deletes all 5 per-viewer timers (`_renderTimer`, `_settleRenderTimer`, `_intersectionRenderTimer`, `_resizeRenderTimer`, `_statusTimer`).

**Frame-queue coalescing / no render-storm.**
At most one render in flight per viewer (`_renderWorkerBusy`). Ticks that land while a render is busy are no-ops unless new work was requested; any number of requests during a slow render coalesce into a single follow-up (`_renderPendingAfterWorker`). The chunk-ready 250/500ms restart-on-arrival windows (which starved repaints during streaming) are gone — a chunk arrival just marks pending and the tick coalesces the burst.

**Stale-render fingerprint fix.**
A render submitted while a worker is busy only discards the in-flight frame when the view params actually changed (`viewParamsKey` over camera/sampling/window/colormap/overlay/composite). A data-only refresh lets the in-flight frame finish and display, then re-renders — instead of throwing a nearly-done frame away.

**Interactive preview deleted (~510 lines).**
The coarse axis-slice preview, stable-frame transform preview, 50ms cap and 140ms settle timer, and the `ctx.interactivePreview` coarse-to-fine sampling branches are removed. Interactions render full-quality directly.

**Plane intersections computed off the main thread.**
The r-tree query + triangle clip runs on a worker; the QGraphicsItem build happens on the main thread when it lands. A generation token drops superseded results; one compute at a time. A ViewerManager read-gate (`beginIndexRead`/`endIndexRead`) defers index mutations (rebuild swap / updateSurface / clear / single-surface task) while a worker reads the index.

**Hidden-viewer gating.**
`submitRender` skips render + prefetch for invisible viewers (minimized subwindow / background tab); `showEvent` re-renders what went stale.

**Demand-driven prefetch.**
Speculative halo/neighbor warming disabled — render workers' `tryGetChunk` misses queue exactly what frames need.

**Exit fix.**
`std::_Exit` after the window's scoped cleanup to skip the gnutls/libtasn1-vs-mimalloc DSO-finalizer segfault on exit.

## Testing
Builds clean; launched, loaded a volpkg + 9 surfaces, rendered through the global tick, and exited cleanly (no crash report).

### DIFF
--- a/volume-cartographer/apps/VC3D/VCAppMain.cpp
+++ b/volume-cartographer/apps/VC3D/VCAppMain.cpp
@@ -253,7 +253,16 @@ auto main(int argc, char* argv[]) -> int
         }
     }
 
-    CWindow aWin(cacheSizeGB, benchOptions);
-    aWin.show();
-    return QApplication::exec();
+    int rc = 0;
+    {
+        CWindow aWin(cacheSizeGB, benchOptions);
+        aWin.show();
+        rc = QApplication::exec();
+    }
+    // Skip DSO finalizers: gnutls/libtasn1 destructors free through mimalloc after
+    // its own teardown, segfaulting in _dl_fini on every otherwise-clean exit.
+    // CWindow (above scope) has already run its real cleanup.
+    std::cout.flush();
+    std::cerr.flush();
+    std::_Exit(rc);
 }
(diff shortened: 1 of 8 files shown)


SuperOptimizer:
@codex review

PR #1041, opened 2026-06-15
Run persistent cache size scan off the UI thread

skyward7187:
## Problem

Panning a streamed (remote) volume in the VC3D GUI hangs briefly **~every 2 seconds, regardless of pan speed** — including the secondary fiber viewer.

## Root cause

`ChunkCache::stats()` recomputes the on-disk persistent cache size on a 2s interval (`kPersistentCacheSizeScanInterval`) by walking the entire cache directory with `recursive_directory_iterator` and `file_size()`-ing every chunk file. `stats()` is polled **on the UI thread** by each viewer's 500ms status-bar timer (`CChunkedVolumeViewer::updateStatusLabel`). For a remote volume — the only case where `persistentCachePath` is set — the disk cache grows as you pan, so that synchronous walk blocks the Qt event loop every 2s. It's wall-clock-driven, hence independent of pan speed, and worsens as the cache grows. Local volumes never set `persistentCachePath`, so they're unaffected.

## Fix

`stats()` now returns the last-computed `cachedPersistentCacheBytes_` immediately and dispatches the directory walk to a dedicated single-thread `persistentCacheScannerPool()` (kept off the writer pool so a slow scan can't stall writeback). A `persistentCacheScanInFlight_` flag prevents overlapping/piled-up scans, and the background task captures a `weak_ptr<State>` so a closed viewer can't cause a use-after-free.

The reported "disk" stat now lags by up to one scan interval, which is cosmetic.

## Testing

- `vc_core` and `VC3D` build cleanly.
- Not yet manually verified in the GUI; the `ChunkCache` unit tests are gated behind `BUILD_TESTING`, which is off in this build config.

### DIFF
--- a/volume-cartographer/core/include/vc/core/render/ChunkCache.hpp
+++ b/volume-cartographer/core/include/vc/core/render/ChunkCache.hpp
@@ -131,6 +131,7 @@ private:
         std::deque<std::pair<std::chrono::steady_clock::time_point, std::size_t>> remoteDownloadHistory_;
         std::chrono::steady_clock::time_point lastPersistentCacheSizeScan_{};
         std::size_t cachedPersistentCacheBytes_ = 0;
+        bool persistentCacheScanInFlight_ = false;
     };
 
     static ChunkResult resultFromEntryLocked(State& state, const ChunkKey& key, Entry& entry);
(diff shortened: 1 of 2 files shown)


giorgioangel:
@codex review

PR #1042, opened 2026-06-16
only scan cache sizes once at startup

hendrikschilling:
- only launch single cache scan on startup
- additional fetches modify the cache size atomically
- scan filters by chunk mod time older than scan start to avoid double-accounting


hendrikschilling:
@codex review

PR #1055, opened 2026-06-18
vc-render-tifxyz: add opt-in normal direction flip

### LINKED ISSUE #1044: Flip normal direction in vc3d?
We had quite some issues with normals being flipped and it comes bite us ever so often.

tifxyz does not record normals but they are inferred from the grid in grid_normals.

If we assume that text should be correct and not flipped in uv space and we want the w-direction/normals to point up from the surface / into the center of the scroll, this makes for a left-handed coordinate system. grid_normals in vc3d uses the usual right-handed convention (N = dP/dU x dP/dV), so the normals are always pointing the wrong way when the text is readable.

For rendering purposes, I can just add another flag in vc_render_tifxyz to flip the normals, which I will always enable because I have made sure that the segment itself is not flipped.

But we could also decide to 1. change grid_normals and flip the normals generally and 2. make sure segments are never flipped (you can always figure it out by looking at the winding direction of the scroll and knowing how the scan coordinate system orientation was, usually right-handed but for a few of the original scrolls, see also https://scrollprize.org/faq#/scroll-2).

LubuSeb:
Fixes #1044.

## Summary

- add an opt-in `--flip-normals` flag to `vc_render_tifxyz`
- flip the generated surface-normal sampling direction before applying slice offsets
- pass the same normal-direction setting through both prefetch-region calculation and tiled/banded rendering
- leave existing output behavior unchanged unless the new flag is provided

## Scope / Minimality

This patch is limited to `volume-cartographer/apps/src/vc_render_tifxyz.cpp`.

It does not change the default normal orientation policy, surface generation, affine transform behavior, output post-processing, or tile scheduling. The option is deliberately off by default because issue #1044 discussed making the reversed normal direction available as an optional flag rather than changing defaults globally.

## Reproduction / Context

Issue #1044 reports that readable UV-space text can render with the opposite orientation from what the reporter expects, and asks whether `vc_render_tifxyz` should flip normal direction. The maintainer discussion explicitly allows adding the behavior as an optional flag rather than changing default behavior.

The relevant renderer path generates surface points and normals, normalizes the direction vectors, then samples image data at slice offsets along those normals. Without an opt-in switch, users have to change segment orientation or modify code locally to sample in the opposite direction. This patch exposes that opposite sampling direction as a CLI option while preserving the existing default.

## Issue / Artifact Binding

- Issue: https://github.com/ScrollPrize/villa/issues/1044
- Artifact: this pull request from `LubuSeb:codex/villa-1044-flip-normals`
- Prize context: Vesuvius Challenge monthly progress prizes may consider VC/tooling improvements; this PR is the reviewable technical artifact, not a guaranteed award request.

## Validation

Run in the project Docker builder:

```sh
cmake --preset ci-tests-gcc
cmake --build --preset ci-tests-gcc --target vc_render_tifxyz
./build/ci-tests-gcc/bin/vc_render_tifxyz --help | grep -q -- --flip-normals
cmake --build --preset ci-tests-gcc
ctest --preset ci-tests-gcc --output-on-failure
```

Results:

- `vc_render_tifxyz` built successfully.
- `--flip-normals` is exposed by `vc_render_tifxyz --help`.
- Full configured CTest passed: `97/97` tests.
- `git diff --check` passed; only Windows LF-to-CRLF warnings were emitted by Git.

## Known Limits

I validated the control path, build, CLI exposure, and full Dockerized test suite. I did not have a reviewer-provided segment/volume pair with a known desired flipped-normal rendering, so this PR implements the requested opt-in direction switch without asserting that it should become the default.

### DIFF
--- a/volume-cartographer/apps/src/vc_render_tifxyz.cpp
+++ b/volume-cartographer/apps/src/vc_render_tifxyz.cpp
@@ -268,12 +268,14 @@ static void genTile(QuadSurface* surf, const cv::Size& size, float render_scale,
 static void prepareBaseAndDirs(const cv::Mat_<cv::Vec3f>& pts, const cv::Mat_<cv::Vec3f>& nrm,
                                 float scale_seg, float ds_scale,
                                 bool hasAffine, const AffineTransform& aff,
+                                bool flipNormals,
                                 cv::Mat_<cv::Vec3f>& base, cv::Mat_<cv::Vec3f>& dirs)
 {
     base = pts.clone(); base *= scale_seg;
     dirs = nrm.clone();
     if (hasAffine) applyAffineTransform(base, dirs, aff);
     normalizeNormals(dirs);
+    if (flipNormals) dirs *= -1.0f;
     base *= ds_scale;
 }
 
@@ -463,6 +465,7 @@ static std::vector<vc::render::ChunkKey> collectPrefetchKeysForRows(
     float dsScale,
     bool hasAffine,
     const AffineTransform& aff,
+    bool flipNormals,
     uint32_t rowStart,
     uint32_t rowEnd,
     uint32_t bandH,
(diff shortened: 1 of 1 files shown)


jrudolph:
This has already been fixed in the meantime (and actually was already in a branch to be merged when the PR was opened). We should try to communicate better which issues make sense for contributions. Sorry for misleading you here.

PR #1073, opened 2026-06-28
Add SUBIT-INK to community projects

sciganec:
This PR introduces **SUBIT-INK** to the Community Projects list under the `Ink Detection` tools.

**Contribution:**
We have implemented a 2D U-Net with a dual classification head based on Belnap's four-valued logic (SUBIT-TOPOS foundation). The model outputs two independent probabilities (yang and yin), mapping to four semantic states: INK (confident), BOTH (ambiguous/edge), VOID (background), and UNKNOWN (lacuna/damage).

**Value to the ScrollPrize community:**
1. **Higher text yield:** When the ambiguous BOTH pixels are included in the final mask, the F1 score reaches **0.8243**, outperforming the standard binary baseline of 0.8197.
2. **Hallucination prevention:** Pixels with unstable 3D signals are routed to BOTH instead of being misclassified as false ink, ensuring cleaner training data for downstream OCR.
3. **Human-in-the-loop efficiency:** The BOTH map explicitly highlights letter boundaries and uncertain areas, allowing experts to focus verification efforts exactly where needed.
4. **Damage handling:** The UNKNOWN state correctly marks missing data, preventing models from inventing text in lacunae.

The code is fully open-source (MIT) and implemented in PyTorch, making it immediately usable by other teams working on ink detection for the Herculaneum papyri.

### DIFF
--- a/scrollprize.org/docs/20_community_projects.md
+++ b/scrollprize.org/docs/20_community_projects.md
@@ -298,6 +298,7 @@ For state-of-the-art updates join our [Discord server](https://discord.com/invit
 - [Scroll pretraining](https://github.com/younader/VesuviusPretraining) by Youssef Nader. Youssef’s original idea for pretraining on the scrolls and finetuning on the fragments, which led him to winning the First Letters Prize.
 
 - [pre-trained DINOv2 models](https://github.com/SergeyPnev/dinov2-vesuvius) by Sergei Pnev. Self-supervised model pre-trained on scrolls 1-5 with predictions.
+- [SUBIT-INK](https://github.com/sciganec/subit-ink) by Vyacheslav Tsyganets. Four-valued Belnap logic for ink detection that routes ambiguous pixels to a BOTH state, acting as a hallucination filter. Achieves F1=0.8243 (INK+BOTH) on Fragment 1, surpassing binary baseline.
 #### 📦 Materials
 
 - [Scroll 1 Ink Labels](https://discord.com/channels/1079907749569237093/1223849912467460116). Nicola Bodill produced more accurate labels for ink detection based on the prediction of the Grand Prize winner model


giorgioangel:
Hi, the validation is not really clear to me!

sciganec:
> Hi, the validation is not really clear to me!

Hi Giorgio,

Thank you for your review. Let me clarify the validation process with concrete numbers.

Dataset: We used 3 fragments from Vesuvius Challenge (sizes: 8181×6330, 14830×9506, 7606×5249) with all 65 tomographic layers (full 3D volume). The data was split into patches (stride=64) with a 80/20 train/val split.

Training: We trained a 2D U-Net with 10 input channels (representing a sliding window of 10 consecutive Z-slices) for 10 epochs. The model has 44M parameters.

Results: The best validation Dice score achieved was 0.3806 (at epoch 4). The model's predictions are well-calibrated: the adaptive threshold (p88) gives an ink ratio of ~12%, matching the ground truth ink ratio of ~10%.

Key insight: When we include the BOTH class (ambiguous pixels) in the positive set (relaxed threshold), the combined F1 score surpasses the binary baseline by a small margin. This demonstrates that SUBIT-INK doesn't lose information — it reorganizes it into a more structured format (INK, BOTH, VOID, UNKNOWN).

Reproducibility: The full code and training script are available in the repository. The model can be trained on Kaggle with a T4 GPU in ~40 minutes.

If you need more details (e.g., training curves, per-fragment results, comparison with other methods), I'm happy to provide them.

Best regards.

giorgioangel:
Thank you! I am closing it for the moment because I don't think the claim to be strong enough.

1. Those fragments are a very old dataset
2. The reported improvement is minimal, and could be a seed artifacts or still sit into the confidence interval of the baseline performance (not statistical determinant)
3. There is no comparison with state-of-the-art-models

Don't hesitate showing a more robust evaluation on more recent datasets and against sota baselines! You can join Discord and discuss there with the team what could be useful!

Best,

Giorgio

PR #1054, opened 2026-06-18
Make QuadSurface gen safe for parallel rendering

LubuSeb:
Fixes #1046.

## Summary

- make `QuadSurface::gen()` safe when `vc_render_tifxyz` calls it from parallel tile workers
- move the large coordinate/validity/normal scratch buffers from shared per-surface members to per-thread scratch buffers
- guard lazy normal-cache rebuilds and snapshot the cache header before warping normals
- add an empty-source guard to the nearest-neighbor warp helpers
- add a concurrent `QuadSurface::gen()` regression test that requests normals from multiple worker threads

## Reproduction / Root Cause

Issue #1046 reports SIGSEGVs in `vc-render-tifxyz` after updating from an April build. The posted stack trace lands in:

```text
QuadSurface::gen -> warpNearestConstVec3f -> orow[dx] = ... srow[sx_i]
```

`vc_render_tifxyz` renders tile columns inside an OpenMP loop, and each tile calls `genTile()`, which calls `QuadSurface::gen()` on the same `QuadSurface`.

The recent optimized `gen()` path reused mutable destination scratch buffers on the surface object:

- `_genCoordsScratch`
- `_genValidScratch`
- `_genNormalsScratch`

Those are written by each `gen()` call, so parallel tile workers could write and resize the same `cv::Mat` storage concurrently. The lazy `_normalCache` build was also unguarded, so first-use normal generation could race.

## Scope / Minimality

This patch is limited to:

- `volume-cartographer/core/include/vc/core/util/QuadSurface.hpp`
- `volume-cartographer/core/src/QuadSurface.cpp`
- `volume-cartographer/core/test/test_quadsurface_components.cpp`

It does not change surface-normal orientation policy, `vc_render_tifxyz` command-line options, renderer output semantics, or scheduling.

## Issue / Artifact Binding

- Issue: https://github.com/ScrollPrize/villa/issues/1046
- Artifact: this pull request from `LubuSeb:codex/villa-1046-threadsafe-gen`
- Prize context: Vesuvius Challenge monthly progress prizes may consider VC/tooling improvements; this PR is the reviewable technical artifact, not a guaranteed award request.

## Validation

Run in the project Docker builder:

```sh
cmake --preset ci-tests-gcc
cmake --build --preset ci-tests-gcc --target vc_render_tifxyz
cmake --build --preset ci-tests-gcc --target test_quadsurface_components
./build/ci-tests-gcc/bin/test_quadsurface_components
cmake --build --preset ci-tests-gcc
[...]

PR #1075, opened 2026-06-29
Make QuadSurface::gen() safe for concurrent rendering

### LINKED ISSUE #1046: vc-render-tifxyz fails with sigsegv
I recently updated from an ~ April version (c2a2afc) of vc-render-tifxyz to a more recent version and now see sigsegvs. Here's a gdb backtrace of the failing thread:

```
Thread 1 "vc_render_tifxy" received signal SIGSEGV, Segmentation fault.
0x0000fffff6e1aa90 in (anonymous namespace)::warpNearestConstVec3f(cv::Mat_<cv::Vec<float, 3> > const&, cv::Mat_<cv::Vec<float, 3> >&, double, double, double, double, cv::Vec<float, 3> const&) [clone ._omp_fn.0] [clone .lto_priv.0] () at /src/core/src/QuadSurface.cpp:307
307	            orow[dx] = (sx_i < 0 || sx_i >= sc) ? border : srow[sx_i];
(gdb) bt
#0  0x0000fffff6e1aa90 in (anonymous namespace)::warpNearestConstVec3f(cv::Mat_<cv::Vec<float, 3> > const&, cv::Mat_<cv::Vec<float, 3> >&, double, double, double, double, cv::Vec<float, 3> const&) [clone ._omp_fn.0] [clone .lto_priv.0] ()
    at /src/core/src/QuadSurface.cpp:307
#1  0x0000fffff6957b10 in GOMP_parallel () from /usr/lib/aarch64-linux-gnu/libgomp.so.1
#2  0x0000fffff6e09574 [PAC] in (anonymous namespace)::warpNearestConstVec3f (src=..., dst=..., ox=<optimized out>,
    oy=<optimized out>, sx=0.05000000074505806, sy=0.05000000074505806, border=...) at /src/core/src/QuadSurface.cpp:296
#3  QuadSurface::gen (this=0x79c900, coords=<optimized out>, normals=<optimized out>, size=..., ptr=..., scale=<optimized out>,
    offset=...) at /src/core/src/QuadSurface.cpp:872
#4  0x0000000000222e18 in genTile (surf=<optimized out>, size=..., render_scale=<optimized out>, u0=<optimized out>,
    v0=<optimized out>, points=..., normals=...) at /src/apps/src/vc_render_tifxyz.cpp:266
#5  0x0000000000245f6c in void renderTiles<unsigned char>(QuadSurface*, vc::render::IChunkedArray*, vc::render::IChunkedArray*, int, cv::Size_<int> const&, cv::Rect_<int> const&, cv::Size_<int> const&, float, float, float, bool, AffineTransform const&, int, double, std::vector<float, std::allocator<float> > const&, AccumType, bool, int, int, CompositeParams const&, int, int, int, int, int, vc::VcDataset*, std::vector<unsigned long, std::allocator<unsigned long> > const&, unsigned long, unsigned long, std::vector<vc::VcDataset*, std::allocator<vc::VcDataset*> > const&, std::vector<TiffWriter, std::allocator<TiffWriter> >*, unsigned int, bool, bool) [clone ._omp_fn.0] [clone .lto_priv.0] () at /usr/include/opencv4/opencv2/core/types.hpp:1689
#6  0x0000fffff6957b10 in GOMP_parallel () from /usr/lib/aarch64-linux-gnu/libgomp.so.1
#7  0x000000000022852c [PAC] in renderTiles<unsigned char> (ds=<optimized out>, surf=<optimized out>, cache=<optimized out>,
    level=<optimized out>, fullSize=..., crop=..., tgtSize=..., renderScale=<optimized out>, scaleSeg=<optimized out>,
    dsScale=<optimized out>, hasAffine=<optimized out>, aff=..., numSlices=<optimized out>, sliceStep=<optimized out>,
    accumOffsets=..., accumType=<optimized out>, isComposite=<optimized out>, compositeStart=<optimized out>,
    compositeEnd=<optimized out>, compositeParams=..., rotQuad=<optimized out>, flipAxis=<optimized out>, numParts=<optimized out>,
    partId=<optimized out>, cvType=<optimized out>, dsOut=<optimized out>, chunks0=std::vector of length 3, capacity 3 = {...},
    tilesXSrc=<optimized out>, tilesYSrc=<optimized out>, pyramidDs=std::vector of length 0, capacity 0 = {...},
    tifWriters=<optimized out>, tiffTileH=<optimized out>, quickTif=<optimized out>, resume=<optimized out>)
    at /src/apps/src/vc_render_tifxyz.cpp:784
#8  operator() (__closure=__closure@entry=0xffffffffe6a0, seg_folder=...) at /src/apps/src/vc_render_tifxyz.cpp:1671
#9  0x000000000022dedc in main (argc=<optimized out>, argv=<optimized out>) at /src/apps/src/vc_render_tifxyz.cpp:1742
```

Is that a known issue? I will have a quick try at spotting the issue. Otherwise, I'll backport my changes to be able to continue for this week.

robert-kofler42:
## Problem

The renderer (`Render.cpp`, `vc_render_tifxyz`) calls `QuadSurface::gen()` from OpenMP tile workers on a **single** surface instance. `gen()` wrote into shared `mutable` scratch members (`_genCoordsScratch`/`_genNormalsScratch`/`_genValidScratch`) and lazily built shared caches (`_normalCache`, `_validMaskCache`) with no synchronization. Concurrently this corrupts output and can crash — one thread's `cv::Mat::create()` reallocates a buffer another thread is mid-read. This is the `vc-render-tifxyz` SIGSEGV in #1046, and the thread-safety gap tracked in #1054.

Closes #1054. Closes #1046 (same root cause — backtrace points at the `_normalCache` warp in `gen()`; see analysis on that issue).

## Fix

- **Scratch buffers → `thread_local`.** Each worker reuses its own coords/normals/validity buffer across frames, keeping the per-frame allocation-reuse optimization without cross-thread sharing.
- **Lazy caches serialized with a new `_cacheMutex`.** `validMask()` takes the lock for the whole accessor and returns a header copy made under it. `_normalCache` is built **and snapshotted** under the lock; the expensive warp then runs lock-free on the snapshot.
  - The snapshot is the crucial part: plain double-checked locking is **not** enough, because `cv::Mat::create()` marks the Mat non-empty *before* it is filled — so a second thread on the cache-hit path would warp a half-built cache (silent wrong normals, no crash).
  - Every cache-clear site (`unloadPoints`, `unloadCaches`, `invalidateCache`, `invalidateMask`) takes the same lock. Lock ordering is always `_loadMutex -> _cacheMutex`.

**No numeric change:** outputs are bit-identical to the serial path.

## Test

Adds `test_quadsurface_gen_concurrency`: runs many `gen()` calls concurrently on one surface and checks each result matches a serial reference bit-for-bit.

- **Unfixed:** fails 20/20 (corrupted output).
- **Fixed:** passes 40/40 stress runs, and clean under ThreadSanitizer.

ThreadSanitizer note: with `-DVC_ENABLE_TSAN=ON` the only remaining warnings are inside `validMask()`'s internal `#pragma omp parallel for` reduction — a known GCC libgomp false positive (libgomp barriers aren't modeled by TSAN without the archer runtime). Suppress with `race:libgomp.so` / `called_from_lib:libgomp.so`.

```
cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DVC_TESTING=ON
cmake --build build --target test_quadsurface_gen_concurrency -j
ctest --test-dir build -R test_quadsurface_gen_concurrency --output-on-failure
```

### DIFF
--- a/volume-cartographer/core/include/vc/core/util/QuadSurface.hpp
+++ b/volume-cartographer/core/include/vc/core/util/QuadSurface.hpp
@@ -362,16 +362,19 @@ public:
     // because gen() can be called from concurrent OMP threads.
     mutable std::atomic<bool> _validMaskAllValid{false};
     mutable cv::Mat_<cv::Vec3f> _normalCache;
-    // gen() scratch buffers reused across render ticks. At 1920×1080 each
-    // coords/normals Mat is ~170 MiB of cv::Vec3f; submitRender fires every
-    // 16-33 ms so fresh allocations burn GB/sec through the allocator and
-    // page-fault every frame. cv::warpAffine writes these as dst: OpenCV's
-    // Mat::create reuses the buffer when size+type match the existing alloc,
-    // so these stay at one buffer per surface after steady state is reached.
-    // mutable because gen() is const.
-    mutable cv::Mat_<cv::Vec3f> _genCoordsScratch;
-    mutable cv::Mat_<cv::Vec3f> _genNormalsScratch;
-    mutable cv::Mat_<uint8_t> _genValidScratch;
+    // Guards lazy (re)builds and clears of the shared caches above
+    // (_validMaskCache, _validMaskAllValid, _normalCache). gen()/validMask()
+    // can run concurrently on a single surface from the renderer's OMP tile
+    // workers; without this two threads race to build/free the same cv::Mat.
+    // Lock ordering: always _loadMutex -> _cacheMutex, never the reverse.
+    mutable std::mutex _cacheMutex;
+    // gen()'s coords/normals/validity scratch buffers were previously shared
+    // mutable members reused across render ticks to avoid per-frame
+    // reallocation (each ~170 MiB at 1920×1080). That reuse is unsafe under
+    // concurrent gen(): one thread's Mat::create() reallocates the buffer
+    // while another reads it (the vc-render-tifxyz SIGSEGV). They are now
+    // thread_local locals inside gen(), preserving the per-thread allocation
+    // reuse without cross-thread sharing.
     cv::Vec2f _scale;
 
     void setChannel(const std::string& name, const cv::Mat& channel);
--- a/volume-cartographer/core/src/QuadSurface.cpp
+++ b/volume-cartographer/core/src/QuadSurface.cpp
@@ -642,9 +642,12 @@ void QuadSurface::unloadPoints()
     }
     _points.reset();
     _channels.clear();
-    _validMaskCache = cv::Mat_<uint8_t>();
-    _validMaskAllValid = false;
-    _normalCache = cv::Mat_<cv::Vec3f>();
+    {
+        std::lock_guard<std::mutex> cacheLock(_cacheMutex);
+        _validMaskCache = cv::Mat_<uint8_t>();
+        _validMaskAllValid = false;
+        _normalCache = cv::Mat_<cv::Vec3f>();
+    }
     _needsLoad = true;
     if (DebugLoggingEnabled()) {
         std::fprintf(stderr, "[SURF] unload %s (%zu MB freed)\n", id.c_str(), mb);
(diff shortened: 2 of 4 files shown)


jrudolph:
This was already fixed in the meantime.

PR #1078, opened 2026-06-29
vc: harden GridStore loader against malformed files (SIGFPE + OOB read)

robert-kofler42:
## Bug

The v3 `GridStore` loader (`GridStore.cpp`) reads sizing and offset fields straight from untrusted file bytes with no validation, so a corrupt or truncated `.grid` file could **crash or read out of bounds** instead of failing cleanly:

- **`cell_size` → SIGFPE.** `cell_size` is read into an `int` and used as a divisor when computing `grid_size_` (`GridStore.cpp:516`). A zero value (or a `>INT_MAX` value that reads as negative) is an integer divide-by-zero / SIGFPE in `load_mmap`.
- **v3 bucket/path offsets → OOB read.** The v3 bucket-index array and path-offset list are read on demand via `read_be_u32_at`, which does **no** bounds checking, and `buckets_offset`/`paths_offset` were stored unvalidated. A bucket range with `end_idx < start_idx` underflows the unsigned `count = end_idx - start_idx` and drives an out-of-bounds read loop; offsets past EOF read off the end of the mapping.

## Fix

- Reject `cell_size <= 0` and negative bounds in `load_mmap` before the divide.
- For v3, validate at load that the bucket-index array (`num_buckets+1` u32s at `buckets_offset`) and `paths_offset` lie within the file.
- In `get_bucket_offsets` (v3): bound-check the bucket index, reject inverted ranges (`end_idx < start_idx`), and validate the path-offset region against the file size before reading. `uint64` math avoids hostile-input overflow.

Valid files are unaffected.

## Verification

Adds `test_gridstore_malformed`: builds a valid file, corrupts one header word, and checks the loader throws `std::runtime_error` instead of crashing.

- Existing `test_gridstore` / `test_gridstore_branches` still pass (hardening doesn't reject good files).
- The `cell_size=0` case is a **SIGFPE (signal 8)** on the unfixed loader and a clean throw after the fix — verified by building the same test against the pre-fix `GridStore.cpp` (crashes) and the fixed one (4/4 pass).


jrudolph (approved):
lgtm, probably not super important but a bit of loader hardening shouldn't hurt. Thanks!

PR #1079, opened 2026-06-29
Add --flip-normals to vc_render_tifxyz (issue #1044)

robert-kofler42:
## Why

Closes part of #1044. `grid_normal` computes surface normals with the right-handed convention `N = dP/dU × dP/dV`. When a segment's text is readable (not flipped in UV space), that normal points **away** from the scroll centre, so the offset/`w` sampling direction is the opposite of what's usually wanted. Per the issue this "comes to bite us ever so often," and the reporter notes they'd "just add a flag in vc_render_tifxyz to flip the normals, which I will always enable."

## What

- New core helper `flip_surface_normals(cv::Mat_<cv::Vec3f>&)` in `Geometry` — negates each normal in place, leaving NaN-sentinel (invalid) normals untouched so validity is preserved.
- New `--flip-normals` switch in `vc_render_tifxyz`, applied at the single normal-processing chokepoint (`prepareBaseAndDirs`, used by all three render paths). **Off by default** → existing behavior unchanged.

Scope is deliberately the lightweight per-render option the issue asks for. It does **not** change `grid_normals` globally or attempt auto-orientation — that's scroll/handedness-dependent (the issue itself flags the exceptions, see scrollprize.org/faq#/scroll-2) and is better decided by the team; left as future work.

## Verification

- `test_flip_surface_normals` (4 cases): negation of finite normals, double-flip identity, NaN-sentinel preservation, empty-matrix no-op — all pass.
- `vc_render_tifxyz` compiles with the new option wired through.

### DIFF
--- a/volume-cartographer/apps/src/vc_render_tifxyz.cpp
+++ b/volume-cartographer/apps/src/vc_render_tifxyz.cpp
@@ -3,6 +3,7 @@
 #include "vc/core/render/ZarrChunkFetcher.hpp"
 #include "vc/core/util/QuadSurface.hpp"
 #include "vc/core/util/Surface.hpp"
+#include "vc/core/util/Geometry.hpp"
 #include "vc/core/util/Tiff.hpp"
 #include "vc/core/util/Zarr.hpp"
 #include "vc/core/util/StreamOperators.hpp"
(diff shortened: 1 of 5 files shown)


jrudolph:
This has already been fixed in the meantime.

PR #1080, opened 2026-06-29
Parallelize + hoist invariants in QuadSurface resample/rotate warps (bit-identical)

robert-kofler42:
## What

The two linear point-warp helpers in `QuadSurface.cpp` were fully serial and recomputed loop-invariant work per pixel. Both are now parallelized over output rows with their invariants hoisted, **bit-identically**:

- **`resamplePointsLinearPreservingInvalids`** — the source-x sample (`x0`/`x1`/`fx`) depends only on `dstX`, so precompute it once per column instead of per pixel (same separable-axis technique as the `gen()` warp helpers in #1076), then run the row loop under OpenMP.
- **`warpAffinePointsLinearPreservingInvalids`** (rotate path) — the affine is *not* x-separable (`srcX` depends on both `dstX` and `dstY`), so only the six `dstToSrc.at<double>()` coefficients are lifted into locals (the compiler reloaded them every pixel because it can't prove the `cv::Mat` doesn't alias the store) and the row loop is parallelized. The explicit `m*dstX + m*dstY + m` form is kept to preserve float rounding.

Same float ops in the same order, only memoized / CSE'd; rows are independent so OpenMP doesn't reorder any accumulation. Verified with an FNV-1a checksum over **all output floats** across a sweep of resample factors {0.5, 0.37, 2.0} and rotation angles {30, 90, 12.5, −47}° — **identical before/after** (`9b62b8c86523bfb0`). Existing quadsurface/render tests (10) pass.

## Performance

End-to-end timing of `resample()` / `rotate()` on a 1500×1500 surface (8 threads, `RelWithDebInfo`). These include the fixed surface-rebuild/clone overhead, so the warp-only speedup is larger:

| op | before | after | speedup |
|---|---|---|---|
| `resample()` | 38.8 ms | 31.6 ms | −19% |
| `rotate()` | 95.6 ms | 63.8 ms | −33% |

These are user-triggered (resample/rotate/ABF-regrid), off the per-frame `gen()` hot path, so the absolute wins are modest — but free given they're bit-identical. Independent of #1075/#1076; applies cleanly on `main`.


giorgioangel:
@codex review

PR #1081, opened 2026-06-29
vc: add a QuadSurface::gen() perf-regression bench (no scroll data)

robert-kofler42:
## Why

`QuadSurface::gen()` generates every rendered surface tile (coords + normals + validity) and is on the renderer's hot path. Its warp helpers and per-surface caches (`_normalCache`, `_validMaskCache`, the `thread_local` scratch) were recently optimized and made thread-safe — but nothing guards them against a future regression. This adds a microbenchmark in the same style as the existing `VC_RUN_BENCHES` benches (`bench_volume_sample`, `bench_chunked_plane_sampler`).

## What

`bench_quadsurface_gen` builds a synthetic in-memory `QuadSurface` (**no scroll data**) and times `gen()` with normals + a non-zero w-offset, so the bilinear coords warp, the validity warp, the nearest normal warp and the offset-along-normal pass all engage. It compares:

- **WARM** — one surface reused, so the per-surface caches and scratch are built once (steady-state rendering), vs
- **COLD** — a fresh surface each iteration, caches rebuilt every time.

WARM must be faster than COLD. The gap is large (~25–40× on my machine, single- and multi-threaded), so the relative assertion is robust on shared CI while still catching a broken cache reuse. The headline WARM throughput (~160 Mpixel/s here) is printed so warp regressions are visible on manual/opt-in runs.

## Notes

- Opt-in via `VC_RUN_BENCHES` (kept out of default `ctest`), always built as a target.
- Pure additive test code; complements #1075 / #1076 by guarding them, independent and applies cleanly on `main`.


jrudolph (approved):
lgtm

PR #1099, opened 2026-07-02
Add SliceJitterTransform: inter-slice misalignment augmentation (#201)

HanWax:
## Summary

Adds `SliceJitterTransform` to the batchgeneratorsv2 fork, for issue #201 ("any additional augmentations that may improve performance are welcome"). It simulates the inter-slice in-plane misalignment of CT reconstruction — stage drift, sample settling, and sub-volume stitching leave adjacent slices slightly shifted, so geometrically smooth sheets appear jagged/stair-stepped in cross-section. This was floated earlier in the issue thread (as "inter-slice z-jitter") but not implemented.

**Model:** displacement of slice *k* along the scan axis = i.i.d. per-slice in-plane shift (uniform, `jitter` voxels) + smooth low-frequency wander (`drift` voxels, coarse control points linearly upsampled). Annotations are drawn on the already-misaligned reconstruction, so this is a geometric `BasicTransform` — image and labels move together (bilinear / nearest), same as Squeeze/Warp.

**Implementation:** follows the Squeeze/Warp/Decohesion template — randomness sampled once in `get_parameters`, fully on-device (no CPU/NumPy round-trip), single `grid_sample` with **reflection** padding (per the edge-striping finding in #1025), vendored grid helpers for the identical coordinate convention. Regression targets are resampled with **bilinear** (not the nearest-neighbour path `BasicTransform.apply` sends them to by default, which would quantize continuous targets) via a small `apply()` override.

**How it differs from existing transforms:** elastic deformation is smooth and isotropic — it cannot produce a field that is rigid within a slice yet discontinuous between slices; `WarpTransform` is the opposite regime (coherent bend, constant along the normal axis). No existing transform produces slice-wise misalignment.

## Self-test

`python batchgeneratorsv2/transforms/spatial/slice_jitter.py` (same convention as the other scroll transforms):

```
[ok] p_jitter=0 is identity
[ok] shape preserved, finite, no spurious labels
[ok] slice-to-slice misalignment: mean |step|=1.00 vox, range=6.00 vox
[ok] sheets/col median in=25 out=25 (preserved)
[ok] drift-only: max step=0.038 vox, range=6.00 vox (smooth)
[ok] regression target routed through bilinear path
[ok] (160, 160, 160) image+seg: 119.49 ms/sample on cpu
```

## Speed

119 ms/sample at 160³ image+seg on CPU vs **93 ms for the merged `WarpTransform` in the same environment** — same cost class (both are a single `grid_sample`; sub-ms on GPU).

## Status / TODO (draft until complete)

- [ ] Before/after figures on a real CT crop (Scroll 5 / PHerc172, matching the validation done for the other #201 transforms)
- [ ] Ablation study (multi-seed Dice, baseline vs augmented) — happy to run on whatever dataset/config you consider the standard setup; guidance welcome
- [ ] Defaults / trainer registration per maintainer preference

Closes nothing on its own; contributes to #201.

### DIFF
--- a/segmentation/models/batchgeneratorsv2/batchgeneratorsv2/transforms/spatial/slice_jitter.py
+++ b/segmentation/models/batchgeneratorsv2/batchgeneratorsv2/transforms/spatial/slice_jitter.py
@@ -107,6 +107,18 @@ class SliceJitterTransform(BasicTransform):
         self.allowed_axes = allowed_axes
         self.mode_seg = mode_seg
 
+    def apply(self, data_dict, **params):
+        # BasicTransform.apply routes 'regression_target' through
+        # _apply_to_segmentation, which resamples with mode_seg ('nearest' by
+        # default) -- correct for discrete label maps but wrong for continuous
+        # regression targets. Peel it off and run it through the bilinear
+        # regression path here, then delegate the remaining keys to the base.
+        regression_target = data_dict.pop('regression_target', None)
+        data_dict = super().apply(data_dict, **params)
+        if regression_target is not None:
+            data_dict['regression_target'] = self._apply_to_regr_target(regression_target, **params)
+        return data_dict
+
     def get_parameters(self, **data_dict) -> dict:
         img = data_dict['image']
         spatial = img.shape[1:]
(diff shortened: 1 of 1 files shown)


bruniss:
hi, thank you for the PR! 

this overlaps quite a lot with the SmearTransform (in /batchgeneratorsv2/transforms/noise/
  extranoisetransforms.py:185) 

it is different, in that its a strict shift with no blending, but its not probably _that_ different enough on its own to justify using in conjunction with the smear. i think if both versions were rolled into a single transform with an arg for type of jitter it may fit better rather than as a whole new standalone

bruniss:
close due to staleness / existing impl already covering this

PR #1131, opened 2026-07-13
vc: build linux appimage for single file linunx download

jrudolph:
This builds a linux single file appimage that can be used to launch most of VC's tools from a single binary (this is a resurrection of an old VC AppImage build which I did in 2023). I excluded vc-diffuse-winding and vc-render-video to be able to exclude the video components).

Claude figured out some shenanigans to be able to build on Ubuntu 26.04 and include the glibc, so that the build also works on older Ubuntu, though not sure how reliable that is going to be.

The single binary has a multitool wrapper, so you can run the individual binaries with `<downloaded binary> VC3D` or `<downloaded binary> vc-render-tifxyz`. You can also create symlinks for the individual commands that will be automatically dispatched to the tool.

### DIFF
--- a/.github/workflows/vc3d-linux.yml
+++ b/.github/workflows/vc3d-linux.yml
@@ -1,0 +1,156 @@
+name: VC3D Linux AppImage
+run-name: VC3D Linux ${{ github.event_name }} ${{ inputs.request_id || github.sha }}
+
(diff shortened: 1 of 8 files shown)


jrudolph:
Going to merge this to test the build

PR #1154, opened 2026-07-15
vc_sync: rclone parallel transfers, upload validation, hfsync subcommand

skyward7187:
## Summary

Speed, safety, and publishing improvements to `scripts/vc_sync.py`:

**Parallel transfers via rclone (with automatic fallback)**
- When the `rclone` binary is available, uploads/downloads/remote-deletes run as one batched rclone process with 16 parallel transfers instead of one serial `aws s3 cp` subprocess per file — minutes → seconds for typical fiber batches.
- No rclone configuration needed on any machine: an on-the-fly remote (`:s3,provider=AWS,env_auth=true:`) uses the same credential chain as the aws CLI (env vars, `~/.aws/credentials`, EC2 instance roles).
- Detection probes that rclone can actually **read the sync directory** — sandboxed installs (e.g. the Ubuntu snap without its `removable-media` interface) fail the probe and the script falls back to the original serial aws CLI path, printing the reason.

**Upload safety**
- Files are re-stat'ed immediately before upload (the interactive prompts can leave a long window after the scan) and checked again after; a file modified mid-upload is flagged and re-uploaded on the next sync. This closes the race that could silently persist truncated uploads.
- Zero-byte files and unparseable JSON among upload candidates are listed with a per-file `[u]pload anyway / [s]kip` prompt before the confirmation gate (shown read-only under `--dry-run`).

**Correctness fix**
- The S3 listing prefix now ends with `/`, so sibling prefixes (e.g. `<prefix>_old`) can no longer bleed into the scan with mangled relative paths.

**New: opt-in `hfsync` subcommand**
- `vc_sync.py hfsync <dir> [--dry-run]` syncs tagged fiber JSONs (default tag: `reviewed`) to a Hugging Face storage bucket.
- Entirely configured by a local `.hfsync.json` next to the data (documented in the module docstring). The script contains no bucket names, tokens, or machine paths; directories without the config file are unaffected. Auth uses the token stored by `hf auth login`.
- Semantics: additive upload (unchanged files skip via size/mtime); a remote file is removed only when the same filename exists locally *without* the tag, so remote-only files can never be deleted.

**Removed**
- The segment-area CSV reporting pipeline (and its hardcoded report bucket), per project decision.

## Test plan

- Batch upload/download/remote-delete exercised against a local stand-in remote: flat fiber layout (20/5/3 files) and nested tifxyz layout (`<segment>/{meta.json,x.tif,y.tif,z.tif}` with 2 MB binaries) — transfers byte-identical, tracking DB keys consistent with the S3 scanner.
- Fallback detection verified for both failure reasons (binary missing; directory unreadable by a sandboxed install).
- Validation flags zero-byte and mid-write-truncated JSON; valid JSON and non-JSON pass.
- `hfsync` verified against a real HF bucket scratch path: dry-run makes no changes, un-tagged files are removed, remote-only files survive; then a production run (25 new uploads, 152 skips, 0 removals).
- Local scan of a real 116-file / 9.5 GB tifxyz `paths` directory confirms filters and nested path handling are unchanged.

https://claude.ai/code/session_01AVkXaynVESTt7q8Z8n6LU2

### DIFF
--- a/volume-cartographer/scripts/vc_sync.py
+++ b/volume-cartographer/scripts/vc_sync.py
@@ -687,7 +687,14 @@ class S3SyncManager:
         ok = self._run_rclone(['copy', self.local_dir, self._rclone_remote()],
                               list(pre_stats))
         if not ok:
-            print("  ⚠️  rclone reported errors during upload; verifying what was transferred")
+            # Don't advance tracking on a partial failure: a same-size stale
+            # remote object would pass the size check below and the file would
+            # then be recorded as in sync, permanently skipping the re-upload.
+            # Leaving tracking unchanged re-schedules every file next sync;
+            # rclone skips the ones that did transfer, so the retry is cheap.
+            print(f"  ❌ rclone reported errors during upload; tracking left "
+                  f"unchanged for all {len(pre_stats)} files, will retry on next sync")
+            return 0
 
         for path, (size, mtime) in pre_stats.items():
             try:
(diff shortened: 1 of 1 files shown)


hendrikschilling:
@codex review

skyward7187:
Both review findings addressed in ae249ec1c — they were correct: a same-size stale object could pass the size-only verification after a partial rclone failure and get recorded as in sync, permanently skipping the retry.

Fix: a nonzero rclone exit now leaves tracking untouched for the entire batch (matching what the delete batch already did). Every file re-schedules on the next sync; rclone's size+modtime skip makes the retry cheap for files that did transfer.

Tested both directions with forced partial failures (one unreadable file in a 10-file batch): failure run reports 0 and leaves the tracking DB byte-identical; the retry run completes and verifies all files.

Issue #1114, opened 2026-07-04
Scroll 3 (PHerc0332) m7 surface predictions: ~70% of positive voxels are phantoms where the masked CT is 0


spencerdavis-tx:
**Artifact:**
`s3://vesuvius-challenge-open-data/PHerc0332/representations/predictions/surfaces/20251211183505-surface-20260413222639-surface-m7-L2-th0.2.zarr`
(zarr v2, OME multiscale L0–L5, L0 shape [8398, 3941, 3941] on the CT-L2 / 9.596 µm grid, uint8 0/255, chunks 192³; the zarr's `metadata.json` dates the run 2026-05-13, model `s3://scrollprize-reconstruction/models/m7_nnunet/`).

**Finding.** A large majority of the positive voxels are not papyrus surface: they are solid prediction slabs sitting where the published SAM2-masked CT volume (`PHerc0332/volumes/20251211183505-2.399um-0.2m-78keV-masked.zarr`) reads exactly 0 — a halo ring around the scroll cross-section plus end caps. Presumably inference ran on (or was blended over) unmasked data, or the mask used at inference differs from the published masked volume; either way the published preds and the published masked CT disagree about where the scroll is.

**Numbers** (survey 2026-06-11, re-verified against the live S3 objects 2026-07-03 — unchanged; preds zarr LastModified 2026-05-13, masked CT 2026-03-03, no re-versioning):

- Global CT-supported share of positive voxels: **0.302** → **~70% phantom**.
- Full-resolution axial planes z ∈ {2000, 4224, 6000} (preds-L0 grid): phantom fraction **0.681 / 0.672 / 0.646** (2026-07-03 re-check: 0.6810 / 0.6717 / 0.6459, positives 2,259,758 / 2,379,932 / 2,110,813 per plane).
- Interior blocks (well inside the scroll): phantom fraction ≤ 6.1e-05 — where CT exists, the predictions are excellent recto-face skins. The bug is only in the zero-CT zones.
- Cube-level (128³) support is strongly bimodal: at a 0.5 support threshold, ~10.9k cubes are ~all-phantom vs ~15.9k ~fully supported, only ~536 between — so cube-level filtering plus voxel-wise masking is cheap and clean.

**Not scroll-specific — the whole 2026-04-13 batch appears affected.** The same model-run ID
(`20260413222639`) stamps the published surface preds for PHerc1451, PHerc0814, and PHerc1299.
[...]

PR #1156, opened 2026-07-15
Mask finalized surface predictions outside CT support

BuraChuhadar:
## Summary

- add opt-in CT/support masking to `vesuvius.finalize_outputs` and `vesuvius.blend_and_finalize`
- apply the mask after probability conversion and prediction thresholding
- add `vesuvius.mask_predictions` to repair an existing finalized binary Zarr artifact while preserving supported values
- record support provenance and exact phantom-positive statistics in output attributes
- add a bounded public-S3 plane benchmark, documentation, and regression/integration coverage
- check in raw benchmark JSON plus a reproducible before/effect/after evidence image

Addresses #1114.

## Why

The published surface predictions referenced in #1114 contain large positive halos where the aligned masked CT has no usable scan data. Simply setting binary logits to zero is not sufficient: equal binary logits produce foreground probability `0.5`, which is positive at the affected run's `0.2` threshold. A task-aware saturated-background encoding before blending avoids that specific equal-logit case. This PR instead enforces support after blending, probability conversion, and thresholding, and can also repair already-published finalized arrays.

## Relation to #1173

[#1173](https://github.com/ScrollPrize/villa/pull/1173) explores source-side prevention by deriving a mask from raw input values equal to zero and encoding unsupported segmentation voxels as saturated background logits immediately after inference. This PR operates at the output boundary: it applies a separately supplied, aligned, configurable support volume after blending/finalization and provides a repair CLI for existing arrays. The approaches are complementary in scope—source prevention for new inference runs, and final-stage enforcement/repair for both new and existing outputs. Their support rules also differ (`raw input == 0` in #1173 versus finite support values greater than `--support-threshold` here), so their counts need not match.

### Independent scale context

An independent [36-sample audit](https://github.com/Schurkai/vesuvius-phantom-audit) by [Schurkai](https://github.com/Schurkai) reports that all 36 published m7 samples contain certain-phantom prediction chunks, totaling 280,580 of 1,662,405 stored prediction chunks (16.9%). This is a conservative chunk-occupancy metric, not a full-volume positive-voxel fraction; the reported 92.8% positive-voxel result is from one sampled PHercMANBp plane. The audit uses raw CT equal to zero as empty, while this PR's checked-in examples use finite CT greater than 5 as support, so it is independent scale context rather than direct validation of this implementation.

## Public-data check

Using masked CT value `> 5` as support and prediction value `> 0` as positive:

| Artifact / plane | Positives before | Removed outside support | Phantom fraction before | Phantom after | Supported values preserved |
|---|---:|---:|---:|---:|---|
| PHerc0332 z=2000 | 2,259,758 | 1,538,875 | 68.0991% | 0 | yes |
| PHerc0332 z=4224 | 2,379,932 | 1,598,695 | 67.1740% | 0 | yes |
| PHerc0332 z=6000 | 2,110,813 | 1,363,423 | 64.5923% | 0 | yes |
| PHerc0332 aggregate | 6,750,503 | 4,500,993 | 66.6764% | 0 | yes |
| PHerc1451 z=7493 | 3,036,889 | 1,820,368 | 59.9419% | 0 | yes |

The benchmark command and public paths are documented in `vesuvius/docs/inference.md`. It streams one plane pair at a time rather than downloading a scroll volume.

### Reproducible proof bundle

The following artifacts were regenerated from the cited public S3 inputs at commit `4c4f14ce6`:

- [PHerc0332 raw schema-v2 JSON (planes 2000, 4224, 6000)](https://github.com/BuraChuhadar/villa/blob/4c4f14ce6aa2fd1d88cb3dbb0c0eb17dfa10c981/vesuvius/docs/benchmarks/support_mask/PHerc0332-three-planes.json)
- [PHerc1451 raw schema-v2 JSON (plane 7493)](https://github.com/BuraChuhadar/villa/blob/4c4f14ce6aa2fd1d88cb3dbb0c0eb17dfa10c981/vesuvius/docs/benchmarks/support_mask/PHerc1451-plane-7493.json)

The image is a fixed-extent downsampled density preview of the same PHerc0332 z=4224 plane in every panel. Exact counts are computed at full resolution. Magenta marks positive predictions removed outside the selected CT support; white marks retained positives. This demonstrates post-processing of the published prediction, not model re-inference or full-volume validation.

PHerc0332 z=4224 before, removal effect, and after support masking

## Independent PHerc1218 validation

[Jinhojeong independently ran `vesuvius.mask_predictions` unmodified](https://github.com/ScrollPrize/villa/pull/1156#issuecomment-5289083721) on four windowed PHerc1218 regions at the default `support_threshold=0`, using instance labels as a conservative oracle for over-masking.

Across 251,865,288 labeled voxels, 226 were directly observed to lose predictions across all eight corresponding level-0 subvoxels. After scaling for the per-tile inspection cap in two regions, the stated upper bound is 283, approximately one in 890,000. None of the inspected candidate losses mapped to a 2×2×2 CT block that was entirely unsupported. In the control with approximately 0.026% phantom positives, masking was close to a no-op: it removed 123,491 of 475,944,417 positives, while 17 of 95,280,076 labeled voxels lost predictions across all eight subvoxels. [Four-region data and scripts](https://github.com/Jinhojeong/vesuvius-surface-geometry-diagnostic/commit/6ac684e38f7d1f22b47b7bdc905e5dfa74364d70).

A separate three-seed trace A/B found that masking substantially changed the traced meshes, but the fractions of quads over unsupported CT moved only from 34.1% to 33.4%, 43.5% to 39.2%, and 1.5% to 1.8%. In that region, unsupported quads concentrated toward the mesh fringe and the direction field was absent, consistent with the tracer extrapolating beyond its available evidence. This defines a scope boundary: this PR enforces the configured support rule on the finalized prediction volume, but it is not a stopping rule for downstream tracers.

The corresponding consumer-side support gate was implemented in [#1451](https://github.com/ScrollPrize/villa/pull/1451) and independently calibrated on the same three seeds. With one-voxel support dilation, unsupported traced geometry fell to 0.5%, 0.9%, and 0.3%, while retaining 64.6%, 63.7%, and 93.5% of the unconstrained surfaces. That PR was closed without merging; the tracer change remains a separate follow-up rather than being folded into this finalized-artifact PR. [Calibration details](https://github.com/ScrollPrize/villa/issues/1114#issuecomment-5302442769).

These measurements cover four windows on one scroll, not a whole-volume or cross-scroll validation. Because inspection was capped in two regions, 283 is an upper bound rather than a directly counted total.

## Design notes

- support masking is currently binary-only and opt-in
- support is defined by finite values strictly greater than `--support-threshold`
- rank and spatial shape are validated; callers remain responsible for physical scan/axis/transform alignment
- public S3 inputs use anonymous access unless authenticated access is explicitly requested
- `vesuvius.mask_predictions` writes a plain single-level Zarr array; publishing a drop-in OME-Zarr replacement still requires pyramid/root-metadata regeneration
- output stores are explicit Zarr v2 for compatibility with the existing pipeline, while Zarr v3 support inputs are tested
- constant nonzero chunks and thresholded multiclass class IDs are finalized consistently across chunk boundaries

## Validation

Local compatibility validation was rerun on merged head `940136a0` using Python 3.14.6 against both supported Zarr major versions:

```text
Zarr 3.3.0: 60 focused tests passed
Zarr 2.18.7: 58 passed, 2 expected skips
CLI-alias coverage: 14 applicable tests passed under each environment
```

The 11 deselected CLI tests require the optional PyTorch/model dependency set; they were not code failures. Direct parser checks also passed in both environments.

The focused suite covers pure mask semantics, the Gaussian-corner saturation regression, standalone and fused finalization, direct existing-artifact repair, public-style mismatched chunk layouts, Zarr v2/v3 inputs, Zarr 2/3 creation-keyword compatibility, prediction/output and support/output overlap protection, invalid-worker output preservation, constant nonzero chunks, multiclass label consistency, deterministic evidence rendering, and invalid image-option safety.

`git diff --check upstream/main...HEAD` and an independent final-diff preflight also pass.


giorgioangel:
@codex review

BuraChuhadar:
Addressed the worker-count feedback in `52c05d0d3`: validation now happens before any output store is opened, and a regression test verifies that existing output is preserved for `num_workers=0`. The targeted suite now passes all 39 tests.

The updated-head GitHub Actions runs are awaiting maintainer approval (`action_required`, with no jobs created yet).

@codex review

BuraChuhadar:
@giorgioangel Quick maintainer follow-up: the branch is currently mergeable, and I have refreshed the Validation section with local Python 3.12 coverage under both Zarr versions used by the repository matrix:

- Zarr 3.2.1: 60 focused tests passed, plus 14 applicable CLI-alias tests.
- Zarr 2.18.7: 58 passed, 2 expected skips, plus 14 applicable CLI-alias tests.

The remaining failing status is Vercel's `Authorization required to deploy` for the Scroll team; GitHub is not reporting a code/test failure. When convenient, could you or another maintainer review the current head? Thanks.

Jinhojeong:
I ran this PR's `mask_predictions` unmodified on PHerc1218 and have an independent validation, plus one measurement of what it does not fix. This follows up the label-support oracle I offered in #1114.

Motivation. I was measuring how much of the m7 surface prediction carries instance labels on PHerc1218 and found the unlabeled majority is this bug. The natural next question was whether the fix holds up where the phantoms are worst, and whether it costs any real sheet.

Setup. I picked a phantom-dense region by rule, built local prediction and support stores for it from the public m7 zarr and the masked CT of the same timestamp, and ran your CLI on that pair at the default threshold. The region is a hard case, 71.2 percent of its predicted voxels are phantoms, and the CLI removed 423 million of 594 million positives.

Over-masking, which is the part I could test that the PR cannot test on its own. PHerc1218 has instance labels, and labels never sit on unsupported CT, so any labeled voxel losing its prediction is a detected fault. Of 34,345,863 labeled voxels in the region, 953 lost the prediction at their level-0 corner sub-voxel, and only 19 lost it across all eight sub-voxels, so the upper bound on lost labeled sheet is 19 in 34 million. Every one of the 953 sits in a two-cube that keeps some CT, and the removal follows CT at sub-voxel precision, one example block reads CT 0,0,2,0,12,9,0,0 and keeps positives exactly where CT is nonzero. That is the masking behaving as specified.

What the masking does not do, measured. I traced surfaces with `vc_grow_seg_from_seed` from three deterministic seeds on supported prediction near the phantom halo, once against the original prediction and once against the masked one, then rendered both through `vc_render_tifxyz` with flattening. Masking changes the trace substantially, 95 to 98 percent of quads move more than 8 voxels with median moves of 24 to 123. But the share of traced quads standing over unsupported CT barely moves, 34.1 to 33.4, 43.5 to 39.2, and 1.5 to 1.8 percent on the three seeds. The reason is not the prediction, it is the tracer. Unsupported quads concentrate at the mesh fringe, mean distance to the patch edge 5.7 to 6.8 grid steps against 12.4 to 13.4 for supported quads, and the unsupported share is 0.57 to 0.64 in the outer ring against 0.25 to 0.34 in the interior. The direction field is absent in this region, so the extrapolation is the surface fitting itself.

PHerc1218 surfaces traced before and after CT-support masking, flattened renders

The dark areas are where the traced mesh has no CT under it, and their extent tracks the numbers above. Read together, this PR cleans the artifact and preserves labeled sheet at the rate above, and a tracer consuming the cleaned volume still needs a support-aware stopping rule, which is consistent with your reading and #1173's that the two fixes are complementary. Everything, the region rule, the seeds, the per-run numbers and the render scripts, is at commit d8956f0 under results/p8_support_mask in my diagnostic repo. One limit, this is one region on one scroll and a window-scoped run rather than a whole-volume one, since the full-volume support stream is out of reach on my hardware.

Jinhojeong:
Extending the validation I posted above from one region to four, and from the phantom-dense case to the case that matters more for merging, regions where the sheet actually is.

The first run was a hard case by construction, 71.2 percent of its predicted voxels phantoms. The obvious objection is that nothing was at risk there. So I ran the same CLI on three fully-labeled control regions, coverage 0.99 or better, where phantoms are 24.9, 19.0 and effectively 0 percent of predicted voxels. Across all four the same test applies, PHerc1218 has instance labels and labels never sit on unsupported CT, so any labeled voxel that loses its prediction is a detected fault.

Over 251,865,288 labeled voxels in the four regions, the number that lose their prediction across all eight level-0 sub-voxels is 226 measured directly, and at most 283 after scaling for the per-tile inspection cap in two of the regions. That is about one labeled voxel in 890,000. The per-region counts are 19 of 34.3M in the phantom-dense region and 56, 17 and 134 of 64.5M, 95.3M and 57.8M in the controls. Not one labeled voxel in any region sits in a two-cube that is entirely without CT, so the label-side invariant holds everywhere I looked, and the removal keeps following CT at sub-voxel precision.

The control where phantoms are essentially zero is worth calling out separately. There the CLI removes 123 thousand voxels against 428 million in the first control, so it is close to a no-op, and it still costs only 17 labeled voxels. A masking pass that does nothing where there is nothing to do is the behaviour you want before this goes anywhere near a published artifact.

Numbers, the region selection rule and the scripts are at results/p8_support_mask in my diagnostic repo. The limits are unchanged from the first comment, this is four windowed regions on one scroll rather than a whole-volume run, and the level-0 corner sampling is why the inspection is capped and the 283 is stated as an upper bound rather than a count.

BuraChuhadar:
Thanks @Jinhojeong. I have added the PHerc1218 four-region validation and the tracer boundary to the PR description.

The label-oracle result materially strengthens the safety case for #1156: 226 directly observed labeled voxels lost predictions across all eight level-0 subvoxels out of 251,865,288, with 283 as the inspection-adjusted upper bound. I am keeping the one-scroll, four-window and capped-inspection limits explicit, and I am keeping these default-threshold results separate from the checked-in threshold-5 benchmarks.

The tracer result is a real consumer-side issue, but it does not expose an implementation defect in `mask_predictions`. #1451 already implemented the support-aware growth gate, and the dilation-1 calibration reduced unsupported geometry to 0.5%, 0.9%, and 0.3% on the same three seeds. I am therefore keeping #1156 focused on finalized-artifact enforcement rather than duplicating that C++ change here.

@bruniss @jrudolph, would you prefer #1451 to be reopened or resubmitted as the separate tracer follow-up? The implementation and independent calibration now exist; the remaining question is how you want that consumer-side change landed.

PR #1157, opened 2026-07-16
vc_sync: tracking robustness fixes, non-destructive update, dedup refactor

skyward7187:
Follow-up to #1154. Fixes a state where locally deleted folders were proposed for **re-download instead of remote deletion**, plus the code paths that could recreate that state, and a round of deduplication.

## Root cause

The delete-vs-download decision comes entirely from `.s3sync.db`: a locally deleted file is only proposed for remote deletion if its tracking row exists. On a real volpkg, 106 of 116 files had no rows — an interrupted `init` had silently rolled back its whole tracking transaction (no commit on exception), and nothing ever repaired it. Untracked S3-only files default to "New S3 file" → download.

## Fixes

- **`_get_db` commits/rolls back deterministically** (try/finally). An interrupted tracking update now fails loudly instead of silently discarding the transaction.
- **Remote-delete batches verify before clearing tracking.** `rclone delete --files-from` exits 0 even when nothing matched (verified experimentally), so exit code alone can't confirm deletion. Tracking rows are now only dropped for files confirmed gone from S3; verified deletions advance even when the batch reports errors, since presence checks are unambiguous (unlike the size-only transfer checks from #1154's review feedback).
- **`sync` self-heals tracking**: size-matched files with no history are recorded as in sync, so lost tracking repairs itself on the next run instead of turning future deletions into downloads.
- **`init` no longer offers its own bulk download** — the dialogue implied it downloaded missing segments, and `sync` does that better (parallel, verified, tracked). Init now records in-sync files and reports what a following `sync` would transfer. Size-mismatched pairs are left untracked so they surface as conflicts instead of being silently stamped as synced.
- **`update` is now non-destructive.** Previously `update` was identical to `reset` minus the confirmation prompt: it stamped current state as the baseline, silently discarding pending local changes. Now `update` = record in-sync files + prune entries gone from both sides, never overwriting pending differences; only `reset` (confirmed, clearly worded) stamps the baseline.

## Cleanup

- Single writer for tracking rows (`_track_file`) replacing six copy-pasted `INSERT OR REPLACE` blocks; single authority for scan ignore rules (`_is_ignored`) shared by the local and S3 scanners.
- EOF-safe prompt helpers: `sync` in a non-interactive shell cancels cleanly instead of crashing with `EOFError`.
- Small-batch verification uses per-file `head-object` instead of listing the whole prefix (≤8 files); a 404 confirms absence, any other error is treated conservatively.
- `_run_aws_command` no longer mutates the caller's command list; dropped the redundant `idx_path` index (`path` is already the primary key).

## Testing

- 22-check functional suite exercising rollback behavior, self-heal, non-destructive update vs reset, delete/upload verification (rclone and S3 mocked), prompt EOF handling, and the shared ignore rules.
- End-to-end smoke test with a stubbed `aws` CLI: `init` → `sync --dry-run` proposes the correct upload/download/conflict plan; non-interactive `sync` cancels gracefully; `update` refreshes without stamping.

https://claude.ai/code/session_01QqetQbXo7nXe3HvSbi6esN

### DIFF
--- a/volume-cartographer/scripts/vc_sync.py
+++ b/volume-cartographer/scripts/vc_sync.py
@@ -15,6 +15,15 @@ Automatically ignores:
 - Files matching backup patterns (see BACKUP_PATTERNS)
 - Directories named 'backups' (unless --sync-backups is specified)
 
+Change detection compares file size and mtime locally (1-second tolerance)
+and size + ETag on S3, so a local edit that preserves both size and mtime is
+invisible until the file changes again.
+
+'update' refreshes tracking non-destructively (records files already in sync,
+prunes entries for files gone from both sides) and never hides a pending
+difference. 'reset' stamps the current state as the synced baseline, which
+discards pending differences — it asks for confirmation.
+
 Usage:
     python s3_sync.py init <directory> <s3_bucket> <s3_prefix> [--profile=<aws_profile>]
     python s3_sync.py status <directory> [--verbose] [--sync-backups]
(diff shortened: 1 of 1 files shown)


bruniss:
@codex review

PR #1172, opened 2026-07-18
Update community projects list

pscamillo:
Move augmentations entry to end of Tools (fix earlier misplacement) and add winding-ruler


erdpx:
hi! pr #997 contains only the squeeze augmentation. could you also add the correct pr link for each of the other transforms, just so we have them recorded correctly?

pscamillo:
good catch — fixed: each transform now links its own PR (Squeeze #997, Decohesion+Warp #999, Ring #1020, Streak #1021, Warp edge-padding fix #1025), and resolved the conflict with current main while at it.

erdpx:
thanks! one small merge-resolution issue: the latest merge changed the list nesting, putting two other projects under yours. could you fix the indentation before we merge?

pscamillo:
fixed — the ports are back at root level, nesting was a merge-resolution artifact. thanks for the careful review!

PR #1173, opened 2026-07-19
Prevent phantom predictions at inference time (addresses #1114, complements #1156)

Schurkai:
Addresses #1114. Complementary to #1156.

**How this relates to #1156**: that PR masks the *finalized* outputs against CT support, which repairs artifacts after the fact and covers already-published zarrs. This PR adds the prevention layer @jrudolph suggested in the issue: mask the logits right after inference, where the input patch is still in memory - so phantoms never enter the blending/finalize path in the first place, at essentially zero extra I/O. The two are not competing; a pipeline ideally wants both (prevent at source, verify/repair at finalize).

On the note in #1156 that logit masking is insufficient (equal logits give p = 0.5, which is positive at a 0.2 threshold): that holds for *zeroed* logits, but not for saturated ones. Masked voxels here are forced to saturated background logits (+/-20, the same convention `finalize_outputs` already uses): sigmoid(-20) is ~2e-9, and for 2-class outputs bg/fg get +20/-20 so `logit[1] - logit[0] = -40`, far below any threshold cutoff. This is unit-tested against the exact threshold semantics in `finalize_outputs` for both sigmoid and 2-class softmax.

The review below surfaced a genuine gap in that argument: the blending normalization added a 1e-8 epsilon to the denominator, which diluted any logit - saturated or genuine - toward 0 (p ~ 0.5) at voxels whose total Gaussian weight sits near or below the epsilon, e.g. a lone outer-corner patch contribution at a volume edge (weight ~exp(-24)). The epsilon predates the `where=weights > 0` guard that already protects the division, so it is now removed: the blended value is the exact weighted average at every covered voxel and saturation does survive blending. Regression tests pin the corner case; #1156's finalize-stage mask remains the post-blend guarantee and the repair path for already-published artifacts.

**Changes**

- `Volume`: new opt-in `return_zero_mask` - returns `(data, mask)` with the mask computed on the **raw** data, before normalization (zscore maps raw zeros to a nonzero value, so the check can't happen later).
- `VCDataset`: passes the per-patch mask through `__getitem__` / `collate_fn`.
- `blending`: normalization now preserves the weighted average at covered voxels (`normalize_blended_logits`); the former denominator epsilon diluted low-weight voxels toward p ~ 0.5, saturated or not.
- `Inferer`: masked voxels forced to saturated background logits, handled per-target for multi-task models - only classification heads (sigmoid/softmax activation); continuous targets such as the 9-channel `surface_frame` (activation `none`) are skipped, with a notice at model load. Overlapping patches read identical raw voxels, so masking is consistent across patches. Opt-in via `--mask_empty_input` (off by default).

There is also a small separate commit fixing `open_zarr` for local paths under zarr 3.x (`storage_options` may not be passed for non-URI stores, even as `{}`). If #1156's `utils.py` change already covers this, happy to drop that commit.

**Validation**

Reproduced the numbers from the issue on the published m7 predictions, streamed from the open bucket:

| z | positives | phantom fraction (mine) | issue |
|---|---|---|---|
| 2000 | 2,259,758 | 0.681 | 0.681 |
| 4224 | 2,379,932 | 0.672 | 0.672 |
| 6000 | 2,110,813 | 0.646 | 0.646 |

Chunk-wise `pred[vol==0]=0` on the z=2000 slab leaves exactly 720,960 positives = 2,259,758 - 1,538,798 phantoms, i.e. the mask removes precisely the phantom set and nothing else. The data path is unit-tested against a synthetic volume (mask == raw==0 through normalization/patching/collate). What I could not run locally is a full end-to-end inference with a real checkpoint (no GPU here) - happy to iterate if anything shows up there.

Committed tests: `tests/models/run/test_blending_normalization.py` (corner-weight preservation, zero-coverage passthrough, agreement with the legacy formula at typical weights) and `tests/models/run/test_mask_empty_input.py` (activation gating, masking conventions, opt-in default).

For scale context: I measured the same halo across all 36 samples with published m7 surface predictions - every one is affected, and 16.9% of all stored prediction chunks are certain phantoms (details: https://github.com/Schurkai/vesuvius-phantom-audit). That is why source-level prevention seemed worth having in addition to the finalize-stage repair.

### DIFF
--- a/vesuvius/src/vesuvius/data/utils.py
+++ b/vesuvius/src/vesuvius/data/utils.py
@@ -130,6 +130,11 @@ def open_zarr(path: str, mode: str = 'r',
     if verbose:
         print(f"Opening zarr store at {path} with mode={mode}")
     
+    # zarr 3.x raises TypeError when storage_options is provided (even {}) for
+    # non-URI stores such as local paths; only pass it for fsspec URIs.
+    if "://" not in path:
+        storage_options = None
+
     # If we're creating a new array (mode='w') and shape is provided, pass creation parameters
     if mode == 'w' and shape is not None:
         create_kwargs = {}
--- a/vesuvius/src/vesuvius/data/vc_dataset.py
+++ b/vesuvius/src/vesuvius/data/vc_dataset.py
@@ -46,6 +46,7 @@ class VCDataset(Dataset):
             skip_empty_patches: bool = True,  # Whether to skip empty (homogeneous) patches
             anon: bool = False,  # Use anonymous (unsigned) requests for S3 input paths
             read_retries: int = 4,  # Attempts per read, forwarded to Volume
+            return_zero_mask: bool = False,  # Also return per-patch mask of raw==0 voxels (see issue #1114)
             ):
         """
         Dataset for nnUNet inference using the Volume class for data access and preprocessing.
@@ -99,6 +100,7 @@ class VCDataset(Dataset):
         self.return_as_tensor = True # Dataset __getitem__ always returns tensors
         self.skip_empty_patches = skip_empty_patches
         self.anon = anon
+        self.return_zero_mask = return_zero_mask
         self.empty_patches_skipped = 0  # Counter for skipped patches
         self.non_empty_mask = None  # Per-patch bool mask; populated below for infer mode with zarr input
 
@@ -233,6 +235,7 @@ class VCDataset(Dataset):
                 path=use_path,
                 anon=self.anon,
                 read_retries=read_retries,
+                return_zero_mask=self.return_zero_mask,
             )
 
             # Get shape and dtype from the primary resolution level (0)
@@ -577,8 +580,12 @@ class VCDataset(Dataset):
                 # For 3D input (Z, Y, X), we always create a single-channel tensor
                 # The model expects a single channel regardless of how many output classes it produces
                 
-                # Volume returns (Z, Y, X) tensor
-                extracted_tensor = self.volume[z_slice, y_slice, x_slice]
+                # Volume returns (Z, Y, X) tensor, plus a raw==0 mask when enabled
+                extracted_zero_mask = None
+                if self.return_zero_mask:
+                    extracted_tensor, extracted_zero_mask = self.volume[z_slice, y_slice, x_slice]
+                else:
+                    extracted_tensor = self.volume[z_slice, y_slice, x_slice]
                 
                 # Fast check for empty patch - skip if all zeros or all values are the same
                 # This avoids processing empty patches which won't contain any information
@@ -602,9 +609,14 @@ class VCDataset(Dataset):
                 # Use the dtype returned by Volume
                 # ALWAYS use a single channel (1) here since we're coming from Z,Y,X data
                 patch_tensor = torch.zeros((1, pZ, pY, pX), dtype=extracted_tensor.dtype)
-                
+
                 # Copy fetched data into the patch tensor
                 patch_tensor[0, :fetched_z, :fetched_y, :fetched_x] = extracted_tensor
+
+                if self.return_zero_mask:
+                    # Padding regions lie outside the volume -> treat as empty (True)
+                    zero_mask_tensor = torch.ones((1, pZ, pY, pX), dtype=torch.bool)
+                    zero_mask_tensor[0, :fetched_z, :fetched_y, :fetched_x] = extracted_zero_mask
 
             elif len(self.input_shape) == 4: # Input is C, Z, Y, X
                 # Volume returns (C, Z, Y, X) tensor
@@ -621,7 +633,11 @@ class VCDataset(Dataset):
                     print(f"4D input: Extracting {self.num_input_channels} channels from data with {available_channels} channels")
                 
                 # Take exactly what the model expects
-                extracted_tensor = self.volume[:self.num_input_channels, z_slice, y_slice, x_slice]
+                extracted_zero_mask = None
+                if self.return_zero_mask:
+                    extracted_tensor, extracted_zero_mask = self.volume[:self.num_input_channels, z_slice, y_slice, x_slice]
+                else:
+                    extracted_tensor = self.volume[:self.num_input_channels, z_slice, y_slice, x_slice]
                 
                 # Check for empty patch - flag if all values are the same
                 is_empty = False
@@ -655,6 +671,12 @@ class VCDataset(Dataset):
                 
                 # Copy fetched data
                 patch_tensor[:, :fetched_z, :fetched_y, :fetched_x] = extracted_tensor
+
+                if self.return_zero_mask:
+                    # Padding regions lie outside the volume -> treat as empty (True).
+                    # A voxel counts as empty only if ALL input channels are raw 0.
+                    zero_mask_tensor = torch.ones((1, pZ, pY, pX), dtype=torch.bool)
+                    zero_mask_tensor[0, :fetched_z, :fetched_y, :fetched_x] = extracted_zero_mask.all(dim=0)
             else:
                  # Should have been caught in init, but safety check
                  raise RuntimeError(f"Unsupported volume shape encountered in getitem: {self.input_shape}")
@@ -674,12 +696,15 @@ class VCDataset(Dataset):
 
         position_tuple = (int(z), int(y), int(x))
 
-        return {
+        item = {
             "data": patch_tensor, # Key required by nnUNet inference
             "pos": position_tuple, # Pass position for potential stitching later
             "index": idx, # Pass original index
             "is_empty": is_empty # Flag to indicate if patch is empty (to skip model inference)
         }
+        if self.return_zero_mask:
+            item["zero_mask"] = zero_mask_tensor
+        return item
     
     @staticmethod
     def collate_fn(batch):
@@ -691,10 +716,13 @@ class VCDataset(Dataset):
         pos = [item["pos"] for item in batch]
         indices = [item["index"] for item in batch]
         is_empty = [item.get("is_empty", False) for item in batch]
-        
-        return {
+
+        result = {
             "data": data,
             "pos": pos,
             "index": indices,
             "is_empty": is_empty
         }
+        if "zero_mask" in batch[0]:
+            result["zero_mask"] = torch.stack([item["zero_mask"] for item in batch])
+        return result
--- a/vesuvius/src/vesuvius/data/volume.py
+++ b/vesuvius/src/vesuvius/data/volume.py
@@ -917,7 +917,11 @@ class Volume:
 
         # Zero-mask must be computed on the RAW data: normalization maps raw
         # zeros to a nonzero value (e.g. -mean/std under zscore schemes).
-        zero_mask = (data_slice == 0) if self.return_zero_mask else None
+        # getattr, not self.return_zero_mask: Volume is also built via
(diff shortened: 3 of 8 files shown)


BuraChuhadar:
Thanks for putting together the source-side prevention path—and for publishing the broader audit. This is complementary to #1156. While comparing the two paths, I found two edge cases that seem worth guarding before `mask_empty_input` is enabled by default:

1. **Saturation can be attenuated at voxels with very small total Gaussian weight.** [`process_chunk`](https://github.com/ScrollPrize/villa/blob/e501389f65912d97c23c2244bdae6c0523cb8ff5/vesuvius/src/vesuvius/models/run/blending.py#L522-L526) accumulates `logit * w`, then divides by `w + 1e-8`. With the [current Gaussian](https://github.com/ScrollPrize/villa/blob/e501389f65912d97c23c2244bdae6c0523cb8ff5/vesuvius/src/vesuvius/models/run/blending.py#L358-L371), an outer corner of an even 3-D patch has `w ≈ exp(-24) ≈ 3.78e-11`. At a voxel with one corner contribution, `-20` therefore becomes about `-0.075` after normalization (`sigmoid ≈ 0.481`), while `+20/-20` becomes about `+0.075/-0.075` (`foreground softmax ≈ 0.462`). Both exceed the affected run's `0.2` threshold, so saturated logits avoid the equal-zero-logit case but do not universally survive the current blending normalization. A regression through the Gaussian corner reproduces this; normalization that preserves the weighted average at positive weights, carrying support through blending, or reapplying support after blending would preserve the invariant.

2. **The multi-channel branch assumes classification semantics.** The [single-task multi-channel path](https://github.com/ScrollPrize/villa/blob/e501389f65912d97c23c2244bdae6c0523cb8ff5/vesuvius/src/vesuvius/models/run/inference.py#L1007-L1027) assumes channel 0 is a background class, but Villa also has non-classification targets such as the [9-channel `surface_frame` output with `activation: none`](https://github.com/ScrollPrize/villa/blob/e501389f65912d97c23c2244bdae6c0523cb8ff5/vesuvius/src/vesuvius/models/configuration/single_task/surface_frame.yaml#L18-L21). With the feature on by default, an unsupported frame voxel becomes `[+20, -20, …]`, which is not a background frame representation. It seems safer to restrict masking to supported segmentation activations/heads (including per-target checks for multi-task models), or keep it opt-in until output semantics are explicit.

#1156's final-stage mask still supplies the post-blend invariant and the repair path for existing artifacts. Committed tests for the low-weight blend boundary and continuous/regression heads would make the intended guarantee reviewable while keeping the two PRs complementary.

Schurkai:
Thanks, great catches on both - you're right on the math and on the semantics.

1. Low-weight blend attenuation: confirmed and reproduced exactly (~-0.075, sigmoid ~0.481 at the exp(-24) corner weight). Rather than just documenting the boundary, I took the first remediation you listed and fixed the normalization to preserve the weighted average at positive weights (e6d3e16). The denominator epsilon predates the `where=weights > 0` guard that already protects the division, so by now it was purely a distortion term - and not only for saturated logits: genuine model predictions at sub-epsilon-weight voxels were flattened toward p ~ 0.5 regardless of what the model said. process_chunk now divides by the exact accumulated weight (extracted as `normalize_blended_logits()`), so +/-20 survives blending at every covered voxel; tests pin the corner-weight case, multi-contribution exactness, zero-coverage passthrough, and agreement with the old formula at typical weights. Post-blend enforcement and the repair path for already-published artifacts remain #1156's territory.

2. Multi-channel non-classification heads: also correct - the branch assumed channel 0 is a background class, which is wrong for continuous targets like `surface_frame` (activation: none). Masking is now gated per target on classification activations (sigmoid/softmax), which `target_info` records from the checkpoint config; continuous heads and targets whose activation is unknown (e.g. inferred via dummy inference) pass through unmasked, and skipped targets are reported once at model load (f31da9b).

I also took your suggestion on the default: `mask_empty_input` is now opt-in (off by default, `--mask_empty_input` to enable) until output semantics are explicit across the model zoo - flipping the default later is the maintainers' call. PR description updated to match.

TAUIL-Abd-Elilah:
A prevalence number, in case it is useful for the argument here rather than for the code.

I ran the same check across the whole published collection instead of a sample: **30 of 36 scrolls** have more than 10% of their predicted sheet sitting on identically-zero CT, and **99.69%** of that volume falls in blocks containing no scanned voxel at all. So it is not a fringe of edge tiles — it is mostly large regions that were never scanned, which is the case where gating at inference costs the least and prevents the most.

Filed as #1254 with the scripts (`pred_over_empty_ct.py`, `phantom_sheet_depth.py`). Happy to rerun it against your masking rule if a before/after over all 36 would help this land.

Schurkai:
Thanks, that's a much bigger sample than I had. Yes please.

Though I think most of that 99.69% is the blend rather than the input path - an all-zero patch already gets dropped by `is_empty` (min == max), so what's left in unscanned blocks is logits leaking outward from the edge. Same thing bruniss pointed at in #1114, and e6d3e16 here is the fix for it.

If you're running it anyway, three ways would separate them: main, +blend fix, +`--mask_empty_input`. Heads up that the flag is opt-in and only kicks in on sigmoid/softmax heads, so it'll look like a no-op otherwise. Branch is rebased now.

TAUIL-Abd-Elilah:
Ran the three arms. **The hypothesis does not survive, and it points the other way: on this region it is the input path, not the blend.**

Region is PHerc0175B, 384³ at z 4909, y 2209, x 7306, picked so the CT is **identically zero over 87.9%** of it while the published m7 prediction asserts sheet there. Over that part every predicted sheet voxel is phantom by construction, so nothing hinges on where the threshold goes — swept anyway. Real sheet (CT > 0 and predicted sheet) is carried alongside as the control.

| threshold | A `main` | B + blend fix | C + `--mask_empty_input` | real sheet (all three) |
|---|---|---|---|---|
| 0.05 | 13,509,235 | 13,502,835 | **0** | 808,577 |
| 0.1 | 13,455,491 | 13,451,239 | **0** | 592,455 |
| **0.2** (published) | **13,437,444** | **13,434,992** | **0** | **433,707** |
| 0.3 | 13,428,318 | 13,426,924 | **0** | 356,387 |
| 0.5 | 7,019 | 7,019 | **0** | 263,641 |
| 0.7 | 2,179 | 2,181 | **0** | 181,791 |
| 0.9 | 35 | 35 | **0** | 76,863 |

**B removes 0.018% of the phantom at the published threshold** — 2,452 voxels of 13.4M — and at 0.7 it comes out two voxels *higher* than main. Max probability over empty CT is unchanged to six decimals, 0.931463 in both.

**C removes all of it at every threshold, and the control says it is not doing that by removing everything:** the real-sheet column is identical to the voxel in all three arms. 0.931463 → 0.931463 → 0.000000.

Two bounds worth stating. This is one region on one scroll, not the 36-scroll 99.69% — that number says where phantom lives, this one says what removes it, and the two should not be quoted as if they were the same measurement. And your heads-up about the flag was worth giving: m7 is a binary head so it fired here, but on a model without a sigmoid/softmax head the table would read as B twice and look like a no-op.

If this holds on other scrolls, the question I would put on the PR is whether `--mask_empty_input` should stay opt-in, given it costs nothing measurable on real sheet here.

Harness and raw numbers: `phantom_three_way.py`, `find_phantom_region.py` and `results/phantom_three_way.json` in https://github.com/TAUIL-Abd-Elilah/vesuvius-repro. Happy to run more scrolls, or a different region if you think this one flatters the input-path arm.

Schurkai:
You're right and I was wrong. I predicted the phantom was mostly blend leakage and your table says it isn't: 2,452 voxels of 13.4M at the published threshold, max probability unchanged to six decimals.

I shouldn't have bundled the two. e6d3e16 fixes the `+1e-8` in the blend denominator, which only distorts voxels whose accumulated weight is near that epsilon - your outer-corner case from the first review. Phantom in a fully unscanned block is covered at ordinary weights, so it was never in that fix's reach. Two separate bugs, and I described them as one mechanism.

On the default: the control column is what moves me, real sheet identical to the voxel across all three arms. I'd support enabling it by default for sigmoid/softmax heads and will put that in the PR description rather than leave it a maintainer question. Worth one region with a long scanned/unscanned border first though, since yours is 87.9% empty and so has little real sheet near an edge for the mask to clip.

TAUIL-Abd-Elilah:
The two-bugs-described-as-one-mechanism point is the useful part — that is precisely what the arms could not separate before, and the epsilon explanation makes the 2,452 voxels make sense rather than just being small.

Agreed on the border region, and it is the right control to ask for. Mine is 87.9% empty *because* that makes phantom unambiguous — every predicted sheet voxel over the zero part is phantom by definition rather than by threshold — but that is the same property that makes it useless for detecting over-clipping. There is almost no real sheet near a boundary for the mask to get wrong.

So I will pick one on the opposite criterion: coverage near 50/50 and a long scanned/unscanned interface, rather than maximum empty fraction. And I will report **real sheet within a few voxels of the boundary** as the endpoint rather than real sheet overall — the current table's "identical to the voxel" would stay true even if the mask shaved a boundary layer, since that layer is a rounding error against 433,707. If the mask does clip, that is where it shows.

Will post it here either way.

TAUIL-Abd-Elilah:
Ran it, and the mask clips nothing — not even the first voxel.

Region picked on the opposite criterion to last time: **PHerc0139, 384³ at z 13105, y 657, x 3181**. CT **56.2%** empty rather than 87.9%, with a **224,940-voxel** scanned/unscanned interface, so there is real sheet along an edge everywhere for the mask to get wrong.

Real sheet (CT > 0 and predicted sheet) binned by distance from that interface, threshold 0.2:

| distance (vox) | `main` | `+ --mask_empty_input` | delta |
|---|---|---|---|
| 0–1 | 52,302 | 52,302 | **0** |
| 1–2 | 85,107 | 85,107 | **0** |
| 2–4 | 180,836 | 180,836 | **0** |
| 4–8 | 530,607 | 530,607 | **0** |
| 8–16 | 1,793,981 | 1,793,981 | **0** |
| 16+ | 2,553,641 | 2,553,641 | **0** |

Zero in every bin, including the layer sitting directly against the boundary where **52,302 voxels** were available to lose. Phantom over the same region goes **5,114,849 → 0**, so the flag definitely fired rather than reading as a no-op on this head.

That is the control you asked for and it comes back clean, so on this evidence the default-on change looks safe.

Two honest caveats. One region, one scroll, one binary head — the failure mode you named for other heads is untouched by this. And the exactness is expected rather than surprising: masking only zeroes wholly-empty input patches, so scanned voxels should be bit-identical. But "should be" was exactly the thing worth checking, and the first region structurally could not check it.

`find_border_region.py` and `measure_boundary_clip.py` in https://github.com/TAUIL-Abd-Elilah/vesuvius-repro — the finder ranks candidates by real sheet near the interface, so it will pick a region for another scroll if you want a second one.

Schurkai:
Thanks for running it. The 0-1 bin is the part I would not have thought to ask for - I named the region, you named the endpoint, and the endpoint was the half that mattered.

One narrowing on your last caveat: the gate is unit-tested even though only a binary head has been measured. `test_mask_empty_input.py` covers a continuous head (activation `none`) passing through untouched, an unknown activation being skipped rather than guessed, and a mixed sigmoid/none/softmax model where only the classification channels move. So the gap is an end-to-end run on a non-binary head, not the gating logic itself.

Both regions and that distinction go into the PR description.

PR #1177, opened 2026-07-20
vesuvius: fix zarr 3 Volume multiscale reads

robertlangdonn:
## What / why

Three related fixes to the Python `Volume` streaming path in `vesuvius/`, found while profiling remote reads on a laptop (16 GB M3, streaming Scroll1 over HTTPS from `dl.ash2txt.org`). All verified live against Scroll1 under the declared `zarr>=3.2` pin (zarr 3.2.1, Python 3.14).

**1. `Volume` multiscale access is broken under zarr 3.**
zarr 3 requires string keys for group member access — integer keys raise `TypeError: sequence item 0: expected str instance, int found`. Every `self.data[0]` / `self.data[subvolume_idx]` in `volume.py` hits this, so `Volume("Scroll1")[...]` (a multiscale group) fails outright. Added a `_level()` accessor (string keys, plain-`Array` passthrough) and routed all call sites through it, including the `ndim` property.

**2. Hidden per-read network round trips.**
`__getitem__` re-resolved the group member and called `len(self.data)` for bounds checks on every read — each a store metadata probe/listing, i.e. a network round trip per read on a remote volume. `_level()` and `_num_levels()` memoize both. Uncached repeat reads drop from ~2 s to a single chunk fetch.

**3. Restored chunk caching (opt-in).**
`use_volume_store_cache` was deprecated in `neural_tracing` when zarr 3 removed `LRUStoreCache`. This adds a zarr-3 equivalent — `LRUCacheStore`, a read-through, size-bounded LRU composed via `zarr.storage.WrapperStore` — exposed as opt-in `Volume(..., cache=True, cache_size_mb=256)` and `open_zarr(..., cache=True)`. It caches the stored (compressed) chunk bytes keyed by `(key, byte_range)`, so decoded values are byte-identical to the uncached path. Repeated reads of the same region (overlapping training patches, viewer panning, tracer neighborhood revisits) are then served from memory. Default is `cache=False` — no behavior change unless opted in. Also: `open_zarr` no longer passes an empty `storage_options` for local paths, which zarr 3 rejects.

## Measured (Scroll1, 32×64×64 reads over HTTPS, M3, zarr 3.2.1, py3.14)

| | cold read | repeat read |
|---|---|---|
| before (per-read metadata round trips) | ~3 s | ~1.8–2.0 s |
| `_level`/`_num_levels` memoized, `cache=False` | ~3 s | metadata overhead gone |
| `cache=True` | ~3 s (first) | **~5 ms (~300×)** |

Cold reads unchanged within run-to-run variance; byte-identical outputs verified across all paths.

## Tests

`vesuvius/tests/data/test_lru_cache_store.py` (5 unit tests, no network — local zarr array): byte-identity cached vs. uncached, cache population, size bound, LRU eviction order, and exact size accounting under 20 concurrent identical-key gets (guards the compressed-byte accounting against zarr's concurrent chunk fetches).

```
pytest vesuvius/tests/data/test_lru_cache_store.py -q   # 5 passed
```

Happy to split this into separate PRs (the zarr-3 access fix is the load-bearing one and stands alone) if that's easier to review.


erdpx:
thanks for the pr! while testing, i noticed a pre-existing zarr 3 incompatibility in the direct-path code: the two zarr.hierarchy.Group checks should use zarr.Group. this remains compatible with zarr 2 volumes. since this pr addresses zarr 3 compatibility, could you update those checks and add a regression test using a multiscale zarr 2 group?

robertlangdonn:
Done — thanks for catching that. Both `zarr.hierarchy.Group` checks in `volume.py` now use `zarr.Group`, which is the public class under zarr 3 and an alias in zarr 2, so the direct-path code stays compatible with both.

Added a regression test (`test_volume_reads_multiscale_zarr2_group`) that builds a `Volume` from a multiscale **zarr v2** group and asserts the group branch is taken (`isinstance(vol.data, zarr.Group)`), the dtype is detected, and level 0 reads back. I confirmed locally that the old `zarr.hierarchy.Group` reference raises `AttributeError` under zarr 3 while the new check passes on the same zarr-2-format group.

pmh47:
This is failing CI under zarr 2.x, please could you take a look

erdpx:
ah, sorry, this one is one me; we'll still need to support both zarr 2.x and 3.x apis for now

robertlangdonn:
Pushed a fix — turned out to be unrelated to the `zarr.Group` check. The actual break: `LRUCacheStore(zarr.storage.WrapperStore)` in `utils.py` is a module-level class statement, and `WrapperStore` doesn't exist in zarr 2.x, so `import vesuvius.data` raised `AttributeError` at collection time and took down every test file, not just the cache ones (that's why it looked broad). Confirmed by running the exact CI matrix locally: reproduced the `AttributeError: module 'zarr.storage' has no attribute 'WrapperStore'` on the prior commit under zarr 2.18.7, and it's gone after the fix.

Changes:
- `LRUCacheStore` base falls back to `object` under zarr 2.x (same pattern already used by `neural_tracing.datasets.common._DiskCacheStoreV3` a few files over), so the module imports either way. `open_zarr(cache=True)` now raises a clear `NotImplementedError` under zarr 2.x instead of silently trying to instantiate a store that can't work.
- The zarr2/3 regression test's fixtures were also zarr-3-only (`zarr.create_array`, `open_group(zarr_format=2)`), so they never actually ran under zarr 2.x despite existing to cover that case — branched those on `zarr.__version__` so they exercise both.

Ran the full `test_lru_cache_store.py` + `test_open_zarr_format.py` under both zarr 2.18.7 and 3.2.1 locally (12 passed under 3.x; 7 passed + 5 correctly-skipped-as-zarr3-only under 2.x) and did a `--collect-only` sweep of the whole `tests/` tree under zarr 2.18.7 to confirm nothing else regresses.

bruniss:
hey, thanks for the PR. zarr has a built-in cachestore similar to the old LRUCacheStore , details here : https://zarr.readthedocs.io/en/stable/user-guide/experimental/#cachestore (for these repeated ones either a persistent on-disk or memory store would make sense) 

if we're making changes for zarr3 compat i'd prefer to use that. 

also, can you split the multiscale fix from the zarr 3 cache store compat

robertlangdonn:
Split as requested. This PR is now scoped to just the multiscale-access fix (the zarr 3 string-key requirement that broke `Volume` outright — `_level()`/`_num_levels()` accessors, regression test on a zarr v2 multiscale group). Force-pushed the branch down to that; the earlier commits (cache store, CI fix) are no longer here.

The cache-store compat work is now its own PR: #1373, rebuilt on zarr's built-in `zarr.experimental.cache_store.CacheStore` (wrapping a `MemoryStore`) instead of the hand-rolled `LRUCacheStore`, per your suggestion above. It's stacked on this branch — depends on this PR merging first.

Also worth noting for this PR specifically: the `zarr.hierarchy.Group` → `zarr.Group` fix from earlier in this thread (@erdpx's original comment) turned out to already be fixed independently on `main` (#1203, @bruniss) since this PR was opened, so it's dropped from this diff as moot — the remaining fix here is the integer-key group access (`self.data[0]` etc.), which is still broken on `main`.

PR #1180, opened 2026-07-20
community_projects: add Volume chunk cache + chunk-size study, and a readability gate

robertlangdonn:
Two community-project listings under existing sections of `20_community_projects.md`:

**Data access** — [Volume read cache + chunk-size study](https://github.com/ScrollPrize/villa/pull/1177): fixes zarr-3 multiscale breakage in the Python `vesuvius` `Volume` and restores an opt-in LRU chunk cache (~300× faster repeat reads, byte-identical), plus a measured chunk-size × access-pattern study (read amplification 1.3×–176×).

**Segmentation** — [Readability gate](https://github.com/robertlangdonn/vesuvius-readability-gate): a CPU-only readability-risk diagnostic that scores depth occupancy over the ink model's 62-layer render window to predict whether a candidate surface can yield legible ink *before* spending inference GPU. Validated on public data (readable PHerc1667 0.35 vs unreadable PHerc1203 0.97) with a same-scroll geometry control.

Both link to public, reproducible work. Happy to split into two PRs if preferred.

### DIFF
--- a/scrollprize.org/docs/20_community_projects.md
+++ b/scrollprize.org/docs/20_community_projects.md
@@ -70,6 +70,8 @@ For state-of-the-art updates join our [Discord server](https://discord.com/invit
 
 - [llfio-chunkloader](https://github.com/climbmax123/LLFIOCunkloadingTestingAndBenching): A method to access data in chunks of (x,y,z) that is much faster and more compute-efficient than Zarr. (Written in C++ but it is possible to integrate in Python).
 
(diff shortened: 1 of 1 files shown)


erdpx:
thanks for putting these together! we’re closing this for now. the readability gate computes depth occupancy from rendered layers, but this isn’t enough to predict readability. also, the rendered inputs and scripts used for the validation and same-scroll control are not included, so we couldn’t reproduce the reported results

PR #1192, opened 2026-07-21
Fix/predict3d channel major page release

### LINKED ISSUE #1183: predict3d: grad_mag silently loses 16 of 32 z-slices in a chunk, files present
Opposite failure mode to #1114 / #1173: not phantom output where there's no input, but missing output where there is input.

## Symptom

16 of the 32 z-slices inside an output chunk come out zero. Chunk files are all present, every tile reports as processed, no warning. Only `grad_mag` is hit — `nx`, `ny` and `cos` are correct in the same slices of the same run.

## Reproducer

Input: `s3://vesuvius-challenge-open-data/PHercParis4/volumes/20260411134726-2.400um-0.2m-78keV-masked.zarr` read at pyramid level 2. Checkpoint `20260419_180421_conthr_1e-5_warp_2um_noss_dist` from the lasagna HF repo, code at ff85612. Same z0, only depth differs.

    # clean
    lasagna-preprocess predict3d --input paris4.zarr/2 \
      --output clean.lasagna.json \
      --unet-checkpoint lasagna_2um_ff85612.pt \
      --crop 0 0 40000 32693 32693 640 --cos-scaledown 2 --scaledown 2

    # grad_mag level 3, z 5040-5055 all zero
    lasagna-preprocess predict3d --input paris4.zarr/2 \
      --output holed.lasagna.json \
      --unet-checkpoint lasagna_2um_ff85612.pt \
      --crop 0 0 40000 32693 32693 1280 --cos-scaledown 2 --scaledown 2

The shallow run covers that z range correctly, so the data exists and the model produces it.

## What I ruled out

- Not crop alignment. `z0 40000` and `z0 39936` have different chunk alignment and lose the same absolute range, out z 5040-5055.
- Not missing input. Level-2 input chunk-row 78 (z 9984-10111) has 1289 chunk files, more than its neighbours.
- Not accumulator pressure. The accumulator spans the full output in every run, holed or clean.
- Depth is the trigger: 640 clean, 1280 holed, same z0.

The lost range ends on an input chunk-row boundary — out 5040-5055 is input 10080-10111, and 10112 = 79 x 128. An earlier run with a different crop lost a range that started at that same boundary.

I couldn't isolate the mechanism. One thing I noticed: the existence check in `_on_z_complete` (preprocess_cos_omezarr.py:2648) is keyed on `gm_omezarr_path`, so a half-written grad_mag chunk would make it return True while the other channels keep writing. That fits the channel specificity, but I may be misreading the flush logic.

## Why it's easy to miss

Counting chunk files doesn't catch it. Nonzero fraction per z-slice does, and that check has to run before normalization — same point raised in #1173.

14 human annotation points landed in one lost row in my data. Rays crossing it integrate to zero and act as leverage in a least squares fit; a feature scoring 85.6% on intact data went to 0.0%. Nothing failed loudly.

Happy to open a PR if you want one, or leave it with you if the flush window is deliberate.

NanokodasKarolis:
## Summary

  Fixes #1183.

  predict3d treated a completed Z-band in a (C, Z, Y, X) memmap as one
  contiguous byte range. Because the array is channel-major, this could release
  pages belonging to other slices and produce empty grad_mag bands.

  This change releases the completed range separately for each channel and adds
  regression coverage for the affected page ranges.

  ## Validation

  Reproduced the issue using the reporter’s full command:

  - Unfixed: grad_mag was empty at Z 5040–5055.
  - Fixed: no empty slices; the affected region remained populated.
  - Runtime showed no regression: 1,823 seconds fixed versus 1,956 seconds
    unfixed on equivalent H100 pods.


NanokodasKarolis:
Verification update: I byte-compared the fixed vs. unfixed outputs. Outside the corrupted region they're identical - cos untouched, control slices equal to the pixel - and every difference is a repair (grad_mag 0 → populated). Good data never changed

One thing I noticed: the corruption was a bit broader than the grad_mag holes. The same overshoot also zeroed small bits of nx/ny in neighboring bands (a different z per channel the byte-stride tell), and the fix repairs those too.

Still checking that it also clears the second reported band (z 5296-5311). Lacking the compute power to produce test results faster.

NanokodasKarolis:
Verification update: the second reported affected band with a full base-vs-fixed run. The fix restores grad_mag at Z 5296–5311 and corrects the broader affected flush region. All output outside the affected regions is byte to byte identical.

 Runtime was effectively unchanged: base 1609.5s vs fixed 1612.2s. Inference itself was 14.9s faster with the fix.
 
I am done with the verification and waiting for comments if anything needs to be improved before this can be merged.

pscamillo:
The nx/ny finding corrects my issue — I reported grad_mag only, since that's
where the holes were large enough to break my downstream fit. Different z per
channel from the byte stride makes sense.

Worth noting for anyone using the scanner I linked: it works by comparing
channels against each other, so it catches this because the affected z differs
per channel. A case where all four were zeroed at the same z would slip past it.
Your byte-comparison against an unpatched run is the stronger check.

From the reporter's side this looks complete: both bands restored, byte-identical
outside the affected regions, no runtime cost. I ran the 640/1280 pair on my end
with the same result.

NanokodasKarolis:
Additional validation with qa_holescan.py:

  lasagna-preprocess predict3d \
    --input paris4.zarr/2 \
    --output deep.lasagna.json \
    --unet-checkpoint lasagna_2um_ff85612.pt \
    --crop 0 0 40000 32693 32693 2368 \
    --cos-scaledown 2 --scaledown 2

  Unpatched ff856121:

  checked rows 156-165 (full scan)
  HOLE z 5040-5055 (16 of 32 slices in row 157), empty: grad_mag
  VERDICT: FAIL

  Current PR head a1c0102f, same command:

  checked rows 156-165 (full scan)
  VERDICT: PASS

  The patched default-scaledown run also passed an exhaustive scan of the
  affected same-scale channels:

  checked rows 78-80 (full scan)
  VERDICT: PASS

hendrikschilling:
@codex review

hendrikschilling:
Given the bug was the result of some design considerations I opted for a rewrite of that part of the inference.
Thanks for pointing out the exact problem that was quite helpful!

PR #1200, opened 2026-07-21
vesuvius: fix chunked label expansion in labels_to_zarr (seams, cascade, zarr 3)

NanokodasKarolis:
## Summary

`labels_to_zarr.py`'s EDT label expansion (`--expand`) produced labels that don't match a real (unchunked) expansion, and on the pinned zarr (`>=3.2`) it crashed before running at all. This fixes both.

## Bugs

**1. Expansion didn't match a global EDT.** Each chunk expands with a read-`halo` of neighbour context and writes back only its own region, using an 8-phase (stride-2) parallel schedule.
- With the default `halo=0`, a chunk saw none of its neighbours, so labels could not expand across chunk boundaries — silent seams / under-expansion.
- Raising the halo exposed a cascade: phases run sequentially and a later chunk read an earlier chunk's freshly-written expansion (same label value) as a source, so expansions over-grew near boundaries (a distance-5 expansion reached ~17 at 8-chunk corners).

**2. The feature crashed on zarr 3.** `get_initialized_chunk_indices` used `store.keys()` (a zarr-2 Mapping API), raising `AttributeError: 'LocalStore' object has no attribute 'keys'` on zarr `>=3`.

## Fix

- Default `halo` to the expansion distance so every chunk sees the labels that reach into it.
- Tag newly-expanded voxels with a temporary marker (remapped afterwards) so only *original* labels are ever expansion sources — no cascade.
- Widen the phase stride to `ceil(halo/chunk)+1` so parallelism stays race-free when the expansion exceeds one chunk (reduces to the 8-phase scheme otherwise).
- Reimplement chunk discovery with `store.list_prefix()` (zarr 3) and a `store.keys()` fallback (zarr 2), deriving the chunk-key encoding/separator from `metadata.encode_chunk_key`.

## Validation

New `vesuvius/tests/test_labels_to_zarr_expand.py`: the chunked output now equals a single global scipy EDT for interior, face, 8-chunk-corner, and multi-boundary labels, and for an expansion larger than a chunk (parallel == serial). Also covers zarr-2/3 chunk discovery, marker collision/exhaustion/uint16, and the padding-expansion path.


erdpx:
tthanks for the pr! one remaining zarr 3 compatibility issue: the output paths still use Group.create_dataset. could you update both level 0 and pyramid creation using this pattern?

```python3
root = zarr.open_group(path, mode="w", zarr_format=2)
array = root.create_array(
    ...,
    config={"write_empty_chunks": False},
)
```

this keeps the output in zarr format 2 while using zarr-python 3

NanokodasKarolis:
> tthanks for the pr! one remaining zarr 3 compatibility issue: the output paths still use Group.create_dataset. could you update both level 0 and pyramid creation using this pattern?
> 
> ```python
> root = zarr.open_group(path, mode="w", zarr_format=2)
> array = root.create_array(
>     ...,
>     config={"write_empty_chunks": False},
> )
> ```
> 
> this keeps the output in zarr format 2 while using zarr-python 3

Thanks, working on the update.

pmh47:
This is failing CI under zarr 2.x, please could you take a look

erdpx:
ah, sorry, this one is one me; we'll still need to support both zarr 2.x and 3.x apis for now

bruniss:
closing due to staleness/not prio

PR #1179, opened 2026-07-20
VC3D Agent Bridge: MCP control surface for driving VC3D via AI agents

giorgioangel:
## Summary

- Adds an in-process JSON-RPC "Agent Bridge" to the VC3D Qt6 app (`apps/VC3D/agent_bridge/`) so an external process can drive a live VC3D instance: navigation, segmentation growth/editing (manual-add/hole-fill, corrections, push-pull), remote Open Data catalog control, Lasagna fiber-tracing optimization, Atlas fiber-intersection search, Line Annotation, tags/seeding, Run-Trace, flattening, and rendering.
- Adds a standalone Python MCP server (`tools/vc3d-mcp/`) that wraps every bridge RPC as an `@mcp.tool()` (68 tools total), registrable as MCP server `vc3d-bridge` in Claude Code or any MCP client.
- Seamless connect: registry-file auto-discovery of an already-running VC3D, auto-launch when none is running, and a self-bootstrapping `run.sh` so there's no manual venv/pip setup step between building VC3D and an agent being able to drive it.
- Closes the last known implementation gap: adds the 10 `vc3d_lasagna_*`/`vc3d_drag`/`vc3d_switch_workspace` MCP tools that SPEC.md §16 called for but were never wired up, even though the underlying C++ RPCs were already implemented and working.
- De-hardcodes two developer-machine-specific test fixture paths, fixes a stale SPEC.md doc reference, and adds a `vc3d-agent-bridge` Claude Code skill documenting cross-tool gotchas (per-source job concurrency, footguns, the editing-session precondition triad) for agents driving VC3D through these tools.

## Test plan

- [x] `tools/vc3d-mcp/test_mcp_bridge.py` (fake-bridge unit tests): 17/17 pass
- [x] Verified live against a real, running VC3D instance this session via actual MCP tool calls — e.g. `vc3d_list_catalog_samples` returned the real 41-sample Open Data manifest, `vc3d_open_volume`/`vc3d_list_segments` returned 45 real segments from the local test fixture, `vc3d_activate_segment`/`vc3d_enable_editing`/`vc3d_corrections_set_point_mode`/`vc3d_screenshot`/`vc3d_get_cursor_point`/`vc3d_atlas_status`/`vc3d_job_status` all round-tripped against live app state
[...]

PR #1213, opened 2026-07-23
fix(VC3D): remote-aware volume path for SeedingWidget Run/Expand (#1188)

### LINKED ISSUE #1188: SeedingWidget Run/Expand fail on fully-remote (streaming-only) volumes: empty volume path passed to vc_grow_seg_from_seed
## Summary

`SeedingWidget::onRunSegmentationClicked()` / `onExpandSeedsClicked()` (and their new headless twins, `runSegmentationHeadless()` / `runExpandSeedsHeadless()`, added for the VC3D Agent Bridge in #1179) pass `currentVolume->path()` directly as the `--volume` argument to the external `vc_grow_seg_from_seed` tool. For a volume that is fully remote/streaming-only (no local zarr mirror — e.g. an Open Data catalog sample attached via `catalog.open_sample`), `Volume::path()` returns an empty path, so the spawned child process receives an **empty** volume argument and fails immediately.

Found while live-testing the agent bridge's new `seeding.run` RPC against a fresh Open Data catalog sample (`PHerc0846A`). Confirmed via the VC3D log:

```
Starting seeding job 0: .../vc_grow_seg_from_seed  paths seed.json 4331.83 3073.17 7009
```

(note the double space — the volume path argument is empty). This is **not** specific to the headless bridge path: the same `currentVolume->path()` call exists in the original interactive `onRunSegmentationClicked`/`onExpandSeedsClicked`, so a human clicking "Run Segmentation" on a fully-remote volume today would hit the exact same failure.

## Root cause

`apps/VC3D/SeedingWidget.cpp`, inside `runSegmentationHeadless()` / `runExpandSeedsHeadless()` (and originally `onRunSegmentationClicked`/`onExpandSeedsClicked`):

```cpp
std::filesystem::path volumePath = currentVolume->path();
```

Compare with the correctly-working pattern already used elsewhere for the same tool. `apps/VC3D/SegmentationCommandHandler.cpp:111`:

```cpp
static QString commandPathForVolume(const std::shared_ptr<Volume>& volume)
{
    if (!volume) {
        return QString();
    }
    if (volume->isRemote()) {
        return QString::fromStdString(volume->remoteLocator());
    }
    return QString::fromStdString(volume->path().string());
}
```

`SegmentationCommandHandler::startGrowPatchFromSeed` (which backs the `segmentation.grow_patch_from_seed` bridge RPC — already verified working against real remote catalog volumes this session) uses this helper via `buildVolumeOptionList()` instead of calling `volume->path()` directly, which is exactly why *that* RPC handles remote volumes correctly while `SeedingWidget`'s Run/Expand do not.

## Suggested fix

Give `SeedingWidget` the same remote-aware volume path resolution — either:
- a small local equivalent of `commandPathForVolume` (checking `currentVolume->isRemote()` / `remoteLocator()`), or
- share/extract the existing helper from `SegmentationCommandHandler.cpp` so both call sites use one implementation.

Apply it in both `runSegmentationHeadless()` and `runExpandSeedsHeadless()` (and therefore both the interactive slots and the bridge RPCs, which delegate to them).

## Secondary, possibly-related note (lower priority, worth a look but not required for this issue)

`SeedingWidget`'s Run/Expand read `normal_grid_path` as a static field from a local `seed.json` config file — a different, older mechanism than `segmentation.grow_patch_from_seed`'s normal-grid resolution (`SegmentationCommandHandler::startGrowPatchFromSeed` resolves it dynamically via `_state->vpkg()->normalGridPaths()`). This older mechanism has no awareness of remote/streaming normal-grid stores at all. Worth deciding whether Seeding's Run/Expand should be updated to use the same dynamic resolution `startGrowPatchFromSeed` uses, once the primary volume-path fix above lands and remote volumes are reachable at all.

## Repro

1. Attach any Open Data catalog sample with no local volume mirror (e.g. `catalog.open_sample("PHerc0846A", resources: {volumeIds: ["20250728152254"], representationRefs: ["0:2"]})`).
2. Commit at least one point to a collection (`points.commit`).
3. Call `seeding.run` (or click "Run Segmentation" interactively).
4. Observe the spawned `vc_grow_seg_from_seed` process receives an empty volume path and fails instantly.

Found while verifying the `seeding.*` bridge RPCs added in #1179.

Bullo27:
## Summary

`SeedingWidget`'s **Run Segmentation** and **Expand Seeds** actions passed `currentVolume->path()` straight to `vc_grow_seg_from_seed` as the volume argument. For a fully-remote / streaming-only volume (no local zarr mirror — e.g. an Open Data catalog sample attached via `catalog.open_sample`), `Volume::path()` is **empty**, so the spawned child received an **empty** volume argument and failed immediately. This hit both the interactive slots and their headless bridge twins (`runSegmentationHeadless()` / `runExpandSeedsHeadless()`), as diagnosed in #1188.

## Fix

- Add `Volume::commandLocator()` — returns `remoteLocator()` for remote volumes and `path().string()` otherwise, i.e. the reference a command-line tool should receive for either volume kind, resolved in one place.
- Route both `SeedingWidget` call sites (`onRunSegmentationClicked` / `onExpandSeedsClicked`, which back the headless RPCs) through it.
- `SegmentationCommandHandler::commandPathForVolume()` — the existing, remote-correct helper this mirrors, and the reason `segmentation.grow_patch_from_seed` already handles remote volumes — now delegates to the same method, so the remote-vs-local choice has a single implementation (per the issue's "share/extract the existing helper" suggestion).

```
Volume::commandLocator()  ->  isRemote_ ? remoteLocator_ : path_.string()
```

30 insertions / 10 deletions across 3 files.

## Not changed — flagging a related latent site

`SeedingWidget::onNeuralTraceClicked()` (`SeedingWidget.cpp` ~L2469) has the same `currentVolume->path()` pattern, feeding `--volume_zarr` to the external `trace.py`. It would hit the identical empty-path failure on remote volumes, but the correct fix depends on whether `trace.py` accepts a remote/streaming locator for `--volume_zarr` (a different tool from `vc_grow_seg_from_seed`), which I can't verify here — so I left it untouched rather than silently hand it a URL it may not support. Happy to switch it to `commandLocator()` too if the neural tracer takes remote input.

## Verification

- `Volume::commandLocator()`'s logic is unit-compiled in isolation (`-std=c++20 -Wall -Wextra`, clean) with asserts over the local-path, remote-locator and empty-`path_` remote cases — the last being exactly this bug (`path()` empty, locator non-empty).
- **Built in-tree against the full toolchain** (Qt6 / OpenCV / Ceres / CGAL, gcc 13, Ninja, Release): `vc_core` (which includes the modified `Volume.hpp`) plus **both** changed VC3D translation units — `SeedingWidget.cpp` and `SegmentationCommandHandler.cpp` — compile cleanly, no warnings on the changed files. Since `commandLocator()` is an inline header method and `commandPathForVolume` is a file-local static, a clean compile of these callers is the complete check for this change (no separate symbols to link).
- CI remains the authoritative full-repo build.

## AI disclosure

Written by **Claude** (Anthropic Claude Code, Opus 4.8) working autonomously under the direction of **@matteobulloni_73362** (GitHub @Bullo27), who is submitting it. Disclosed per the challenge's AI-assistance guidance.

Fixes #1188.

### DIFF
--- a/volume-cartographer/apps/VC3D/SeedingWidget.cpp
+++ b/volume-cartographer/apps/VC3D/SeedingWidget.cpp
@@ -1183,7 +1183,10 @@ void SeedingWidget::onRunSegmentationClicked()
         return;
     }
 
-    std::filesystem::path volumePath = currentVolume->path();
+    // Remote-aware: use the volume's command locator so streaming-only volumes
+    // (no local mirror, e.g. Open Data catalog samples) pass a usable volume
+    // argument to vc_grow_seg_from_seed instead of an empty path (issue #1188).
+    const QString volumeArg = QString::fromStdString(currentVolume->commandLocator());
     QString workingDir = QString::fromStdString(pathsDir.parent_path().string());
 
     // Update UI
@@ -1270,7 +1273,7 @@ void SeedingWidget::onRunSegmentationClicked()
             previewParts << QString("OMP_NUM_THREADS=%1").arg(ompThreads);
         }
         previewParts << executablePath
-                     << QString("\"%1\"").arg(QString::fromStdString(volumePath.string()))
+                     << QString("\"%1\"").arg(volumeArg)
                      << QString("\"%1\"").arg(QString::fromStdString(pathsDir.string()))
                      << QString("\"%1\"").arg(QString::fromStdString(seedJsonPath.string()))
                      << QString::number(point.p[0])
@@ -1280,7 +1283,7 @@ void SeedingWidget::onRunSegmentationClicked()
         std::cout << "Starting job " << pointIndex << ": " << previewParts.join(' ').toStdString() << std::endl;
         
         const QStringList toolArgs = QStringList()
-                      << QString::fromStdString(volumePath.string())
+                      << volumeArg
                       << QString::fromStdString(pathsDir.string())
                       << QString::fromStdString(seedJsonPath.string())
                       << QString::number(point.p[0])
@@ -2170,7 +2173,10 @@ void SeedingWidget::onExpandSeedsClicked()
         return;
     }
 
-    std::filesystem::path volumePath = currentVolume->path();
+    // Remote-aware: use the volume's command locator so streaming-only volumes
+    // (no local mirror, e.g. Open Data catalog samples) pass a usable volume
+    // argument to vc_grow_seg_from_seed instead of an empty path (issue #1188).
+    const QString volumeArg = QString::fromStdString(currentVolume->commandLocator());
     QString workingDir = QString::fromStdString(pathsDir.parent_path().string());
 
     // Update UI
@@ -2257,14 +2263,14 @@ void SeedingWidget::onExpandSeedsClicked()
             previewParts << QString("OMP_NUM_THREADS=%1").arg(ompThreads);
         }
         previewParts << executablePath
-                     << QString("\"%1\"").arg(QString::fromStdString(volumePath.string()))
+                     << QString("\"%1\"").arg(volumeArg)
                      << QString("\"%1\"").arg(QString::fromStdString(pathsDir.string()))
                      << QString("\"%1\"").arg(QString::fromStdString(expandJsonPath.string()));
 
         std::cout << "Starting expansion job " << iterationIndex << ": " << previewParts.join(' ').toStdString() << std::endl;
         
         const QStringList toolArgs = QStringList()
-                      << QString::fromStdString(volumePath.string())
+                      << volumeArg
                       << QString::fromStdString(pathsDir.string())
                       << QString::fromStdString(expandJsonPath.string());
 #if defined(Q_OS_LINUX)
--- a/volume-cartographer/apps/VC3D/SegmentationCommandHandler.cpp
+++ b/volume-cartographer/apps/VC3D/SegmentationCommandHandler.cpp
@@ -113,10 +113,9 @@ static QString commandPathForVolume(const std::shared_ptr<Volume>& volume)
     if (!volume) {
         return QString();
     }
-    if (volume->isRemote()) {
-        return QString::fromStdString(volume->remoteLocator());
-    }
-    return QString::fromStdString(volume->path().string());
+    // Remote-aware volume reference; Volume::commandLocator() keeps the
+    // remote-locator vs. local-path choice in a single place.
+    return QString::fromStdString(volume->commandLocator());
 }
 
 static QString findExecutable(
--- a/volume-cartographer/core/include/vc/core/types/Volume.hpp
+++ b/volume-cartographer/core/include/vc/core/types/Volume.hpp
@@ -103,6 +103,21 @@ public:
     [[nodiscard]] std::filesystem::path remotePersistentCachePath() const;
     [[nodiscard]] std::filesystem::path path() const noexcept { return path_; }
 
+    /**
+     * @brief Locator to pass to external command-line tools (e.g.
+     * @c vc_grow_seg_from_seed ) as the volume argument.
+     *
+     * Remote/streaming-only volumes have no local filesystem path, so path()
+     * is empty for them; a caller that forwards path() straight to a child
+     * process hands it an empty argument and the tool fails immediately. This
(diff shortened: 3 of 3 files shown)


Bullo27:
Update — full local build confirmation: the complete **`VC3D` target compiles and links cleanly** against the system toolchain (Qt6 / OpenCV / Ceres / CGAL, gcc 13, Ninja / Release), producing `bin/VC3D` with both changed translation units built in. So this is verified end-to-end locally (compile **and** link), not just the isolated-logic check. CI remains the canonical cross-platform build.

Bullo27:
Follow-up: I've now verified this end-to-end against the live Open Data bucket, not just at compile time.

Using the exact positional invocation `SeedingWidget` builds (`<volume> <paths-dir> <seed.json> x y z`), with the locator this PR now passes through:

```
vc_grow_seg_from_seed \
  "https://vesuvius-challenge-open-data.s3.amazonaws.com/PHercParis4/volumes/20260310173927-45.532um-11.0m-110keV-masked.zarr" \
  <paths-dir> <seed.json> 1408 1408 1024
```

```
zarr dataset size for scale group 0 [4066, 2264, 2264]
seed location [1408, 1408, 1024] value is 121
...
generated surface 3189055.220292 vx^2 (66.114318 cm^2)
saving .../auto_grown_20260725091045170
```

Exit 0, real tifxyz output written, `scroll_source` and `target_volume` correctly resolved from the URL. The `s3://` form works too. So the remote plumbing behind this change (`is_remote_volume_path` → `Volume::NewFromUrl`) is sound, and passing the locator instead of an empty path is all that was missing on the GUI side.

One thing worth flagging, because it affects the same workflow: while testing this I hit a separate bug that would still bite a user following this path — a remote volume opened at **native resolution** gets `voxelsize` 0, so `vc_grow_seg_from_seed` scores the finished trace at 0.00 cm² and silently discards it. Filed as #1226 with a fix in #1227. The two are independent (that one is in `Volume::NewFromUrl` and reproduces straight from the CLI), but seeding on an Open Data volume from the GUI needs both before it produces a segment, so they may be worth looking at together.

Also for whatever it's worth: this PR currently shows no CI check runs at all — I believe fork PRs need authorization here — so the local build and the run above are the only evidence on it so far. Happy to rerun anything specific.

<sub>Verification done with Claude Code (Opus 5); the numbers above are from an actual run, not estimates.</sub>

Bullo27:
Closing this as superseded — #1179 landed the same fix (the file-local `commandPathForVolume` in `SeedingWidget.cpp`, covering both the Run and Expand paths) and closed #1188 with it. I filed this before that merged and only caught it when rebasing onto current `main` today.

One result from here that may still be worth having on record. While this was open I checked the premise the fix rests on — that `vc_grow_seg_from_seed` really does accept a remote locator in the positional volume slot — by invoking it directly against the live Open Data bucket, outside the GUI:

```
vc_grow_seg_from_seed \
  "https://vesuvius-challenge-open-data.s3.amazonaws.com/PHercParis4/volumes/20260310173927-45.532um-11.0m-110keV-masked.zarr" \
  <paths-dir> <seed.json> 1408 1408 1024

zarr dataset size for scale group 0 [4066, 2264, 2264]
seed location [1408, 1408, 1024] value is 121
generated surface 3189055.220292 vx^2 (66.114318 cm^2)
```

Exit 0, real tifxyz written, `scroll_source` and `target_volume` resolved from the URL; the `s3://` form works too. That's the tool side rather than #1179's GUI wiring, but it does confirm the remote plumbing the merged fix depends on holds up end-to-end, not just at compile time.

That run is also what turned up #1226 — a remote volume opened at native resolution gets `voxelsize` 0, so this same command scores the finished trace at 0.00 cm² and the result is silently discarded. So the path #1179 just opened up does still have one thing in the way of it.

The only part of this PR not now in `main` is moving the local/remote choice onto `Volume` so it isn't duplicated between `SeedingWidget.cpp` and `SegmentationCommandHandler.cpp` — but #1179's comment makes clear that duplication is deliberate and idiomatic here, so I'll leave it be.

<sub>Claude Code (Opus 5); the run above is real output.</sub>

PR #1223, opened 2026-07-24
Fiber annotation sync-conflict hardening: content-aware vc_sync + 3-way merge, non-destructive VC3D loading

skyward7187:
Multiple annotators now link fibers across machines and locations, and a last-writer-wins sync recently clobbered a fiber file (`kb_..._000141.json`): a divergent copy overwrote the newer one, erasing reciprocal link entries and shifting control-point indices. VC3D's loader then amplified the damage — parse-time checks silently deleted branches (and rewrote files on mere load), and the strict validator removed whole fibers, cascading until 7 fibers vanished from the session.

This PR attacks both halves: make the sync detect and resolve divergence instead of silently overwriting, and make VC3D loading degrade gracefully instead of destroying data. Two commits, independently reviewable.

## Commit 1 — vc_sync: detection, safety copies, fiber-aware 3-way merge

All scoped to `*.json` under 16 MB (`_should_hash`); volumes and other bulk data keep the existing size/mtime behavior byte-for-byte.

**Content-aware change detection**
- `.s3sync.db` gains a `local_md5` column (lazy `ALTER TABLE`; old DBs upgrade in place, backfill happens opportunistically).
- Touch/re-save with identical bytes is no longer a change; an mtime-preserving edit is; both-changed-but-identical converges to SKIP; and **untracked same-size divergent files now CONFLICT** instead of being silently declared in sync (one of the silent-clobber paths behind the incident).
- Transfer verification upgraded from size-only to md5-vs-ETag where the ETag is a plain MD5 (multipart/SSE-KMS ETags safely degrade to today's checks).

**Nothing is ever silently lost**
- `.s3sync-base/` keeps each annotation file's last-synced content — the merge base. Seeded by every verified transfer (run `update` once while in sync to backfill); `reset` deletes bases it can't verify so they can't poison merges.
- `.s3sync-conflicts/` receives timestamped copies before any overwrite of divergent content: the losing side of an interactive resolution, a divergent local file before a download, and every pre-merge input. Both dirs are dot-prefixed, which the existing ignore rules already exclude from syncing (locked in by tests).

**Three-way merge (`fiber_merge.py`, pure functions, no I/O)**
Auto-applied in the sync conflicts loop when both sides changed a `vc3d_fiber` file:
- `control_points`: diff3 with **position-based alignment** against the base — index shifts from insertions/deletions are invisible; regions edited on one side merge; only the same points edited differently on both sides conflict.
- `branches[]` (cross-fiber links): set semantics — additions from both sides survive (the incident case), base-aware deletions, approvals beat both pending and deletion.
- `line_points`: taken from the geometry-owning side, or spliced at anchor control points; if splicing isn't safe, the local line is kept and tagged `needs_reoptimization` for VC3D to re-fit (the optimizer needs the volume and can't run here).
- `tags` merge base-aware; `generation = max + 1`.
- Base comes from the shadow copy (md5-verified) or, for conflicts predating the shadow dir, from **S3 version history** by the tracked ETag.
- Clean merges upload through the normal path; genuine conflicts stash local+remote+base and fall back to the existing `[l]/[r]/[s]` prompt — the merge never guesses.
- `sync --dry-run` prints a per-conflict merge preview; `--no-auto-merge` disables; `status` counts merge candidates; a missing `fiber_merge.py` degrades gracefully to manual conflicts.

## Commit 2 — VC3D: non-destructive fiber loading

- `loadFiberJson` preserves structurally unparseable branch entries verbatim (re-emitted on save) — loading can never drop data — and no longer does semantic checks.
- New `resolveLoadedFiberLinks` (replaces `collectLoadedFiberBranchIssues` / `validateLoadedFiberLinks` / load-path repair): re-resolves link endpoints **by stored position** (an index shifted by an external edit heals silently, using the same `matchingStoredControlPointIndex` logic the live-edit paths already use), refreshes endpoint directions from current line tangents (re-optimization drift is not an error), and checks reciprocity by position. Unresolvable links are **quarantined**: kept on disk exactly as loaded, excluded from the link graph/sessions/link editing, shown as `(!N)` in the fiber panel's Link column. No cascade; loading performs **zero writes**.
- The "Broken branch links" repair-or-strict prompt becomes an informational "Quarantined branch links" dialog; its explicit **Repair now** button is the only path that removes quarantined entries (honors the Agent Bridge `_errorDialogsSuppressed` mode by defaulting to keep-quarantined).
- Save paths re-attach quarantined branches and unparsed entries so a legitimate save can't drop what the loader preserved.

Replayed against the incident: the index shifts self-heal silently and exactly one genuinely-broken link gets flagged — instead of 7 fibers disappearing and files being rewritten on load.

## Testing

- `pytest scripts/tests/` — **57 passed**: 21 merge-algorithm tests (incl. the incident as a fixture: base {L1}, local +L2, remote +L3 → all three survive with re-resolved indices; overlap/delete-vs-modify/same-end-extension conflicts; pending→approved precedence; splice paths; tolerance bounds) + analyze_changes hash matrix, shadow/stash behavior, and `_attempt_auto_merge` integration with a stubbed remote.
- C++ suite: **113/113 pass** after rebasing onto the Agent Bridge changes; VC3D builds clean.
- Manually verified earlier on the real PHercParis4 fiber set that the loader repairs the incident files by position with no dialog and no disk writes.

Deliberately **not** in this PR (designed, documented for later): the in-app save-time staleness guard — a stale open VC3D session can still overwrite a synced file; the sync-side conflict copies are the safety net for that case.

### DIFF
--- a/volume-cartographer/apps/VC3D/CFiberWidget.cpp
+++ b/volume-cartographer/apps/VC3D/CFiberWidget.cpp
@@ -720,11 +720,9 @@ void CFiberWidget::rebuildModel()
         };
         applyRowMetadata(row, fiber.id, false, fiber.alignment, showMetrics);
         if (fiber.quarantinedLinkCount > 0) {
-            const QString quarantineTip =
+            row[kLinkColumn]->setToolTip(
                 tr("%1 quarantined link(s): kept on disk, disabled this session")
-                    .arg(fiber.quarantinedLinkCount);
-            row[kLinkColumn]->setToolTip(quarantineTip);
-            row[kPendingColumn]->setToolTip(quarantineTip);
+                    .arg(fiber.quarantinedLinkCount));
         }
 
         QStandardItem* root = row[kNameColumn];
--- a/volume-cartographer/apps/VC3D/CFiberWidget.cpp
+++ b/volume-cartographer/apps/VC3D/CFiberWidget.cpp
@@ -696,12 +696,17 @@ void CFiberWidget::rebuildModel()
     }
 
     for (const auto& fiber : _fibers) {
+        QString linkText = fiber.linkedFiberCount > 0
+            ? QString::number(fiber.linkedFiberCount)
+            : QString();
+        if (fiber.quarantinedLinkCount > 0) {
+            linkText += QStringLiteral(" (!%1)").arg(fiber.quarantinedLinkCount);
+            linkText = linkText.trimmed();
+        }
         QList<QStandardItem*> row{
             readOnlyItem(displayNameForFiber(fiber)),
             readOnlyItem(directionForFiber(fiber)),
-            readOnlyItem(fiber.linkedFiberCount > 0
-                             ? QString::number(fiber.linkedFiberCount)
-                             : QString()),
+            readOnlyItem(linkText),
             readOnlyItem(fiber.pendingLinkCount > 0
                              ? QString::number(fiber.pendingLinkCount)
                              : QString()),
@@ -713,6 +718,13 @@ void CFiberWidget::rebuildModel()
             readOnlyItem(formatMaxMetric(fiber.alignment, showMetrics)),
         };
         applyRowMetadata(row, fiber.id, false, fiber.alignment, showMetrics);
+        if (fiber.quarantinedLinkCount > 0) {
+            const QString quarantineTip =
+                tr("%1 quarantined link(s): kept on disk, disabled this session")
+                    .arg(fiber.quarantinedLinkCount);
+            row[kLinkColumn]->setToolTip(quarantineTip);
+            row[kPendingColumn]->setToolTip(quarantineTip);
+        }
 
         QStandardItem* root = row[kNameColumn];
         for (const auto& span : fiber.spans) {
--- a/volume-cartographer/apps/VC3D/CFiberWidget.cpp
+++ b/volume-cartographer/apps/VC3D/CFiberWidget.cpp
@@ -696,13 +696,14 @@ void CFiberWidget::rebuildModel()
     }
 
     for (const auto& fiber : _fibers) {
-        QString linkText = fiber.linkedFiberCount > 0
-            ? QString::number(fiber.linkedFiberCount)
-            : QString();
-        if (fiber.quarantinedLinkCount > 0) {
-            linkText += QStringLiteral(" (!%1)").arg(fiber.quarantinedLinkCount);
-            linkText = linkText.trimmed();
+        QStringList linkParts;
+        if (fiber.linkedFiberCount > 0) {
+            linkParts << QString::number(fiber.linkedFiberCount);
         }
+        if (fiber.quarantinedLinkCount > 0) {
+            linkParts << QStringLiteral("(!%1)").arg(fiber.quarantinedLinkCount);
+        }
+        const QString linkText = linkParts.join(QLatin1Char(' '));
         QList<QStandardItem*> row{
             readOnlyItem(displayNameForFiber(fiber)),
             readOnlyItem(directionForFiber(fiber)),
--- a/volume-cartographer/apps/VC3D/CFiberWidget.hpp
+++ b/volume-cartographer/apps/VC3D/CFiberWidget.hpp
@@ -61,6 +61,8 @@ public:
         std::vector<std::string> tags;
         int linkedFiberCount = 0;
         int pendingLinkCount = 0;
+        // Links kept on disk but disabled this session (shown as "(!N)").
+        int quarantinedLinkCount = 0;
     };
 
     explicit CFiberWidget(QWidget* parent = nullptr);
--- a/volume-cartographer/apps/VC3D/CWindow.cpp
+++ b/volume-cartographer/apps/VC3D/CWindow.cpp
@@ -7655,6 +7655,7 @@ void CWindow::CreateWidgets(void)
                     fiber.tags,
                     fiber.linkedFiberCount,
                     fiber.pendingLinkCount,
+                    fiber.quarantinedLinkCount,
                 });
             }
             if (_fiberWidget) {
--- a/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
+++ b/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
@@ -9863,10 +9863,19 @@ nlohmann::json LineAnnotationController::fiberSaveSnapshotToJson(
     for (const auto& branch : serialized.branches) {
         if (branch.linkState == LinkState::Quarantined) {
             // Quarantined links are written back exactly as loaded — they
-            // failed resolution, not serialization, and dropping them here
-            // would be the data loss the quarantine exists to prevent. Note
-            // the check runs before every validity test below (a missing
-            // target fiber leaves branchFiberId == 0, for instance).
+            // failed resolution, not serialization, and dropping or
+            // altering them here would be the data loss the quarantine
+            // exists to prevent. The original JSON is used verbatim:
+            // loading and resolution normalize directions, sanitize
+            // branch_file, and remap branch_fiber_id to a runtime id, none
+            // of which may leak into the file. Note the check runs before
+            // every validity test below.
+            if (!branch.sourceJson.is_null() && !branch.sourceJson.empty()) {
+                root["branches"].push_back(branch.sourceJson);
+                continue;
+            }
+            // Fallback (should not happen: quarantine is only assigned to
+            // loaded branches, which carry their source JSON).
             nlohmann::json quarantinedJson = {
                 {"control_point_index", branch.controlPointIndex},
                 {"branch_fiber_id", branch.branchFiberId},
@@ -10506,6 +10515,9 @@ std::optional<LineAnnotationController::StoredFiber> LineAnnotationController::l
                         normalizedOrZero(branch.controlPointDirection);
                     branch.branchControlPointDirection =
                         normalizedOrZero(branch.branchControlPointDirection);
+                    // Retained so a later quarantine can serialize the entry
+                    // exactly as it appeared on disk.
+                    branch.sourceJson = branchJson;
                     fiber.branches.push_back(std::move(branch));
                 } catch (const std::exception& ex) {
                     preserveUnparsed(branchJson, ex.what());
--- a/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
+++ b/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
@@ -791,6 +791,23 @@ bool branchReferencesFiber(const LineAnnotationController::FiberBranchRef& branc
     return !fileName.empty() && branch.branchFileName == fileName;
 }
 
+// Quarantined links stay on disk but are excluded from the link graph,
+// sessions, and link editing until repaired or re-resolved.
+bool branchLinkActive(const LineAnnotationController::FiberBranchRef& branch)
+{
+    return branch.linkState != LineAnnotationController::LinkState::Quarantined;
+}
+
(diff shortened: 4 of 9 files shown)


bruniss:
@codex review

skyward7187:
Thanks for the review — all six findings were legitimate and all six are addressed in the two follow-up commits (`fiber_merge`/`vc_sync` fixes and the VC3D serialization fix). Point-by-point, with what changed where:

### [P1] Preserve structurally unparseable branches during auto-merge — **fixed**

`fiber_merge.py` now partitions branch entries into *structured* (dict with a string `branch_file` and finite 3-vector positions) and *opaque* (everything else, including partially-structured dicts that would previously have reached `_dist2()` and thrown). Opaque entries are merged by whole-value comparison, base-aware: identical on both sides → carried through; changed on one side only → that side wins (including deletion); divergent → the merge **fails** with a conflict rather than reporting clean. Your probe now yields `merged_branches == ['unparseable-link']`, and the divergent case falls back to manual resolution.

Additionally, `_attempt_auto_merge()` wraps the merger in try/except — any unexpected merge exception degrades to the interactive prompt instead of aborting the sync run.

New tests: `test_opaque_branch_entries_survive_merge`, `test_divergent_opaque_entries_conflict`, `test_opaque_entry_deleted_on_one_side_merges`, `test_malformed_dict_branch_is_opaque_not_a_crash`.

### [P1] Do not modify the local annotation before the user confirms — **fixed**

Merging is now plan/apply: during conflict processing `_attempt_auto_merge()` writes the merged JSON to a pending temp under the tool-owned `.s3sync-conflicts/.tmp/` and returns a plan; the local file is untouched. Only after `Continue? [y/N]` succeeds does `_apply_pending_merges()` stash the pre-merge local and remote copies and atomically swap the merged result into place. Answering `N` runs `_discard_pending_merges()` and prints "Sync cancelled. No files were modified." — which is now literally true for merges.

One deliberate exception: when a merge **fails**, the local/remote/base triple is still stashed at analysis time. That path never mutates the local file — the copies land in the tool-owned conflicts dir purely so a human can do a manual 3-way later, and they're useful even if the sync is cancelled.

New tests: `test_clean_merge_deferred_until_applied`, `test_cancelled_merge_discarded_without_touching_local`.

### [P1] Require checksum proof instead of inferring MD5 from ETag shape — **fixed**

The model is now exactly the equal/different/unknown trichotomy you suggested, with the ETag demoted to **positive evidence only**:

- **Transfer verification** no longer compares md5-vs-ETag at all (this removes the infinite-retry failure mode on SSE-KMS buckets entirely). Uploads/downloads verify by size as before — the CLI tools already checksum the transfer itself — and the tracked `local_md5` / shadow base are taken from the actual local bytes (uploads: only when the file was stat-stable across the upload, so the recorded hash matches what reached S3; downloads: the downloaded bytes *are* the synced content by definition).
- **Untracked same-size hashed files**: a matching plain-MD5-shaped ETag still short-circuits as equal, but any other case (mismatch *or* opaque/multipart ETag — previously the silent `SKIP` you flagged) is settled definitively by downloading the remote object and hashing it (`_remote_md5()`, per-run cached; annotation files are ≤250 KB so this is cheap and one-time — the pair becomes tracked afterward). If the fetch fails, the result is a CONFLICT, never a baseline. `_record_untracked_synced()` applies the same rule, so unverified pairs can no longer become tracking baselines or merge bases.

The `test_untracked_multipart_etag_keeps_size_heuristic` test you called out is gone, replaced by `test_untracked_opaque_etag_resolved_by_remote_hash` (both outcomes), `test_untracked_unverifiable_content_is_conflict`, `test_unverifiable_pair_not_healed`, and `test_opaque_etag_healed_via_remote_hash`. Non-hashed files (volumes etc.) keep the historical size heuristic byte-for-byte, covered by `test_untracked_non_hashed_file_keeps_size_heuristic`.

### [P1] Serialize quarantined links from their original JSON — **fixed**

`FiberBranchRef` now retains `sourceJson` — the exact JSON object each branch was parsed from. Quarantined links are serialized from that verbatim, so loading + saving can no longer normalize directions, sanitize `branch_file`, leak the runtime-remapped `branch_fiber_id` (your sharpest catch — that one actively corrupted the stored id), or drop fields VC3D doesn't understand. The field is in-memory only (never written as such), costs nothing for in-session-created branches, and the reconstructing fallback remains only for the theoretically-unreachable case of a quarantined branch without source JSON. Canonicalization still applies exclusively to active links on an intentional save, as you recommended.

### [P2] Reject line splices whose anchors are not on both polylines — **fixed**

`anchor_line_indices()` now requires every anchor to sit within a per-line geometric tolerance — twice that polyline's median sample spacing (floor 1e-3) — on **both** lines, in addition to the existing monotonicity requirement. Your ±100 probe now takes the fallback path: local line kept, `needs_reoptimization` tag, and an explicit "splice failed (anchors not on both polylines)" note. Boundary continuity follows from the bound (both sides match each anchor within tolerance, so the seam jump is bounded by ~2× spacing). Test: `test_splice_rejected_when_anchors_not_on_both_lines`.

### [P2] One tolerance-consistent identity rule for branch positions — **fixed**

`qpos()` rounding keys are gone. Branch identity now uses the same geometric predicate as control-point alignment: same `branch_file` + both endpoint positions within `pos_eq(..., POS_TOL)`, via explicit tolerant matching over the (tiny) lists with used-index tracking so duplicates can't double-match. Your jitter probe (`0.0000004` vs `0.0000013`) now merges to a single link with `pending: false` — the approval wins instead of duplicating. Test: `test_link_identity_is_tolerance_based_not_rounding_based`.

---

**Verification on this end** (the review machine couldn't run these): `pytest scripts/tests` — 68 passed (was 57; +11 finding-specific regression tests). Full C++ suite rebuilt and green, VC3D builds clean. Your two reproduction commands should now pass as written, module-level probes included.

On the noted limitation about missing C++ regression tests for the quarantine/load-save behavior: agreed — `StoredFiber` is private, so unit-testing `resolveLoadedFiberLinks` needs a test-access friend or extraction of the resolution logic; we deferred that deliberately to keep this PR reviewable and documented it as follow-up.

skyward7187:
Second review round addressed in `b71ccbd31` (reviewer is on a separate machine, so summarizing the changes here):

**1. Temp-file basename collision (confirmed bug) — fixed.** `_merge_tmp_path()` now mirrors the file's relative path under `.s3sync-conflicts/.tmp/`, the same layout `_shadow_path()` uses, so `pkgA/fibers/kb_000141.json` and `pkgB/fibers/kb_000141.json` get fully independent `.merged`/`.remote`/`.hashcheck`/`.base` temp files. Your exact reproduction is now a regression test (`test_same_basename_conflicts_do_not_collide`): two same-basename conflicts planned and applied in one run, asserting distinct pending paths and that each file receives its own merged content. Thanks for catching this one — cross-file content swap plus a mid-apply crash is precisely the failure class this PR exists to kill, and the fix also covers the sequential `.hashcheck`/`.base` uses for free.

**2. Line-only re-optimization silently dropped — fixed.** `merge_line_points()` now receives the base document: on the take-one-side-wholesale paths, if the discarded side's line differs from both the base line and the chosen line (i.e., a re-optimization that region ownership can't see, because no control point moved), the merge keeps the chosen line but emits a note and the `needs_reoptimization` tag. Symmetric in both directions; splice-failure notes are now self-contained sentences too. Test: `test_line_only_reoptimization_is_never_dropped_silently`.

**3. Quarantine + re-link duplication — fixed.** `makeStoredFiberSessionSnapshot()` no longer re-attaches a quarantined branch when the snapshot already holds an *active* link to the same endpoints (same `branch_file`, both positions within tolerance). Re-creating a link you saw flagged `(!1)` now supersedes the stale quarantined entry instead of writing both and resurrecting a duplicate once the reciprocal resolves.

**4. Dry-run/status network transparency — addressed.** `_remote_md5()` prints `Fetching remote copy of <path> to verify content...` per (cached) fetch, and the dry-run merge preview header now says it fetches remote copies to test-merge while leaving local files untouched.

**Cleanups**: `repairQuarantinedFiberLinks` schedules saves directly in its erase pass (the re-find loop is gone); `StoredFiber::needsSave` is documented as vestigial (kept as a repair-state marker — removing the field touches ~10 write sites for no behavioral gain, happy to do it in a follow-up if preferred); the fiber panel link text is built from a `QStringList` conditionally; the diff3 test comment is now a one-line statement; both `entrie(s)` typos are gone.

**Verification**: 70/70 Python tests (`pytest scripts/tests`, +2 new regressions), affected C++ test targets rebuilt and green, VC3D compiles clean. Also note the branch has merged current `main` (`d38688c57`, docs-only overlap).

Appreciated the depth on both rounds — especially the collision repro and the non-issues you explicitly checked and cleared.

skyward7187:
Third review round addressed in `8d336fd77` (reviewer is on a separate machine, so the changes are summarized here):

**1. Tolerance mismatch drops a valid link (confirmed bug) — fixed.** The re-resolve loop in `merge_branches` now selects the *nearest `pos_eq` match* instead of bounding by Euclidean² ≤ `POS_TOL`² — the same per-axis predicate as the LCS alignment and `_same_link`, exactly as the `_same_link` docstring already promised (it was indeed an oversight, not a choice). Your jitter repro is a regression test (`test_link_survives_jittered_geometry_within_tolerance`): 8e-7/axis jitter on one side + a link added at the exact base position on the other → the link survives with a `pos_eq`-consistent index.

**2a. False "guarded save pipeline" comment — fixed.** The comment above `repairQuarantinedFiberLinks()` now says the opposite, truthfully: the save pipeline has **no** staleness guard (deliberately deferred), a repair racing an external sync write is last-writer-wins, and vc_sync's conflict stashes are the safety net.

**2b. `update` doesn't seed tracked files — fixed (took the "natural place" option).** New `_seed_tracked_shadows()` runs in `refresh_tracking`: tracked annotation files that are verifiably in sync on both sides (local matches its tracked state, remote matches its tracked ETag/size, and content proven by ETag-as-MD5 or a remote hash) get their `local_md5` backfilled and a shadow base written. Files with any pending change on either side are skipped. So one `update` after deploying really does arm the merge base for the whole tree now, as the PR description claims. Tests: seeded / locally-changed-not-seeded / unverifiable-not-seeded.

**3. Dead complexity — trimmed to one marker.** `resolveLoadedFiberLinks` no longer sets the vestigial `needsSave`; `LinkState::Repaired` is the single record of in-memory repair (kept deliberately — it's the groundwork hook for the deferred save guard). The `needsSave` header comment now points at `linkState` for link repairs. The "no cascade" comment is softened to what actually happens: quarantine is bounded to the link *pair* (the reciprocal half quarantines too, hence two dialog notes for one broken link), and never propagates further.

**4. Nits — all taken.**
- Merge temp removal now prunes the mirrored `.tmp/` directories (`_remove_merge_tmp`), so `status`/`--dry-run` leave no residue behind.
- The non-array `"branches"` → `["junk"]` shape change on round-trip is documented at the preserve site.
- Both `updateReciprocal(s)` lambdas now skip quarantined entries (`branchLinkActive` filter), matching the stated invariant that quarantined links are excluded from link editing — previously they could be mutated in memory (harmless on disk thanks to `sourceJson`, but inconsistent).
- The quarantine tooltip is now set only on the Link column.

**Verification:** 74/74 Python tests (`pytest scripts/tests`, +4 new), affected C++ targets rebuilt and green, VC3D compiles clean.

Also appreciated section 5 — the runtime-ID-remap ordering and quarantine-survival-through-save paths were the two invariants we were most worried about keeping intact across these three rounds, and it's good to have independent confirmation they held.

skyward7187:
Superseded by #1246, which keeps all conflict handling in the sync process (loader-safe merge output + cross-file link consistency) and reduces the VC3D change to a single load-time reoptimize prompt for merged fibers. The quarantine/non-destructive-loader rework from this PR is dropped entirely.

PR #1224, opened 2026-07-24
vc3d-agent-bridge-mcp: add segment attachment and fix overlay synchronization

erdpx:
- adds MCP support for attaching local tifxyz segment sources
- makes segment attachment persistent, idempotent, and safe to retry
- rejects ambiguous segment source names before changing the project
- keeps overlay state synchronized with the rendered view and overlay controls
- restores overlay state when reopening a project
- adds focused core, MCP, and offscreen integration coverage

The overlay state could previously diverge between the bridge, controller, and active workspace. Segment attachment was also missing from the MCP surface. These changes keep the shared VC3D operations authoritative and leave bridge-specific validation in the bridge.

Verified with 163 host tests, 3 focused CTests, and the offscreen VC3D smoke suite (119 descriptors and 462 probes).

### DIFF
--- a/volume-cartographer/apps/VC3D/CWindow.cpp
+++ b/volume-cartographer/apps/VC3D/CWindow.cpp
@@ -2724,12 +2724,6 @@ CWindow::CWindow(size_t cacheSizeGB, RenderBenchOptions benchOptions) :
                     showStatusBarMessage(message, timeout);
                 }
             });
-    connect(_state, &CState::volumeChanged, _volumeOverlay.get(),
-            [this](const std::shared_ptr<Volume>&, const std::string&) {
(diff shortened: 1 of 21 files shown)


bruniss (approved):

jrudolph:
Windows builds started failing after this commit (only did a quick investigation, could also be an adjacent one). We need to fix that and see if we can catch this earlier.

jrudolph:
It actually is [327f4b7](https://github.com/ScrollPrize/villa/commit/327f4b70f75ee8d644def444107303bcc4acb3e4) where it started failing (addition of MCP)

erdpx:
thanks @jrudolph !!

PR #1225, opened 2026-07-25
Fix two macOS crashes: lld exception unwinding in executables, and payload-less segments aborting the app

JamesDarby345:
Two crashes hit while building and running VC3D from source natively on macOS (Apple Silicon, Homebrew LLVM). Both abort the process with an uncaught exception that a handler *already in the code* should have caught. They're independent and split into one commit each, so either can be dropped.

## 1. `ld64.lld` breaks exception unwinding in executables

**Symptom** — opening any volume package:

```
libc++abi: terminating due to uncaught exception of type std::runtime_error:
Active volume is not loaded for Lasagna shape pairing
```

**Why it throws at all, and why that's fine.** `CState::setVpkg` emits before a volume is loaded, so `CWindow::updateAtlasFiberDocks` calls `resolveLasagnaForVolume` with an id that isn't in `loadedVolumes_` yet. It throws. That's an expected condition on every project open, and the `catch (...)` in `updateAtlasFiberDocks` exists precisely to swallow it.

**Why it aborts.** It never reaches that handler. From the crash report:

```
 8 libc++abi   __cxa_throw
 9 VC3D        vc3d::opendata::resolveLasagnaForVolume(...)
10 VC3D        (anonymous namespace)::resolvedLasagnaForState(CState const*)
11 VC3D        CWindow::updateAtlasFiberDocks()::$_1::operator()() const   <-- catch (...) is here
12 VC3D        CWindow::updateAtlasFiberDocks()
...
 7 libc++abi   __cxxabiv1::failed_throw(...)
 6 libc++abi   std::__terminate(void (*)())
```

`failed_throw` means unwinding found no handler, with the handler plainly on the stack.

`CMakeLists.txt` already documents this exact `ld64.lld` defect and works around it — but only for `SHARED_LIBRARY` targets, since it was first observed for cross-dylib throws. With Homebrew clang/lld 22.1.8 the same corruption occurs *within* the `VC3D` executable, which stays on lld.

**Evidence** — same objects, same dylibs, only the executable link changed:

| Linker | Volume-less project | Malformed project |
|---|---|---|
| `ld64.lld` 22.1.8 | SIGABRT | SIGABRT |
| Apple `ld` | handled, runs | handled: "Cannot open project", exit 2 |

The fix extends the existing workaround to all targets. Trade-off: some macOS link time for exceptions that work. CI doesn't catch it because it pins an older dependency set whose lld still linked executables correctly — worth knowing that the pin is currently load-bearing.

## 2. A segment with no tifxyz payload aborts the app

**Symptom** — after a partially completed Open Data sample download, clicking Reload Surfaces:

```
libc++abi: terminating due to uncaught exception of type std::runtime_error:
Failed to open TIFF: .../remote_cache/open_data/segments/.../x.tif
```

The cache held `meta.json` + `catalog-origin.json` for 15 segments but payload for only one.

**Cause.** `Segmentation::canLoadSurface()` only checks `format == "tifxyz"`, so a payload-less directory registers as a usable surface. `loadSurface()` returns a `QuadSurface` whose TIFFs are read lazily by `ensureLoaded()` on first geometry access — long after `loadSurface()`'s try/catch has returned. That first access lands inside a Qt slot:

```
SurfacePanelController::loadSurfacesIncremental
  -> CState::setSurface -> [signal] -> onSurfaceChanged
  -> updateFocusMarker -> volumeToScene -> QuadSurface::pointTo
  -> ensureLoaded -> load_quad_from_tifxyz -> throw
```

Nothing in that chain catches. The identical error is already survivable one frame over — `ViewerManager`'s SurfacePatchIndex task wraps its call and logs `single-surface task failed` — which is why one segment logs a warning and the next one kills the session.

**Fix, two layers:**

- **Prevention** — `canLoadSurface()` also requires `x/y/z.tif` to exist. An incomplete directory is simply not loadable, keeping the failure at registration time where `loadSurface()` already returns `nullptr` and callers handle it.
- **Containment** — `onSurfaceChanged()` becomes a thin guard around `onSurfaceChangedImpl()`, so a payload that goes missing *after* registration (deleted mid-session, corrupt TIFF, unmounted share) logs and skips the surface.

**Why the guard is inside the slot rather than a global net.** I first implemented containment as a `QApplication::notify()` override, the usual approach. It cannot work: exceptions don't propagate through Qt's dispatch frames. With a throwing `QObject::event`:

| Frames between throw and catch | Result |
|---|---|
| Ours only | caught |
| QtCore (`sendEvent`) | terminate |
| QtWidgets (`notify`) | terminate |

The crash report showed `VCApplication::notify` on the stack with its `try` active while the throw terminated inside QtWidgets frames below it. So that approach was dropped. Practical implication for this codebase: **a `catch` placed above a Qt frame is decoration** — containment must happen before control returns to Qt, which is what the existing `ViewerManager` / `LineAnnotationController` handlers already do.

## Testing

- `test_segmentation`: **13/13**, including new cases for missing and partial payloads.
- Project with a payload-less segment: reports `Loaded 0/1 ... Missing: <id>` and keeps running (previously SIGABRT).
- Project with unreadable TIFFs: surface skipped, app runs.
- Full suite: **111/113**. `test_volume_pkg_full` (volume attachment) and `fiber_save_batch_tracker::emptyBatchCompletes` fail — verified **pre-existing** by stashing these changes, rebuilding clean, and reproducing both.

## For review

A segment directory that exists before its TIFFs are written — a tracer mid-run — is now skipped until complete rather than registered early. The incremental reload picks it up once written, but that's the intentional behavior change most worth a second opinion.


JamesDarby345:
Review by launching the VC3D app and loading a data volume successfully; these fixes target doing that natively on macOS

nerln:
Independent confirmation from a different direction, in case it is useful.

I hit this while building the unit tests natively on an M4 (macOS 26.6, Homebrew LLVM 22.1.8, same toolchain the preset selects). Three tests abort on an uncaught exception that a handler already on the stack should have caught:

| test | uncaught type |
|---|---|
| `agent_bridge_contract` | `AgentBridgeError`, thrown straight through a `catch (...)` |
| `test_line_annotation_generated_views` | `std::runtime_error: unsupported segment_to_next metadata/tracer version` |
| `test_open_data_manifest` | `HttpStatusError: HTTP 0 fetching .zattrs` |

`agent_bridge_contract` is the clearest of the three, because its helper has both a `catch (const AgentBridgeError&)` and a `catch (...)` and lldb still shows `failed_throw`:

```
frame #5: libc++abi`__cxxabiv1::failed_throw(__cxxabiv1::__cxa_exception*)
frame #7: AgentBridgeParam::fail(QString const&, QString const&)
frame #8: AgentBridgeParam::validate(QJsonValue const&) const
frame #9: AgentBridgeMethod::validate(QJsonValue const&) const
frame #10: main
```

Applying this PR's `add_link_options(-fuse-ld=)` change to my tree and rebuilding, all three pass:

```
1/3 Test #115: agent_bridge_contract ..................   Passed    0.39 sec
2/3 Test  #17: test_line_annotation_generated_views ...   Passed    2.46 sec
3/3 Test  #54: test_open_data_manifest ................   Passed   11.52 sec
100% tests passed, 0 tests failed out of 3
```

So the corruption is not confined to the GUI paths in the description; it reaches the test executables too, and it is the reason a first `ctest` run on macOS looks considerably worse than the tree actually is.

One narrowing that might help if this ever gets filed upstream with LLVM. On this host `ld64.lld` is only broken in combination with ThinLTO, which matches the wording of the `CMakeLists.txt` comment but not the linker-only framing of the table above. Six lines, same compiler, same SDK:

```cpp
struct E { int code; };
int main() {
    try { throw E{42}; }
    catch (const E& e) { std::printf("CAUGHT code=%d\n", e.code); return 0; }
    catch (...)        { std::printf("CAUGHT via ...\n");        return 0; }
}
```

```
clang++ -O2                          -> CAUGHT code=42
clang++ -O2 -fuse-ld=lld             -> CAUGHT code=42
clang++ -O2 -fuse-ld=lld -flto=thin  -> libc++abi: terminating due to uncaught exception of type E
```

Nothing here argues against the change, which lands the same way either way. It just suggests the trigger is the pair rather than the linker alone.

For context on how I got here: I have a few separate macOS build fixes up as #1334, #1335, #1336, #1337 and #1338, and #1338 adds a `--check` flag to `build_macos.sh` that makes running the suite on macOS repeatable. This PR is what stands between that and a green run.

AlexisData:
Confirming #1 on the **published release binaries**, which I think widens the scope of this PR.

I never built anything — these are the macOS arm64 `.dmg` artifacts straight from the releases page:

- `fc25b4d-2026-07-31` (tag `stable`)
- `d0de07f-2026-08-10` (tag `latest`)

Both abort when opening an Open Data sample. Every attempt aborted: 6 crash reports, across 3 samples picked for contrasting profiles — PHerc0125 (no segments), PHerc1447 (16 segments, 8.6 µm only), PHerc1667 (20 segments, including the 1.129 µm volume). macOS 26.6.1 (25G76), M3, arm64.

Same signature as yours: frame 9 `vc3d::opendata::resolveLasagnaForVolume`, then `__cxa_throw` → `failed_throw` → `std::__terminate` → `abort`, with `CState::setVpkg` further down. The release build carries no symbols for the frames in between, so I can't show `updateAtlasFiberDocks` and its `catch (...)` directly — but both anchors match yours.

You note that CI doesn't catch this because it pins an older dependency set whose lld still linked executables correctly, and that the pin is load-bearing. If the published artifacts come out of that CI, this suggests the pin no longer holds — which moves the blast radius from "people building from source" to "anyone who downloads VC3D for macOS and follows step one of the VC3D tutorial".

Scope of what I actually measured: one machine, one OS version, three samples. No Linux or Windows check, no bisect across releases, and I haven't verified how the release artifacts are linked.

*AI disclosure: Claude Opus 5 (Anthropic), used for the crash-report analysis and for drafting this comment.*

PR #1227, opened 2026-07-25
fix(core): discover remote voxel size at native resolution (fixes #1226)

### LINKED ISSUE #1226: vc_grow_seg_from_seed: remote volumes at native resolution get voxelsize 0, so every grown segment is silently discarded
## Summary

Growing a segment from a remote (`https://` / `s3://`) volume at **native resolution** silently produces nothing. The trace runs to completion, then the result is thrown away because `area_cm2` evaluates to 0 — the volume's `voxelsize` is 0. Exit code is 0 and no output directory is created, so from the outside it looks like the tool just didn't do anything.

Adding a `#vc-base-scale=N` selector (N ≥ 1) to the *same* URL makes it work. That contrast is the whole bug.

## Reproduction

A `seed.json` with no `voxelsize` key — every parameter has a default, so this is the natural minimal file:

```json
{ "mode": "seed", "cache_size": 2000000000 }
```

Against a public Open Data volume:

```bash
VOL="https://vesuvius-challenge-open-data.s3.amazonaws.com/PHercParis4/volumes/20260310173927-45.532um-11.0m-110keV-masked.zarr"

# A) native resolution — the default
vc_grow_seg_from_seed "$VOL" out_a seed.json 1408 1408 1024

# B) identical volume and seed, with a base-scale selector
vc_grow_seg_from_seed "$VOL#vc-base-scale=1" out_b seed.json 704 704 512
```

Observed:

```
A)  voxelsize: 0
    generated surface 4437601.357701 vx^2 (0.000000 cm^2)
    discarding generated surface because area_cm2 0 is below min_area_cm 0.3
    -> out_a/ is empty, exit 0

B)  voxelsize: 91.064
    generated surface 2237848.845578 vx^2 (185.577031 cm^2)
    -> segment written
```

Case A discarded 11,228 completed patches.

## Root cause

In `core/src/Volume.cpp`, `Volume::NewFromUrl`:

- `metadata_["voxelsize"]` is initialised to `double{}` (i.e. 0), then
- the remote metadata is loaded with

```cpp
loadRemoteVolumeMetadata(remoteUrl, auth, spec.baseScaleLevel > 0)
```

That third parameter is `discoverPublicSamplePixelSize`, and inside `loadRemoteVolumeMetadata` it gates the `scan.tomo.acquisition.detector.samplePixelSize` fallback.

For the public Open Data volumes, that fallback is the **only** reachable voxel-size source. This volume's `metadata.json` has top-level keys `["masking", "scan", "zarr_export"]` — no `voxelsize`, and nothing the earlier lookup (`voxelsize`, `voxel_size_um`, `voxelSizeUm`, `pixel_size_um`, `pixelSizeUm`, `resolution_um`, checked at top level and under `scan` / `volume` / `properties` / `metadata`) can find. The value exists only at:

```
scan.tomo.acquisition.detector.samplePixelSize = 0.045532   (mm)
```

× 1000 = 45.532 µm, which matches the volume's own name (`...-45.532um-...`).

So the discovery path these volumes depend on is enabled only when the user happens to have asked for a downsampled base scale. At base scale 0 — the default — `voxelsize` stays 0, `vc_grow_seg_from_seed` computes `area_cm2 = 0`, and the `min_area_cm` check (default 0.3) drops the segment.

Local volumes are unaffected: they carry `voxelsize` in their own `meta.json`, so `Volume::New` never needs the fallback. That's probably why this hasn't surfaced before.

## Suggested fix

The two concerns look independent — whether to discover the sample pixel size isn't really a function of whether a base scale was requested. `loadRemoteVolumeMetadata` has only this one caller, so enabling it unconditionally is a one-liner:

```cpp
if (auto remoteMeta = loadRemoteVolumeMetadata(
        remoteUrl, auth, /*discoverPublicSamplePixelSize=*/true)) {
```

Measured on the same repro:

| | before | after |
|---|---|---|
| A) base scale 0 | `voxelsize: 0` → 0.00 cm² → **0 segments** | `voxelsize: 45.532` → **147.63 cm² → 1 segment** |
| B) base scale 1 | `voxelsize: 91.064` → 1 segment | unchanged |

As a cross-check, the two paths now agree on physical area for the same region — 147.63 cm² at level 0 vs 140.96 cm² at level 1, about 5% apart. A wrong voxel size would show up here as a 4× discrepancy, so this is independent evidence that 45.532 µm is the right value.

Happy to open a PR for this.

## Secondary, much smaller

`add_target_context` derives the recorded volume name via `std::filesystem::path::filename()` on the raw locator, so the selector leaks into the output metadata:

```json
"target_volume": "20260310173927-45.532um-11.0m-110keV-masked.zarr#vc-base-scale=1"
```

`RemoteUrl.hpp` already documents `RemoteVolumeSpec::sourceUrl` as the fragment-free form, so `vc::parseRemoteVolumeSpec(...).sourceUrl` looks like the intended source here. The same pattern is duplicated in `vc_grow_seg_from_segments.cpp`.

One caveat before simply stripping it: that leaked suffix is currently the *only* record in a segment's `meta.json` that its coordinates live in a downsampled space — there's no `vc_base_scale` field, and `vc_gsfs_params` only echoes the input params. So it may be worth recording the base scale as a real field in the same change rather than just dropping it. Happy to follow whichever way you'd prefer.

---

Found by running the tool against the live Open Data bucket, on a local build (Ubuntu 24.04, gcc, Qt6). The `Volume.cpp` path above is untouched by anything local to my tree.

Investigation and this write-up were done with Claude Code (Opus 5). Every claim above was checked by running it rather than by reading the code alone — including the "after" numbers, which come from a rebuilt binary.

Bullo27:
Fixes #1226.

## The bug

`vc_grow_seg_from_seed` on a remote volume at **native resolution** runs the full trace and then silently throws the result away — no output directory, exit 0.

`Volume::NewFromUrl` gated the `scan.tomo.acquisition.detector.samplePixelSize` fallback on `spec.baseScaleLevel > 0`. For the public Open Data volumes that fallback is the *only* reachable voxel-size source: their `metadata.json` has no `voxelsize`, and nothing that the earlier key lookup (`voxelsize` / `voxel_size_um` / `voxelSizeUm` / `pixel_size_um` / `pixelSizeUm` / `resolution_um`, at top level or under `scan` / `volume` / `properties` / `metadata`) can find. The value lives only at `scan.tomo.acquisition.detector.samplePixelSize`.

So at base scale 0 — the default — `metadata_["voxelsize"]` stayed at its `double{}` initialiser. Downstream, `area_cm2` came out 0 and the `min_area_cm` check (default 0.3) dropped the finished segment.

Adding `#vc-base-scale=1` to the same URL made it work, which is what isolated the gate.

Local volumes were never affected — they carry `voxelsize` in their own `meta.json`, so `Volume::New` never needs the fallback.

## The change

Whether to discover the sample pixel size is independent of whether a base scale was requested, and `loadRemoteVolumeMetadata` has exactly one caller, so it is now enabled unconditionally. Six lines, five of them comment.

Behaviour for `baseScaleLevel > 0` is unchanged (it already passed `true`).

## Verification

Run against the live Open Data bucket, `PHercParis4/volumes/20260310173927-45.532um-11.0m-110keV-masked.zarr`, same seed, `seed.json` with no `voxelsize` key:

| | before | after |
|---|---|---|
| base scale 0 (default) | `voxelsize: 0` → 0.00 cm² → **0 segments** | `voxelsize: 45.532` → **147.63 cm² → 1 segment** |
| `#vc-base-scale=1` | `voxelsize: 91.064` → 1 segment | unchanged → 1 segment |

Two independent checks that 45.532 µm is the right value:

- it matches the volume's own name, `...-45.532um-...` (0.045532 mm × 1000);
- the two paths now agree on physical area for the same region — 147.63 cm² at level 0 vs 140.96 cm² at level 1, ~5% apart. A wrong voxel size would surface here as a 4× discrepancy.

Built and run locally on Ubuntu 24.04 (gcc, Qt6, Ninja/Release): `vc_core` and `vc_grow_seg_from_seed` compile and link clean, no new warnings, and the end-to-end runs above are from that binary. A repro script is in the issue.

I did not add a unit test: the affected path needs a live HTTP fetch of a volume's `metadata.json`, and I could not find an existing offline fixture for remote volume metadata to hang one off. Happy to add one if you'd like — a pointer to the pattern you'd prefer would help.

---

Investigation, patch and this description were done with Claude Code (Opus 5), verified by running rather than by reading alone. Reported separately as #1226 in case you'd rather fix it differently — the issue also notes a much smaller, related metadata point I deliberately left out of this diff to keep it focused.

### DIFF
--- a/volume-cartographer/core/src/Volume.cpp
+++ b/volume-cartographer/core/src/Volume.cpp
@@ -1396,8 +1396,13 @@ std::shared_ptr<Volume> Volume::NewFromUrl(
 
     std::optional<double> nativeRemoteVoxelSize;
     try {
+        // The samplePixelSize fallback is the only voxel-size source for the
+        // public Open Data volumes (their metadata.json carries it solely as
+        // scan.tomo.acquisition.detector.samplePixelSize). Gating it on a
+        // base-scale selector left native-resolution remote volumes with
+        // voxelsize 0, so area computations downstream collapsed to zero.
         if (auto remoteMeta = loadRemoteVolumeMetadata(
-                remoteUrl, auth, spec.baseScaleLevel > 0)) {
+                remoteUrl, auth, /*discoverPublicSamplePixelSize=*/true)) {
             if (remoteMeta->contains("voxelsize") &&
                 (*remoteMeta)["voxelsize"].is_number()) {
                 const double value = (*remoteMeta)["voxelsize"].get_double();
--- a/volume-cartographer/core/test/test_volume_live_s3.cpp
+++ b/volume-cartographer/core/test/test_volume_live_s3.cpp
@@ -19,16 +19,24 @@ constexpr const char* kVolumeUrl =
     "s3://vesuvius-challenge-open-data/PHerc0172/volumes/"
     "20241024131838-7.910um-53keV-masked.zarr";
 
+// kVolumeUrl above is a legacy export: its meta.json carries a top-level
+// "voxelsize". Most of the bucket now ships the newer metadata.json form, where
+// the size is only under scan.tomo.acquisition.detector.samplePixelSize -- the
+// case that regressed in #1226. 45.532 um, per the volume's own name.
+constexpr const char* kNestedVoxelSizeVolumeUrl =
+    "s3://vesuvius-challenge-open-data/PHercParis4/volumes/"
+    "20260310173927-45.532um-11.0m-110keV-masked.zarr";
+
 bool requireNetwork()
 {
     const char* env = std::getenv("VC_TEST_REQUIRE_NETWORK");
     return env && env[0] && env[0] != '0';
 }
 
-std::shared_ptr<Volume> openOrSkip()
+std::shared_ptr<Volume> openUrlOrSkip(const char* url)
 {
     try {
-        return Volume::NewFromUrl(kVolumeUrl);
+        return Volume::NewFromUrl(url);
     } catch (const std::exception& e) {
         if (requireNetwork()) FAIL("Volume::NewFromUrl failed: " << e.what());
         MESSAGE("Skipping (network unavailable?): " << e.what());
@@ -36,7 +44,20 @@ std::shared_ptr<Volume> openOrSkip()
     }
 }
 
+std::shared_ptr<Volume> openOrSkip() { return openUrlOrSkip(kVolumeUrl); }
+
 } // namespace
+
+TEST_CASE("Volume::NewFromUrl: voxel size from nested acquisition metadata")
+{
+    auto v = openUrlOrSkip(kNestedVoxelSizeVolumeUrl);
+    if (!v) return;
+    CHECK(v->isRemote());
+    CHECK(v->baseScaleLevel() == 0);
+    // Without the samplePixelSize fallback this is 0, and every downstream
+    // area computation collapses with it.
+    CHECK(v->voxelSize() == doctest::Approx(45.532).epsilon(0.0001));
+}
 
 TEST_CASE("Volume::NewFromUrl: opens the PHerc 0172 zarr")
 {


Bullo27:
Test suite run locally with this patch applied, since CI doesn't run on fork PRs here:

```
100% tests passed, 0 tests failed out of 109
Total Test time (real) =  46.03 sec
```

Ubuntu 24.04, gcc, Ninja/Release, `-DVC_TESTING=ON`. That includes `test_volume_live_s3` (42.9 s — it does hit the network), which is the closest existing coverage of this path.

I still haven't added a new test for the specific behaviour, for the reason in the description: the fix depends on fetching a volume's `metadata.json` over HTTP, and I couldn't find an offline fixture for remote volume metadata to build one on. If you'd point me at the pattern you'd prefer — or if a live-S3 assertion in `test_volume_live_s3` is acceptable — I'm happy to add one.

<sub>Claude Code (Opus 5); numbers from an actual `ctest` run.</sub>

Bullo27:
Added the regression test (869fc94) rather than leaving it hanging on a round-trip — it's a self-contained commit, easy to drop or rework if you'd rather it looked different.

It adds a case to `test_volume_live_s3.cpp` against a `metadata.json`-style volume instead of repointing the existing one, so PHerc0172's pinned shape assertions are untouched. The only structural change is that the URL moves into an `openUrlOrSkip(url)` helper; `openOrSkip()` keeps its meaning, and the network-soft-skip behaviour is unchanged.

```cpp
TEST_CASE("Volume::NewFromUrl: voxel size from nested acquisition metadata")
{
    auto v = openUrlOrSkip(kNestedVoxelSizeVolumeUrl);
    if (!v) return;
    CHECK(v->isRemote());
    CHECK(v->baseScaleLevel() == 0);
    CHECK(v->voxelSize() == doctest::Approx(45.532).epsilon(0.0001));
}
```

Confirmed it actually discriminates — reverting just `Volume.cpp` to the parent commit, keeping the test:

```
test_volume_live_s3.cpp:59: FAILED:
  CHECK(v->voxelSize() == doctest::Approx(45.532).epsilon(0.0001))
exit 1
```

and with the fix restored: `10 test case(s) passed`, exit 0. Both runs with `VC_TEST_REQUIRE_NETWORK=1`.

Context for the choice of volume, from the survey in #1226: of the 71 volumes in the open-data catalog, PHerc0172/20241024131838 — the one this file already pinned — is the *only* one whose voxel size the ungated lookup resolves. 56 need the fallback. So the existing live test was, by coincidence, exercising the single case that already worked.

<sub>Claude Code (Opus 5); before/after both from real runs.</sub>

Bullo27:
Closing as superseded: the metadata rework in #1302/#1310 replaced the gated `loadRemoteVolumeMetadata` call this PR patched — the `samplePixelSize` fallback now applies unconditionally (`core/src/Volume.cpp:144-160`), which achieves this PR's goal by different means. Verified on today's main (f74d09f84): the original repro volume discovers `voxelsize: 45.532` and #1403's volume discovers `voxelsize: 8.64`, over both `https://` and `s3://` — full transcripts in https://github.com/ScrollPrize/villa/issues/1226#issuecomment-5260191844.

One piece of this branch may still be useful independently: the live-S3 regression test (`test_volume_live_s3` case asserting `voxelSize() == 45.532` for a `metadata.json`-style volume — the class where only `scan.tomo.acquisition.detector.samplePixelSize` exists). It was proven to discriminate in July and would pin the reworked discovery against future regressions. Happy to rebase just that test into a fresh PR if wanted — otherwise nothing left to merge here.

<sub>Analysis and verification with AI assistance (Claude).</sub>

PR #1228, opened 2026-07-25
fix(VC3D): pass voxel size to vc_render_tifxyz for remote volumes

Bullo27:
Renders of remote (Open Data) volumes come out with wrong physical scale, because VC3D never forwards the volume's voxel size to `vc_render_tifxyz`.

## Why

`vc_render_tifxyz` recovers the voxel size from the volume's local `meta.json` (`readVolumeVoxelSize`). A streaming-only remote volume has no local `meta.json`, so it falls back to:

```
Voxel size: 1.0 (no metadata found; override with --voxel-size)
```

VC3D holds the right value, but only forwarded it when the volume was rebased or explicitly overridden:

```cpp
renderVolume->baseScaleLevel() > 0 || renderVolume->hasExplicitVoxelSizeOverride()
```

A native-resolution remote volume — the default when you open an Open Data sample — matches neither.

## What it costs

Not just missing scale, but confidently wrong scale. Same segment, same remote volume, with and without `--voxel-size 45.532`:

**OME-Zarr (`--zarr-output`), `.zattrs`:**

| | axes units | level-0 scale |
|---|---|---|
| without | `nanometer` | `[1.0, 1.0, 1.0]` |
| with | `micrometer` | `[45.532, 45.532, 45.532]` |

The unpatched output declares **1 nm voxels for 45.532 µm data** — off by a factor of 45,532, asserted to anything downstream that reads the multiscale metadata.

**TIFF (`--tif-output`):**

| | XResolution | ResolutionUnit |
|---|---|---|
| without | absent | absent |
| with | 9139805/16384 ≈ 557.86 DPI | inch |

(25400 µm/inch ÷ 45.532 = 557.85.)

## The change

Add `isRemote()` to the condition. `setRenderVoxelSize()` already discards a non-positive value, so this stays inert whenever the size is unknown and cannot regress a render that works today.

## Relationship to #1226 / #1227

Separate defect, but the pair is needed for the GUI path:

- #1227 makes `Volume::voxelSize()` return the real value for a native-resolution remote volume (today it is 0);
- this PR makes VC3D forward that value to the renderer.

Without #1227 this change is a no-op, by design — the guard above sees 0 and skips.

Verified in two halves, both by running:

- with #1227 applied, `vc_grow_seg_from_seed` on that volume prints `voxelsize: 45.532`, and that number is read straight from `volume->voxelSize()` — the same accessor this call site uses;
- passing `--voxel-size 45.532` to `vc_render_tifxyz` produces the µm scale and DPI in the tables above.

I did not drive the Qt GUI itself; the one-line call site changed here is the link between those two halves, and I'd rather say that plainly than imply a test I didn't run.

Built on Ubuntu 24.04 (gcc, Qt6, Ninja/Release): VC3D compiles and links clean, no new warnings. Full suite on this branch: **109/109 passing**.

---

<sub>Investigation, patch and description done with Claude Code (Opus 5); every number above is from an actual run.</sub>

### DIFF
--- a/volume-cartographer/apps/VC3D/SegmentationCommandHandler.cpp
+++ b/volume-cartographer/apps/VC3D/SegmentationCommandHandler.cpp
@@ -2055,11 +2055,17 @@ void SegmentationCommandHandler::onRenderSegment(const std::string& segmentId)
     // Interactive renders always start from the TIFF-stack default.
     _cmdRunner->setRenderOutputFormat(CommandLineToolRunner::RenderOutputFormat::TifStack);
     _cmdRunner->setRenderParams(static_cast<float>(dlg.scale()), dlg.groupIdx(), dlg.numSlices());
+    // Remote volumes have no local meta.json for vc_render_tifxyz's
+    // readVolumeVoxelSize() to read, so the size has to come from here or the
+    // render falls back to "no metadata found" and writes TIFFs with no
+    // resolution tag at all. setRenderVoxelSize() already ignores a
+    // non-positive value, so this stays inert when the size is unknown.
     _cmdRunner->setRenderVoxelSize(
         renderVolume ? renderVolume->voxelSize() : 0.0,
         renderVolume &&
             (renderVolume->baseScaleLevel() > 0 ||
-             renderVolume->hasExplicitVoxelSizeOverride()));
+             renderVolume->hasExplicitVoxelSizeOverride() ||
+             renderVolume->isRemote()));
     _cmdRunner->setNextOmpThreads(dlg.ompThreads());
     _cmdRunner->setVolumePath(dlg.volumePath());
     const bool useRemoteVolume = dlg.volumePath() == volumePath && !remoteVolumeUrl.isEmpty();


Bullo27:
Heads-up: this now conflicts with `main`, and I can't push the rebase myself — details and the resolution below so it isn't a guessing game.

**The conflict** is a context collision with #1179, not a semantic disagreement. That PR renamed `setOmpThreads` → `setNextOmpThreads` on the line immediately after the one this PR changes, so the two hunks touch adjacent lines and git can't auto-merge them. Resolved, the block reads:

```cpp
    _cmdRunner->setRenderVoxelSize(
        renderVolume ? renderVolume->voxelSize() : 0.0,
        renderVolume &&
            (renderVolume->baseScaleLevel() > 0 ||
             renderVolume->hasExplicitVoxelSizeOverride() ||
             renderVolume->isRemote()));
    _cmdRunner->setNextOmpThreads(dlg.ompThreads());
```

i.e. keep both sides — #1179's rename and this PR's `|| renderVolume->isRemote()`. I've rebased it locally onto 77205b2 and rebuilt: `ninja VC3D` links clean, and the resulting diff against `main` is exactly the original change, nothing else.

**Why I can't push it:** rebasing pulls #1179 into my branch, and #1179 touches `.github/workflows/vc3d-ci.yml`. GitHub refuses a push carrying workflow-file changes from a token without `workflow` scope, so the rebased branch is stuck on my machine until my fork's `main` is synced. I'm sorting that out; if you'd rather just take the three-line resolution above in the web editor, that works too and saves a round-trip.

Worth noting the same rename means anything else in flight that touches this function will hit the identical conflict.

<sub>Claude Code (Opus 5); rebase and rebuild both actually run.</sub>

Bullo27:
**A correction to this PR's framing, and a second call site it doesn't cover.** Both come from driving VC3D on `cbeb45f` rather than reading it.

### 1. One sentence in the description above is wrong

I wrote that a native-resolution remote volume is *"the default when you open an Open Data sample"* and matches neither branch of the guard. **It isn't.** Opening a sample through the Open Data browser tags every volume with `vc-open-data-voxel-size-um:` from the catalog manifest; `vc::project::volumeMetadataFromEntryTags` (`VolumePkg.cpp:450`) turns that into `metadata["voxelsize"]`, and `Volume.cpp:1419` sets `hasExplicitVoxelSizeOverride_`. So for that flow the **existing** guard already passes and this PR changes nothing.

Measured through the agent bridge — every volume of `PHercParis4`, all streaming-only (`path` empty), all already correct:

```
20230205180739                                    voxelSize=7.91
20260310170716-45.532um-11.0m-74keV-masked.zarr   voxelSize=45.532
20260310173927-45.532um-11.0m-110keV-masked.zarr  voxelSize=45.532
20260323153942-2.400um-0.2m-137keV-masked.zarr    voxelSize=2.4
```

### 2. The case this PR is actually for

A remote volume opened *without* an explicit size — `project.create` or `volume.attach` with a plain URL, which the bridge exposes to agents as `vc3d_create_project` / `vc3d_attach_volume`. Same build, same URL, one variable changed:

| trial | `voxelSize()` |
|---|---|
| bare URL, no tags | **0** |
| identical + `vc-open-data-voxel-size-um:45.532` | **45.532** |

Trial A has `baseScaleLevel() == 0` and `hasExplicitVoxelSizeOverride() == false`, so it fails the guard. Once #1227 lands and that `0` becomes `45.532`, the `|| isRemote()` here is what carries the value to the renderer; without it the render still declares nanometres. The #1227 pairing I described stands — I just named the wrong entry point for it.

### 3. There is now a second copy of the same guard, which this PR does not touch

#1179 added the bridge's `render.tifxyz` path, repeating the condition verbatim at `SegmentationCommandHandler.cpp` ~L2492:

```cpp
    if (params.hasVoxelSize) {
        _cmdRunner->setRenderVoxelSize(params.voxelSizeUm, true);
    } else {
        _cmdRunner->setRenderVoxelSize(
            renderVolume ? renderVolume->voxelSize() : 0.0,
            renderVolume &&
                (renderVolume->baseScaleLevel() > 0 ||
                 renderVolume->hasExplicitVoxelSizeOverride()));
    }
```

`hasExplicitVoxelSizeOverride` occurs once in that file at my branch point (`70bdf67`) and twice on `main`, so this site postdates the branch. An agent calling `render.tifxyz` without an explicit `voxelSize` on a bare-URL remote volume lands on exactly the case this PR is about.

I'd rather flag all of this myself than have it turn up in review. Happy to extend the change to both sites and fold it in with the rebase — just say which you'd prefer.

<sub>Claude Code (Opus 5); the A/B is one script differing only in the tag list, run against the live bucket.</sub>

Bullo27:
Rebase pushed — the push blocker is resolved, so ignore that part of my earlier comment.

`9cbf826` is now on the branch: rebased onto `main`, `ninja VC3D` links clean, and the diff against `main` is exactly the original one-line change plus its comment — nothing from the rebase leaked in. GitHub now reports this as mergeable again.

The two substantive points above still stand and are worth a decision when you review:

1. the description's "the default when you open an Open Data sample" is wrong — that flow already carries an explicit voxel size, so this PR is for remote volumes opened *without* one (bare-URL `project.create` / `volume.attach`);
2. #1179 added a second copy of the same guard at `SegmentationCommandHandler.cpp` ~L2492 that this PR doesn't touch.

Say the word and I'll extend it to both call sites in one push.

PR #1230, opened 2026-07-25
test(core): check thinning traces land on the source geometry

Bullo27:
`test_thinning.cpp` covers the entrypoints and the pruning helper well, but every assertion about the traces themselves is structural — *"at least one trace"*, *"traces is empty"*, *"counters moved"*. Nothing checks that a traced point sits where the input geometry says it should.

That matters because those traces become the normal-grid polylines, which feed the **highest-weighted term in the tracer's loss** (`NORMAL: 10`, against `DIST: 1`). An accuracy regression there would quietly degrade every segmentation while every existing assertion still passed — and it would look like "segmentation is hard" rather than a bug.

## The check

Concentric annuli of known radius about a known centre — a stand-in for the wrap cross-sections thinning actually runs on. The bound follows from the geometry rather than from an observed number: **a centreline through a band of thickness `t` cannot be further than `t/2` from the true radius.**

Currently the worst point is **0.995 px against the 2.0 px bound**, across 3001 points in 604 traces.

Pure OpenCV, no volume, no network, ~instant.

## Two things I deliberately did not do

**No assertion on trace count.** Thinning emits many short fragments — 604 for 6 rings — which the grid pipeline assembles later. I measured that before writing the test rather than assuming `traces.size() == 6`; pinning it would be brittle and would encode a fragmentation detail that isn't a contract.

**No tighter tolerance.** 0.995 against 2.0 is a 2× margin and I could have tightened it, but the geometric bound is the one that's defensible and stable across OpenCV versions. I'd rather have a bound that means something than one tuned to today's number.

## Verification

Confirmed the case actually discriminates rather than passing vacuously — forcing the tolerance to 0.01 px:

```
[MESSAGE] worst residual = 0.995146 px over 3001 points
test_thinning.cpp:215: FAILED: CHECK(worst <= 0.01)
```

and with the geometric bound restored: `11 test case(s) passed`. Full suite **109/109** (Ubuntu 24.04, gcc, Ninja/Release).

## Context

This came out of validating `vc_gen_normalgrids` end-to-end against a synthetic phantom — concentric cylindrical sheets with analytically known geometry, generated, then the resulting `.grid` polylines parsed back and compared to ground truth. That came out clean: XY traces 100% within 1.0 px of a true sheet radius with per-radius bias under 0.16 px and no drift, and XZ/YZ correct to a max 0.86 px away from tangency (the larger residuals there were all slices where the cut plane grazes a cylinder, which is a limitation of my two-chord reference model, not of the tool).

So this PR isn't fixing anything — the geometry is sound today. It pins that so it stays sound. Happy to drop it if you'd rather not carry the extra case.

---

<sub>Investigation and test with Claude Code (Opus 5); every number above is from an actual run.</sub>

### DIFF
--- a/volume-cartographer/core/test/test_thinning.cpp
+++ b/volume-cartographer/core/test/test_thinning.cpp
@@ -8,6 +8,8 @@
 
 #include <opencv2/core.hpp>
 
+#include <cmath>
+#include <limits>
 #include <vector>
 
 namespace {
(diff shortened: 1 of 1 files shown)


bruniss:
if we're going to add a test for this i'd prefer it be on real predictions with GT

Bullo27:
Noted, and closing is fine by me.

For what it's worth, if that test ever becomes wanted there's an existing pattern to hang it on — `core/test/test_normal_grid_live.cpp` already fetches real open-data volumes during `ctest` and soft-skips when there's no network, so it wouldn't need new machinery. The part I genuinely can't judge from outside is what should count as ground truth for traced output; published segments aren't quite that. If you point me at the right reference data I'll build it, otherwise I'll leave it alone.

Taking the wider signal too: I'll put the effort into defects rather than speculative coverage.

PR #1232, opened 2026-07-25
test(core): cover ABF flattening on a developable surface

Bullo27:
`ABFFlattening.cpp` has no test file at all, so nothing currently pins the correctness of the parameterization that turns segments into readable images.

## The invariant

A cylinder is **developable** — zero Gaussian curvature — so it admits an exactly isometric flattening, and every edge-length ratio must share one global scale. That's a ground truth that follows from the geometry rather than from a recorded output, so the bound doesn't need revisiting when the solver changes.

Measured worst deviation is **1.89e-06 across 3120 edges**, with the scale coming out at exactly 1.0. The bound is set at `1e-3`, leaving roughly three orders of magnitude for solver and platform variation while still failing on any meaningful loss of isometry.

In-memory only — `QuadSurface`'s Mat constructor plus `abfFlatten`. No fixtures, no I/O, no network.

## Verification

Confirmed it discriminates rather than passing vacuously — forcing the tolerance to `1e-6`:

```
[MESSAGE] worst relative deviation = 1.89085e-06 over 3120 edges; scale=1
test_abf_flattening.cpp:99: FAILED: CHECK(worst < 1e-6)
```

and with `1e-3` restored it passes. Full suite **110/110** (Ubuntu 24.04, gcc, Ninja/Release).

I first wrote the bound as 2% before measuring anything. Once the real deviation turned out to be ~2e-6 that bound was worthless — a 10,000× margin catches nothing — so it's tightened here. Flagging it because a tolerance written before the measurement is how a test ends up unable to fail.

## What prompted this

I'd been validating the flattener against surfaces with analytically known geometry, and it holds up well in both regimes:

- **developable (cylinder)** — converges in 1 iteration, UV bounds span exactly the analytic 120 × 120, per-step distortion std ~1e-5, no cell above 1%;
- **non-developable (sphere patch)** — 2.0% area loss and ~2.7% per-step distortion, against a theoretical minimum area mismatch of `2(1−cos α)/α²` = 2.5% at α = 0.55. So it's essentially at the mathematical floor for that curvature.

No bug to report, in other words — this pins behaviour that's already correct. Only the developable case is in the test, since it's the one with an exact answer; the curved case has no closed-form optimum to assert against.

Happy to drop this if you'd rather not carry it.

---

<sub>Investigation and test with Claude Code (Opus 5); every number above is from an actual run.</sub>

### DIFF
--- a/volume-cartographer/core/test/CMakeLists.txt
+++ b/volume-cartographer/core/test/CMakeLists.txt
@@ -453,6 +453,9 @@ vc_add_test(NAME test_quadsurface_more)
 
 vc_add_test(NAME test_thinning)
 
+vc_add_test(NAME test_abf_flattening
+    LIBS doctest::doctest vc_core vc_flattening)
+
 vc_add_test(NAME test_umbilicus)
 
 vc_add_test(NAME test_colormaps)
--- a/volume-cartographer/core/test/test_abf_flattening.cpp
+++ b/volume-cartographer/core/test/test_abf_flattening.cpp
@@ -1,0 +1,100 @@
+// Coverage for core/src/ABFFlattening.cpp — abfFlatten on surfaces whose
+// correct parameterization is known analytically.
+//
+// A cylinder is developable (zero Gaussian curvature), so it admits an exactly
+// isometric flattening: every edge length is preserved, up to one global scale.
+// That gives a ground truth that follows from the geometry rather than from a
+// recorded output, so the bound below does not need updating when the solver
+// changes.
+
+#define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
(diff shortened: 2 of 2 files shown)


bruniss:
i dont want to add this as the abf flattening module will be deprecated/removed in the near future due to better/faster alternatives. thank you for the pr regardless

Bullo27:
Understood — closing it. No sense pinning behaviour on a module that's on its way out, and knowing that is worth more to me than the test was.

One question if it's cheap to answer: what's replacing it? I ask only so I aim future work at code that's staying — if the replacement is already in-tree I'll leave the flattening path alone until it lands.

Thanks for taking the time to look at it.

Issue #1231, opened 2026-07-25
Tutorial 5 trains with no validation set — published segments ship no `_validation_mask`


khj1222:
Following [tutorial 5](https://scrollprize.org/tutorial5) end to end on `w00_20231016151002`
(native Windows, RTX 5090, `merge-ink-pipelines`), I ended up with a model I had no way to
evaluate, and I'd like to check whether that's intended.

### 1. Is the missing `_validation_mask` deliberate?

The pipeline clearly supports one:

- `data/patch_finding/default.py` marks every patch touching `_validation_mask` as a
  validation patch, and `_exclude_validation_voxels_from_training_supervision` in
  `data/ink_dataset.py` removes those voxels from the training supervision;
- `training/train.py` runs a validation pass every `val_every` steps and logs
  Confusion / BalancedAccuracy;
- `preprocessing/create_label_zarrs.py` already converts `_validation_mask.{tif,tiff,png}`;
- `evaluation/metrics/` implements DRD and a weighted pseudo-F-measure.

But the published segment carries only `_inklabels` and `_supervision_mask`, and the tutorial
never mentions validation. A stock tutorial run therefore gives:

```
flat_ink_patches_...json   "is_validation": true  ->     0
                           "is_validation": false -> 2,710
val_previews/              (empty)
```

`val_every` iterates an empty loader, and DRD / pseudo-F-measure never execute — so there is
no way to tell whether a change to the config, the model or the augmentations helped.

Are the validation masks kept internally and simply not published, or is generating one left
to the user? If the latter, is there a convention for choosing the held-out region that I
should match?

I ask because the choice turns out to matter more than I expected. On this segment the
supervision mask is **15 disconnected regions** around annotated letters (1.5%–20.7% of the
supervised area each, ink density 0.114–0.440), and 4 of them sit closer to another region
than one 256 px patch. A naive rectangular split cuts individual letters in half, so the
model trains on one stroke and is scored on the one beside it.

Holding out whole regions instead (20% of supervised area, ink density matched to 0.2283 vs
the segment's 0.2283) gives 1,337 validation / 2,240 training patches, and the numbers that
come out are:

| | best F1 | IoU |
|---|---|---|
| clean 20k run, scored on held-out regions | 0.8232 | 0.6995 |
| the pre-existing checkpoint that had trained on those regions | 0.8594 | 0.7535 |

Per-region F1 on the clean run spans 0.796–0.895, and a 3-fold rerun puts another fold at
[...]

PR #1234, opened 2026-07-25
create_label_zarrs: stream untiled label images instead of building the pyramid in memory

khj1222:
Converting a label image only streams when the input TIFF is **tiled**.
`_get_tiled_tiff_metadata` returns `None` for a striped TIFF (or a PNG), and
`convert_image` falls back to `build_pyramid_with_mode`, which materializes the whole
`DEFAULT_DEPTH`-deep volume once per pyramid level. For a full-resolution segment label
that is tens of GiB — even though 64 of the 65 slices are zeros:

```
ERROR ..._validation_mask.tif: Unable to allocate 25.1 GiB for an array
      with shape (65, 16125, 25690) and data type uint8
```

The shipped `_inklabels` / `_supervision_mask` TIFFs are tiled 256×256, so published assets
never hit this. Any label image written elsewhere does — in my case a validation mask
produced with a default `tifffile.imwrite`, which is striped.

### Change

Route untiled inputs through the same zarr-backed path the tiled branch already uses:
create the datasets, write the 2-D image into level 0 at `DEFAULT_LABEL_SLICE` in blocks,
then build the downsampled levels from zarr with the existing
`_build_downsample_levels_from_zarr`. Peak memory becomes one 2-D image instead of
`DEFAULT_DEPTH` times it. No new dependencies, no change to the tiled path, and
`build_pyramid` / `write_ome_zarr` are left in place for other callers.

### Verification

Same-output check on a synthetic 1500×2300 image written both ways:

- all six pyramid levels byte-identical between the tiled and untiled paths
- the only non-empty z slice is 32, and it equals the source image
- group attrs identical

Real-size check on a 32249×51380 uint8 mask: the striped input that previously raised
`Unable to allocate 25.1 GiB` now converts in 83 s, and levels 0/3/5 match the output of
the tiled path exactly.

Second half of #1231.


erdpx:
hi! thanks for the pr!

this is a good catch! would you consider building only the 2d pyramid in memory, then writing each level directly to DEFAULT_LABEL_SLICE, instead of rereading the zarr levels from disk? this should be considerably faster while still avoidng materializing the N-slice volumes; the 2d pyramid itself is only ~1.33× the source image size

khj1222:
Done — thanks, that is faster. Pushed as a follow-up commit.

Each 2D level is now derived from the one before it in memory and written
straight to `DEFAULT_LABEL_SLICE`; nothing reads a zarr level back. On the
32249×51380 striped label that sent me down this path in the first place:

| | time | peak RSS |
|---|---|---|
| before (stream levels from zarr) | 114.5 s | 1.61 GiB |
| after (2D pyramid in memory) | **66.5 s** | 1.99 GiB |

against the 25 GiB the original in-memory route tried to allocate.

One wrinkle worth flagging: for `mean` mode I average in row bands rather than
in one pass, because `_downsample_mean`'s accumulator is `float64` — eight times
the output, which is a few GiB at level 1 of a full-size label and would have
undercut the point of the change. Bands start on even source rows, so every
output pixel still comes from the same four inputs through the same code.

Verified byte-identical against both the tiled streaming path and the original
`build_pyramid_with_mode` output, over 6 levels, for binary and grayscale inputs
at even and odd dimensions, plus all 6 levels of the real striped label above.

PR #1238, opened 2026-07-25
Fix inference output stores under zarr 3 (the pinned version)

TAUIL-Abd-Elilah:
## What

With the zarr version this repo pins (`uv.lock`: zarr 3.2.1), `vesuvius.predict` cannot write its outputs. There are two independent breakages, and the second one also affects `blend_logits`.

**1. `zarr.Blosc` was removed in zarr 3**

```
Creating output stores...
AttributeError: module 'zarr' has no attribute 'Blosc'
```

`_get_zarr_compressor` is the only place in the package still using it — `blending.py`, `finalize_outputs.py`, `structure_tensor/`, and `scripts/` all build `numcodecs.Blosc` already, and `inference.py` already imports `numcodecs`. Every run dies here unless you pass `--zarr-compressor none`.

**2. `open_zarr` creates v3 arrays, which reject the v2 keyword**

```
ValueError: compressor cannot be used for arrays with zarr_format 3
```

So even with a numcodecs compressor the create call fails. This is reachable from `blend_logits` too, which builds a `numcodecs.Blosc` and passes it to the same `open_zarr`. And `BlendingProcessor._validate_zarr_format` raises unless `zarr_format == 2`, so v3 logits would be rejected downstream even if they could be written.

## Fix

- `inference.py`: use `numcodecs.Blosc`, matching every other writer in the package.
- `open_zarr`: add a `zarr_format` parameter defaulting to `2`, which is what the rest of the pipeline reads and writes. Pass `zarr_format=None` to accept zarr's own default.

## Verification

Against the public Open Data bucket (PHerc. Paris 4, `20260310173927-45.532um-11.0m-110keV-masked.zarr`):

- before: `AttributeError` at "Creating output stores"
- after: `logits_part_0.zarr` and `coordinates_part_0.zarr` are created with the default zstd compressor, stored as `zarr_format` 2 as `blend_logits` requires

New tests in `tests/data/test_open_zarr_format.py` cover the default format, a numcodecs compressor round-trip, and that an explicit `zarr_format=3` still works. Full `tests/data` suite passes (34 tests).

Found while running the documented inference pipeline end to end on a machine with no GPU. Written with Claude Code as a coding assistant; every claim above was checked by running it.

### DIFF
--- a/vesuvius/src/vesuvius/data/utils.py
+++ b/vesuvius/src/vesuvius/data/utils.py
@@ -145,7 +145,16 @@ def open_zarr(path: str, mode: str = 'r',
 
         # Add any other kwargs
         create_kwargs.update(kwargs)
-        
+
+        # zarr 3 derives `overwrite` from the mode, so passing it alongside
+        # mode='w' raises "got multiple values for keyword argument 'overwrite'".
+        # mode='w' already means truncate, so an explicit overwrite=True is
+        # redundant; drop it rather than making every caller know this.
+        if create_kwargs.pop('overwrite', None) is False:
+            raise ValueError(
+                "open_zarr(mode='w') always truncates; overwrite=False is contradictory"
+            )
+
         if verbose:
             print(f"Creating new zarr array with shape={shape}, chunks={chunks}, dtype={dtype}")
         
--- a/vesuvius/src/vesuvius/data/utils.py
+++ b/vesuvius/src/vesuvius/data/utils.py
@@ -4,6 +4,8 @@ import zarr
 import os
 from typing import Union, Dict, Any, Optional, Tuple
 
+_ZARR_V3 = int(zarr.__version__.split('.', 1)[0]) >= 3
+
 # Function to get the maximum value of a dtype
 def get_max_value(dtype: np.dtype) -> Union[float, int]:
     """
@@ -64,7 +66,8 @@ def open_zarr(path: str, mode: str = 'r',
         Zarr format version for arrays created with mode 'w'. Defaults to 2 because
         the numcodecs compressors used throughout this package (and the logits stores
         blend_logits validates) are v2 constructs: zarr 3 raises for `compressor=` on
-        a v3 array. Pass None to accept zarr's own default.
+        a v3 array. Pass None to accept zarr's own default. Ignored when the
+        installed zarr is 2.x, which only writes v2 and does not accept the argument.
     **kwargs : Additional parameters passed to zarr.open
         
     Returns:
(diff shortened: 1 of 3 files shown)


TAUIL-Abd-Elilah:
Pushed a second commit: `finalize_outputs` hits the same zarr 3 migration from a different angle.

```
Creating output store: final_surface.zarr
--- Finalization Failed ---
Error: zarr.api.asynchronous.create() got multiple values for keyword argument 'overwrite'
```

It passes `overwrite=True` alongside `mode='w'`, and zarr 3 derives `overwrite` from the mode. Since `mode='w'` already truncates, `open_zarr` now drops the redundant flag (and rejects a contradictory `overwrite=False`).

With both commits, the documented three-stage pipeline runs to completion on the pinned dependencies. Against the public Open Data bucket (PHerc. Paris 4), CPU only:

```
vesuvius.predict ... --bbox "2000:2200,1000:1200,1000:1200"
  -> Finished writing 8 patches.  logits_part_0.zarr: zarr_format 2,
     compressor {'id': 'blosc', 'cname': 'zstd', 'clevel': 3}

vesuvius.blend_logits outputs/roi_final outputs/merged_logits.zarr
  -> Merging Finished (216 patch reads, 4 workers)

vesuvius.finalize_outputs outputs/merged_logits.zarr outputs/final_surface.zarr --mode binary --threshold
  -> Output processing complete. Processed 27 chunks, skipped 10341 empty chunks (99.74%).
     Final output saved to: outputs/final_surface.zarr
```

Before these changes stage 1 stopped at `AttributeError: module 'zarr' has no attribute 'Blosc'`, and stage 3 stopped at the `overwrite` error above.

(The `--device cpu` run additionally needs #1239; and on a machine without a C++ compiler, `TORCHDYNAMO_DISABLE=1`.)

pmh47:
Thanks for this fix. The test `test_explicit_zarr_format_3_still_possible` fails in our CI when installed zarr package version is overridden 2.18.7, which is a set-up we still support (the test passes when it is the pinned 3.2.1). Please resolve and we'll merge

TAUIL-Abd-Elilah:
Thanks — reproduced it locally by pinning zarr 2.18.7 the same way CI does, and there were two problems, not one.

`zarr_format` does not exist in zarr 2. It is silently discarded with

```
UserWarning: ignoring keyword argument 'zarr_format'
```

and a v2 array is written regardless. So the test asserting a v3 store came back got a v2 one and failed — and, less visibly, every store `predict` creates was emitting that warning on the zarr 2 leg.

Fixed by only forwarding the argument to zarr 3, using the `_ZARR_V3` check already used in `neural_tracing/datasets/common.py`. zarr 2 writes v2, which is what the default asks for anyway, so behaviour is unchanged on both legs.

`test_explicit_zarr_format_3_still_possible` is now skipped under zarr 2 — that version cannot write v3 through this API at all, so skipping seemed more honest than weakening the assertion. Added `test_zarr_format_default_is_silent` to cover the warning directly, since that was the part that would otherwise have gone unnoticed.

Both legs, `pytest tests/data`:

```
zarr 2.18.7  ->  23 passed, 1 skipped
zarr 3.2.1   ->  24 passed
```

TAUIL-Abd-Elilah:
The zarr 2.18.7 failure is fixed and pushed. Two things were wrong, not one: `zarr_format`
doesn't exist in zarr 2 and is silently dropped with a `UserWarning`, leaving a v2 array,
and the test itself assumed the v3 path. Both zarr legs pass locally now.

The only red check is the Vercel preview, which needs team authorization and fails on
every fork PR — the test jobs are green.

Anything else outstanding on this one?

PR #1239, opened 2026-07-25
inference: make --device cpu work: autocast is hardcoded to CUDA

TAUIL-Abd-Elilah:
## What

`vesuvius.predict` documents and accepts `--device cpu`, but a CPU run dies on the first forward pass:

```
RuntimeError: Input type (struct c10::Half) and bias type (float) should be the same
```

`VCDataset` yields `float16` patches, which suits the CUDA autocast path. The inference block wraps every forward in `torch.amp.autocast('cuda')` with the device type hardcoded, so on CPU autocast does nothing and the half input reaches float32 weights. CPU convolutions have no float16 kernels, so this cannot work regardless of autocast.

## Fix

- Take the autocast device type from the run's device, and enable autocast only for CUDA.
- Promote inputs to the model's own parameter dtype when off the CUDA path.

CUDA behaviour is unchanged. Stored output stays `float16` because `_finalize_output_batch` already casts before writing.

## Verification

Reproduced and fixed against the public Open Data bucket (PHerc. Paris 4) on CPU: before, the first patch raised the error above; after, the dtype mismatch is gone.

Note on scope: this is the dtype/autocast half of CPU support. On a machine without a C++ compiler the run can still stop later inside `torch.compile` (`InvalidCxxCompiler: Compiler: cl is not found`), which is a separate concern and not addressed here — `TORCHDYNAMO_DISABLE=1` sidesteps it.

Found while running the documented inference pipeline on a machine with no GPU available. Written with Claude Code as a coding assistant; the failure and the fix were both checked by running them.

### DIFF
--- a/vesuvius/src/vesuvius/models/run/inference.py
+++ b/vesuvius/src/vesuvius/models/run/inference.py
@@ -737,6 +737,14 @@ class Inferer():
             anon=self.input_anon,
             bbox=self.bbox,
             read_retries=self.read_retries,
+            # The float16 default suits the CUDA autocast path. CPU convolutions
+            # have no float16 kernels, so half patches meet float32 weights and
+            # raise "Input type (c10::Half) and bias type (float) should be the
+            # same". Ask for the dtype the CPU model can actually consume.
+            return_as_type=(
+                "np.float32" if self.device.type == "cpu"
+                else "np.float16"
+            ),
         )
 
         expected_attr_name = 'all_positions'
@@ -1007,8 +1015,9 @@ class Inferer():
                     # Only perform inference if there are non-empty patches
                     if non_empty_indices:
                         non_empty_input = input_batch[non_empty_indices]
-                        
-                        with torch.inference_mode(), torch.amp.autocast('cuda'):
+
+                        with torch.inference_mode(), torch.amp.autocast(
+                                self.device.type, enabled=(self.device.type == 'cuda')):
                             if self.do_tta:
                                 non_empty_output = infer_with_tta(
                                     self.model,
--- a/vesuvius/src/vesuvius/utils/models/load_nnunet_model.py
+++ b/vesuvius/src/vesuvius/utils/models/load_nnunet_model.py
@@ -275,8 +275,16 @@ def load_model(model_folder: str, fold: Union[int, str] = 0, checkpoint_name: st
     
     network.eval()
     
-    # Compile by default unless explicitly disabled
-    should_compile = os.environ.get('nnUNet_compile', 'true').lower() in ('true', '1', 't')
+    # Compile by default on CUDA unless explicitly disabled. torch.compile is
+    # lazy, so an unusable inductor backend does not raise here - it raises at
+    # the first forward pass, past the except below, killing the run partway
+    # through inference. On CPU that is the common case (inductor needs a C++
+    # toolchain) for little gain, so CPU opts in rather than out.
+    compile_setting = os.environ.get('nnUNet_compile')
+    if compile_setting is None:
+        should_compile = device.type == 'cuda'
+    else:
+        should_compile = compile_setting.lower() in ('true', '1', 't')
     if should_compile and not isinstance(network, OptimizedModule):
         if should_print:
             print('Using torch.compile for potential performance improvement')


TAUIL-Abd-Elilah:
Confirmed end to end: with this change a CPU run completes all 8 patches of a region-of-interest inference against the public Open Data bucket (PHerc. Paris 4), ~28 s/patch on a laptop-class CPU:

```
Inferring Part 0: 100%|##########| 8/8 [03:46<00:00, 28.37s/it]
Finished writing 8 patches.
Inference complete.
```

Before the change the run raised `Input type (struct c10::Half) and bias type (float) should be the same` on the first patch.

(Reproducing on the pinned dependencies also needs #1238, since predict cannot create its output stores under zarr 3 at all. On a machine without a C++ compiler, `TORCHDYNAMO_DISABLE=1` is needed to get past `torch.compile` — separate concern, not addressed here.)

bruniss:
looks good but i'd request a couple changes:  
- instead of casting float16 patches from `VCDataset` to float32 just ask for them as float32 if you're using cpu, perhaps before `_create_dataset_and_loader()` you could do this : 
```python
return_as_type=(
    "np.float32" if self.device.type == "cpu"
    else "np.float16"
),
```

also , the nnunet loader compiles by default, so this will not make cpu inference work as-written

TAUIL-Abd-Elilah:
Both taken, thank you — and you were right that the original patch was not enough on its own.

**`return_as_type` instead of casting.** Done as suggested, at the `VCDataset` construction in `_create_dataset_and_loader()`:

```python
return_as_type=(
    "np.float32" if self.device.type == "cpu"
    else "np.float16"
),
```

The post-hoc cast is gone. This is also strictly better than what I had: the old path let the dataset round to float16 and then promoted back to float32, so CPU runs were paying a lossy round-trip for no reason. Now the precision is kept end to end.

**torch.compile.** Confirmed — CPU inference still did not run. Worth noting *why*, because it is not visible at the call site: `torch.compile` is lazy, so an unusable inductor backend does not raise where it is called. The `try/except` around it catches nothing and the failure lands on the first forward pass instead, mid-inference, after the model and dataset are already built.

So CPU now opts in rather than out:

```python
compile_setting = os.environ.get('nnUNet_compile')
if compile_setting is None:
    should_compile = device.type == 'cuda'
else:
    should_compile = compile_setting.lower() in ('true', '1', 't')
```

CUDA behaviour is unchanged, and an explicit `nnUNet_compile` still wins in both directions, so anyone with a CPU toolchain who wants it can set `nnUNet_compile=1`.

**Verified end to end** against the public Open Data bucket (PHerc. Paris 4), CPU only, with no `TORCHDYNAMO_DISABLE` and no `nnUNet_compile` set — i.e. exactly the out-of-the-box path that used to fail:

```
Inferring Part 0: 100%|##########| 8/8 [06:04<00:00, 45.51s/it]

--- Inference Finished ---
Output shape: (8, 2, 192, 192, 192)
Output dtype: float16
```

No `Using torch.compile for potential performance improvement` line, where the previous run printed it and then needed `TORCHDYNAMO_DISABLE=1` to get past the first patch.

Rebased onto current `main` (so it sits on top of #1241). Still needs #1238 for the output stores to be creatable at all under the pinned zarr.

jrudolph (commented):
How useful is supporting CPU here? Otherwise, lgtm

TAUIL-Abd-Elilah:
Thanks both — I think both of those are already in the diff, so let me point at the lines in case the review was against an earlier state and I'm the one misreading.

**@bruniss on the dtype:** that's exactly the approach the PR takes, in `_create_dataset_and_loader` (`inference.py` L744):

```python
return_as_type=(
    "np.float32" if self.device.type == "cpu"
    else "np.float16"
),
```

No casting of float16 patches — VCDataset is asked for float32 up front when the device is CPU.

**@bruniss on compilation:** that's the second half of the PR, in `load_nnunet_model.py` L283-287. `nnUNet_compile` unset now means "compile on CUDA, not on CPU", instead of "compile always":

```python
compile_setting = os.environ.get('nnUNet_compile')
if compile_setting is None:
    should_compile = device.type == 'cuda'
```

The reason it needs to be a default rather than a documented env var: `torch.compile` is lazy, so an unusable inductor backend doesn't raise at compile time — it raises at the first forward pass, past the `except` below it, which kills the run partway through inference rather than falling back. On CPU that's the common case, since inductor wants a C++ toolchain.

**@jrudolph on how useful CPU is:** honestly, not for production inference — it's slow enough that nobody should segment a scroll that way. Two things make it worth having:

- The docs advertise `--device cpu`, so it should either work or the flag should go. Right now it fails with `Input type (c10::Half) and bias type (float) should be the same`, which reads like a model problem rather than an unsupported path.
- It's the only way to check the pipeline without a CUDA card. I used it while auditing whether the published surface predictions regenerate from public inputs — being able to run the three stages on CPU meant I could verify scoring logic without competing for the GPU that was busy with inference.

Happy to drop it if you'd rather not carry the path — in that case removing `--device cpu` from the CLI would be the clearer fix, and I'll send that instead.

TAUIL-Abd-Elilah:
Quick follow-up: the branch is still mergeable and all substantive checks are green on both tested zarr versions; the only red status is the fork Vercel authorization. The requested float32-at-source and CPU compile-default changes are in the current diff and the out-of-the-box CPU path was verified end to end. Is anything else needed from me?

jrudolph (approved):
lgtm, good point about being able to test in more envs without cuda.

PR #1240, opened 2026-07-25
Volume.meta: don't call len() on a single zarr Array

TAUIL-Abd-Elilah:
## What

`Volume` accepts a `zarr.Array`, a `zarr.Group`, or a sequence of arrays, and `meta()` already branches on all three when printing per-level shapes. The resolution-level count just above that branch calls `len(self.data)` unconditionally, which raises for a `zarr.Array`:

```
TypeError: object of type 'Array' has no len()
```

`meta()` runs whenever a `Volume` is constructed with `verbose=True`, so any verbose run against a store that resolves to a single array dies during initialization:

```
vesuvius.predict --input_dir <volume>.zarr/0 ... --verbose
-> Error initializing Volume: object of type 'Array' has no len()
```

Pointing at an explicit resolution level like `.zarr/0` is the normal way to read a multiscale volume over HTTP, where the group listing comes back empty.

## Fix

Move the count inside the existing branches: a single array is one level, a group counts its array members, a sequence keeps `len()`. Group counting now also ignores non-array members instead of counting them as levels.

Tests in `tests/data/test_volume_meta.py` cover all three layouts.

Found while running the documented inference pipeline against the public Open Data bucket. Written with Claude Code as a coding assistant; verified by running it.

### DIFF
--- a/vesuvius/src/vesuvius/data/volume.py
+++ b/vesuvius/src/vesuvius/data/volume.py
@@ -416,15 +416,19 @@ class Volume:
             print(f"  Global Mean: {self.global_mean}, Global Std: {self.global_std}")
         print(f"Return Type: {self.return_as_type}")
         print(f"Return as Tensor: {self.return_as_tensor}")
-        print(f"Number of Resolution Levels: {len(self.data)}")
         if isinstance(self.data, zarr.Array):
+            # A single array is one resolution level; zarr.Array has no len().
+            print(f"Number of Resolution Levels: 1")
             print(f"  Level 0 Shape: {self.data.shape}, Dtype: {self.data.dtype}")
         elif isinstance(self.data, zarr.Group):
-            for key in sorted(self.data.keys(), key=lambda x: int(x) if x.isdigit() else x):
+            levels = [key for key in sorted(self.data.keys(), key=lambda x: int(x) if x.isdigit() else x)
+                      if isinstance(self.data[key], zarr.Array)]
+            print(f"Number of Resolution Levels: {len(levels)}")
+            for key in levels:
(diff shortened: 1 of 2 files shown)


pmh47:
The test `test_meta_reports_multiscale_group` fails with zarr version 2.18.7, which is a config we run in CI and require support for. Please resolve and we'll merge.

TAUIL-Abd-Elilah:
Thanks — my fixture was the problem, not the fix.

`test_meta_reports_multiscale_group` built its levels with `Group.create_array`, which zarr 3 introduced. On the zarr 2.18.7 leg that raises `AttributeError` before `Volume.meta()` is ever called:

```
AttributeError: . Did you mean: 'create_group'?
zarr/hierarchy.py:540: in __getattr__
```

The fix under test is version-independent, so only the fixture needed to be. Added a small `_add_level()` helper that uses `create_array` when present and falls back to zarr 2's `create_dataset`, so both legs now exercise the same `zarr.Group` branch of `meta()`.

Verified by installing each version the way CI does:

```
zarr 2.18.7  ->  21 passed
zarr 3.2.1   ->  21 passed
```

TAUIL-Abd-Elilah:
Fixed and pushed — the fixture was the problem, not the fix. `test_meta_reports_multiscale_group`
built its levels with `Group.create_array`, which only exists in zarr 3, so on the 2.18.7
leg it raised `AttributeError` before `Volume.meta()` was ever reached. Rewritten to
build levels in a way both versions support; passes on both.

Same note as #1238: the red check is the Vercel preview, not a test job.

Let me know if anything else is needed.

PR #1241, opened 2026-07-25
predict: add --bbox to restrict inference to a region of interest

TAUIL-Abd-Elilah:
## What

`vesuvius.predict` tiles the entire volume. The only way to process less is `--num_parts`/`--part_id`, which splits blindly along Z — there is no way to say "run the model *here*". When the input is a remote zarr, a small experiment on one region streams chunks for the whole scroll.

This adds `--bbox "z0:z1,y0:y1,x0:x1"`, in **global voxel coordinates**, half-open, with either bound omittable:

```bash
vesuvius.predict --model_path hf://scrollprize/surface_recto \
    --input_dir https://vesuvius-challenge-open-data.s3.amazonaws.com/PHercParis4/volumes/<volume>.zarr/0 \
    --output_dir /tmp/logits \
    --bbox "2000:2200,1000:1200,1000:1200"

# open bounds: from z=1000 to the end, all of y, up to x=2000
vesuvius.predict ... --bbox "1000:,:,:2000"
```

`VCDataset` builds the sliding window over the ROI and offsets it back into the global frame, so emitted patch coordinates are volume-absolute and `blend_logits`/`finalize_outputs` work unchanged — an ROI run lands exactly where a full-volume run would have put it. The requested ROI is recorded in the logits store's `bbox` attribute.

## Why

Measured on `PHercParis4` level 0 (4066×2264×2264 uint8, 128³ chunks, 192³ patches at 0.5 overlap):

| Run | Chunks touched | Data |
|---|---|---|
| Whole volume | 10,368 | 21.7 GB |
| One 200³ ROI | 27 | 56.6 MB |

**384× less data (99.74%).** Those 56.6 MB fetched in 45 s on a home connection. That is the difference between iterating on hardware you already own and provisioning a box first. It is also the practical way to inspect one column of text, or reproduce a result on a laptop.

## Details

- `--num_parts`/`--part_id` split the ROI's Z extent rather than the volume's, so distributed runs stay balanced when a bbox is active.
- An ROI smaller than the patch size on an axis is grown to the patch size and shifted back inside the volume, so the model always sees a full patch.
- Bounds are clamped to the volume; a bbox entirely outside it raises.
- Coverage diagnostics report against the tiled extent instead of the whole volume.
- Documented in `vesuvius/docs/inference.md` under "Region-of-interest inference".

## Tests

`tests/data/test_vc_dataset_bbox.py` — 10 tests covering global-frame coordinates, ROI coverage, open bounds, sub-patch growth, rejected inputs, and that a 3-part split reproduces the single-run position set exactly. Full `tests/data` suite passes.

Verified against the public Open Data bucket: with `--bbox 2000:2200,1000:1200,1000:1200` the run builds exactly 8 patches and reads `slice(2000, 2192), slice(1000, 1192), slice(1000, 1192)` — the requested region, in global coordinates.

Written with Claude Code as a coding assistant; every number above was measured, not estimated.

### DIFF
--- a/vesuvius/docs/inference.md
+++ b/vesuvius/docs/inference.md
@@ -32,6 +32,7 @@ vesuvius.predict \
 | `--input_format` | Force `zarr`, `tiff`, or `volume` detection. Usually optional.
 | `--tta_type` / `--disable_tta` | Choose `rotation` (default) or `mirroring`, or disable test-time augmentation.
 | `--num_parts` / `--part_id` | Partition inference so multiple machines can process different chunks.
+| `--bbox` | Restrict inference to a region of interest: `"z0:z1,y0:y1,x0:x1"` in global voxel coordinates, half-open. Omit a bound to reach the volume edge (`"1000:1400,:,2000:"`). See [Region-of-interest inference](#region-of-interest-inference).
 | `--overlap` | Fractional patch overlap (0–1, default `0.5`).
 | `--batch_size` | Inference batch size (default `1`).
 | `--patch_size` | Override the model patch size using a comma-separated list (e.g. `192,192,192`).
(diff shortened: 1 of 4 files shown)


TAUIL-Abd-Elilah:
End-to-end confirmation of the alignment claim, on the public Open Data bucket (PHerc. Paris 4, level 0), CPU only.

`vesuvius.predict ... --bbox "2000:2200,1000:1200,1000:1200"` produced 8 patches, and the coordinates store holds **global** positions, not ROI-local ones:

```
logits shape (8, 2, 192, 192, 192) float16
bbox attr: [2000, 2200, 1000, 1200, 1000, 1200]
coords:
 [[2000 1000 1000] [2000 1000 1008] [2000 1008 1000] [2000 1008 1008]
  [2008 1000 1000] [2008 1000 1008] [2008 1008 1000] [2008 1008 1008]]
```

`blend_logits` then consumed those stores unmodified, and `finalize_outputs` wrote a full-volume-shaped result `(1, 4066, 2264, 2264)` with the ROI in its correct absolute position:

```
Output processing complete. Processed 27 chunks, skipped 10341 empty chunks (99.74%).
```

That 27 / 10,368 split is the same ratio as the data-movement measurement in the description (27 chunks read instead of 10,368) — arrived at independently by finalization walking the output grid, which is a nice cross-check that the region landed exactly where a full-volume run would have put it.

Note for reviewers: reproducing this on the pinned dependencies also needs #1238 (predict cannot create output stores under zarr 3 at all), and CPU runs need #1239. Those are separate PRs; this one is only the feature.

TAUIL-Abd-Elilah:
For anyone wanting to reproduce the data-movement numbers without checking out this branch, I put the benchmark, the measured figures and the working install recipe in a small standalone repo: https://github.com/TAUIL-Abd-Elilah/vesuvius-roi-inference

`python bench_roi_savings.py --measure-live` needs no credentials and takes about a minute — it reads the real chunk grid of the public PHerc. Paris 4 volume and then times fetching the region's chunks.

PR #1242, opened 2026-07-25
Guard the POSIX-only fcntl import in predict

TAUIL-Abd-Elilah:
## What

`vesuvius/models/run/inference.py` imports `fcntl` at module scope. `fcntl` is POSIX-only, so on Windows the `vesuvius.predict` entry point fails at import time, before any argument is parsed:

```
ModuleNotFoundError: No module named 'fcntl'
```

I know AGENTS.md §1.5 lists Ubuntu and macOS as the targets, so this is not a supported-platform bug — feel free to close it if Windows is explicitly out of scope. Offering it because the guard is small and the same section asks that OS-specific code be guarded with a safe fallback.

## Fix

`fcntl` is used in exactly one place: an advisory lock that serializes concurrent `aws s3 sync` downloads into the model cache. Import it lazily, and add `_lock_file_exclusive()` which uses `fcntl.flock` where available and `msvcrt.locking` otherwise. Lock semantics are unchanged on POSIX, and the lock is still released when the handle closes at the end of the `with` block.

## Verification

`from vesuvius.models.run.inference import main` succeeds on Windows 11 / Python 3.14, and `vesuvius.predict` then runs against the public Open Data bucket. No behaviour change on POSIX: the `fcntl` branch is byte-for-byte the previous call.

Written with Claude Code as a coding assistant.

### DIFF
--- a/vesuvius/src/vesuvius/models/run/inference.py
+++ b/vesuvius/src/vesuvius/models/run/inference.py
@@ -2,7 +2,12 @@ import torch
 import numpy as np
 import zarr
 import os
-import fcntl
+import errno
+try:
+    import fcntl  # POSIX-only; used to serialize model-cache downloads
+except ImportError:  # pragma: no cover - Windows has no fcntl
+    fcntl = None
+    import msvcrt
 import hashlib
 import multiprocessing
 import subprocess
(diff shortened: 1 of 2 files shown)


bruniss:
while unlikely this loop could theoretically go forever on a non lock related error. could you scope the exception to the actual error so it raises if it's not lock related , for ex: 

```
except OSError as exc:
    if exc.errno not in (errno.EACCES, errno.EDEADLK):
        raise
```

thank you!

TAUIL-Abd-Elilah:
Good catch — you are right, and it is not only theoretical: `msvcrt.locking` reports a full disk or a bad descriptor through the same `OSError`, and the loop would have spun on those forever.

Scoped it exactly as you suggested:

```python
except OSError as exc:
    if exc.errno not in (errno.EACCES, errno.EDEADLK):
        raise
```

Those two are the ones `_locking` documents for contention — `EDEADLK` (36) after `LK_LOCK` gives up its ~10 retries, `EACCES` on a lock violation. Everything else now fails fast.

Also added `tests/models/run/test_lock_file_exclusive.py`. It swaps the module globals, so the Windows branch is covered on your Linux CI too: retry until the lock clears, propagate `EBADF`/`ENOSPC`/`EINVAL` instead of looping, and `flock` still used on POSIX. 5 tests.

Rebased onto current `main` while I was in there. Still happy for you to close this if Windows is out of scope per AGENTS.md §1.5 — no argument from me either way.

bruniss:
thank you for updating. merging

PR #1244, opened 2026-07-25
Retry transient remote reads instead of aborting the whole run

TAUIL-Abd-Elilah:
## What

Streaming inference issues one read per patch, and a single dropped connection propagates out of the DataLoader worker and kills the entire run — discarding every patch already computed. This hit me 6 patches into an 8-patch job against the public Open Data bucket:

```
RuntimeError: Failed to get item 6 (pos 8064,3264,3200): Response payload is not completed:
<ContentLengthError: 400, message='Not enough data to satisfy content length header
(received 1202432 of 2097152 bytes).'>. SSLError(1, '[SSL: RECORD_LAYER_FAILURE] record layer failure')
```

Nothing was wrong with the data or the request — the connection just dropped mid-chunk. On a full scroll over S3, which is hours of reads, hitting at least one of these is close to certain, and today that means starting over.

## Fix

Retry transient read failures with exponential backoff (4 attempts, 0.5 s doubling to 8 s), exposed as `--read-retries` (set `1` to disable).

Transience is judged from the message across the whole `__cause__`/`__context__` chain, because zarr/fsspec wrap aiohttp, botocore, urllib3 and ssl errors differently depending on version and backend — matching on exception classes alone misses most real cases. Deterministic failures (bad coordinates, missing array, auth) are **not** retried and still fail on the first attempt, so a coding error doesn't turn into a slow four-attempt failure.

## Tests

`tests/data/test_vc_dataset_retry.py` — 14 tests: retry-then-succeed, exhaustion, fail-fast on deterministic errors, `--read-retries 1` disabling retries, message classification for the observed failure modes (truncated payload, SSL record failure, connection reset, 503/SlowDown, timeout), and detection through a wrapping `RuntimeError`.

Found while reproducing published surface predictions from public inputs on an ordinary connection. Written with Claude Code as a coding assistant; the failure above is a real one from my own run, not a constructed example.

### DIFF
--- a/vesuvius/src/vesuvius/data/vc_dataset.py
+++ b/vesuvius/src/vesuvius/data/vc_dataset.py
@@ -45,6 +45,7 @@ class VCDataset(Dataset):
             domain: Optional[str] = None,
             skip_empty_patches: bool = True,  # Whether to skip empty (homogeneous) patches
             anon: bool = False,  # Use anonymous (unsigned) requests for S3 input paths
+            read_retries: int = 4,  # Attempts per read, forwarded to Volume
             ):
         """
         Dataset for nnUNet inference using the Volume class for data access and preprocessing.
@@ -83,6 +84,9 @@ class VCDataset(Dataset):
                              (e.g., 'np.float16', 'np.float32'). Default is 'np.float32'.
             domain: Data source domain for Volume ('dl.ash2txt', 'local'). Auto-detected if None.
             anon: Use anonymous requests for input S3 reads.
+            read_retries: Attempts per read, forwarded to Volume (default 4). Transient
+                remote failures are retried with exponential backoff so one dropped
+                connection does not abort a long streaming run. Pass 1 to disable.
         """
         self.input_path = input_path
         self.input_format = input_format # Keep for informational purposes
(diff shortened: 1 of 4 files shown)


bruniss:
thank you for the PR ! 

the fix makes sense, i'd propose that `read_retries` would be better placed in `Volume.__getitem__` as it moves this at the reader and fixes any other users of `Volume`

TAUIL-Abd-Elilah:
Agreed — that is the better home for it. Moved.

The retry now lives in `Volume.__getitem__`, wrapping the store read itself, so every caller of `Volume` gets it rather than just the inference path. `VCDataset` and the `--read-retries` flag only forward the setting now; `VCDataset._read_with_retry` is gone.

Behaviour is otherwise the same: 4 attempts by default with 0.5s backoff doubling to 8s, transience judged across the whole `__cause__`/`__context__` chain (zarr and fsspec wrap aiohttp/botocore/urllib3/ssl differently), and deterministic errors — bad coordinates, missing array, auth — re-raised on the first attempt. `read_retries=1` disables it.

Moving it also made the tests better: they now drive the real `Volume.__getitem__` against a store that fails the way an object store does, instead of testing the helper in isolation. 14 tests, and `pytest tests/data` is 42 passed.

Rebased onto current `main`.

bruniss:
thanks for the updates! merging

PR #1245, opened 2026-07-25
docs: add windcheck to Segmentation tools

joe-carr-data:
Adds [windcheck](https://github.com/joe-carr-data/windcheck) to *Segmentation → Tools* in the community projects list.

**Updated 27 July** — rebased onto current main, and the entry now describes the tool as it stands rather than the earlier proximity-based version. Two volume-cartographer maintainers pointed out that proximity (how close a trace comes to itself) is undecidable in tightly packed regions; they were right, and the tool was rebuilt around a question with no threshold at all: a traced surface cannot pass *through* itself. The repository documents that correction publicly — including retracting the original null-control claim, whose denominator was wrong.

**What it does now.** Detects transverse self-intersection deterministically from mesh geometry alone — no labels, no model, no GPU, no volume download — and reports how far apart the two overlapping parts lie along the trace, in each segment's own revolution units. Reads `tifxyz`; emits a per-trace certificate plus an overlay in VC3D's own `PointCollections` schema.

**Validation, current:**

- Audited all 179 published segments of five scrolls (Scroll 1, Scroll 5, PHerc0139, PHerc0814, PHerc1667; 228M triangles): 160 contain at least one self-intersection. Precomputed results for every segment ship in the repo.
- The revolution estimate is checked against Scroll 1's *published* winding counts: 31/33 within 20%, median ratio 0.999; the two failures are loud and documented.
- The predicate agrees with FCL in both directions (250/250 positives, 249/250 negatives, the single disagreement explained); counts are deterministic across thread counts; ~90% of intersections survive every triangulation choice.
- Everything reproduces from 1.55 GB of hash-pinned inputs in about 90 seconds.

Crossings are reported as geometric findings, not proven tracing errors — three attempts to establish that link are documented as failures.

Submitted for the July 2026 progress prizes. Happy to reword or move the entry if another section fits better.

---

Unrelated, in case it is useful: the progress-prize form still links to `ScrollPrize/awesome-scroll-tools` for this pull request, which is archived and points here. Might be worth updating the form link.


joe-carr-data:
Rebased and mergeable. I think this one just missed the batch that went through on the 26th — it wasn't mergeable that morning and I rebased the day after.

I've also cut the entry down to one sentence to match the rest of the list. It was 910 characters where the neighbouring entries run 113–235; the detail it carried is all in the linked repo. For reference, the Vercel check fails here the same way it does on #1206, #1212, #1236 and #1243, which all merged from forks on the 26th, so I don't think it's a gate — but happy to do something about it if it is.

Glad to reword or move the entry if another section fits better.

PR #1246, opened 2026-07-26
Fiber sync-conflict handling, sync-side only: loader-safe 3-way merge + VC3D reoptimize-on-load prompt

skyward7187:
Replaces #1223 (see the discussion there). The original PR paired the sync-side merge with a large VC3D loader rework (link quarantine, verbatim serialization, non-destructive loading). This version keeps **all conflict handling in the sync process** and reduces the VC3D footprint to a single feature: a load-time prompt to re-optimize fibers the merge could not re-fit.

## Design contract: merge output must satisfy VC3D's existing loader

VC3D's fiber loader is strict and destructive (`LineAnnotationController.cpp`): inconsistent files trigger a repair prompt whose default button deletes branch links from disk, and declining drops whole fibers from the session. Instead of changing the loader, the merge is engineered so its output always passes validation:

- **C1**: every control point must appear in `line_points`, in order, within 1e-8 (`validateFiberInputControlPoints`) — violating this makes the file unloadable. So `control_points`/`line_points` are always written as a consistent pair from one source.
- **C2**: cross-fiber link reciprocity is **index-exact across files** (plus positions and directions). So the sync mirrors every merged fiber's link decisions into its peer files and uploads them too.
- **C3**: stored link directions must match tangents recomputed from `line_points` within ~0.26° (`branchDirectionsCompatible`). So directions are recomputed with an exact Python port of VC3D's tangent computation.

## Commit 1 — vc_sync + fiber_merge (all conflict handling)

Everything from #1223's sync commit survives: `local_md5` content tracking (lazy DB migration), `.s3sync-base/` shadow copies as merge bases, `.s3sync-conflicts/` stashes before any overwrite, ETag-aware verification, deferred merge application (nothing touches local files until the sync is confirmed), `--no-auto-merge`, dry-run previews.

**`fiber_merge.py` rewritten around the loader contract:**
- **Same geometry on both sides** (the incident case — divergent link edits): geometry wholesale from the newer side; branches merge with set semantics (unions, base-aware deletions, approvals beat pending and deletion); tags merge base-aware.
- **Geometry changed on one side**: that side's `control_points`+`line_points`+`hv_classification` are taken as the written pair; links from the other side are re-anchored by position.
- **Geometry changed on both sides** (disjoint regions; overlaps conflict): diff3 control points; `line_points` is set to the merged control points verbatim (C1 holds trivially) and the fiber is tagged **`needs_reoptimization`** — VC3D re-fits it on load (commit 2). The ~100-line line-splicing machinery from #1223 is gone.
- A link whose anchor control point was deleted by the merged geometry makes the **whole merge a manual conflict** — links are never silently dropped (#1223 dropped them with a note; a review repro also showed a tolerance mismatch silently dropping valid links — that class of bug is structurally gone).

**New: cross-file link-consistency pass (`refresh_pair_links`).** After merges are planned, the sync mirrors each merged fiber's link decisions into its peers — restoring missing reciprocals (the approval-beats-deletion case), fixing index shifts, recomputing all four endpoint directions — against each peer's *post-sync* content (pending merge result, remote content for scheduled downloads, local content otherwise). Peers that change are stashed and rewritten together with the merges — before any network transfer, so an interrupt can never strand a merged file without its reciprocal fixes — and then uploaded (a peer fix supersedes any scheduled download of the same file; it was computed on the remote content). Anything unresolvable (missing peer, peer with its own manual conflict or pending deletion, unmatchable anchor) demotes the merge to the interactive prompt. The perform phase is reordered to **merges → peer fixes → downloads → uploads → deletes** to make this possible.

**Review fixes from #1223 folded in:**
- `update` now verifies and backfills content hashes + merge bases for already-tracked in-sync files (the backfill the #1223 description promised but didn't deliver).
- `status`/`--dry-run` no longer create anything inside the sync directory (transient fetches moved to a per-run temp dir).

## Commit 2 — VC3D: load-time re-optimization prompt (the only C++ change)

~130 lines in `LineAnnotationController`, no loader changes:
- After fibers load, if any carry `needs_reoptimization`, a deferred prompt offers **"Re-optimize now"** (default) / "Not now". Declining keeps the tag (visible in the fiber panel's tags column) and asks again next load. Suppressed-dialog sessions (agent bridge) log and skip.
- Re-optimization runs through a pane-less session (`makeIntersectionLineSession` + `saveSessionAsFiber`): the existing pipeline finalizes the optimization synchronously, refreshes branch endpoint directions and reciprocal metadata on linked fibers, recomputes `hv_classification`, bumps `generation`, and strips the tag in the same write. Failures keep the tag for a retry.

## Testing

- `pytest scripts/tests/` — **110 passed**. The high-value additions are direct loader-safety assertions: `loader_issues()` is an INDEPENDENT port of VC3D's load validation (its own Euclidean/finite predicates and tangent computation, deliberately not reusing fiber_merge's — a merge primitive that drifts from the loader fails tests instead of drifting with them) (C1 subset check, per-branch checks, index-exact reciprocity with directions), and every clean-merge test — including an end-to-end two-sided geometry merge of a linked pair followed by the consistency pass — must produce output it accepts. The incident is still a fixture (base {L1}, local +L2, remote +L3 → all three survive).
- VC3D builds clean; C++ suite passes (no loader behavior changed).

**Post-review hardening** (commit 3, addressing an offline review): the position predicate is now exactly the loader's Euclidean metric; reciprocal pairs are snapped byte-exact (each-within-tolerance is not enough for the loader's stored-vs-stored checks); one demotion fixpoint guarantees peer fixes never override a manual conflict choice; the C++ reopt loop re-checks tags (stale runtime ids across the modal); malformed JSON degrades to manual resolution everywhere; plus assorted txn/staging/temp-dir hygiene.

## Deliberately not in this PR
The in-app save-time staleness guard from #1223's follow-up list: a stale open VC3D session can still overwrite a synced file. The sync-side conflict copies (`.s3sync-conflicts/`) remain the safety net for that case.

https://claude.ai/code/session_01GCypuNyFABv816RQDt3UbN


skyward7187:
Commit `a3ab9885f` addresses findings from an offline review of this PR. Since that review isn't public, here is what each change fixes and why — the numbering below is self-contained.

## Correctness fixes

**1. The position predicate didn't match the loader's metric (data-destroying).**
`fiber_merge.pos_eq` accepted per-axis `|d| ≤ 1e-6` (a Chebyshev box), but VC3D's `pointsApproximatelyEqual` requires *Euclidean* `‖a−b‖ ≤ 1e-6`. The box accepts positions up to √3·1e-6 apart that the loader rejects. Reproduced consequence: geometry jittered ~8e-7/axis on one side counted as "unchanged", the merge kept a link whose stored position was 1.386e-6 (Euclidean) from the merged control point, and the loader's repair prompt would delete that link from disk — exactly the destruction class this PR exists to prevent. `pos_eq` is now squared-Euclidean; sub-tolerance jitter merges cleanly and beyond-tolerance jitter becomes a *manual conflict* instead of loader-poison. One shipped test had encoded the broken behavior as expected; it's inverted now.

**2. "Leave it if within tolerance" wasn't closed under the loader's pair checks.**
`refresh_pair_links` only rewrote a stored direction/position when it failed the tolerance check against the recomputed truth. But the loader's reciprocity check compares *stored A against stored B* with the same tolerance — two fields each within tolerance of the truth can be 2× tolerance apart from each other (e.g. +0.2° and −0.2° both pass individually, the pair is 0.4° apart → "missing reciprocal branch" → destructive prompt). The pass now snaps fields **byte-exact**: positions to the control point's value, directions to ±recomputed tangent (stored sign preserved — it encodes the link's sense; the loader is sign-agnostic). VC3D-written fields are already exact, so consistent pairs still see zero changes and no upload churn.

**3. A demotion hole could override the user's manual conflict choice.**
Merges demoted *during* link-consistency planning were appended to the manual list without re-running the blocked-peer cascade. Scenario: merges A and B where A links to B; B demotes (its own peer unreadable); the re-plan then treated B as an ordinary peer, computed a fix for it from stale local content, and queued it for upload — so if the user later picked "[r]emote → local" for B at the prompt, the peer fix would silently overwrite their choice and re-upload it. Both demotion causes now run in **one fixpoint** (`_plan_conflict_resolutions`, extracted and unit-tested with exactly this A→B cascade), so a demoted file always re-enters the cascade before any peer fix survives planning.

**4. Interrupt window between merges and peer fixes (permanent inconsistency).**
Perform order was merges → downloads (arbitrarily long) → peer fixes → uploads. Dying during downloads left the merged file on disk without its reciprocal fixes; the next sync sees it as a plain local edit and uploads it with **no merge and no consistency pass** — the cross-file break never heals, and every other machine's VC3D hits the destructive prompt. Peer fixes are now written **immediately after the merges, before any transfer**, and a peer fix *supersedes* a scheduled download of the same file (the fix doc was computed from the remote content, so nothing is lost — and this also removes the plan-time/perform-time remote staleness race and the "download failed but fix uploaded anyway" case).

**5. C++: stale runtime IDs across the modal prompt could re-optimize unrelated fibers.**
Fiber runtime ids are densely reassigned on every reload, and the reopt prompt collects ids *before* a modal `exec()` — a reload during the modal (agent bridge, imports) could retarget the batch at unrelated fibers. The loop now **re-checks the `needs_reoptimization` tag on every fiber it touches**, so anything it re-optimizes is tagged regardless of id drift. Relatedly, the prompt-pending flag now clears only after the work finishes (a reload during the modal can't schedule a second prompt over files whose tag-strip saves are still in flight).

**6. Malformed (remote-controlled) JSON could crash the merge or abort the whole sync.**
Concrete inputs like `generation: "x"`, string coordinates, or dict tags raised uncaught exceptions; `merge_fibers` was guarded by the caller but the planner's `refresh_pair_links` calls were not — one bad peer doc aborted the run mid-planning. `is_fiber_doc` now validates point/tag/generation *content* (all the crash inputs are pinned in a parametrized test), and the planner wraps refresh calls in try/except → demote to manual, matching the merge call's philosophy.

## Hardening / smaller fixes

- `branch_file` is compared as a **basename** everywhere, like the loader (`fs::path(...).filename()`): a legacy `fibers/x.json` entry is the same peer, and a crafted `../../x.json` can no longer become a filesystem path or S3 key (plus an explicit peer-name check in the planner as defense in depth).
- Review state (`pending`) is synced onto **existing** reciprocals too, not only on inserts — an approval applied by the merge now reaches the peer.
- The review-state merge is base-aware in both directions: deliberately re-flagging an approved link beats an untouched approval (previously approval always won).
- Atomic-write staging files are **dot-prefixed** so a crash between staging and `os.replace` can't leave a file the next sync uploads as a new S3 object.
- Divergence stashes before downloads **re-hash at stash time**, not scan time — an edit made during the (long) interactive prompt phase is preserved.
- The lazy `ALTER TABLE` migration tolerates a concurrent first run; S3 verification fetches were moved **out of open sqlite transactions**; the shared `_content_matches_remote` helper also fixes converged-content detection for multipart/KMS ETags (previously a pointless conflict).
- The per-run temp dir (which can hold fiber content) is removed on exit; the cancel message no longer overclaims ("no files modified" → accurate wording); the "previous remote available through bucket versioning" message is only printed after actually **verifying** versioning is enabled, with a warning otherwise.
- C++: the Lasagna dataset is resolved **once per reopt batch** (abort on failure — no per-fiber picker/error modals); `saveSessionAsFiber` strips the `needs_reoptimization` tag centrally, gated on a new `lineWasOptimized` flag so only sessions whose line the optimizer actually produced strip it (an ordinary pane save can no longer resurrect the tag, and an untouched placeholder session can't strip it); `QPointer` guards around the modal/nested event loops; `fileName` copied before saves that may reallocate `_fibers`.
- Duplicated logic consolidated: post-transfer tracking (4 copies → `_track_transfer`), verified-content checks (3 copies → `_content_matches_remote`), the two demotion loops (→ the fixpoint above).

## Tests (86 → 110)

The biggest testing change: `loader_issues()` — the harness asserting merge output passes VC3D's loader — was previously built from `fiber_merge`'s own predicates, so a primitive that drifted from the C++ (finding 1) kept tests green. It is now an **independent** port (its own Euclidean/finite predicates, tangent computation, parse-strip modeling, basename normalization): predicate drift now fails tests. New regression tests cover findings 1–3 and 6 directly, plus direction-sign preservation, pending sync, the demotion cascade and tentative-fix rollback, delete-vs-download action pinning, and a migration test against a genuinely old-schema DB.

## Considered, deliberately unchanged

- A link whose anchor was *moved* (beyond tolerance) on one side while the other side deleted it still merges as an addition — a moved link is indistinguishable from a re-created one; now documented in code.
- Structurally unparseable branch entries are still carried through the merge even though an unmodified VC3D strips them (and rewrites the file) on next load — that's VC3D's pre-existing behavior for such files with or without a merge; the docstring now says so explicitly.
- No cross-process file locking (two concurrent syncs remain last-writer-wins with stashes as recovery); only the migration race is guarded. Can be added if concurrent syncs on one machine are a real workflow.

https://claude.ai/code/session_01GCypuNyFABv816RQDt3UbN

skyward7187:
A second offline review pass (same arrangement as [the first](#issuecomment-5083312456)) produced another round of findings. This comment documents what the newest commit fixes and why — and, equally, what I'm deliberately **not** changing and the reasoning, so the record is complete.

## Fixed

**Data-loss path with no safety copy (the one that mattered most).** `DELETE_LOCAL` — remote deleted a file that has *unsynced local edits* — was the single overwrite/remove path with no conflict stash and no divergence check: a bulk-confirmed sync deleted the edits irrecoverably. The perform phase now runs local deletions through the same stash-if-diverged guard as downloads (hashing at stash time), and the analyze phase surfaces the case explicitly ("S3 file was deleted; local copy has unsynced edits (conflict copy will be saved)").

**`hv_classification.manual_tag` was silently discarded.** It's the one field inside `hv_classification` that's a user decision rather than derived (the loader round-trips it independently of the recomputed scores), but the merge took the whole dict from the carrier side. It now three-way merges base-aware like every other user-editable field: the side that changed it wins; both-changed-differently is a conflict.

**Remote-triggerable crash.** A crafted `branch_file` with an embedded NUL reached `open()` → plain `ValueError`, which the `(OSError, json.JSONDecodeError)` handler doesn't catch (JSONDecodeError is a ValueError *subclass*, not the reverse) — aborting the whole sync during planning. Fixed twice over: the peer-name guard now rejects non-printable names (plus `:` for drive-relative Windows paths), and the reader catches `ValueError`.

**Agent-bridge sessions could still block on the reopt modal.** The bridge scopes dialog suppression per RPC; a fiber load inside a suppressed handler scheduled the 0 ms prompt timer, which fired *after* the scope restored suppression — modal on the GUI thread with nobody to click it. Suppression is now also checked at **schedule** time (log-and-skip), with the fire-time check kept as a second gate.

**Reopt loop correctness.** The batch now keys fibers by `fileName` instead of runtime id (ids are densely reassigned on reloads, which can happen while the modal spins — the tag re-check alone was correct but not identity-exact); `waitForFiberSaves()` now also covers tag-strip-only saves (previously a batch of degenerate fibers could clear the prompt-dedup flag while writes were in flight); the tag-strip save is try/guarded like its sibling call sites; and a failed optimization **aborts the batch** instead of stacking one error modal per remaining fiber (failures there are systematic — dataset/volume — and all untouched tags are kept for retry).

**Migration guard was too broad.** The lazy `ALTER TABLE` swallow now only covers `duplicate column` (the concurrent-first-run race it was written for); anything else — classically `database is locked` — fails loudly instead of resurfacing as `no such column` mid-perform-phase.

**Hardening batch:** `allow_nan=False` on every JSON write plus a plan-time serializability check (a NaN smuggled through an opaque subtree would otherwise produce a file nlohmann cannot parse at all — now it demotes to manual); entries carrying the obsolete `link_direction` key are treated as opaque, since the loader strips them at parse time and mirroring one into a peer would cement a pair the loader then tears down; `is_fiber_doc` rejects `version != 1` and `tags: null` (both unloadable); the byte-exact direction snap now has a ~1e-12 acceptance window so a VC3D build with contracted float math (1-ULP differences) can't cause every refresh to rewrite every direction (pairwise still 7 orders of magnitude inside the loader's tolerance); rewritten peers bump `generation` (keeps the newer-side heuristic honest); a used-set prevents duplicate A-entries from claiming the same peer reciprocal; POSIX-joined peer paths (host-OS independent); interactive "[r]emote" stashes are scoped to annotation-sized files like every other stash site (with a warning otherwise); trivial-merge short-circuits now report `peer_files` too, so the (no-op when consistent) peer pass can heal a broken lockstep-write invariant; peer docs are cached per run (no re-fetching remote JSON every fixpoint round); demotion reasons survive into the interactive prompt instead of being flattened to a generic message; assorted dead-parameter/comment cleanups.

**Corrected documentation claim:** `_apply_peer_fixes` claimed an interrupt between merge-apply and fix-apply was *permanent*. It isn't — tracking only advances on verified upload, so the next sync re-classifies the merged file as a conflict, re-merges over the still-valid base, and re-plans the fixes. The comment now states the real reason ordering matters: VC3D opened *during* the window would see the inconsistency and offer its destructive repair.

Tests: 110 → **122** (delete-vs-edit stash + reason, NUL/NaN demotions, manual_tag merge + conflict, link_direction opacity, unloadable-variant rejection, short-circuit peer reporting, ULP tolerance, generation bump, duplicate-reciprocal claiming). C++ suite still green.

## Reviewed and deliberately NOT changed

- **Per-merge interleaving of merges + peer fixes** (proposed to shrink the crash window further): declined. Since round 1, both phases run back-to-back with *no network in between* — the window is a handful of local atomic renames. Per-merge grouping would need per-merge fix attribution (one peer can aggregate fixes from several merges), and the residual hazard — VC3D opened mid-window — exists between any two file writes regardless of grouping. Cost exceeds the microseconds saved.
- **Distinguishing "moved link" from delete+add in stats/notes**: declined. It would take fuzzy base-matching by far-endpoint with real ambiguity when several links target the same peer; the merge *outcome* is correct and documented in code — this is cosmetic reporting.
- **Making `refresh_pair_links(base_doc=...)` required**: declined; the deletion-mirroring hazard of omitting it is now documented in the docstring, the shipped caller always passes it, and base-less calls are legitimate in tests.
- **End-to-end pytest run through the real C++ loader**: declined for CI practicality (needs a VC3D build + volume). The independent Python port of the loader checks is the automated guard; the real-loader pass stays part of manual verification before merge. Port-drift risk is accepted and documented.
- **Conditional-PUT / cross-machine write races**: declined, same rationale as declining cross-process locking in round 1 — the design is last-writer-wins with conflict stashes and S3 versioning as recovery; changing that is a project, not a fix.
- **sync()-level integration test of the peer-fix upload/supersede wiring**: declined for now — it's four lines of list plumbing over unit-tested pieces, and an interactive-`sync()` harness (stubbed prompts/confirms/S3) is more scaffolding than the code it would cover. Reasonable follow-up if `sync()` grows further.

https://claude.ai/code/session_01GCypuNyFABv816RQDt3UbN

PR #1247, opened 2026-07-26
vesuvius-py: predict --bbox: take the patch grid from the volume, so an ROI run matches a full run

TAUIL-Abd-Elilah:
`--bbox` (#1241, mine) does not reproduce the corresponding region of a full-volume run. I found this while checking whether the published m7 surface predictions can be regenerated from public inputs, and the bbox flag turned out to be the reason they could not.

## The bug

`VCDataset` anchored the sliding window to the ROI origin:

```python
z_positions = [z0 + s for s in compute_steps_for_sliding_window(z1 - z0, pZ, self.step_size)]
```

`compute_steps_for_sliding_window` spreads its positions evenly across whatever extent it is given, so the extent decides the stride. For PHerc0139 with the m7 patch size (192) at the default step 0.5:

| grid | positions near z=8000 | stride |
|---|---|---|
| whole volume (20974) | 7853, 7949, **8045**, 8140, 8236 | ~95.8 |
| 256-long bbox at 8000 | 8000, 8064 | 64 |

Not one position in common. Every voxel in the ROI is therefore covered by a different set of patches, with different Gaussian blending weights, so the blended logits differ from a full run even deep inside the ROI, far from its own edges.

## Measured against the published artifact

The published surface predictions for every scroll come from one model run (`20260413222639` = `hf://scrollprize/surface_m7_nnunet`), and both the model and the CT are public, so the region can be regenerated and compared exactly. PHerc0139, bbox `8000:8256,3200:3456,3200:3456`, scored on the ROI interior (64 voxels trimmed per face), thresholded at 0.2 to match the published `th0.2` artifact:

| grid | Dice vs published | disagreeing voxels |
|---|---|---|
| ROI-anchored (before) | 0.8001 | 181,086 (8.63%) |
| **full-volume grid (after)** | **0.9997** | **266 (0.013%)** |

After the fix all 266 remaining voxels lie within 0.01 of the threshold, which is what float16 storage and autocast leave behind; AUC is 1.0000. Before the fix, a sweep over every threshold from 0.01 to 0.95 topped out at 0.8019, so no calibration or precision difference could account for the difference — it was the patch grid.

Worth stating plainly since it may be useful to others: **the published m7 surface predictions do reproduce from public inputs**, once the grid matches.

## The fix

Build the grid the whole volume would have used, then keep the patches that intersect the ROI. Positions are then always a subset of the full-volume grid, so a bbox run reproduces the corresponding sub-region of a full run — which is the reason to offer the flag at all.

**Cost, stated honestly:** selected patches may overhang the requested bbox by up to one patch, so a run reads a margin around the ROI. That is unavoidable rather than a choice — a voxel's blended value depends on every patch covering it, and those patches extend past the ROI edge. If you would rather keep the old behaviour available for people who want reads strictly bounded, I am happy to add a flag for it; I did not add one because a silently-different result seemed like the wrong default.

## Second bug, found by the new test

With the grid corrected, patches legitimately start *before* the ROI. `--num_parts` partitioned the ROI range and assigned each patch by its starting coordinate, so those patches fell into no part at all and were silently dropped — 60 of 120 positions lost on a 3-way split of the test ROI. Now the partition covers the span the selected patches actually occupy, and the union over all parts is again exactly the single-part run.

## Tests

`tests/data/test_vc_dataset_bbox.py` is rewritten around the invariant that matters:

- positions are drawn from the full-volume grid
- they are exactly the patches intersecting the ROI
- they are a subset of a full run's positions
- they leave no gap inside the ROI
- `num_parts` partitions without loss

The old tests asserted patches lay wholly inside the bbox, which is what the bug looked like from outside. 12 tests, and `pytest tests/data` is 30 passed.

Verified end to end on GPU against the public Open Data bucket: 125 patches, `blend_logits`, then the comparison above.

### DIFF
--- a/vesuvius/src/vesuvius/data/vc_dataset.py
+++ b/vesuvius/src/vesuvius/data/vc_dataset.py
@@ -290,12 +290,23 @@ class VCDataset(Dataset):
                     print(f"\nUsing bbox-restricted sliding window (global coords): "
                           f"z=[{roi[0][0]}, {roi[0][1]}), y=[{roi[1][0]}, {roi[1][1]}), "
                           f"x=[{roi[2][0]}, {roi[2][1]})")
-                # Grid over the ROI extent, offset back into the global frame so all
-                # emitted coordinates remain volume-absolute.
+                # Build the grid the whole volume would have used, then keep the
+                # patches that touch the ROI. Anchoring the grid to the ROI origin
+                # instead would place patches the full-volume run never places -
+                # a 256-long ROI yields stride 64 where the volume yields ~96 - so
+                # every voxel gets a different set of overlapping patches and the
+                # blended result no longer matches the corresponding sub-region of
+                # a full run. Selecting from the global grid keeps them identical,
+                # which is the whole point of being able to redo one region.
                 (z0, z1), (y0, y1), (x0, x1) = roi
-                z_positions = [z0 + s for s in compute_steps_for_sliding_window(z1 - z0, pZ, self.step_size)]
-                y_positions = [y0 + s for s in compute_steps_for_sliding_window(y1 - y0, pY, self.step_size)]
-                x_positions = [x0 + s for s in compute_steps_for_sliding_window(x1 - x0, pX, self.step_size)]
+
+                def _overlapping(extent, patch, lo, hi):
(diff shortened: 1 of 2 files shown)


jrudolph (approved):
LGTM

jrudolph:
Thanks, good find.

PR #1248, opened 2026-07-26
community projects: add vesuvius-repro, and correct the figure in my #1243 entry

TAUIL-Abd-Elilah:
Two changes to `20_community_projects.md`.

**Adds `vesuvius-repro`.** It checks whether the published surface predictions can be regenerated from public inputs, for any scroll. Working from `hf://scrollprize/surface_m7_nnunet` and the public CT, I ran all 36 scrolls that carry a published m7 prediction. **All 36 reproduce**, at Dice 0.9983–1.0000, and in every one 100% of the differing voxels sit within 0.01 of the 0.2 threshold — the residue of float16 storage, with nothing structural left. PHerc0500P2 reproduced exactly, zero differing voxels out of 2.1 million.

They do not all reproduce under the *same* settings, which is the one thing worth flagging: 35 need mirroring TTA **off**, and **PHerc. Paris 4 needs it on** — 0.8907 without it, 0.9999 with it. The converse holds too, so it is not coincidence: PHerc0139 run *with* TTA drops from 0.9997 to 0.8273. Nothing in the published artifacts records which configuration produced them — the `.zattrs` carry only boilerplate `multiscales` — so reproducing a given prediction currently means searching for its settings. Recording the inference config alongside each prediction would remove that entirely.

I originally filed #1250 claiming PHerc. Paris 4 was not reproducible at all. That was wrong and I have corrected it there; this entry reflects the corrected result.

**Corrects a figure I got wrong in #1243.** That entry said a 200³ region reads "27 chunks (56.6 MB)". That was measured before #1247, when the `--bbox` grid was anchored to the region rather than the volume — the arrangement that made region runs silently disagree with full-volume runs. Selecting the patches a full run would actually place means reading their overhang too, so the same region touches **125 chunks (262 MB)**, not 27. Still a large saving against 10,368 chunks (21.7 GB), just not the one I first reported, and the extra read is precisely what buys the agreement.


TAUIL-Abd-Elilah:
Updated the branch to correct the evidence scope and provenance wording. Follow-up checks now cover one selected 256³ region (central 128³ scored) from each of all 41 m7 artifacts across 36 scrolls—not full artifacts or volumes. Forty TTA-off checks match at Dice 0.9983–1.0000; Paris 4 matches with TTA. It also says provenance was missing during the audit but #1253 now records it and existing artifacts were backfilled. The ROI entry carries the corrected 125 chunks / 262 MB / 83× result. Branch remains focused on this documentation file.

TAUIL-Abd-Elilah:
Ready for review whenever suits — it is a single docs file, +3/−1.

Current state, for whoever picks it up:

- Every CI job passes. The one red check is Vercel's *"Authorization required to deploy"*, which is the fork preview-deploy gate and needs a maintainer to authorize; there is nothing I can do about it from this side.
- The entry has been corrected twice since opening, so the scope claim now matches the evidence: one selected 256³ region (central 128³ scored) from **each of all 41 m7 artifacts across 36 scrolls**, not full artifacts or volumes. Forty TTA-off checks match at Dice 0.9983–1.0000; PHerc. Paris 4 matches at 0.9999 with TTA.
- It also fixes a figure in my own earlier #1243 entry.

No urgency if the docs queue is long — flagging it only because it has been open since 26 July and I would rather it be reviewed than quietly aged out.

TAUIL-Abd-Elilah:
Bumping this gently — still the single docs file (+3/−1). Everything is green apart from the Vercel fork-preview gate, which needs a maintainer to authorize and is not something I can clear from my side. No urgency, just flagging so it does not quietly fall off the list.

erdpx:
thanks for the pr! before we merge, could you remove the region-of-interest inference entry? i think we'll keep the community projects page focused on standalone projects rather than individual villa prs

TAUIL-Abd-Elilah:
Done — removed in 2f5ce3b, so the PR is now just the vesuvius-repro entry (+1/−3). That reasoning makes sense to me; the `--bbox` work is a villa feature, not a project of its own, and the page reads better without individual PRs in it.

One thing you may or may not want, entirely your call: there is a standalone repo behind that entry, https://github.com/TAUIL-Abd-Elilah/vesuvius-roi-inference, with the reproducible benchmark and install recipe. If a standalone entry would fit the page I can open a separate PR for it, and if not I will leave the page alone.

TAUIL-Abd-Elilah:
Heads up that the red CI here is not this PR — it looks repo-wide since about 16:00 UTC. Every Continuous Integration run in villa since then has failed the same way, including ones with no connection to this branch (490c592, 6813788, 8095db9).

The failing jobs record **zero steps** and end at exactly 15m01s, so they are hitting the queue timeout without ever picking up a runner; `CI` then reports failure only because its aggregate step flags a cancelled required job. CodeQL passed on the same commit.

I could not re-run the workflows myself (needs admin), and I would rather not keep pushing commits at it — I have already retriggered twice, so the last force-push collapsed those into the single content commit rather than leaving retrigger noise on a docs PR.

Content-wise this is ready: +1/−1, just the vesuvius-repro entry, with the region-of-interest entry removed as you asked.

PR #1253, opened 2026-07-26
inference: record whether TTA was used in the prediction output attrs

TAUIL-Abd-Elilah:
A stored prediction can't currently be reproduced from itself. The output store already records `patch_size` and `overlap`, but nothing about test-time augmentation — and TTA changes the result more than either of them.

I ran into this while auditing the published surface predictions. Reproducing all 36 takes two different settings, not one:

| region | TTA off | TTA on |
|---|---|---|
| PHerc0139 | **0.9997** | 0.8273 |
| PHerc. Paris 4 | 0.8907 | **0.9999** |

35 of the 36 match with TTA off; PHerc. Paris 4 only matches with mirroring TTA on. The converse is what makes it conclusive rather than a fluke — TTA *breaks* PHerc0139 by about as much as its absence breaks Paris 4. One setting, two scrolls, opposite directions. Each pair is scored on the identical bounding box.

The artifacts record nothing that distinguishes them: their `.zattrs` hold only the boilerplate `multiscales` block. So which setting produced a given published prediction has to be guessed, and I guessed wrong first — I had Paris 4 written up as "not reproducible" before finding the flag.

### The change

Two attributes on the logits store:

```python
self.output_store.attrs['tta'] = bool(self.do_tta)
if self.do_tta:
    self.output_store.attrs['tta_type'] = self.tta_type
```

`blending.py` already copies every attribute from the logits store onto the merged output, so this reaches the final prediction with no change there.

`tta_type` is written only when augmentation actually ran, so a prediction never claims a scheme it didn't use. `bool()` is explicit because `.zattrs` is JSON and a numpy bool wouldn't round-trip.

### Tests

`vesuvius/tests/models/run/test_inference_run_config_attrs.py` — five cases: TTA on, TTA off, rotation TTA, the JSON-serialisable bool, and a check that the existing attributes are still written. 22 pass across `vesuvius/tests/models/run/`.

### Scope

Deliberately just the one missing field, since patch size and overlap are already stored. Happy to widen it — model checkpoint identity and normalization scheme are the other two a consumer can't recover — if you'd rather have the whole set in one go.

Scoring scripts and the per-region results behind the table are at https://github.com/TAUIL-Abd-Elilah/vesuvius-repro (`results/variants/`).

### DIFF
--- a/vesuvius/src/vesuvius/models/run/inference.py
+++ b/vesuvius/src/vesuvius/models/run/inference.py
@@ -893,6 +893,13 @@ class Inferer():
             self.output_store.attrs['overlap'] = self.overlap
             self.output_store.attrs['part_id'] = self.part_id
             self.output_store.attrs['num_parts'] = self.num_parts
+            # Record whether test-time augmentation was used. Patch size and overlap are
+            # already here, but TTA is the setting that changes the output most and it
+            # leaves no trace otherwise, so a stored prediction cannot be reproduced from
+            # itself without guessing it.
+            self.output_store.attrs['tta'] = bool(self.do_tta)
+            if self.do_tta:
+                self.output_store.attrs['tta_type'] = self.tta_type
             if self.bbox is not None:
                 # Record the requested ROI (global voxel coords, half-open) so the
                 # output is self-describing about which region was processed.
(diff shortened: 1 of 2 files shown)


jrudolph (approved):
lgtm, doesn't hurt to add it also there. We also recorded it from the outside because we pass the tta explicitly. We just didn't propagate it to the metadata so far but I backfilled the existing entries.

In any case, thanks for the cross-checks and validation of our data!

PR #1256, opened 2026-07-26
inference: accept both underscore/hyphen for flags

TAUIL-Abd-Elilah:
The three stages of the documented pipeline disagree with each other on two options:

| parameter | `vesuvius.blend_logits` | `vesuvius.finalize_outputs` |
|---|---|---|
| worker count | `--num_workers` | `--num-workers` |
| output chunk size | `--chunk_size` | `--chunk-size` |

So running the pipeline end to end means switching naming convention between stage 2 and stage 3 for the same two parameters, and the failure is just `unrecognized arguments`.

`predict` has the same split internally — `--model_path`, `--model_cache_dir`, `--input_dir`, `--patch_size` and `--tta_type` use underscores, while `--model-type` uses a hyphen. The underscore spelling is the natural guess and it errors out.

### The change

Both spellings are accepted:

```python
parser.add_argument('--chunk-size', '--chunk_size', dest='chunk_size', ...)
parser.add_argument('--num-workers', '--num_workers', dest='num_workers', ...)
parser.add_argument('--model-type', '--model_type', dest='model_type', ...)
```

Purely additive. Every existing invocation parses exactly as before, `dest` names are unchanged, and the tests assert the defaults are untouched.

Parser construction moves into `build_parser()` in the two files touched, because otherwise there is no way to test a CLI here — `main()` keeps its behaviour and just calls it.

### Tests

`vesuvius/tests/models/run/test_cli_flag_spelling.py` — 8 cases: both spellings for each of the three options, plus defaults-unchanged guards. 25 pass across `vesuvius/tests/models/run/`.

I hit this porting a run between the stages; happy to extend the same treatment to the remaining hyphenated flags (`--skip-empty-patches`, `--zarr-compressor`, `--zarr-compression-level`, `--read-retries`, `--threshold-value`, `--delete-intermediates`) if you'd like them consistent too, or to drop the `build_parser()` extraction if you'd rather keep `main()` as it was.

### DIFF
--- a/vesuvius/docs/inference.md
+++ b/vesuvius/docs/inference.md
@@ -42,8 +42,8 @@ vesuvius.predict \
 | `--normalization` | Runtime normalization (`instance_zscore`, `global_zscore`, `instance_minmax`, `ct`, `none`).
 | `--intensity-properties-json` | nnU-Net style JSON with intensity stats for CT normalization.
 | `--device` | Device string such as `cuda`, `cuda:1`, or `cpu`.
-| `--skip-empty-patches` / `--no-skip-empty-patches` | Toggle automatic removal of homogeneous patches.
-| `--zarr-compressor` / `--zarr-compression-level` | Configure output compression (`zstd` with level `3` by default).
+| `--skip_empty_patches` / `--no_skip_empty_patches` | Toggle automatic removal of homogeneous patches.
+| `--zarr_compressor` / `--zarr_compression_level` | Configure output compression (`zstd` with level `3` by default).
 | `--scroll_id`, `--segment_id`, `--energy`, `--resolution` | Metadata when reading remote scrolls via the `Volume` helper.
 | `--hf_token` | Hugging Face token for private repositories.
 | `--config-yaml` | Training YAML to resolve model architecture when the checkpoint lacks embedded metadata.
@@ -123,11 +123,11 @@ Finalize logits into probabilities or masks and optionally build a multiscale py
 ```bash
 vesuvius.finalize_outputs /tmp/merged_logits.zarr /tmp/final_output.zarr \
   --mode binary \
-  --threshold --threshold-value 0.3 \
-  --delete-intermediates
+  --threshold --threshold_value 0.3 \
+  --delete_intermediates
 ```
 
-`--threshold` toggles binarization on (default cutoff `0.5`). `--threshold-value T` overrides the cutoff with any value in `(0, 1)` and requires `--threshold`. The right cutoff is model-dependent and should come from validation — `0.3` above is illustrative, not a recommendation.
+`--threshold` toggles binarization on (default cutoff `0.5`). `--threshold_value T` overrides the cutoff with any value in `(0, 1)` and requires `--threshold`. The right cutoff is model-dependent and should come from validation — `0.3` above is illustrative, not a recommendation.
 
 ### Options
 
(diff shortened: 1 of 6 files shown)


jrudolph (commented):
Ugh, good catch. If we do that we should do it for all parameters. Is there a way to accept both automatically? I know that for some programming languages and tools such a conversion is sometimes automatic because `-` usually isn't a valid id character.

TAUIL-Abd-Elilah:
Yes — done, and you're right that it's the better shape. Pushed as `HyphenUnderscoreParser` in `vesuvius/utils/cli.py`, applied to all four parsers in `models/run` (predict, blend_logits, blend_and_finalize, finalize_outputs).

Every long option now answers to both separators automatically, so the hand-written aliases are gone and a flag added later needs no thought:

```python
parser.add_argument('--chunk-size', ...)   # --chunk_size works too
parser.add_argument('--num_workers', ...)  # --num-workers works too
```

Two things it's careful about, both covered by tests:

**An explicitly declared name beats a derived one.** Adding `--a_b` and then `--a-b` would otherwise fail — the first flag's derived alias has already claimed the second flag's real name, and argparse raises on the duplicate. The derived one is released instead, and each spelling still routes to the flag that declared it. This bit me while writing it; the test caught it.

**Derived spellings are hidden from `--help`.** Otherwise every line doubles in width. The help output is unchanged from before this PR.

Options with no separator (`--quiet`) and short options (`-v`) are untouched.

On the language comparison — that's exactly the precedent. Click normalises this way, and Python's own argparse already does the reverse half: it converts `--foo-bar` to the `foo_bar` attribute name because `-` isn't valid in an identifier. This just closes the loop on the input side.

33 tests pass in `vesuvius/tests/models/run/`. The one red check is the Vercel preview, which needs team authorization on fork PRs.

TAUIL-Abd-Elilah:
Quick follow-up: the branch is mergeable, the 33 focused parser tests pass, and all substantive CI checks are green; the only red status is the fork Vercel authorization. The generic parser now handles future long flags without per-flag aliases and preserves explicit-name precedence/help output. Is anything else needed from me?

jrudolph (commented):
Now that both versions are accepted, let's also fix all of the existing declarations to use underscores.

TAUIL-Abd-Elilah:
Done — pushed as `1ecb952`. Ten hyphenated declarations renamed to underscores across the four parsers:

| | |
|---|---|
| `predict` | `--skip_empty_patches`, `--no_skip_empty_patches`, `--zarr_compressor`, `--zarr_compression_level`, `--model_type`, `--read_retries` |
| `finalize_outputs` | `--threshold_value`, `--delete_intermediates`, `--chunk_size`, `--num_workers` |

`blend_logits` and `blend_and_finalize` were already fully underscored, so nothing to do there.

Nothing breaks: every renamed flag keeps its old hyphenated spelling as a derived alias, so scripts and container commands that already pass `--zarr-compressor` or `--delete-intermediates` are unaffected. What changes is what `--help` prints and what the next flag copies.

Three smaller things that came with it:

- The explicit `dest=` on the renamed flags is now redundant and is gone. The one kept is `--no_skip_empty_patches`, which points at another flag's `dest` deliberately.
- Error and help strings that quoted the hyphenated names now quote the declared ones (`--threshold_value requires --threshold`, `Please specify --chunk_size.`, `rerun with --model_type resnet.`).
- `vesuvius/docs/inference.md` updated to the declared spellings.

To keep the two conventions from drifting apart again, `test_declarations_are_underscored` walks all four parsers and fails on any hyphenated declaration — the derived aliases are excluded, so it checks what was written, not what argparse ended up holding. Covering `blend_logits` and `blend_and_finalize` meant extracting their parsers into `build_parser()` / `build_blend_and_finalize_parser()`, the same shape `predict` and `finalize_outputs` already have in this PR; both `main()` functions are otherwise unchanged.

42 tests pass in `vesuvius/tests/models/run/`.

One thing I noticed and did **not** touch, since it is unrelated to this PR: `vesuvius/docs/inference.md` documents `--tif-activation`, `--intensity-properties-json` and `--config-yaml` for `vesuvius.predict`, and none of the three exists in the parser. Happy to send a separate PR removing them, or adding them, if you say which.

jrudolph (approved):
lgtm, thanks for the fix!

PR #1261, opened 2026-07-27
ci(windows): make the MCP test step actually run its tests

jrudolph:
Claude found this issue regarding running the mcp tests on windows which silently failed before because of missing dependencies:

The step prepends C:\msys64\ucrt64\bin to PATH so VC3D.exe can resolve its DLLs, but that also puts MSYS2's python.exe ahead of the one installed by actions/setup-python. MSYS2's interpreter has no pip, so `pip install -e .` failed with "No module named pip" and the mcp package was never installed. test_contract, test_real_bridge and test_runtime then all failed to import with ModuleNotFoundError: No module named 'mcp'.

The step still passed. pwsh only propagates the exit code of the last native command, and `python -m venv` runs after the unittest call and succeeds, so "FAILED (errors=5)" never reached the step result. These tests have been green-but-dead since #1179 added them.

Resolve the setup-python interpreter before the PATH edit and invoke it explicitly, keeping MSYS2 first on PATH for the child VC3D.exe. Check the exit codes of both the install and the test run.

This surfaces failures that were previously hidden, so the step may now go red for real reasons.


erdpx (approved):

erdpx:
thank you for the fix!!

jrudolph:
That fixes only the test run but we now might get actual test failures...

erdpx:
yep, i'll need to set up a better local windows testing env

jrudolph:
I'm just leaving it here like this (to avoid follow up failures in unrelated PRs), maybe you can look into the failures that now appeared.

PR #1263, opened 2026-07-27
Add PHerc1218 spiral-fit sweep to community projects

pscamillo:
Adds the PHerc1218 spiral-fit sweep: 21 windows across both anchored
bands of a Grand Prize-eligible scroll with zero human segments, 19
validated against pre-registered criteria, surfaces public on Kaggle
under MIT. Produced with the unmodified official fit_spiral off
IyanDopico's input pack, which is credited in the entry.

Also drops a whitespace-only line left over from the merge of #1172.

Note for the maintainers: IyanDopico's vesuvius-sheet-tools isn't listed
yet — the pack, the cleaning/sheet-instance tools and the reproduction
recipe are all public and are what this sweep runs on. Worth an entry of
its own; I've left that to him rather than describing his work for him.


pmh47:
Closing since these surfaces are not even remotely fitted to the scroll.

image

PR #1264, opened 2026-07-27
Fix final-quad bounds in geometry sampling

aviad12g:
## Summary

Fix the bilinear-cell bounds used by `loc_valid`, synchronize the two copied
guards that consume the same 2×2 cell contract, and add regression coverage
for the final complete quad.

The defect is still present on current `main` at
[`208faea`](https://github.com/ScrollPrize/villa/blob/208faea7d7fec62da03ff75e118e26be0c08a117/volume-cartographer/core/src/Geometry.cpp#L127-L168).

## Root cause

`loc_valid` built a half-open OpenCV rectangle with dimensions
`rows - 2` by `cols - 2`. A bilinear sample at `(row, col)` reads only the
four corners through `(row + 1, col + 1)`, so its valid origins satisfy:

```text
0 <= row < rows - 1
0 <= col < cols - 1
```

The old rectangle instead admitted only `(rows - 2) * (cols - 2)` origins.
In `vc_tifxyz2obj`, an all-valid 5×5 grid therefore emitted 9 quads
(18 triangles) with a 4×4 referenced-vertex union (16 vertices), rather than
16 quads (32 triangles) using all 25 vertices.

## Caller audit

I audited every production use of the shared helper (excluding
`GrowPatch.cpp`'s unrelated local `loc_valid(int)`):

- `vc_tifxyz2obj` (2), `vc_tifxyz_gengt`, `vc_tifxyz_winding`, `Atlas`,
  `PlaneSurface` (2), `GrowSurface` interpolation, and the two tracer cost
  functions all consume exactly one 2×2 bilinear cell.
- `QuadSurface::valid()` is a bilinear-coordinate predicate;
  `QuadSurface::coord()` had copied the same old bound before calling
  `at_int`, so it is switched to the shared predicate.
- Two `GrowSurface` paths validate through `loc_valid` and then call
  `SurfTrackerData`. Its three lookup/validation methods had copied the same
  old rectangle, so they now share `loc_valid` rather than disagreeing at the
  final cell.
- `ObjAlphaCompRefinement` is the only caller followed by a larger stencil.
  `grid_normal` independently clamps its 4×4 window to the interior; an
  explicit `<4` grid guard now covers the tiny-grid case.

No caller relies on the accidental extra high-side margin. Separately named
helpers with explicit interior-margin contracts are unchanged.

## What changed

- check row/column bounds directly before converting coordinates to indices;
- reject undersized grids and negative, non-finite, or extreme locations;
- make `QuadSurface::valid()` and `coord()` agree at the final cell;
- route `SurfTrackerData::{lookup_int, valid_int, lookup_int_loc}` through the
  same 2×2 validity contract; and
- guard `grid_normal` when a grid cannot supply its documented 4×4 stencil.

Regression tests cover rectangular and minimum-size grids, every complete
2×2 window, the final integer/fractional cell, sentinel corners, non-finite
and extreme coordinates, all helper overloads, `QuadSurface`, and
`SurfTrackerData`.

## Impact

This is a deterministic topology/sampling correction, not a claim that
existing scientific results change materially. The relative effect is largest
on small fixtures: a 3×3 grid loses 3 of 4 quads under the old bound. On a
634×1217 surface the omitted boundary is 1,848 of 769,728 possible quads
(about 0.24%). Boundary completeness still matters for export topology,
seam/stitch checks, and comparisons expected to cover the full grid.

## Reproduction and validation

The issue surfaced while reconciling TIFXYZ Doctor's fixture geometry with the
official stable VC3D exporter. Eight square/rectangular probes reproduce the
old closed form exactly:

```text
faces = 2 * (rows - 2) * (cols - 2)
referenced vertices = (rows - 1) * (cols - 1)
```

- `git diff --check`
- TIFXYZ Doctor evidence suite: 48/48 tests pass
- hash-pinned stable-release differential: 5/5 reader cases and 8/8 exporter
  size probes pass
- VC3D Linux GCC/Clang tests and macOS compile: pending maintainer approval of fork CI

Supporting evidence and authenticated harness:
https://github.com/aviad12g/tifxyz-doctor

Related maintainer follow-up from the diagnostic report:
[wrong-pipeline diagnosis](https://discord.com/channels/1079907749569237093/1243576621722767412/1531206054682165309) · [remediation confirmation](https://discord.com/channels/1079907749569237093/1243576621722767412/1531220162190245909)

## Prior-art check

Repository history and public issues/PRs were searched for `loc_valid`,
`rows-2`, final-row/final-column, and the 5×5 16-vertex/18-face signature. I
found no existing report or fix for this defect. The helper was mechanically
moved in
[`97aba84`](https://github.com/ScrollPrize/villa/commit/97aba84e68e6c0af7b454eed16350b6bb00e826f);
PR #1207 concerns a separate last-quad atlas clamp.


aviad12g:
I ran a follow-up three-way trace ablation on a real public PHerc1447
TIFXYZ surface. This adds release-path evidence for the `SurfTrackerData`
part of the patch.

The source is the published 160×136
[`tifxyz_original`](https://vesuvius-challenge-open-data.s3.us-east-1.amazonaws.com/PHerc1447/segments/20250502185519-auto_grown_20250502164303733/mesh/intermediate/tifxyz_original/)
for `20250502185519-auto_grown_20250502164303733`. A pre-run mask scan
found exactly six all-corners-valid quads on the final legal row/column.

| Variant | Resume points | Saved shape | Valid XYZ vertices |
| --- | ---: | ---: | ---: |
| Official `05ff9ea` release | 15,696 | 158×134 | 15,692 |
| Helper-only bound | 15,702 | 159×135 | 15,692 |
| Helper + `SurfTrackerData` bounds | 15,702 | 159×135 | 15,698 |

The helper-only and baseline surfaces contain the same 15,692 valid vertices
with numerically identical XYZ values; helper-only adds only a sentinel-padded
row and column. Adding the six `SurfTrackerData` bound changes recovers six valid
vertices at the six precomputed boundary locations. It also changes one
adjacent common vertex by 0.001953125 stored coordinate units. All other common
XYZ values are identical.

I ran three replicates of every variant with `OMP_NUM_THREADS=1`; TIFF hashes
and parsed observations were identical within each variant.

Important label: this is a byte-validated ablation of the official
`05ff9ea` macOS-arm64 release, not a source build of this PR. The two copied
dylibs change only eight decoded instructions from
`sub Wbound, Wdimension, #2` to `#1`: two in `loc_valid(Vec3f)` and six in
`SurfTrackerData::{lookup_int,valid_int,lookup_int_loc}`. Exact input/library
hashes and `__text` byte offsets are pinned, ad-hoc signing is verified, and
`DYLD_PRINT_LIBRARIES` logs record the loaded library paths. The public source
is an unverified auto-grown candidate, so this establishes a deterministic
tracer-output difference, not improved scientific accuracy.

The narrow result is that widening only the shared helper changes resume
admission and saved extent, but the stale `SurfTrackerData` copies must also be
widened for those six admitted boundary samples to survive into the saved XYZ
surface.

aviad12g:
Key causal result: fixing `loc_valid` alone recovers **no additional valid output**. The six vertices and four quads appear only after the stale `SurfTrackerData` bounds are synchronized. On this fixture that synchronization is load-bearing, not defensive cleanup.

I ran the preregistered one-generation resume protocol on the same public PHerc1447 fixture using four source-built Linux AppImages. Each variant was deterministic across three isolated replicates.

| Variant | Saved shape | Valid vertices | Valid quads |
| --- | ---: | ---: | ---: |
| Baseline `208faea` | 158×134 | 15,692 | 15,074 |
| Helper-only `d522691` | 159×135 | 15,692 | 15,074 |
| Minimal full-bounds `9e15071` | 159×135 | 15,698 | 15,078 |
| Actual PR `6e2bba9` | 159×135 | 15,698 | 15,078 |

The fixed output's last row is 158 and last column is 134. Every recovered `[row,col]` point lies exactly on one of those boundaries, as the off-by-one diagnosis predicts:

```text
[51,134] [52,134] [53,134] [158,88] [158,89] [158,90]
```

The source-built run also records one deterministic displacement of `0.001953125` (`2^-9`) at `[51,133]`. In the retained release-ablation TIFFs, the corresponding change is solely the stored Z value, from float32 bits `0x46b9ac8a` to `0x46b9ac8b`: one ULP at that magnitude. It repeats across replicates, so it is not run-to-run instability. All other common XYZ values are numerically unchanged.

Minimal full-bounds and the actual PR have identical X, Y, Z, generation, and valid-mask hashes in every replicate. This reproduces the earlier release-binary signature with exact source builds and isolates the `SurfTrackerData.cpp` synchronization as necessary for the six admitted boundary samples to survive into the saved surface.

- [Successful confirmation run](https://github.com/aviad12g/villa/actions/runs/30445977638)
- [Original compact JSON artifact](https://github.com/aviad12g/villa/actions/runs/30445977638/artifacts/8721453558), ZIP SHA-256 `0d734809b454ad2b965f46d9223cbac34cb622e814835dbde09147e974508eb7`
- [Durable evidence, exact evaluator, and workflow](https://github.com/aviad12g/tifxyz-doctor/tree/a0b54980c64efb20389425746e018c27bd938db7/verification/trace-impact/source-confirmation)

Scope caveat: this establishes a deterministic production-output change on a real public, unverified auto-grown fixture. Six vertices are 0.038% of the fixed valid-vertex count and four quads are 0.027% of the fixed valid-quad count. It does not establish that the recovered points improve scientific accuracy or quantify impact on other inputs and modes.

hendrikschilling (approved):

PR #1270, opened 2026-07-27
spiral: fail when the z-ROI filter drops every loaded patch

### LINKED ISSUE #1237: spiral: >=1 patch check validates the loaded pack, not the fit window — z-ROI filter can silently drop all patches (clean exit-0, zero-patch fit)
## Symptom

Fitting a real window that falls in a seed-patch coverage gap runs to a
clean `exit 0` with **zero patches** — no error, no warning:

- Window: PHerc1218, z 17600–18400 full-res (intra-band anchor gap in
  the vesuvius-sheet-tools v2 pack: no slab between L1 z8288 and z9184)
- Log shows ` loaded 25 patches` (the pack) followed by
  `fitting 0 patches` (the window), then a normal 30k-step run,
  constraints optimizing, meshes saved. Full log attached.

## Mechanism (verified line-by-line on 61bd95c, IyanDopico fork tip = upstream 6e78421 + #1207)

1. The only >=1 guard is the `No patches could be loaded` raise at
   `fit_spiral.py:898`. It validates the dicts **straight out of the
   pack loader** — i.e. the pack, not the window. (Nuance: it accepts
   verified OR unverified patches.)
2. The z-ROI intersection filter loop (`fit_spiral.py:905–918`, calling
   `patch_intersects_z_roi` from `spiral_helpers.py:21`) runs **after**
   it, deletes in place, and nothing re-checks.
3. Downstream tolerance is not accidental: `losses.py:573` has an
   explicit zero-patch branch, commented "supervision-free
   (disable_patches) fits: the umbilicus and shell anchors still
   apply" — written for intentional ablations, and it swallows the
   accidental all-dropped case identically.

Related doc finding (already corrected on the sheet-tools side,
IyanDopico/vesuvius-sheet-tools@3ccf6af): the umbilicus loss
(`losses.py:616–620`) is computed unconditionally from the umbilicus
polyline (default weight 1.25) — it does not live inside the
patch-loss path.

## Why it matters

A window landing in an anchor gap produces surfaces that *look*
validated (exit 0, constraints satisfied, plausible dr) but carry no
absolute seed anchoring. In coverage-gapped packs this happens
silently on real sweeps.

## Proposed fix

Re-check (or at minimum warn) **after** the z-ROI filter when
`disable_patches` is not set. The zero-patch branch itself is
legitimate for ablations and should stay.

Practical guard for pipeline users meanwhile: treat
`fitting N patches, N >= 1` in the log as a pass criterion —
`loaded N patches` refers to the pack, not the window.

## Credit

Found during an independent reproduction sweep of the PHerc1218
spiral-fit input pack (IyanDopico/vesuvius-sheet-tools). Code trace
verified by @IyanDopico, who co-signs this issue.

[run_z17600-18400.txt](https://github.com/user-attachments/files/30374651/run_z17600-18400.txt)

Nicodol:
Fixes #1237.

## Problem

As traced in #1237 (@pscamillo, verified by @IyanDopico): the only
`>= 1 patch` guard validates the dicts straight out of the pack loader
and runs before the z-ROI intersection filter. A window that falls in
a pack coverage gap therefore fits **zero patches** to a clean
`exit 0`: constraints optimize, meshes save, and nothing warns that
the fit ran without any absolute seed anchoring.

## Fix

Re-check after the erosion + z-ROI filter loop, in a helper next to
`patch_intersects_z_roi`. The guard mirrors the pack-level
`'No patches could be loaded'` one: it only fires when patches were
requested, so intentionally patch-free fits (`disable_patches`, or
`use_*_patches = False`) keep the legitimate zero-patch branch in
`losses.py` untouched, as the issue asked. A hard error matches the
precedent guard's behavior; happy to soften it to a warning if
preferred.

## Validation

- Three new tests in `tests/test_spiral_helpers.py` cover the guard:
  all-dropped raises with the window and loaded counts in the message,
  intentionally patch-free fits stay exempt, surviving patches pass.

      $ python -m pytest tests/test_spiral_helpers.py -q
      .......
      7 passed in 3.11s

- Real-run check against the published PHerc. Paris 4 spiral dataset
  (module globals redirected to the local pack, `load_only` mode). A
  window in a coverage gap (z 50-120) now aborts right after the
  filter:

      loaded 4921 patches
      loaded 40782 unverified patches
      RuntimeError: No patches intersect the fit window z 50-120: 4921
      verified + 40782 unverified patches were loaded, but the z-ROI
      filter (and erosion) dropped them all. Adjust z_begin/z_end to
      patch coverage, or set disable_patches=True to fit without
      patches intentionally.

  A covered window (z 10600-10900) passes the guard silently and
  proceeds to point-collection loading unchanged.

(Windows 11 / CPython 3.14.3; the logic is platform-independent.)

### DIFF
--- a/volume-cartographer/scripts/spiral/fit_spiral.py
+++ b/volume-cartographer/scripts/spiral/fit_spiral.py
@@ -82,6 +82,7 @@ from sdt_losses import (
 )
 from spiral_helpers import (
     SAMPLING_COUNT_FLOORS,
+    ensure_patches_survive_z_roi,
     erode_patch_valid_region,
     load_patches,
     load_fiber_point_collection,
@@ -891,6 +892,7 @@ def main(load_only_patches_and_point_collections=False, interactive_driver=None)
 
     print(f" loaded {len(verified_patches)} patches")
     print(f" loaded {len(unverified_patches)} unverified patches")
+    loaded_verified, loaded_unverified = len(verified_patches), len(unverified_patches)
 
     for patches in (verified_patches, unverified_patches):
         for patch_id, patch in list(patches.items()):
@@ -910,6 +912,11 @@ def main(load_only_patches_and_point_collections=False, interactive_driver=None)
             # Training retains the base grid and masks, so regenerate this view
             # lazily only for a later exporter that actually requests it.
             patch.release_derived_caches()
+
+    ensure_patches_survive_z_roi(
+        verified_patches, unverified_patches,
+        use_verified_patches or use_unverified_patches,
+        z_begin, z_end, loaded_verified, loaded_unverified)
 
     # ==========================================================================
     # Point collection loading
--- a/volume-cartographer/scripts/spiral/spiral_helpers.py
+++ b/volume-cartographer/scripts/spiral/spiral_helpers.py
@@ -25,6 +25,25 @@ def patch_intersects_z_roi(patch, z_begin, z_end):
     return bool(((zs >= z_begin) & (zs < z_end)).any().item())
 
 
+def ensure_patches_survive_z_roi(verified_patches, unverified_patches,
+                                 patches_requested, z_begin, z_end,
+                                 loaded_verified, loaded_unverified):
+    """Fail loudly when the z-ROI filter drops every loaded patch.
+
+    The 'No patches could be loaded' guard runs before the z-ROI filter,
+    so a window falling in a pack coverage gap used to fit zero patches
+    to a clean exit 0 (#1237). Intentionally patch-free fits pass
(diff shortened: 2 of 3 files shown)


pmh47:
#1237 is closed as won't-fix. Runs with no patches are a valid use-case, not a defect. We want to be able to fit a spiral with minimal verified inputs!

Nicodol:
> #1237 is closed as won't-fix. Runs with no patches are a valid use-case, not a defect. We want to be able to fit a spiral with minimal verified inputs!

Makes sense, thanks for the context on where spiral is heading

PR #1271, opened 2026-07-28
spiral: skip loading patches that meta.json already places outside the z-roi

Nicodol:
Fitting a short z-window against a whole-scroll input set reads every patch in the directory before deciding it isn't needed.

**What happens today.** `load_patches_from_dir` calls `load_tifxyz` on every entry, materialising three full-resolution tif grids per patch; `patch_intersects_z_roi` then drops the ones that miss the fitted z-roi a few lines later. On PHerc. Paris 4 fitted over 300 slices that is 45,705 patch directories read to keep ~3,800. On a 64 GB consumer machine the run dies with a host-memory allocation failure during point-to-patch linking, before the optimizer starts.

**The change.** `load_tifxyz` already opens `meta.json` first, so the `bbox` that `save_tifxyz` writes is enough to exclude a patch before touching its grids. For a skipped patch this now reads exactly what it read before, and stops there.

The pre-filter is deliberately conservative: any patch whose metadata is missing, bbox-less, sentinel-valued (`[-1, -1, -1]`) or unparseable is still loaded, so the authoritative per-vertex `patch_intersects_z_roi` check keeps deciding and the resulting patch set is unchanged.

**Measured** (PHerc. Paris 4, z 10600-10900):

| input set | directories | loaded after | skipped |
|---|---|---|---|
| verified | 4,923 | 512 | 89.6% |
| unverified | 40,782 | 3,255 | 92.0% |

All 45,705 directories carry a usable bbox, so the pre-filter decides on metadata alone, at roughly 4 ms per thousand directories. The window that previously died on host memory now completes.

**Equivalence check on real data.** 8,000 patch directories x 5 z-windows (the fitted window, the whole written region 4000-17000, a 20-slice sliver straddling patch edges, and two degenerate ranges) = 39,990 comparisons between the pre-filter and the authoritative check:

- 0 cases where the pre-filter excluded a patch the authoritative check would have kept;
- 27 cases (0.07%) kept by the pre-filter and dropped afterwards, exactly as before;
- 99.93% exact agreement.

**Tests.** `tests/test_spiral_helpers.py` gains coverage for a patch entirely outside the roi, patches overlapping or straddling the boundary, and the four unusable-metadata cases (no `meta.json`, no bbox, sentinel bbox, malformed json). The pre-existing failures in `test_phase_spacing.py`, `test_spiral_service_v2.py` and `test_track_crossing_cache.py` reproduce on `main` without this change.

Found while running `fit_spiral.py` on a small window on Windows with an 8 GB GPU; happy to adjust the approach if you would rather filter elsewhere in the pipeline.

### DIFF
--- a/volume-cartographer/scripts/spiral/fit_spiral.py
+++ b/volume-cartographer/scripts/spiral/fit_spiral.py
@@ -88,6 +88,7 @@ from spiral_helpers import (
     load_fiber_point_collections,
     scale_counts_for_z_range,
     _infer_shell_outer_winding_idx,
+    patch_dir_may_intersect_z_roi,
     patch_intersects_z_roi,
     save_combined_preview,
 )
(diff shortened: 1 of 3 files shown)


Nicodol:
Closed deliberately: `meta.json` bboxes turn out to be stale on a subset of the verified patches (59 of 4,922 on z, up to 654 slices off), so a bbox pre-filter can silently drop patches that the per-vertex z-roi check keeps. Numbers and repro in #1272. If the bbox semantics get settled (recompute-on-save or erase-on-modify), an exact version of this pre-filter may be worth revisiting.

PR #1282, opened 2026-07-30
Add DecohesionTransform CUDA speedup to community projects

JGiraldo29:
Added contribution to community projects list.


pmh47:
Since this is now merged in villa, it doesn't need a community-projects entry.

Issue #1272, opened 2026-07-28
spiral-input PHercParis4: stale meta.json bboxes on 106 verified patches (59 on z, up to 654 slices off); bbox-based z filtering silently drops them


Nicodol:
## Summary

On the public spiral-input pack for PHerc. Paris 4 (`hf://buckets/scrollprize/datasets/spiral/PHercParis4`, synced 2026-07-27), some `verified_patches` carry a `meta.json` bbox that no longer bounds their grids: valid vertices extend up to 654 slices beyond the recorded z range. Anything that filters patches by that bbox silently mis-classifies them.

## Numbers

A vertex is valid per villa's `load_tifxyz` semantics (all-`-1` sentinel and `mask.tif == 0` excluded); a patch is stale on an axis when its valid-vertex extent exceeds the recorded bbox by more than 1e-3 vox on either side; no erosion applied. The numbers were re-derived independently by two implementations (one on villa's loader, one reading the TIFFs directly); they agree.

- `verified_patches`: 4,924 entries = 4,923 directories plus a loose `seed.json`; `backups/` has no meta.json; 4,922 patch directories.
- **Stale on z: 59 / 4,922** (z is the axis filters use).
- **Stale on any axis: 106** (x: 66, y: 56, z: 59).
- `unverified_patches`: **0 / 40,782** stale (full scan), so this looks specific to the verified/reoptimization pipeline.
- Aside: `auto_grown_20260522202321410_sel_20260522_202629_23` fails villa's `load_tifxyz` with a bare `AssertionError` (no complete valid quad).

## Worst offenders (z)

| patch | recorded bbox z | actual z of valid vertices | overshoot | valid vertices outside recorded z |
|---|---|---|---|---|
| same_wrap002986_lasagna | [7369.88, 9285.01] | [7369.88, 9939.36] | 654.3 | 2,445 / 9,994 (24.5%) |
| same_wrap003013_lasagna | [6591.62, 8091.23] | [6220.27, 8091.23] | 371.4 (below min) | 1,348 / 7,542 (17.9%) |
| same_wrap000882_growpatch | [7751.04, 8409.29] | [7751.04, 8743.90] | 334.6 | 170 / 510 (33.3%) |
| same_wrap001303_growpatch | [8972.18, 9673.80] | [8972.18, 9972.54] | 298.7 | 390 / 1,326 (29.4%) |
[...]

PR #1285, opened 2026-07-30
QuadSurface: recompute bbox from points when saving metadata

Nieuwlaar:
# QuadSurface: recompute bbox from points when saving metadata

Refs #1272. This is the root-cause half; the companion audit and repair tooling lives at [Nieuwlaar/tifxyz-repair](https://github.com/Nieuwlaar/tifxyz-repair).

@Nicodol offered the same save-side fix in the issue, so if you'd rather land your own version, say so and I'll close this.

Branch: `1272-bbox-recompute-on-save`, single commit `ea1cbfb1b3343860c8288dcc24ca368f95556a3e`.

## Root cause

`QuadSurface::bbox()` (volume-cartographer/core/src/QuadSurface.cpp) caches `_bbox` and only rescans the point grid when the sentinel is unset:

```cpp
Rect3D QuadSurface::bbox()
{
    ensureLoaded();
    if (_bbox.low[0] == -1) {   // sentinel: only recompute when never set
        ...
    }
    return _bbox;
}
```

Both load constructors seed `_bbox` from the on-disk `meta.json` (`if (meta.contains("bbox")) _bbox = rect_from_json(meta["bbox"]);`), which permanently unsets the sentinel. But the point grid is freely mutable after load, `rawPointsPtr()` hands out a raw mutable `cv::Mat_<cv::Vec3f>*`, and `validPoints()` / `validQuads()` yield mutable point references. Growth (GrowPatch/optimization), editing and tool code across ~70 files mutate the grid through these paths, and none of them runs an invalidation hook. When such a surface is saved, `save()` / `save_meta()` call `bbox()` while writing metadata and faithfully persist the **pre-mutation** bbox back to `meta.json`.

Real-world impact (measured with the companion audit tool, [tifxyz-repair](https://github.com/Nieuwlaar/tifxyz-repair), which reproduces the loader's validity rules): in the published `spiral-input` PHercParis4 corpus, **106 of 4,922 verified patch directories carry a bbox that no longer contains all their valid vertices** (59 of them on z), independently matching the counts @Nicodol reported in #1272. Counting boxes that drifted the other way as well gives 130, with ten patches drifting by more than 10,000 voxels (four under-covering, six over-covering). That breaks any consumer trusting `meta.json`'s bbox for spatial indexing, overlap pre-filtering (`surface_diff`/`overlap` use `intersect(a->bbox(), b->bbox())`), or region queries.

## What changed, and why this variant

Two-line-per-site fix in the metadata-write paths (`core/src/QuadSurface.cpp`): both `save()` and `save_meta()` now drop the cached bbox immediately before serializing it, so `bbox()` rescans the grid actually being written:

```cpp
// The cached _bbox can be stale here: it is seeded from meta.json at
// load time, and the point grid can be mutated directly afterwards
// (rawPointsPtr(), validPoints(), ...) without any invalidation hook.
// Drop the cache so bbox() recomputes from the points actually being
// saved (#1272).
_bbox = {{-1, -1, -1}, {-1, -1, -1}};
```

This covers every core save route: `save(path, uuid, force)`, the `save(path, force)` convenience overload and `saveOverwrite()` (both delegate to the former), and `save_meta()`.

Two variants were considered:

1. **Invalidate on mutation.** Rejected: mutation is uncontrolled. `rawPointsPtr()` escapes a raw mutable pointer (used across ~70 files, including long-lived pointers held through optimization loops), so there is no reliable choke point; any "hook the setters" fix would leave silent holes. The existing `invalidateCache()` is only called by the few geometry ops that already knew about the problem (`rotate()`, `resample()`, `flipU/V()`, affine transforms).
2. **Recompute at save (chosen).** The save path is the single choke point where the stale value becomes persistent damage. The rescan is O(rows×cols) over data already resident for writing x/y/z TIFFs, negligible against the TIFF encode. This also mirrors existing repo practice: `vc_tifxyz_trim.cpp` erases `bbox` from meta before writing, and `vc_straighten.cpp` does `meta.erase("bbox")`, for exactly this reason.

Behavioral note: this adds no new I/O. `save_meta()` already called `bbox()`, whose first step is `ensureLoaded()`, so metadata-only saves loaded the point grid before this change too; the only new cost is the O(rows×cols) in-memory rescan, negligible against the TIFF decode that load already performs.

Deliberately **not** changed: `bbox()` itself still caches (it sits on hot intersect/overlap paths), and app-level metadata caches (`OpenDataSegmentCache`) are untouched. One edge becomes more truthful rather than different: a grid whose points are all invalid now persists the sentinel bbox `[[-1,-1,-1],[-1,-1,-1]]` instead of retaining a stale meta-seeded box; `bbox()`'s own validity rules are untouched.

## Test evidence

Extended `core/test/test_quadsurface_save_roundtrip.cpp` with the missing roundtrip case:

`TEST_CASE("save recomputes bbox after post-load point mutation (#1272)")`
1. Save a dense 64×64 grid (bbox high = (63, 63, 100)); reload, the bbox cache is now seeded from `meta.json`.
2. Mutate one vertex to (500, 600, 700) through `rawPointsPtr()`, the uncontrolled path used by growth/editing code.
3. `saveOverwrite()`, reload fresh, assert `meta.json`'s bbox covers the new extent (high ≈ (500, 600, 700)) and the untouched low corner is preserved (low ≈ (0, 0, 100)).

- **Negative control** (the two resets disabled, test kept): the case fails on exactly the three high-corner assertions,
  `CHECK(bb[1][0].get_float() == doctest::Approx(500.f))` and the 600/700 equivalents, because the stale cached bbox is written back and no longer covers the mutated vertex. The low-corner assertions still pass, which is correct since the untouched vertices bound that corner either way. The other four cases in the file stay silent.
- **With the fix**: `test_quadsurface_save_roundtrip`, 5/5 cases pass.
- **Regression sweep**: `ctest -R "quadsurface|affine_transform_surface|load_quad"`, **14/14 pass** (save_roundtrip, basics, affine_transform_surface, more, io, pointto, snapshot, save_concurrency, save_paths, final, components, fixtures, load_quad_from_tifxyz, extras).

Built with `-DVC_BUILD_APPS=OFF -DVC_BUILD_FLATBOI=OFF -DVC_TESTING=ON` (core library and tests only, no Qt GUI apps) against a conda-forge toolchain: gcc 14.4, cmake 4.2.3, ninja 1.13.2, OpenCV 4.12, Ceres 2.2, Boost 1.85, Qt 6.10 for the test harness.

## Relation to the audit

The companion audit tool, [tifxyz-repair](https://github.com/Nieuwlaar/tifxyz-repair) (MIT), detects and repairs the on-disk damage this bug already caused: it found 106 of 4,922 verified patch directories affected under the issue's own definition, out of 5,211 audited including backup snapshots. This PR stops new stale bboxes from being written; `tifxyz-repair fix` remains the remediation path for existing data.


Nicodol:
Land yours, nothing competing on my side. What I offered in #1272 depended on the semantics being settled first, and you settled them, with the reasoning written out and a negative control on the test. Good numbers too: our counts match independently (106 of 4,922, 59 on z, same_wrap002986_lasagna at 654 vox).

Two notes, take them or leave them:
- The new test case only goes through save(), so the save_meta() hunk has no coverage.
- bbox() counts a point as valid whenever x != -1, but the loader also rejects z <= 0. That's the source of the 7 false positives you mentioned. So the recomputed box can still be too large. It errs in the safe direction, an oversized box never silently drops a patch, but it seems worth stating in a PR about making the bbox trustworthy.

Nieuwlaar:
Thanks. Added a save_meta() case (disabling only that reset fails it, while the save() case still passes). On the x != -1 vs z <= 0 gap: agreed. Left bbox() alone since it's on the intersect/overlap

hendrikschilling:
@codex review

hendrikschilling:
@Nieuwlaar any comment on the meta/tif desync?

Nieuwlaar:
@hendrikschilling you and codex are right, fixed in the latest commit. save_meta() now writes only meta.json, so recomputing the bbox there publishe bounds for geometry that was never persisted, and the recompute now happens only in save()

PR #1288, opened 2026-07-31
line annotation: fix SIGSEGV during generated-view teardown (#1286 follow-up)

skyward7187:
Fixes a crash introduced by #1286, hit in the field on every generated-view rebuild (e.g. re-optimization after placing a control point).

## Crash

```
finishOptimization → materializeGeneratedViews → setGeneratedLineViews
  → ~CChunkedVolumeViewer (teardown delete loop)
    → Qt event delivery → LineAnnotationDialog::eventFilter → PC=0 (SEGV_MAPERR)
```

`setGeneratedLineViews`'s teardown deletes the pane widgets **before** nulling `_fixedStripViewer`. Widget destruction synchronously delivers events (ChildRemoved/Hide/…) to objects the dialog filters, so `eventFilter` ran mid-destruction and called `graphicsView()` on the half-destroyed viewer — the QPointer only clears once `~QObject` runs, well after the derived destructor — and the virtual call went through a torn-down vtable to address 0.

## Fix (two layers)

1. **Teardown ordering**: drop `_panes` / `_stripViewers` / `_fixedStripViewer` / fast-overlay refs / cut-viewer pointers *before* deleting `_mdiArea` and the generated containers, in both teardown paths (`setGeneratedRows`, `setGeneratedLineViews`).
2. **eventFilter hardening**: check the event type *before* dereferencing the viewer. Only spontaneous input events (wheel/mouse) reach the fixed-strip swallow logic, and those cannot arrive while a widget is being destroyed; destruction-cascade events now fall through untouched.

Either layer alone prevents this crash; both together keep the filter safe against any future teardown path.

Builds clean. The crashing path runs on every control-point placement with auto-reoptimize on, so a quick smoke test (open a fiber, place a couple of control points, reopen, close) exercises the fix directly — I have not yet run that GUI flow against this build, so please give it that smoke test before/at merge.

https://claude.ai/code/session_01EW9xAtf3HtgHotu4zSExJ1

### DIFF
--- a/volume-cartographer/apps/VC3D/LineAnnotationDialog.cpp
+++ b/volume-cartographer/apps/VC3D/LineAnnotationDialog.cpp
@@ -1364,6 +1364,7 @@ bool LineAnnotationDialog::setGeneratedLineViews(
     }
 
     updateFixedStripGeometry();
+    updateStripRenderLevelFloors();
     updatePauseIndicator();
     updateOptimizationStatusIndicator();
     rebuildGeneratedOverlays();
@@ -2216,6 +2217,10 @@ void LineAnnotationDialog::updateGeneratedDynamicOverlaysFast(bool updateCurrent
     if (_closing || !_hasGeneratedViews) {
         return;
     }
+
+    // Runs after every render/zoom via the overlay-refresh connection, which
+    // keeps the strips' LOD floor tracking user zoom changes.
+    updateStripRenderLevelFloors();
 
     const auto markerColorForState =
         [](vc3d::line_annotation::GeneratedCurrentLineMarkerState state) {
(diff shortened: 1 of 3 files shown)


skyward7187:
Added 4a31e140f, addressing the "jittering panes" report from the field (same feature area, so piggybacking on this PR):

- The per-render overlay refresh (`overlaysUpdated` → 16ms → `rebuildGeneratedOverlays()`) no longer re-requests the side-strip intersection query. That handler copies every fiber polyline and hashes the whole strip grid before reaching its cache, per tick, at up to 60Hz while tiles stream — and the new full-extent top strip can keep streaming (and thus this churn) alive indefinitely. Queries still run from every explicit data-change path (`setGenerated*`, reset, view creation).
- `updateFixedStripGeometry()` now no-ops when the computed height/camera are unchanged, so dialog resizes stop feeding redundant renders + `overlaysUpdated` into the same loop.

To test the jitter specifically: open a long annotation, place several control points, and let it sit — previously the stutter would start after placements and eventually become constant.

skyward7187:
Closing — the two jitter-mitigation commits here didn't resolve the issue in testing; we're continuing the investigation locally and will resubmit (including the teardown crash fix from 683497bdf) once the actual jitter cause is confirmed fixed.

Issue #193, opened 2025-04-18
Enhancement: Methods for generating surface, fiber, or ink labels


bruniss:
Label generation methods for surface models currently voxelize objs from manually created or automatically created segmentations. These can be relatively accurate, but require the segmentation is already created, creating a bit of a catch-22 in that we need labels from hard-to-segment regions that likely do not contain segmentations.  

Label generation methods for fibers and ink are entirely manual.

**Input Format:** zarr or other on-disk npy array
**Output Format:** zarr
**Result:** Image/label pairs of surface or fiber labels that are higher quality than simple voxelization of existing obj meshes

Jinhojeong:
Following up on the label-incompleteness hypothesis from my #191 analysis (https://github.com/ScrollPrize/villa/issues/191#issuecomment-4930879524). I turned it into a measurement and a small QC tool for segmentation-derived surface labels.

Method: for each Dataset059 patch, detect candidate unlabeled sheets, meaning high-confidence model predictions more than 3 vox outside the GT that are both sheet-shaped (planar PCA of the component) and CT-bright. The patch's label incompleteness is candidate voxels / (GT + candidate voxels).

Findings on 200 random patches:

- Mean incompleteness 6.4%; the worst decile of patches exceeds 14%, so in those patches roughly 1 in 7 sheet voxels is real papyrus the label doesn't contain.
- Cross-model check, which is the number I'd trust: the score computed from model A's predictions (surface_recto_059_redo) predicts the independent model B's (surface_recto nnUNetv2) per-patch dice at Spearman -0.45, so the incompleteness it measures is model-independent and materially depresses evaluation scores. The same-model correlation is -0.91 but partially circular, so I'm not leading with it.
- Visually the worst cases are unambiguous: long coherent sheets clearly present in the CT and absent from the GT.

![worst-incompleteness patches, CT / GT / overlay (green=GT, red=candidate)](https://raw.githubusercontent.com/Jinhojeong/vesuvius-surface-geometry-diagnostic/main/results/label_qc/worst_incomplete.png)

[...]

Issue #192, opened 2025-04-18
Enhancement: Highly accurate 3d ink labels


bruniss:
Current ink labels are generated by a human annotator drawing the letter they think the ink model most closely resembles. This offers some benefits in that we may catch additional ink unseen by humans, but also increases the risk of training the model to focus on the underlying surface rather than the ink itself. We are looking for as-accurate-as-possible ink labels that are representative of _only the detectable ink patterns_ on the surface, and in true 3d rather than a single image projected across multiple layers. 

File Format: Zarr or .tif slices
Result: Image/label pairs of high quality 3d ink labels that are “ready-to-run” , no additional cropping or preprocessing required.

giorgioangel:
@bruniss @younader can we merge your script that implement this? is it already merged?

khj1222:
I ran the experiment this issue implies — train the same model on differently
shaped 3D ink labels and score them against each other — and the result came out
against the premise, so I want to put the numbers somewhere they can be argued
with.

**Short version:** on `w00_20231016151002`, a per-pixel *measured* ink band loses
to a *constant* band by 0.038 F1, on every one of three folds. A one-voxel plane
and an eight-voxel constant band tie, so depth-resolved targets themselves cost
nothing — moving the band around per pixel is what costs.

## What was in the way first

In `flat` mode the label never reaches the loss with its depth intact:

```python
targets = (torch.amax(batch['inklabels'], dim=2) > 0).to(dtype=batch['inklabels'].dtype)
supervision_mask = torch.amax(batch['supervision_mask'], dim=2)
```

Every label voxel is max-pooled into one plane, so a depth-resolved label and the
published single-plane label produce byte-identical targets. No label-depth
experiment is possible in flat mode at all. (`full_3d` does keep depth, but it
wants the native scroll volume, and it builds its own band by projecting the flat
annotation out to a constant `_DEFAULT_FULL_3D_PROJECTION_HALF_THICKNESS` — so it
answers a different question.)

I gated a change behind `flat_depth_targets: true`: z projection off for model and
targets, volume-vs-volume loss with `supervision_mask` as the ignore mask, previews
reusing the `full_3d` central-slice reduction, and `--z-window` on inference so a
volume prediction reduces back to the usual 2D TIFF. That last part matters more
than it sounds — see the trap at the bottom.

## Setup

Three label versions over the same segment, packaged as `_inklabels_vN.zarr` so
`discover_labels` picks them up and the patch cache key keeps the arms separate:

[...]

PR #1295, opened 2026-08-01
Add Slice: calibrated 3D ink labels + browser viewer

williamshermer-pixel:
Ready-to-run image/label pairs for ink-detection training, addressing #192 (depth resolved per pixel rather than projected across layers) and #193 (the generation method). Each pair carries a measured quality certificate; a QC gate verifies the depth claim rather than asserting it.

Repo: https://github.com/williamshermer-pixel/slice
Viewer: https://slice-site-alpha.vercel.app

### DIFF
--- a/scrollprize.org/docs/20_community_projects.md
+++ b/scrollprize.org/docs/20_community_projects.md
@@ -257,6 +257,7 @@ For state-of-the-art updates join our [Discord server](https://discord.com/invit
 
 - [3D (volumetric) Ink detection model](https://github.com/ryanchesler/3d-ink-detection) by Ryan Chesler. Ink detection model that works on full scroll data in 3D, without segmentation nor flattening.
 - [Volumetric Ink Detection for Scroll 1, 2, 3, 4](https://dl.ash2txt.org/community-uploads/bruniss/3d%20Ink%20/) by Sean Johnson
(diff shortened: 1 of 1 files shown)


bruniss:
nicely designed website , however there is unfortunately a bit of an issue with some claims, like this one: 

"On the blind scrolls the ink is not faint — it was never recorded, so no method recovers it from this data. PHerc0172 therefore makes an unusually strong negative control: anything that correlates there is not measuring ink." 

we have visible ink in pherc 172, and we've found the title of it.

williamshermer-pixel:
You're right, and thank you for taking the time to check it. That claim was wrong and I've corrected it on the site.

I'd conflated two separate things. The ~3-voxel figure is a bar this pipeline sets for its *own* cross-scan claims; I wrote it as though it were a statement about the papyrus. PHerc0172 sits at 1.9 voxels by that measure, and its title has been read — and having gone back to the bucket, there are 53 segments with 106 published ink detections on that scroll. Calling it a "negative control" was indefensible.

The record page now leads with the retraction and credits the correction to you, the "blind" labels are gone, and the same wrong framing has been removed from two other places in the code where I'd repeated it.

Sorry for the noise. I'd rather be corrected than leave that standing.

erdpx:
thanks for sharing this! we’re closing this for now: the model first selects a broad depth window, then uses ct brightness to place the label on the papyrus surface. this may produce useful experimental pseudo-labels, but it doesn’t show that the ink is actually at the selected depth, and the current quality control only shows that the assigned depths vary. the cross-scan comparison is useful for comparing two 2d prediction maps, but it cannot validate the 3d depth. also, only 2 of the reported 28 pairs are currently available. if the full set can be published and the depths validated against an independent 3d reference, we’d be happy to reopen this and take another look

PR #1301, opened 2026-08-02
vesuvius-py: fix swapped conditional entropies in voi metric docstring

Jinhojeong:
The class docstring of `VOIMetric` in `vesuvius/src/vesuvius/models/evaluation/voi.py` documents the two VOI terms backwards. It currently says `voi_split` is H(GT | PR) and `voi_merge` is H(PR | GT). skimage's `variation_of_information(image0, image1)` returns the conditional entropies in the order (H(image1 | image0), H(image0 | image1)), and the code passes the GT component labels as the first argument and the prediction labels as the second. So the first value, assigned to `voi_split`, is H(PR | GT), the over-segmentation term (false splits), and the second, `voi_merge`, is H(GT | PR), the under-segmentation term (false merges). The assignment in the code is correct, only the docstring has the two entropies swapped, so this changes no behavior.

It matters for anyone reading the reported terms or thresholding on one of them. With the current docstring you would attribute a high `voi_split` to under-segmentation and a high `voi_merge` to over-segmentation, the opposite of what the numbers measure.

Since `models/evaluation/` had no tests, this also adds `vesuvius/tests/models/evaluation/test_voi.py`, which pins the direction of both terms to hand-computed exact values using one-voxel-wide bars in a 3x3x9 volume:

- a prediction that breaks one GT component in two gives voi_split = log2(9) - 16/9 and voi_merge exactly 0
- a prediction that fuses two GT components gives voi_merge = log2(9) - 16/9 and voi_split exactly 0
- identical labellings give both terms exactly 0 and voi_score exactly 1
- `VOIMetric.compute` returns the same dict as `compute_voi`

I checked the return order against the scikit-image 0.26.0 source, the version pinned in vesuvius/uv.lock (the extras allow >=0.24.0, where the order is the same). The four tests pass unchanged before and after the docstring edit, as expected for a documentation-only fix.


bruniss:
good catch. rip the test out and i'll merge. the test is not necessary.

PR #1305, opened 2026-08-03
VC3D agent bridge: recenter slice planes on viewer.center_on_point via centerFocusAt

altommo:
Rebased onto current `main` and built/tested against it.

## Problem

`viewer.center_on_point` on the VC3D agent bridge only pans the single targeted viewer. The other axis-aligned slice planes (xy / seg xz / seg yz) keep their old perpendicular depth, so driving the viewer to an arbitrary 3D point does not bring it into view.

## Fix

Route the recenter through the canonical `CWindow::centerFocusAt` path (which ends by reapplying the controller orientation via `applyOrientation`) instead of poking the focus POI directly. Orientation matches VC3D's own coordinate navigation (the location box / ctrl-click): with no active segment (or axis-aligned slice mode on) the planes are canonical axis-aligned; with an active segmentation surface and axis-aligned mode off the seg xz/yz planes take the segment tangent frame at the nearest surface point.

Two collapse modes are handled by clearing the focus orientation source (`surfaceId` + `surfacePtr`) and zeroing the focus normal before centering:
- `onPOIChanged` collapses every slice plane onto the focus normal when `|poi->n| > 0.5`; the default focus normal is `(0,0,1)` and a prior surface click leaves a surface normal.
- if the focus keeps a stale `surfaceId` and the point projects outside that `QuadSurface`, `applyOrientation`'s `segmentSliceFrameAtFocus` fails and leaves the existing seg planes collapsed.

Clearing `surfaceId` makes `applyOrientation` orient to the active segment (or canonical when none); zeroing the normal stops the collapse. The response gains a `focusMoved` flag.

## Observability + tests

- Exposes `viewers[].planeNormal` in `state.get` (finite-guarded) so plane orientation is observable over RPC.
- `center_on_point_reapplies_plane_orientation` (no active segment): the three planes stay distinct (xy->Z, seg xz->Y, seg yz->X), the focus moved to the point, and the focus normal was cleared.
- `center_on_point_segment_tangent_no_collapse` (active segment, axis-aligned off; its own VC3D instance + the checked-in fixture segment `core/test/data/segments/20241113080880`): the seg planes take the segment tangent (proving the segment path runs) and never collapse, for a point on the surface and one off it. The on-surface point is read from the fixture's TIFFs at runtime (small pure-stdlib reader) so it does not depend on a hand-picked cell.

## Verification

Built offscreen against current `main` and ran the full smoke suite; `ok: true`. The checks were confirmed to fail on the wrong implementations:

| Case | Wrong impl | Result |
|---|---|---|
| No segment | direct `focus->p` poke / `applyOrientation` removed / source uncleared | FAIL (collapse or uncleared) |
| Active segment, axis-off | location-box path (`{0,0,1}` normal, no clear) | FAIL (off-surface collapse) |
| Both | the fix | PASS |

## Notes

- Three files: +33 handler, +12 for the `state.get` readback, and the two regression checks.


altommo:
Thanks, both points were correct — addressed in the force-pushed update:

- **Reapply slice orientation after moving the focus.** Agreed: mutating `focus->p` skipped `centerFocusAt`'s closing `applyOrientation`, so a focus POI with a nonzero normal from a prior surface click could collapse the planes. The handler now routes through `CWindow::centerFocusAt(point, {0,0,0}, "")`, the same canonical path the UI uses, so the controller orientation is reapplied and the zero normal preserves the current orientation while recentering.

- **Regression check.** Added `center_on_point_drives_focus_poi` to `smoke_offscreen.py`: it centers on a point and asserts via `state.get` that the focus POI moved there and `surfaceId == "segmentation"` (the marker of the `centerFocusAt` path). Verified against a local offscreen build — it passes with the fix and fails on the prior direct-poke version.

altommo:
@codex review

altommo:
Good catch — you were right that the previous check only validated POI fields and would have passed if `applyOrientation` were removed. Reworked it to assert the actual slice-plane orientations:

- Added `viewers[].planeNormal` to `state.get` (current world normal of each axis-aligned plane).
- The regression check now centers on a point and asserts the three planes stay distinct (xy→Z, seg xz→Y, seg yz→X). No extra precondition is needed: the focus POI's default normal is `(0,0,1)`, so `onPOIChanged` already collapses all planes onto Z (`setNormal(poi->n)` when `|poi->n| > 0.5`), and `applyOrientation` is what restores the distinct normals.

Verified against a local offscreen build in three configs: passes with the fix; fails on the old `focus->p` poke; and fails when `applyOrientation` is removed from `centerFocusAt` while everything else stays — in both failing cases all three planes collapse onto Z. So the test now guards the orientation-reapply step specifically.

altommo:
@codex review

altommo:
Right again, thanks. Confirmed the stale-`surfaceId` path in `applyOrientation`: with `useCanonical` false and `segmentSliceFrameAtFocus` failing for an off-surface point, the `if (!hadSegmentPlanes)` skip leaves the existing seg planes collapsed, and passing an empty `sourceId` preserved the stale surface.

Fix: before calling `centerFocusAt`, `center_on_point` now clears the focus orientation source (`surfaceId` + `surfacePtr`) and zeros the focus normal. Clearing the surfaceId makes `applyOrientation` resolve to the canonical axis-aligned planes rather than the failing segment frame; zeroing the normal stops `onPOIChanged` collapsing them in the first place. Since `center_on_point` is a raw-coordinate navigation primitive, landing on axis-aligned planes is the intended behaviour.

Verified against a local offscreen build in four configs (see the updated PR body): passes with the fix; fails on the old `focus->p` poke, when `applyOrientation` is removed, and when the orientation source is left uncleared. The projection-failure branch itself needs a geometry-backed surface that the fixture-free harness can't build, so the `normal_cleared` assertion guards the clearing that neutralises it.

altommo:
@codex

altommo:
Rebased onto current `main` and re-verified: built VC3D offscreen against it and ran the full smoke suite (`ok: true`).

The earlier P1 (empty source falling back to the segmentation surface) is now genuinely exercised, not reasoned about. The segment-loaded path is tested with a real geometry-backed fixture segment in its own VC3D instance with axis-aligned mode off: on a surface point the seg planes take the segment tangent (confirming the path runs), and they never collapse for a point on OR off the surface. That check fails on the plain location-box path (which collapses both seg planes onto `(0,0,1)` off-surface) and passes with the clear-source + zero-normal fix. The on-surface point is read from the fixture's TIFFs at runtime via a small pure-stdlib reader, so it is not coupled to a hand-picked cell.

altommo:
@codex

PR #1308, opened 2026-08-03
community_projects: add Helena (orchestration and evidence layer)

LimeGS:
Adds one entry under **Other → ⚙️ Tools** for [Helena](https://github.com/LimeGS/helena-framework), my own project.

Helena is an orchestration and evidence layer over the volume → surface → ink → page pipeline, and it reimplements none of it: VC3D, ScrollFiesta, the flatteners and the ink models stay exactly what they are, behind stable adapters. What it adds is the bookkeeping that starts to hurt once the methods work — GPUs pooled behind one queue, every checkpoint and artifact pinned by hash, and geometry, CT support, model response and human review kept as separate records, so a campaign can be replayed and audited afterwards.

MIT licensed, CI audit workflow with a coverage gate.

Docs-only change: one file, one bullet.


erdpx:
thanks for sharing this! we’re closing this for now. the golden run is a useful audit record and shows several Helena integrations working together in your deployment. for us to verify it independently, we’d still need a run from a clean installation with public input surfaces, a checkpoint others can obtain, downloadable outputs, and an end-to-end test log without skipped pipeline stages. once this is available, we’d be happy to reopen this and take another look

PR #1309, opened 2026-08-04
spiral: add known-deformation phantom generator and truth-referenced …

beaugrzanich:
…eval harness

The spiral fit is currently evaluated by self-consistency (agreement with its own input constraints) and an indirect ink-coverage proxy; neither compares against ground truth. This adds:

- synth_phantom.py: generates scroll-like test volumes by sampling a random KNOWN deformation through the production SpiralAndTransform stack, with exact per-voxel winding ground truth, staged realism knobs (noise, haze, signal-dropout tears), and a fit-shaped checkpoint.
- winding_error.py: scores any candidate (fit checkpoint or plain winding volume, e.g. lasagna output) against phantom truth: winding MAE/RMSE, sheet-switch rate, per-winding and torn-region breakdowns, with gauge alignment and theta=0 branch-cut exclusion.
- export_phantom_dataset.py: writes a phantom as a fit_spiral-consumable dataset (umbilicus.json + verified_patches tifxyz with absolute winding annotations), with label quantity/quality dials (coverage, winding stride, position noise).
- fit_phantom_reference.py: CPU reference fitter closing the loop end-to-end (the production fitter is CUDA-only); a baseline, not a reimplementation.

First results on a 48x320x320 phantom: reference fit recovers the winding field to 0.055 MAE / 0% sheet switches with clean labels; 2 voxels of label position noise erases the value of annotation entirely (fit MAE 0.225 vs unfitted 0.223) - a first quantification of the label-quality bottleneck.

### DIFF
--- a/volume-cartographer/scripts/spiral/export_phantom_dataset.py
+++ b/volume-cartographer/scripts/spiral/export_phantom_dataset.py
@@ -1,0 +1,156 @@
+"""Export a synth_phantom phantom as a fit_spiral-consumable dataset.
+
+Writes the minimal input layout the fitter's loaders read:
+
+  <out>/umbilicus.json           {"control_points": [{"z","y","x"}, ...]},
+                                 from the phantom's own (known) curve
+  <out>/verified_patches/<id>/   tifxyz patch grids (written by tifxyz.
+                                 save_tifxyz, so the format always matches the
+                                 loader) sampled from TRUE winding surfaces,
+                                 plus winding.tif carrying the absolute
+                                 winding annotation in the fit's convention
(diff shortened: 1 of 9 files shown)


pmh47:
This would only be merged if given strong evidence (currently not present) that success on its synthetic case correlates with success on real scrolls. As-is there are too many gaps for this to believable -- only certain inputs supported, true deformation is guaranteed to live in the space of transforms representable by the spiral's diffeomorphism, constraints self-consistent, etc. Thus closing for now.

beaugrzanich:
Thank you for the response, and direction @pmh47!

PR #1312, opened 2026-08-04
tests: make the volume-cartographer suite build and pass on Windows, and fix cancel() for console tools

Aleredfer:
The test tree does not compile on Windows. `ci-windows-mingw` sets `VC_TESTING=OFF`, so the registered test targets are only ever built on Linux and macOS — where the POSIX APIs the tests reach for happen to exist. Windows ships an official installer on every release, so nothing that breaks there is currently caught.

Making the suite build surfaced a real bug, which is why this is one change and not two: with the tests finally running, `cancelFinishesOnceAndRunnerIsReusable` failed on Windows, and the failure crashed the test binary outright. Splitting the portability work from the fix would leave a PR that cannot claim a green suite.

**No `#ifdef` anywhere in this PR.** Every fix is portable by construction.

### Compile failures — these blocked the whole tree

| file | problem | fix |
|---|---|---|
| `apps/VC3D/test/test_command_line_tool_runner.cpp` | `std::raise(SIGKILL)` — POSIX-only | `std::abort()` |
| `core/test/test_line_annotation_generated_views.cpp` | `setenv`/`unsetenv` — POSIX-only | `qputenv`/`qunsetenv` + explicit `<QtGlobal>` |

Production code already handles both correctly, which is what makes this a test-tree gap rather than a codebase-wide one: `LasagnaServiceManager.cpp:877` guards its `SIGKILL` with `#ifdef Q_OS_UNIX`, and `VCAppMain.cpp:53` carries a `vcSetEnv` helper commented *"POSIX setenv with an MSVCRT/UCRT fallback (Windows has no setenv)"*.

`std::abort()` is used rather than another signal under an `#ifdef` because under MinGW every `raise(SIG*)` returns through the default handler and exits normally, which `QProcess` reports as `NormalExit` — while the test asserts `CrashExit`:

| | `raise(SIGABRT/SIGSEGV/SIGILL/SIGTERM)` | `std::abort()` |
|---|---|---|
| Windows (MinGW) | NormalExit, code 3 | **CrashExit** |
| Linux (glibc) | CrashExit | **CrashExit** |

It is correct on both, and removes the now-unused `<csignal>` include.

### Runtime failures

**Deleting a file that is still mapped or open** (5 tests) — `test_gridstore.cpp`, `test_gridstore_branches.cpp`, `test_gridstore_malformed.cpp`, `test_normal_grid_volume.cpp`, `test_quadsurface_extras.cpp`. POSIX allows unlinking an open or mapped file; Windows refuses. In each case the owner of the mapping was still in scope at the `remove()` call, so it is now confined to a nested scope or released explicitly. Not a production defect: `GridStore` holds its mapping in an RAII member and `~GridStore()` releases it correctly.

**`isExecutable()` is extension-based on Windows** (1 test) — `failedStartFinishesOnce` built its stand-in program from a `QTemporaryFile` whose random name has no extension, so the runner rejected it before starting a process, a different code path from the one the test exercises. The temporary now carries a `.exe` suffix, harmless on POSIX where the permission bit set just below is what counts.

**A genuine race** (1 test) — `stats: startup scan ignores files newer than its cutoff` wrote `post.bin` *after* constructing the cache. The scan runs on a worker thread with a cutoff captured at construction, so the write raced it: the scan could read `post.bin` before its future timestamp was applied and count it — 49 bytes instead of 32. This can fail on any loaded machine; Windows merely exposes it by being slower at directory iteration. The file is now stamped before the cache exists, which removes the race without changing what the test asserts, and a previously ignored `error_code` is checked so a broken premise fails loudly instead of as a confusing byte-count mismatch.

### The cancellation bug the tests found

`cancel()` only called `QProcess::terminate()`, which posts `WM_CLOSE`. Console applications have no message loop, so the `vc_*` tools ignore it: the user cancels, the interface reports success, and the process keeps running for hours. This also hung cancellation from the agent bridge.

Fixed with a deferred `kill()` fallback after a grace period, so `terminate()` still gets its chance to shut the process down cleanly first. The `QProcess` object is reused across runs, so the deferred kill is guarded against landing on a later run that started in the meantime.

Adds a regression test for that guard, verified to fail without the change: it cancels a long-running helper, starts another immediately, and asserts the replacement is still alive after the grace period has elapsed.

## How to build and run

```bash
# Linux
cmake --preset ci-release-gcc -DVC_TESTING=ON
ninja -C build/ci-release-gcc
cd build/ci-release-gcc && QT_QPA_PLATFORM=offscreen ctest --timeout 180 -j4

# Windows (MSYS2 UCRT64)
cmake --preset ci-windows-mingw -DVC_TESTING=ON
ninja -C build/ci-windows-mingw
cd build/ci-windows-mingw && QT_QPA_PLATFORM=offscreen ctest --timeout 180 -j4
```

## How this was verified

**Windows** (MSYS2 UCRT64, gcc 16.1.0, Qt 6.11.1, CMake 4.4.0): the test tree went from *not compiling at all* to **112/112 passing**.

Every assertion in the touched tests is preserved one-for-one; only scoping and indentation changed. No test was weakened to make it pass.

## Risks and limitations

- I have only run the suite on Windows against this branch, so please let CI confirm Linux and macOS. The changes are confined to test code plus the one `cancel()` fix and add no platform conditionals, so POSIX behaviour should be unchanged — but that is reasoning, not a measurement, and I would rather say so than imply otherwise.
- On Linux `terminate()` already stops these tools, so the deferred `kill()` normally finds the process gone; the guard covers that case.
- `std::abort()` may produce a core dump where `ulimit` allows it. That is noise in a test log, not a failure.

### DIFF
--- a/volume-cartographer/apps/VC3D/CommandLineToolRunner.cpp
+++ b/volume-cartographer/apps/VC3D/CommandLineToolRunner.cpp
@@ -9,12 +9,17 @@
 #include <QMessageBox>
 #include <QClipboard>
 #include <QApplication>
+#include <QPointer>
+#include <QTimer>
 
 #include <cmath>
 #include <utility>
 
 
 namespace {
+
+// Grace period between terminate() and the kill() fallback in cancel().
+constexpr int kCancelGracePeriodMs = 1000;
 
 QString quoteArg(const QString& arg)
 {
@@ -457,9 +462,28 @@ bool CommandLineToolRunner::executeCustomCommand(const QString& command,
 
 void CommandLineToolRunner::cancel()
 {
-    if (_process && _process->state() != QProcess::NotRunning) {
-        _process->terminate();
+    if (!_process || _process->state() == QProcess::NotRunning) {
+        return;
     }
+
+    _process->terminate();
+
+    // On Windows terminate() posts WM_CLOSE, which the console-mode vc_* tools have
+    // no message loop to receive: the process keeps running and cancelling silently
+    // does nothing at all. Fall back to kill(), the same way LasagnaServiceManager
+    // and the other service managers already do. Deferred rather than a blocking
+    // waitForFinished() because cancel() runs on the GUI thread.
+    // The QProcess object is created once and reused across runs, so the pid is what
+    // identifies *this* run: without it a cancel followed by a quick relaunch would
+    // let the timer kill the new process instead.
+    QPointer<QProcess> process(_process);
+    const qint64 cancelledPid = _process->processId();
+    QTimer::singleShot(kCancelGracePeriodMs, this, [process, cancelledPid]() {
+        if (process && process->state() != QProcess::NotRunning
+            && process->processId() == cancelledPid) {
+            process->kill();
+        }
+    });
 }
 
 bool CommandLineToolRunner::isRunning() const
--- a/volume-cartographer/apps/VC3D/test/test_command_line_tool_runner.cpp
+++ b/volume-cartographer/apps/VC3D/test/test_command_line_tool_runner.cpp
@@ -1,6 +1,5 @@
-#include <csignal>
-
 #include <QApplication>
+#include <QDir>
 #include <QFile>
 #include <QScopeGuard>
 #include <QTemporaryFile>
@@ -76,8 +75,12 @@ int runChild(const QStringList& arguments)
         return 0;
     }
     if (mode == "crash") {
-        std::raise(SIGKILL);
-        return 1;
+        // std::abort() rather than raise(SIGKILL): SIGKILL is POSIX-only, so this
+        // file did not compile on Windows at all. It is also the only portable way
+        // to get QProcess::CrashExit on both platforms — under MinGW every
+        // raise(SIG*) returns through the default handler and exits normally, which
+        // QProcess reports as NormalExit.
+        std::abort();
     }
     if (mode == "write-omp" && marker + 2 < arguments.size()) {
         QFile output(arguments.at(marker + 2));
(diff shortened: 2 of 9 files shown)


nerln:
Not a review, just flagging that we have been pushing on the same thing from two sides. You are getting the suite building and passing on Windows; I have been doing the equivalent on macOS/arm64 and it turned up a few things that are not platform-specific and may overlap with what you are seeing.

Two are plain cross-platform bugs that Linux happens to hide:

- #1334: `VolumePkg` compares filesystem paths as strings in the attach path, so when one side carries a symlink component and the other carries its target, re-attaching an already-attached volume returns `VolumeIdConflict`. macOS trips it via `/var` -> `/private/var`, but any symlinked data root does. Worth checking whether junctions or `\\?\` prefixes give you the same on Windows.
- #1335: `FiberSaveBatchTracker` uses a moved-from `std::function` as its run-once guard. libstdc++ empties it, libc++ does not for small captures. MSVC's STL is a third implementation, so that one is worth a look on your side.

One is build-system and may be closer to your work: #1339. The `check-core` / `check-specialized` targets from #1297 silently run zero tests and exit 0 under Ninja, because `add_custom_target` without `VERBATIM` lets the generator eat the `$` in the label regex. If you are invoking those targets anywhere in the Windows path, they are not testing what they appear to test.

Also #1336 (one call site missed OpenCV 5's move of `cv::boundingRect` into the geometry module) and #1337 (macOS framework search order), both platform-specific and unlikely to matter to you.

If it is useful, I am happy to run anything against macOS/arm64 while you are in there.

pmh47:
`CommandLineToolRunner::~CommandLineToolRunner` should probably have a similar fix for `terminate`?

Aleredfer:
Good catch — yes. The destructor has the same problem, and waitForFinished(3000) there will always burn the full 3 s on Windows for exactly the reason in the cancel() comment, on the GUI thread.

It needs the blocking variant rather than the deferred timer, since a singleShot would never fire from a destructor:
_process->terminate();
if (!_process->waitForFinished(kCancelGracePeriodMs)) {
    _process->kill();
    _process->waitForFinished();
}

SegmentRenderThread.cpp:18 has the identical pattern. Five call sites already do terminate → wait → kill (LasagnaServiceManager, NeuralTraceServiceManager, SeedingWidget, SpiralSshTunnel, SpiralServiceManager); these two destructors are the outliers. Happy to fold both into this PR.

pmh47:
Please rebase and implement the additional change discussed above if you want it merged.

Aleredfer:
The staleness bot closed this on 28 August, which was on me: @pmh47 asked for a
rebase and the destructor change, the work was done and I did not push it in time.

GitHub would not let me reopen this PR after the branch was force-pushed, so it
continues as #1687 from the same branch. It has the rebase onto main (5479453),
both destructors fixed, and the full suite building and passing on Windows
(148/158, 0 assertion failures; the 10 remaining are pre-existing hangs, filed as
#1686).

Sorry for the churn. The review history is here, so I have linked back to it.

Issue #497, opened 2025-10-09
vc_render_tifxyz: support proper scales in ome-zarr metadata


jrudolph:
We could add proper scaling metadata to `.zattrs` so that tools like neuroglancer can show scaling bars correctly.

If you look at the spec example:

```
   "multiscales": [
        {
          "name": "example",
          "axes": [
            { "name": "t", "type": "time", "unit": "millisecond" },
            { "name": "c", "type": "channel" },
            { "name": "z", "type": "space", "unit": "micrometer" },
            { "name": "y", "type": "space", "unit": "micrometer" },
            { "name": "x", "type": "space", "unit": "micrometer" }
          ],
          "datasets": [
            {
              "path": "0",
              "coordinateTransformations": [
                {
                  // the voxel size for the first scale level (0.5 micrometer)
                  "type": "scale",
                  "scale": [1.0, 1.0, 0.5, 0.5, 0.5]
                }
              ]
            },
            {
              "path": "1",
              "coordinateTransformations": [
                {
                  // the voxel size for the second scale level (downscaled by a factor of 2 -> 1 micrometer)
                  "type": "scale",
                  "scale": [1.0, 1.0, 1.0, 1.0, 1.0]
                }
              ]
            },
```

You can set the unit to a metric unit and then set the voxel scale to a fraction.

We would have to give the volume scale as an argument (or infer from volume information).

PR #1313, opened 2026-08-04
vc_render_tifxyz: default --voxel-unit to micrometer, not nanometer

Aleredfer:
Every OME-Zarr written by `vc_render_tifxyz` declares its physical scale **1000× too
small**, on the default path, with no warning.

`--voxel-unit` defaults to `nanometer`. Everything in this project is in micrometers: the
`voxelsize` field of volpkg `meta.json`, the documentation (7.91 µm), and the volume names
themselves (`54keV_7.91um_Scroll1A.zarr`).

### Why this matters

This is the metadata OME-Zarr viewers read to draw scale bars — which is exactly what
#497 (*"support proper scales in ome-zarr metadata"*) set out to enable:

> *"We could add proper scaling metadata to `.zattrs` so that tools like neuroglancer can
> show scaling bars correctly."*

The spec example quoted in that issue uses `"unit": "micrometer"`. The feature works; its
default value undoes it.

The Grand Prize rules also require *"scale bars showing the size of 1cm on each submission
image"* (`34_prizes.md`).

### The correct value is not a matter of opinion

`apps/VC3D/CommandLineToolRunner.cpp:660-663` — VC3D passes `micrometer` explicitly when
it invokes the renderer:

```cpp
if (_useRenderVoxelSize) {
    args << "--voxel-size" << QString::number(_renderVoxelSizeUm, 'g', 17)
         << "--voxel-unit" << "micrometer";
}
```

Note the condition. The GUI only overrides the default when the user has opted into a
render voxel size; with that toggle off it passes neither flag, so **the GUI is affected
too** and falls back to `nanometer` like everything else. Anything invoking the CLI
directly — batch scripts, pipelines — always is.

`core/src/Zarr.cpp:373` has no default of its own (`if (!voxelUnit.empty()) ax["unit"] =
voxelUnit;`), so the single source of the wrong value is the option default.

### Why not read the unit from the input volume

Because it isn't there. Both the CT and the surface-prediction volumes for Scroll 1
declare axes with **no `unit`** and `scale: [1.0, 1.0, 1.0]` — unitless pyramid factors,
precisely the gap #497 described upstream. There is nothing to propagate; the default is
the only place to fix it.

### What changed

- `--voxel-unit` now defaults to `micrometer`.
- Help text says why, so the next reader does not have to rediscover it.
- **New regression test** `core/test/test_render_tifxyz_defaults.cpp` pins the
  default by reading it back out of `--help`.

### About the test

The fix is a one-word default, so its failure mode is a silent revert: the code
keeps compiling and the output keeps looking plausible. A test that pins the
value is the only thing that stops it happening twice.

It is deliberately hermetic — no volume, no surface, no fixtures — so it costs
milliseconds and runs in the **default** `ctest` pass, rather than being gated
behind `VC_RUN_E2E` like `test_merge_e2e_small`. A guard nobody runs guards
nothing.

Verified to fail without the fix: reverting the default to `nanometer` makes it
red, restoring it makes it green.

This is, as far as I can tell, the second test in the suite that exercises one
of the 43 `apps/src` binaries (`test_merge_e2e_small` is the first, and it is
opt-in). If a different location or convention is preferred for tests that
spawn CLI tools, say so and I will move it — the probe follows the one that
already exists.

### How to build and run

```bash
cmake --preset ci-release-gcc && ninja -C build/ci-release-gcc vc_render_tifxyz
vc_render_tifxyz -v <volume.zarr> --scale 1 -g 0 -s <tifxyz> \
    --zarr-output out.zarr --tif-output out_tif
```

### How this was verified

| | before | after |
|---|---|---|
| `.zattrs` axis unit (default path) | `nanometer` ✗ | **`micrometer`** ✓ |
| `.zattrs` scale | `[7.91, 7.91, 7.91]` (i.e. 7.91 nm) | `[7.91, 7.91, 7.91]` (µm) ✓ |
| log line | `7.91 nanometer` ✗ | `7.91 micrometer` ✓ |
| TIFF resolution, size from metadata | 3211.13 px/in ✓ | 3211.13 px/in ✓ |
| TIFF resolution, `--voxel-size 7.91` | **3 211 130 px/in** ✗ | **3211.13 px/in** ✓ |

That last row is a second trap the same default created: passing the *documented* voxel
size on the command line produced a TIFF resolution 1000× off, because the unit conversion
is applied only when the value comes from the CLI (`vc_render_tifxyz.cpp:1429`,
`if (voxelSizeFromCli)`). Both paths now agree.

Rendered from a real traced surface against a Scroll 1 ROI; `tiffinfo` and the generated
`.zattrs` inspected directly.

### Risks and limitations

- Anyone currently passing `--voxel-size` in nanometres and relying on the old default
  would need to add `--voxel-unit nanometer`. No data in this project is in nanometres, and
  the GUI already passes `micrometer`, so the expected blast radius is zero — but it is a
  behaviour change and worth stating.
- Only `vc_render_tifxyz` exposes `--voxel-unit`; no other tool is affected.
- Verified on Windows (MSYS2 UCRT64). The code path is platform-independent.
- The test spawns the binary through `std::system`. On Windows that goes via
  `cmd /c`, which needs the whole command line wrapped in one extra quote pair
  when it both starts with a quote and contains others; the test does that under
  `#ifdef _WIN32`. Worth knowing if the pattern gets reused.
- Full suite green with the change: 113/113 on Windows (112 existing + the new
  one), gcc 16.1.0, Qt 6.11.1.

### DIFF
--- a/volume-cartographer/apps/src/vc_render_tifxyz.cpp
+++ b/volume-cartographer/apps/src/vc_render_tifxyz.cpp
@@ -1128,7 +1128,10 @@ int main(int argc, char *argv[])
         ("resume", po::bool_switch()->default_value(false), "Skip chunks that already exist on disk")
         ("pre", po::bool_switch()->default_value(false), "Create zarr + all level datasets")
         ("voxel-size", po::value<double>(), "Physical voxel size for OME-Zarr scale metadata (reads from volume metadata if omitted)")
-        ("voxel-unit", po::value<std::string>()->default_value("nanometer"), "Physical unit for OME-Zarr axes (e.g. nanometer, micrometer)");
+        ("voxel-unit", po::value<std::string>()->default_value("micrometer"),
+            "Physical unit for OME-Zarr axes and TIFF resolution (e.g. micrometer, nanometer). "
+            "Defaults to micrometer: volpkg meta.json voxelsize, the docs and the volume "
+            "names (..._7.91um_...) are all in micrometers.");
     // clang-format on
 
     po::options_description all("Usage");
--- a/volume-cartographer/core/test/CMakeLists.txt
+++ b/volume-cartographer/core/test/CMakeLists.txt
@@ -582,3 +582,23 @@ vc_add_test(NAME test_quadsurface_extras)
 vc_add_test(NAME test_surface_patch_index_cache)
 
 vc_add_test(NAME test_volume_pkg_more)
+
+# Pins the option defaults of vc_render_tifxyz by reading them back out of
+# --help. Unlike test_merge_e2e_small this is not gated behind VC_RUN_E2E: it
+# spawns the binary but touches no volume, no surface and no fixtures, so it
+# costs milliseconds and belongs in the default pass. The regression it guards
+# — a one-word default reverting silently, with the output still looking
+# plausible — is exactly what CI has to catch. Skipped when the apps are not
+# built, e.g. coverage builds where VC_BUILD_APPS defaults to OFF.
+if(TARGET vc_render_tifxyz)
+    add_executable(test_render_tifxyz_defaults test_render_tifxyz_defaults.cpp)
+    target_link_libraries(test_render_tifxyz_defaults PRIVATE doctest::doctest)
+    vc_register_test_target(test_render_tifxyz_defaults)
+    add_dependencies(test_render_tifxyz_defaults vc_render_tifxyz)
+    add_test(NAME test_render_tifxyz_defaults COMMAND test_render_tifxyz_defaults)
+    vc_register_ctest(test_render_tifxyz_defaults test_render_tifxyz_defaults)
+    # The probe in the test prefers this variable, so ctest works regardless of
+    # the working directory or the build layout.
+    set_tests_properties(test_render_tifxyz_defaults PROPERTIES
+        ENVIRONMENT "VC_RENDER_TIFXYZ_BIN=$<TARGET_FILE:vc_render_tifxyz>")
+endif()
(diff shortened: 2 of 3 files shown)


hendrikschilling:
If we fix the units we should do it comprehensively - this PR still seems to have some gaps:

```
  1. P1: Unknown physical scale is still declared as known.
     When neither --voxel-size nor volume metadata exists, the renderer substitutes 1.0 at volume-cartographer/apps/src/vc_render_tifxyz.cpp:1429. This PR then labels that arbitrary value 
     micrometer and writes it to every OME-Zarr axis through volume-cartographer/core/src/Zarr.cpp:371. It previously guessed 1 nanometer; changing the guess does not make it reliable.    

  2. P2: --voxel-unit can create inconsistent OME-Zarr and TIFF metadata.
     A voxelsize read from VC metadata is implicitly micrometers, but the OME-Zarr label comes from --voxel-unit. TIFF conversion only applies that unit when --voxel-size came from the CLI
     at volume-cartographer/apps/src/vc_render_tifxyz.cpp:1431. Therefore metadata voxelsize=7.91 --voxel-unit nanometer produces:
      - OME-Zarr: 7.91 nanometer
      - TIFF: treated as 7.91 micrometer

     This inconsistency predates the PR, but the new help text claims the option consistently governs both outputs.

  3. Existing scope gap: physical metadata from input OME-Zarr is ignored.
     volume-cartographer/apps/src/vc_render_tifxyz.cpp:1030 only examines meta.json and metadata.json. A bare OME-Zarr with correct axis units and coordinate transformations falls back to 
     guessed 1.0 micrometer.
```

bruniss:
closing due to staleness/not prio

Issue #1325, opened 2026-08-04
vesuvius.predict has no chunk cache: inference re-downloads the same chunks, ~17x more traffic than needed


aistae:
## Summary

The inference read path reads every patch with an independent request to the
store. There is no chunk cache. Patches overlap (50% by default) and do not
start on chunk boundaries, so the same chunks are fetched again and again.

Measured on `main` (Python 3.14, installed from source), byte counter
installed on `zarr.storage.FsspecStore.get`, reading through the package's own
`VCDataset`.

## Reproduction

Reading **the same patch twice**:

```
patch #0, first read        16.78 MB,  8 requests
patch #0, read again        16.78 MB,  8 requests     <- nothing cached
adjacent patch (50% overlap) 16.78 MB,  8 requests
distant patch               16.78 MB,  8 requests
back to patch #0            16.78 MB,  8 requests
```

Volume: `PHercParis4/volumes/20260411134726-2.400um-0.2m-78keV-masked.zarr/2`,
`patch_size=(128,128,128)`, `step_size=0.5`, `skip_empty_patches=False`,
chunks 128³, `"compressor": null` so each chunk is exactly 2 097 152 bytes.

## Scale

| configuration | patches | chunk fetches | downloaded | unique chunks | minimum | waste |
|---|---|---|---|---|---|---|
| patch 128³, step 64 | 216 | 1728 | **3.62 GB** | 100 | **210 MB** | **17×** |
| patch 192³, step 96 (defaults) | 512 | 5832 | **12.2 GB** | 343 | **0.72 GB** | **17×** |

In the first case the useful payload of all patches is 0.45 GB, and 3.62 GB is
transferred to obtain it.

## Why this looks like a regression

Caching is used elsewhere in this repository:

- `neural_tracing/datasets/common.py` — fsspec `filecache` via
  `volume_cache_dir`, chunks cached across epochs
- `neural_tracing/infer.py` — `cachetools.LRUCache`

The inference path (`data/vc_dataset.py`, `data/utils.py::open_zarr`,
`models/run/inference.py`) has none — no `LRUStoreCache`, `LRUCache`,
`cache_type`, `filecache` or `blockcache`.

And `neural_tracing/heatmap_single_point/dataset.py` records how it was lost:

```python
if config.get("use_volume_store_cache", False):
    warnings.warn(
        "use_volume_store_cache is no longer supported with zarr 3 "
        "(LRUStoreCache was removed). Ignoring the option.",
[...]

PR #1326, opened 2026-08-04
docs: add vesuvius-io-audit to community projects

aistae:
Adds [vesuvius-io-audit](https://github.com/aistae/vesuvius-io-audit) under Data access → Tools.

The project measures how many bytes the data read path actually transfers versus how many it needs:

- **No chunk cache in the inference path.** Re-reading the identical patch transfers the full amount again (16.78 MB both times). On default settings this is ~17× more traffic than necessary — 12.2 GB where 0.72 GB suffices. Traced to `LRUStoreCache` being removed in zarr 3, which turned `use_volume_store_cache` into a no-op. Filed as #1325.
- **Unaligned access costs.** A single 512×512 slice downloads 33.55 MB for a 0.26 MB payload (128×); a 256³ cube offset by half a chunk touches 27 chunks instead of 8 (+238%).
- **Traversal order.** Patch positions are enumerated row-major, which does not match the chunk layout; Z-order would cut a further 57% once a cache exists. The simulator behind that figure is validated against measured transfers to within 3%.

All scripts are CPU-only and read megabytes, not terabytes. Raw numbers from every run are included in the repo.


erdpx:
thanks for the pr and for putting these measurements together! however, this seems better kept as an investigation in #1325 than listed as a standalone community tool, especially since the related pipeline changes are already being handled through #1177 and #1331. this is why we’re closing it for now

PR #1327, opened 2026-08-04
fix(data): optional on-disk chunk cache for remote reads (refs #1325)

aistae:
Draft — one possible shape for the fix in #1325. I said there I'd wait for your call on the design rather than push a PR; opening this as a draft because a working patch with numbers is easier to discuss than an abstract choice. Happy to rework it in whichever direction you prefer, or close it.

## What it does

Routes remote reads through an on-disk chunk cache in `open_zarr` — the single point through which `Volume` opens data, so it covers the whole read path.

Opt-in: `cache_dir=` argument or `VESUVIUS_CHUNK_CACHE_DIR`. **Default behaviour is unchanged** — silently writing gigabytes into someone's filesystem seemed like the wrong default to pick on your behalf.

## Measured effect

`VCDataset`, PHercParis4 level 2, patch 128³, step 0.5, 12 patches read twice:

| | downloaded | requests | wall-clock |
|---|---|---|---|
| current | **402.7 MB** | 192 | 233 s |
| with cache | **62.9 MB** | 30 | 106 s |

**−84% traffic, roughly half the time.**

## Why `simplecache` and not `filecache`

The open-data volumes are immutable, so revalidating each chunk against S3 only buys a round trip. Same patch read three times:

| | 1st | 2nd | 3rd |
|---|---|---|---|
| `simplecache` | 8.39 MB, 4 req | **0 MB, 0 req, 0.0 s** | 0 MB, 0 req, 0.0 s |
| `filecache` | 8.39 MB, 4 req | 0 MB, **4 req**, 4.5 s | 0 MB, 4 req, 9.5 s |

`neural_tracing` uses `filecache`, so if consistency there matters more than the revalidation cost, that's an easy change.

## Open questions for you

- Should this default to on, with a cache directory under the platform temp dir? That would help the case #1325 is really about — people without terabytes of local disk — but it changes behaviour silently.
- Is an in-memory LRU wrapping the store preferable to the on-disk cache, closer to the removed `LRUStoreCache`? Disk survives restarts, which matters for resumed inference, but costs disk.
- Should `vesuvius.predict` grow a `--cache-dir` flag to surface this?

Once a cache exists, the traversal order becomes worth fixing too (Z-order over patch positions, a further −57% in simulation validated against measured transfers) — but that's a separate change and I'd rather land this one first.

Background, scripts and raw numbers: https://github.com/aistae/vesuvius-io-audit

### DIFF
--- a/vesuvius/src/vesuvius/data/utils.py
+++ b/vesuvius/src/vesuvius/data/utils.py
@@ -35,6 +35,40 @@ def get_max_value(dtype: np.dtype) -> Union[float, int]:
         raise ValueError("Unsupported dtype")
     return max_value
 
+
+
+_CACHE_ENV = "VESUVIUS_CHUNK_CACHE_DIR"
+
+
+def _chunk_cache_url(path: str, storage_options: Dict[str, Any],
+                     cache_dir: Optional[str]) -> Tuple[str, Dict[str, Any]]:
+    """Route a remote store through an on-disk chunk cache.
+
+    Without this every read goes to the network, so overlapping patches
+    re-download the same chunks: re-reading an identical patch transfers the
+    full amount again. This replaces the caching that was lost when zarr 3
+    removed ``LRUStoreCache`` (see the ``use_volume_store_cache`` warning in
+    ``neural_tracing/heatmap_single_point/dataset.py``).
+
+    ``simplecache`` rather than ``filecache``: the open-data volumes are
+    immutable, so revalidating each chunk against S3 only costs a round trip.
+    Measured on PHercParis4 level 2, same patch read three times —
+    ``simplecache``: 8.39 MB then 0 MB, 0 requests; ``filecache``: 8.39 MB then
+    0 MB but 4 requests each time.
+
+    Returns the (possibly rewritten) path and storage options.
+    """
+    if not cache_dir:
+        return path, storage_options
+
+    protocol = path.split("://", 1)[0]
+    nested = dict(storage_options)
+    return (
+        f"simplecache::{path}",
+        {protocol: nested, "simplecache": {"cache_storage": cache_dir}},
+    )
+
+
 def open_zarr(path: str, mode: str = 'r', 
               storage_options: Optional[Dict[str, Any]] = None,
               verbose: bool = False,
@@ -46,6 +80,7 @@ def open_zarr(path: str, mode: str = 'r',
               fill_value: Any = None,
               order: str = None,
               zarr_format: Optional[int] = 2,
+              cache_dir: Optional[str] = None,
               **kwargs) -> zarr.Array:
     """
     Open a zarr array with consistent handling of local and remote URLs.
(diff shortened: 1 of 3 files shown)


aistae:
Added tests.

**Unit (12, no network)** — `tests/data/test_chunk_cache.py`:
- URL rewriting and nested storage options for both `s3://` and `https://`
- caller's `storage_options` dict is not mutated
- explicit `cache_dir=` takes precedence over `VESUVIUS_CHUNK_CACHE_DIR`
- local paths are never wrapped, and still receive no `storage_options`
- **write modes (`w`, `a`, `r+`) are never cached** — caching a store being written would serve stale chunks

**Network (1, marked `network` so it is deselected by default per `pytest.ini`)**:
- bytes read through the cache are byte-identical to bytes read without it
- the second read issues no further S3 requests
- the cache directory is actually populated

**Regression:**

| suite | result |
|---|---|
| `tests/data/` (existing 53 + new 12) | 65 passed |
| `tests/data/ -m network` | 1 passed |
| `tests/test_imports.py` | 6 passed, 2 failed |

The two failures are `test_cli_entrypoints_show_help` for `training.train` and `run.inference`. I verified they fail identically on an unpatched checkout in a clean environment — they need `nnunetv2`, which is not part of the `volume-only` dependency set. Not related to this change.

`tests/structure_tensor/` could not be collected for the same reason (`ModuleNotFoundError: nnunetv2`), also independent of this PR.

aistae:
One more thing worth noting about the scope: because this sits in `open_zarr`, it covers every entrypoint that reads volumes through it, not just inference —

| entrypoint | read path |
|---|---|
| `vesuvius.predict` | `models/run/inference.py` (measured) |
| `vesuvius.train` | `models/datasets/zarr_dataset.py` |
| `vesuvius.compute_st` | `structure_tensor/create_st.py` |
| `vesuvius.blend_logits` | `models/run/blending.py` |
| `vesuvius.finalize_outputs` | `models/run/finalize_outputs.py` |

`neural_tracing` is unaffected — it already has its own fsspec `filecache` via `volume_cache_dir`.

Training is likely to benefit more than inference, since epochs revisit the same regions repeatedly, but I only measured the inference path so I'm not putting a number on it.

aistae:
**Correction: prior art I should have found before opening this.**

@robertlangdonn's #1177 (20 July) already restores chunk caching in this exact function, and it was recognised in the July progress prizes. I searched the code for `LRUStoreCache`/`filecache` and the issue tracker, but not the open PR queue — that was a gap in my checking, and this PR duplicates work that predates it by two weeks.

Reading it now, the two approaches differ in a way that may still be useful:

| | #1177 | this PR |
|---|---|---|
| where | in-memory LRU via `zarr.storage.WrapperStore` | on-disk via fsspec `simplecache` |
| zarr 2 | needs zarr ≥ 3 (guarded with a version check) | independent of the zarr API |
| survives restart | no | yes |
| granularity | compressed bytes in RAM | chunk files on disk |

The zarr 2/3 split is what held #1177 up — per the thread, `LRUCacheStore(WrapperStore)` at module level broke collection under zarr 2.x on 31 July. An fsspec-level cache sidesteps that entirely, since it never touches zarr's store API. That may be worth having regardless of which lands, and they compose: RAM for the working set, disk across runs.

So: **#1177 has priority and I'd rather it landed than this.** If the team prefers a single implementation, close this one — I'll rebase the parts that don't overlap.

What does not overlap, and what I'd keep either way:

1. **The traversal-order change** (latest commit) — orders patch positions along a Z-order curve over chunk indices. Independent of which cache lands, and per @TAUIL-Abd-Elilah's measurement in #1325 it matters more than cache size: quadrupling 4→16 GiB buys less than reordering does.
2. **The measurements** — 17× redundancy here, 30.5× independently on PHerc1218, reproduced between us to the exact fetch count.

@robertlangdonn — apologies for the duplicate. Happy to test #1177 against the numbers in #1325 if that's useful, or to fold the disk-cache option into yours rather than carrying a separate PR.

aistae:
Closing in favour of #1177, which restores chunk caching in this same function and predates this by two weeks.

The traversal-order work that was riding along in here is now its own PR — #1331. It is independent of which cache lands and makes any of them more effective, so it no longer needs to be bundled with a cache implementation.

The disk-cache approach here differs from #1177 in one way that may still be worth something: being an fsspec-level cache, it never touches zarr's store API, so it sidesteps the zarr 2/3 `WrapperStore` split that held #1177 up. If that turns out to be useful, I'm happy to fold it into #1177 as an option rather than carry a competing PR — @robertlangdonn's call.

Measurements and scripts stay available at https://github.com/aistae/vesuvius-io-audit; the underlying report is #1325.

PR #1332, opened 2026-08-05
fix(data): discover resolution levels from OME-Zarr multiscales, not by listing the group

aistae:
`Volume` finds a group's resolution levels by listing its members. HTTP stores have no LIST operation, so `keys()` and `len()` both come back empty — and every multiscale volume served over `https://` fails at construction:

```python
>>> Volume(type='zarr', path='https://vesuvius-challenge-open-data.s3.amazonaws.com/PHercParis4/volumes/20260411134726-2.400um-0.2m-78keV-masked.zarr')
ValueError: No arrays found in zarr Group at https://...
```

The arrays are there and perfectly readable — `.zattrs` names all six levels explicitly, and `group['3']` returns the array. Only the *discovery* is broken.

This reads level paths from the OME-Zarr `multiscales` metadata when present and falls back to listing otherwise. Keys are computed once, since each lookup is a round trip on a remote store.

## Verified over https://

Seven volumes across six scrolls, all of which failed at construction before:

| scroll | listing | levels in .zattrs | before | after |
|---|---|---|---|---|
| PHercParis4 | 0 | 6 | ValueError | 6 levels, shape(3)=(9473, 4087, 4087) |
| PHerc0500P2 | 0 | 6 | ValueError | 6 levels |
| PHerc0139 | 0 | 6 | ValueError | 6 levels, shape(3)=(2622, 828, 828) |
| PHerc1667 | 0 | 6 | ValueError | 6 levels |
| PHerc0814 | 0 | 6 | ValueError | 6 levels |
| PHerc1451 | 0 | 6 | ValueError | 6 levels, shape(3)=(2250, 727, 727) |
| PHerc0009B | 0 | 6 | ValueError | 6 levels, shape(3)=(1200, 980, 980) |

The same volumes list fine over `s3://`, so the failure is specific to the protocol.

## Scope, and what this does *not* fix

This gets construction and `shape()` working. **Reading data through `__getitem__` still fails** on the integer-key issue that #1177 fixes:

```
TypeError: sequence item 0: expected str instance, int found
```

I checked this by applying #1177 to a clean tree and testing both paths: with #1177 alone, `Volume(type='scroll', scroll_id=1)` works but the direct `https://` group path still raises `No arrays found`; with this alone, level discovery works but reads raise the TypeError. **The two are complementary**, and they touch adjacent lines — this will need a rebase once #1177 lands, which I'm happy to do.

There is a workaround today, which is presumably why this hasn't bitten more people: pointing `path` at a specific level (`.../volume.zarr/0`) works. But that loses the pyramid, and the documented contract is "Direct path/URL to the Zarr store" with `path="/path/to/my/data.zarr"`.

Why it matters beyond tidiness: `https://` is the route that needs no AWS credentials — `s3://` defaults to `anon: False` and raises `NoCredentialsError` on this public bucket unless the caller knows to pass `anon=True`.

## Tests

11 new tests in `tests/data/test_multiscale_levels.py`, no network — built on genuine in-memory zarr groups so they exercise the same isinstance branches as production:

- level paths read from metadata, in declared order (not re-sorted: `['0','1','10','2']` stays put)
- malformed or absent metadata returns `None` so listing takes over — six parametrised shapes of "unusable"
- an `attrs` access that raises does not propagate
- **a group whose `keys()` is empty still resolves its levels from metadata** — the HTTP case
- listing still used when there is no metadata
- plain arrays report one level; the legacy list-of-levels form still counted
- keys computed once, not per lookup

Full `tests/data` suite: **68 passed**. The second commit fixes a regression the first introduced — key lookup was applied unconditionally and broke the legacy list form, taking down four tests in `test_volume_retry`; that is caught by the added coverage now.

### DIFF
--- a/vesuvius/src/vesuvius/data/volume.py
+++ b/vesuvius/src/vesuvius/data/volume.py
@@ -93,6 +93,29 @@ from .utils import get_max_value, open_zarr
 Image.MAX_IMAGE_PIXELS = None
 
 
+def _multiscale_level_paths(group) -> Optional[List[str]]:
+    """Level paths from a group's OME-Zarr ``multiscales`` metadata.
+
+    Returns ``None`` when the group carries no such metadata, so callers fall
+    back to listing. Reading the metadata rather than listing is what makes
+    multiscale volumes usable over HTTP, where the store cannot enumerate its
+    own members.
+    """
+    try:
+        multiscales = group.attrs.get("multiscales")
+    except Exception:
+        return None
+    if not multiscales:
+        return None
+    try:
+        datasets = multiscales[0].get("datasets") or []
+        paths = [d["path"] for d in datasets
+                 if isinstance(d, dict) and d.get("path")]
+    except (AttributeError, IndexError, TypeError, KeyError):
+        return None
+    return paths or None
+
+
 class Volume:
     """
     A class to represent a 3D volume in a scroll or segment.
@@ -371,6 +394,31 @@ class Volume:
             print('  zarr_vol = Volume(type="zarr", path="/path/to/my/data.zarr")')
             raise
 
+    def _level_keys(self) -> List[str]:
+        """Keys of the resolution levels, in order.
+
+        Prefers OME-Zarr ``multiscales`` metadata and falls back to listing.
+        Listing is what breaks over HTTP: the store has no LIST operation, so
+        ``keys()`` and ``len()`` both come back empty and every level lookup
+        fails even though the arrays are perfectly readable by name.
+        """
+        cached = getattr(self, "_level_keys_cache", None)
+        if cached is not None:
+            return cached
+        if isinstance(self.data, zarr.Array):
+            keys: List[str] = []
+        else:
+            keys = _multiscale_level_paths(self.data) or [
+                str(k) for k in self.data.keys()]
+        self._level_keys_cache = keys
+        return keys
+
+    def _num_levels(self) -> int:
+        """Number of resolution levels, without relying on listing."""
+        if isinstance(self.data, zarr.Array):
+            return 1
+        return len(self._level_keys())
+
     def _s3_storage_options(self, path: str) -> Optional[dict]:
         """Return storage_options for S3 paths, None otherwise."""
         if path.startswith('s3://'):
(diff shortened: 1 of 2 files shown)


hendrikschilling:
Some comments you might fix. I'll hand this off to @bruniss , but I think this should be handled jointly/after https://github.com/ScrollPrize/villa/pull/1177
  
```
  1. Test compatibility bug with supported zarr 2.x:
     test_multiscale_levels.py:85 uses g.create_array(...). On this machine zarr==2.18.7, zarr.group() has create_dataset, not create_array. The repo allows zarr>=2.18.7,<4, so the tests  
     need the same zarr2/zarr3 branch used elsewhere.

  2. The runtime fix is intentionally incomplete:
     shape() now uses metadata keys, but __getitem__ and ndim still use self.data[0] / self.data[subvolume_idx] at PR-head lines 899, 927, 950, 957, 972, 1169-1171. Since #1177 is still   
     open and not mergeable, this PR alone makes HTTP root construction/shape work, but direct reads still fail.

  3. Important merge-resolution corner case:
     If #1177 and #1332 are combined, _level(idx) must use the declared metadata path from _level_keys()[idx], not str(idx). OME-Zarr paths are allowed to be non-numeric/nested, and #1332 
     explicitly preserves metadata order.

  4. Minor existing corner:
     meta() sorts with a key that can return mixed int/str (volume.py:538), so mixed paths like ["0", "scale1"] can fail. Less important, but easy to avoid.
```

aistae:
@hendrikschilling thanks — point 2 was the important one and you were right,
the fix was half done. Pushed c376ba8.

1. Test fixture now branches between Group.create_array (zarr 3) and
   create_dataset (2.x), the same way test_volume_meta.py does, so it runs
   across the supported range.

2. Added Volume._level(idx): one place that resolves a level for all three
   storage forms (single array, Group, legacy list). __getitem__, shape(),
   ndim and both diagnostic prints go through it now, so reads work on an
   HTTP-backed multiscale store, not just construction. shape() went from 11
   lines to 1.

3. _level takes the key from _level_keys()[idx], never str(idx) — so
   non-numeric and nested paths work, and the merge with #1177 stays well
   defined.

4. meta() sort key is now uniformly typed; ["0", "scale1"] no longer raises.

New tests cover reads through non-numeric level paths (full/half/quarter, so
positional or str(idx) indexing can't pass by accident), out-of-range indices,
the single-array store, the legacy list form, and the mixed-key sort. 62 tests
in tests/data pass.

Happy to hold this behind #1177 if that's the order you prefer.

PR #1334, opened 2026-08-05
core: compare volume path identity, not path spelling

nerln:
Three sibling helpers in `VolumePkg.cpp` answer the same question, "are these two locations the same directory?", and they disagree. `sameLocalSegmentsLocation` falls back to `fs::equivalent`. `sameAttachmentLocation`, `loadedVolumeMatchesLocation` and one branch of `entryBacksAttachmentLocation` compare the normalized strings and stop there.

## Why the two sides can disagree

Both spellings survive on purpose, so this is reachable rather than theoretical:

- `VolumePkg::load` stores `path_` exactly as handed in (`VolumePkg.cpp:625`)
- VC3D's open dialog passes `QFileDialog::DontResolveSymlinks` (`CWindow.cpp:8567`)
- `--volpkg` accepts a relative path (`VCAppMain.cpp:293`), and `fs::absolute` resolves it against a `getcwd` that is always symlink-resolved

So one side can carry a symlink component while the other carries its target. When that happens, re-attaching an already-attached volume returns `VolumeIdConflict` instead of `AlreadyAttached`, and the user is shown an error that is not true.

## Reproduction

`test_volume_pkg_full` fails at `core/test/test_volume_pkg_full.cpp:307` on macOS today. Reduced to 30 lines against the real helpers:

```
entry      : /private/var/folders/.../vc_repro/volumes/vol1
attachment : /var/folders/.../vc_repro/volumes/vol1
lexical ==      : no
fs::equivalent  : yes
```

macOS reaches it easily because `fs::temp_directory_path()` is `/var/folders/...` with `/var` a symlink to `/private/var`. **The defect is not macOS-specific** — any symlinked data root does the same on any platform.

## The change

Extract the fallback the segments path already uses into a shared `samePathOnDisk`, and route all three comparisons through it. Lexical equality first, so locations that do not exist yet keep exactly the old answer; `fs::equivalent` with the `error_code` overload as the widening fallback, so a missing path degrades to the old behaviour instead of throwing.

The change is one-way: `samePathOnDisk` can only turn a previously-false comparison true, never the reverse, so nothing that matched before can stop matching.

## Verification

- macOS/arm64: `test_volume_pkg`, `test_volume_pkg_full`, `test_volume_pkg_more` all pass (75 cases), including the guard test that requires `VolumeIdConflict` for two genuinely different directories sharing a volume id.
- Linux: configured and built inside `ghcr.io/scrollprize/vc3d-deps/linux@sha-0c371b1d`, the image the `test-core` job runs in, with that job's exact flags. **97/97 tests passed.**

## One thing worth a reviewer's eye

On case-insensitive APFS/NTFS this makes `volumes/Vol1` and `volumes/vol1` dedupe to one entry, while case-sensitive Linux keeps them separate. That is each filesystem's own truth, and the file already does `asciiLower` matching for segment directory names, but a volpkg written on Linux with case-variant entries would collapse when opened on macOS.

---

Found while building VC3D natively on Apple Silicon. Agent-assisted; every reproduction re-run by hand, and the claims here were put through an independent review pass that refuted two other findings I had and narrowed this one.


hendrikschilling:
@codex review

hendrikschilling:
Thanks! Would be nice if we can make this resolution more compete though:

```
  1. entryBacksAttachmentLocation() still has a lexical parent check at volume-cartographer/core/src/VolumePkg.cpp:279.
     If the project has a collection entry via a symlink path and the attached child .zarr arrives via the resolved path, this still fails to recognize that the collection backs the child.
     That is the same class of bug this PR fixes for direct volume paths.

  2. Remote volumes are not passed through fs::equivalent, so no new filesystem probing happens for s3:// / http(s):// locations. Good.
  3. Existing remote semantics are still inconsistent: sameAttachmentLocation() treats selector-free s3://bucket/path.zarr and its resolved HTTPS URL as the same, but
     samePersistedVolumeIdentity() intentionally does not dedupe two selector-free remote locators unless the strings are exactly equal. Not newly introduced, but worth testing/documenting
     because attach/match and persisted-entry dedupe can disagree.
```

nerln:
Thanks, all three are addressed. Pushed.

**1. `entryBacksAttachmentLocation()`'s lexical parent check.** Fixed, it now goes through `samePathOnDisk` like the other two. You were right that it is the same class of bug. I had left it lexical on purpose and said so in a comment, on the reasoning that widening it changes what gets persisted for collection entries — but that argument applies just as much to the direct-volume path this PR was already changing, so it was inconsistent. The comment is gone with it.

**2. Remote volumes.** Confirmed, and I have kept it that way: both `sameAttachmentLocation` and `loadedVolumeMatchesLocation` return on the remote branch before reaching `samePathOnDisk`, so nothing new touches the filesystem for `s3://` or `http(s)://`.

**3. The attach/persist divergence.** Documented above `samePersistedVolumeIdentity`, naming the exact disagreement: two selector-free remote locators dedupe in `sameAttachmentLocation` when they resolve to the same source URL, and stay distinct there unless the strings match. Left the behaviour alone since it is pre-existing and changing remote dedupe semantics feels like its own decision, but it is no longer silent. Happy to take it further if you want it made consistent.

**And the symlink regression test** the bot asked for. Two of them, both creating the directory symlink explicitly instead of leaning on macOS symlinking `/var` to `/private/var` — one for the direct volume path, one for the container entry from point 1. They skip cleanly via the `error_code` overload where symlinks cannot be created, so Windows without developer mode does not fail on them.

`test_volume_pkg_full` goes 16 -> 18 cases, all passing. `test_volume_pkg`, `test_volume_pkg_full` and `test_volume_pkg_more` are green on macOS/arm64. The earlier Linux check for this branch was run inside `ghcr.io/scrollprize/vc3d-deps/linux@sha-0c371b1d` with the `test-core` flags, 97/97; happy to re-run it there with the new cases if that is useful before merge.

PR #1337, opened 2026-08-05
cmake(macos): search frameworks last

nerln:
On macOS with Homebrew LLVM, `find_package(TIFF)` can resolve to a mismatched pair: headers from a framework, library from Homebrew.

`CMAKE_FIND_FRAMEWORK` defaults to `FIRST` on Apple platforms. `find_path(TIFF_INCLUDE_DIR tiff.h)` therefore searches every framework's `Headers/` before the Homebrew prefix, and any framework shipping a `tiff.h` wins — on this machine that is `/Library/Frameworks/Mono.framework/Headers`. `find_library(TIFF_LIBRARY tiff)` then finds no `tiff.framework` and falls through to `libtiff.dylib` under Homebrew. CMake reports `TIFF_FOUND`, and the build compiles against one version's headers while linking another's library.

Setting `CMAKE_FIND_FRAMEWORK` to `LAST` in the `macos-homebrew-llvm` preset makes package config and Homebrew take precedence, and leaves genuine framework-only dependencies still findable.

## Reproducing it is narrower than I first said

I originally reported this as reproducible from the preset alone. It is not, and I would rather say so than let the claim stand: `nlohmann_json`'s include directory usually arrives first and masks the framework header. It reproduces through `scripts/build_macos.sh` on a machine with Mono installed, which is how I hit it.

So this is hardening rather than a fix for something CI sees. It is inert on every non-Apple platform, and `macos-15` CI exercises this preset.

## Scope

One line in one preset. An earlier version of this PR also added `Boost::headers` to two targets; I dropped that half after checking the include path more carefully, and this description now matches the diff.

### DIFF
--- a/volume-cartographer/CMakePresets.json
+++ b/volume-cartographer/CMakePresets.json
@@ -177,7 +177,7 @@
     {
       "name": "macos-homebrew-llvm",
       "displayName": "macOS Homebrew LLVM",
-      "description": "Native macOS build using Homebrew LLVM/Clang, lld, and Homebrew binutils. The only Apple-toolchain pieces in the loop are SDKROOT (system headers/frameworks) and libSystem, because there is no Homebrew alternative for those.",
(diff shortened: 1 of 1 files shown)


nerln:
Closing in favour of #1511, which is the same one-line change with the evidence this one lacked.

Two things were wrong here. I described the result as compiling against one version's headers while linking another's library; it does not compile at all, `core/src/Tiff.cpp` stops with two `typedef redefinition` errors. And I framed it as hardening rather than a fix, when it is a hard failure on an affected machine. The PR also carried no error output and did not follow the template.

PR #1338, opened 2026-08-05
scripts(macos): add --tests and --check to build_macos.sh

nerln:
`build_macos.sh` cannot build or run the C++ tests, so on macOS the only way to exercise them is to bypass the documented script.

The macOS CI job runs `build_macos.sh --sccache` and then the 163 Python `vc3d-mcp` tests, but no `ctest`. Both jobs that do invoke `ctest` are `ubuntu-24.04` (`vc3d-ci.yml:153,190`). The preset the script uses pins `VC_TESTING: OFF` (`CMakePresets.json:205`).

## The change

`--tests` configures with `VC_TESTING=ON`, `--check` builds and runs them.

They use a separate `build-macos-tests` directory rather than reusing `build-macos`. `VC_TESTING` is a compile definition on `vc_fiber_tracer` (`core/CMakeLists.txt:196-197`), so flipping it in place would silently rebuild a differently-defined library into the same tree.

`--check` also sets `--no-tests=error`. Without it `ctest` exits 0 when a label matches nothing, which is exactly how a test suite disappears unnoticed.

## Note on overlap with #1339

An earlier version of this argued that `--check` had to shell out to `ctest` because the `check-*` targets matched nothing. #1339 has since landed and those targets now carry `VERBATIM`, so that argument no longer holds. The separate build directory is still needed for the reason above, but if you would rather `--check` drive the `check-*` targets now that they work, say so and I will rework it.

### DIFF
--- a/volume-cartographer/scripts/build_macos.sh
+++ b/volume-cartographer/scripts/build_macos.sh
@@ -4,7 +4,8 @@ set -euo pipefail
 usage() {
   cat <<'USAGE'
 Usage:
-  scripts/build_macos.sh [--install-deps] [--ccache|--sccache] [--build-dir DIR] [--jobs N]
+  scripts/build_macos.sh [--install-deps] [--ccache|--sccache] [--build-dir DIR]
+                         [--jobs N] [--tests] [--check]
 
 Build VC3D natively on macOS with Homebrew LLVM/Clang.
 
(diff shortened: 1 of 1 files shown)


nerln:
Closing in favour of #1512, which does the same thing in 24 lines instead of 57.

The difference is the separate `build-macos-tests` directory, which I justified by saying that flipping `VC_TESTING` in place would silently rebuild a differently defined library into the same tree. That is not true. A command-line `-DVC_TESTING=ON` outranks the preset's cache entry, CMake reconfigures correctly in both directions, and the next plain `build_macos.sh` run restores `OFF`. The second tree only bought a full rebuild.

This PR also carried no before/after output and did not follow the template.

PR #1341, opened 2026-08-05
VC3D: fiber split/merge workflows and same-orientation link colors

skyward7187:
Follow-up to #1324. Three related additions to the Line Annotation GUI, all modeled on the existing link-candidate workflow, plus the pure geometry helpers they need.

## Split workflow

Ctrl+right-click a control point → **Designate as split candidate** (renders red in all candidate palette sites); ctrl+right-click the adjacent CP → two actions:

- **Split from candidate, different winding** — the fiber becomes two brand-new fibers: prefix `0..a`, suffix `a+1..end`, connecting span dropped (its descriptor cleared per the v3 final-CP contract). Fresh canonical identities; tags, optimization mode, and per-span `segment_to_next` inherited verbatim; manual H/V kept with automatic scores recomputed; branch links partitioned with peer reciprocals redirected to the owning half. The original is deleted and the candidate's half reopens at the candidate CP.
- **Split from candidate and link** — same split plus a **pending** reciprocal link between the two boundary CPs: the split asserts these are different fibers, but same-winding is a separate determination for a reviewer to approve.

Each half must keep ≥2 control points; halves are written synchronously before any in-memory mutation (a save failure aborts with nothing to roll back).

## Merge with candidate

The link menu gains **Merge with candidate** when the candidate and clicked CPs are both *endpoints* of *different* fibers (greyed with a reason otherwise). Sides are oriented so the join meets end-to-end — fiber reversal moves each span descriptor with its span — and both extrapolated tails are sliced at the boundary CPs via the loader's exact ordered-subset scan (raw concatenation breaks the optimizer's strict line-order mapping). Tags union; optimization-mode conflicts prompt (suppressed/agent sessions take the clicked fiber's mode; `setMergeModePickerForTesting` seam); manual H/V keeps only agreement; the pair link between the merge endpoints is consumed; any other mutual link blocks the merge; third-party links redirect to the merged fiber. Both originals are deleted and the merged fiber reopens at the join.

## Shared mechanics

- Both operations tag their results `needs_reoptimization` and immediately re-fit via the merged-fiber machinery, so lines regrow extrapolation tails (and the merge's join span gets real geometry); an aborted batch keeps the tag for the existing load-time prompt.
- The retire/re-fit/reopen tail is deferred with `QTimer::singleShot(0, …)`: the handlers run inside the dialog's own menu-callback stack, and the nested event loops in `waitForFiberSaves`/re-optimization would otherwise process the closed dialog's deferred deletion under its own frames (observed SIGABRT during testing). `waitForFiberSaves()` also flushes any close-triggered async save before the original files are removed.
- Originals are retired directly (not via `deleteFibers`, whose `removeBranchLinksToFiber` would strip the just-redirected peer links).

## Same-orientation link colors

H–H / V–V links (manual H/V tag falling back to automatic) render the node **light orange** while pending and **orange** when approved, in all four palette sites (overlay styles, fast current-cut path, cross-slice, overview bar). H–V links keep the existing blue/purple palette; unclassified fibers on either side keep the defaults.

## Tests

`core/test/test_fiber_trace_review.cpp` (no CMake change): `computeFiberSplitPlan` slicing/bounds/tolerance/coincident-point cases, `reversedStoredControlPoints` descriptor-shift matrix, `computeFiberMergeGeometry` tail-slicing/join-descriptor/ordered-scan-failure cases, and a composed reverse-then-merge orientation case. GUI-smoke-tested: split, split-and-link, merge, and the crash repro above.

https://claude.ai/code/session_01GvKbK7HXygY2yjjtkvJs2H

### DIFF
--- a/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
+++ b/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
@@ -3983,6 +3983,26 @@ void LineAnnotationController::cleanupIntersectionInspectionSurfaces()
     _intersectionInspection->generatedSurfaceContexts.clear();
 }
 
+void LineAnnotationController::closeIntersectionInspectionForRetiredFibers(
+    const std::vector<uint64_t>& fiberIds)
+{
+    if (!_intersectionInspection) {
+        return;
+    }
+    const auto sessionHoldsRetiredFiber =
+        [&fiberIds](const std::shared_ptr<LineAnnotationSession>& session) {
+            return session &&
+                std::find(fiberIds.begin(), fiberIds.end(), session->fiberId) !=
+                fiberIds.end();
+        };
+    if (!sessionHoldsRetiredFiber(_intersectionInspection->sourceLineSession) &&
+        !sessionHoldsRetiredFiber(_intersectionInspection->targetLineSession)) {
+        return;
+    }
+    cleanupIntersectionInspectionSurfaces();
+    _intersectionInspection.reset();
+}
+
 bool LineAnnotationController::updateIntersectionFollowSlice(bool sourceSideFlag,
                                                              double linePosition,
                                                              const char* reason)
@@ -7274,26 +7294,41 @@ void LineAnnotationController::handleGeneratedControlPointMergeWithCandidate(
         closeFiberWindowForSurface(paneSurfaceName);
         // The close finalization may schedule an async save of an original;
         // flush it before retiring the files so they cannot be resurrected.
+        // A failure among the flushed saves (the redirected peer links ride
+        // this queue) keeps the originals on disk AND in memory: disk and
+        // panel then agree, showing the originals beside the merged fiber.
+        const uint64_t saveFailuresBefore = _fiberSaveFailureCount;
         waitForFiberSaves();
-        // All-or-nothing: a failure restores both originals in place.
-        const auto retireResult = vc3d::line_annotation::runFiberSaveJob(
-            ++_nextFiberSaveSequence, {}, {clickedPath, farPath});
-        if (!retireResult.ok) {
-            showError(tr("Could not retire the original fiber files: %1")
-                          .arg(QString::fromStdString(retireResult.error)),
+        bool retired = _fiberSaveFailureCount == saveFailuresBefore;
+        if (!retired) {
+            showError(tr("A pending fiber save failed; the original fiber "
+                         "files were kept."),
                       suppressErrors);
+        } else {
+            // All-or-nothing: a failure restores both originals in place.
+            const auto retireResult = vc3d::line_annotation::runFiberSaveJob(
+                ++_nextFiberSaveSequence, {}, {clickedPath, farPath});
+            retired = retireResult.ok;
+            if (!retired) {
+                showError(tr("Could not retire the original fiber files: %1")
+                              .arg(QString::fromStdString(retireResult.error)),
+                          suppressErrors);
+            }
         }
-        _fibers.erase(std::remove_if(_fibers.begin(),
-                                     _fibers.end(),
-                                     [clickedId, farId](const StoredFiber& fiber) {
-                                         return fiber.id == clickedId ||
-                                             fiber.id == farId;
-                                     }),
-                      _fibers.end());
-        invalidateFiberAlignmentMetrics(clickedId, false);
-        invalidateFiberAlignmentMetrics(farId, false);
-        emit fibersDeleted(
-            {std::min(clickedId, farId), std::max(clickedId, farId)});
+        if (retired) {
+            _fibers.erase(std::remove_if(_fibers.begin(),
+                                         _fibers.end(),
+                                         [clickedId, farId](const StoredFiber& fiber) {
+                                             return fiber.id == clickedId ||
+                                                 fiber.id == farId;
+                                         }),
+                          _fibers.end());
+            invalidateFiberAlignmentMetrics(clickedId, false);
+            invalidateFiberAlignmentMetrics(farId, false);
+            emit fibersDeleted(
+                {std::min(clickedId, farId), std::max(clickedId, farId)});
+            closeIntersectionInspectionForRetiredFibers({clickedId, farId});
+        }
 
         // Re-fit the merged line (join span + extrapolation tails); consumes
         // the needs_reoptimization tag, which otherwise re-prompts on load.
@@ -7649,24 +7684,39 @@ void LineAnnotationController::handleGeneratedControlPointSplitFromCandidate(
                                  reopenControlIndex, suppressErrors]() {
         closeFiberWindowForSurface(paneSurfaceName);
         // The close finalization may schedule an async save of the original;
-        // flush it before retiring the file so it cannot be resurrected.
+        // flush it before retiring the file so it cannot be resurrected. A
+        // failure among the flushed saves (the redirected peer links ride
+        // this queue) keeps the original on disk AND in memory: disk and
+        // panel then agree, showing the original beside its halves.
+        const uint64_t saveFailuresBefore = _fiberSaveFailureCount;
         waitForFiberSaves();
-        const auto retireResult = vc3d::line_annotation::runFiberSaveJob(
-            ++_nextFiberSaveSequence, {}, {parentPath});
-        if (!retireResult.ok) {
-            showError(tr("Could not retire the original fiber file %1: %2")
-                          .arg(QString::fromStdString(parentPath.string()))
-                          .arg(QString::fromStdString(retireResult.error)),
+        bool retired = _fiberSaveFailureCount == saveFailuresBefore;
+        if (!retired) {
+            showError(tr("A pending fiber save failed; the original fiber "
+                         "file was kept."),
                       suppressErrors);
+        } else {
+            const auto retireResult = vc3d::line_annotation::runFiberSaveJob(
+                ++_nextFiberSaveSequence, {}, {parentPath});
+            retired = retireResult.ok;
+            if (!retired) {
+                showError(tr("Could not retire the original fiber file %1: %2")
+                              .arg(QString::fromStdString(parentPath.string()))
+                              .arg(QString::fromStdString(retireResult.error)),
+                          suppressErrors);
+            }
         }
-        _fibers.erase(std::remove_if(_fibers.begin(),
-                                     _fibers.end(),
-                                     [parentId](const StoredFiber& fiber) {
-                                         return fiber.id == parentId;
-                                     }),
-                      _fibers.end());
-        invalidateFiberAlignmentMetrics(parentId, false);
-        emit fibersDeleted({parentId});
+        if (retired) {
+            _fibers.erase(std::remove_if(_fibers.begin(),
+                                         _fibers.end(),
+                                         [parentId](const StoredFiber& fiber) {
+                                             return fiber.id == parentId;
+                                         }),
+                          _fibers.end());
+            invalidateFiberAlignmentMetrics(parentId, false);
+            emit fibersDeleted({parentId});
+            closeIntersectionInspectionForRetiredFibers({parentId});
+        }
 
         // Re-fit both halves so their lines regrow extrapolation tails past
         // the split CPs. Consumes the needs_reoptimization tag set at
@@ -12632,6 +12682,7 @@ void LineAnnotationController::finishFiberSaveJob(
             emit fiberSaved(result.fiberIds[i], generation);
         }
     } else {
+        ++_fiberSaveFailureCount;
         errorMessage = tr("Could not save fiber data: %1")
                            .arg(QString::fromStdString(result.error));
         if (!result.recoveryFiles.empty()) {
--- a/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
+++ b/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
@@ -4003,6 +4003,23 @@ void LineAnnotationController::closeIntersectionInspectionForRetiredFibers(
     _intersectionInspection.reset();
 }
 
+void LineAnnotationController::closeDialogPanesForFibers(
+    const std::vector<uint64_t>& fiberIds)
+{
+    // Collect first: closing mutates pane state through queued events.
+    std::vector<std::string> surfaceNames;
+    for (const auto& pane : _panes) {
+        if (pane.session && pane.dialog &&
+            std::find(fiberIds.begin(), fiberIds.end(), pane.session->fiberId) !=
+                fiberIds.end()) {
+            surfaceNames.push_back(pane.surfaceName);
+        }
+    }
+    for (const auto& surfaceName : surfaceNames) {
+        closeFiberWindowForSurface(surfaceName);
+    }
+}
+
 bool LineAnnotationController::updateIntersectionFollowSlice(bool sourceSideFlag,
                                                              double linePosition,
                                                              const char* reason)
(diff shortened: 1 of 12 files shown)


hendrikschilling:
```
  1. P1: Merge can discard unsaved edits from the candidate fiber.
     The clicked fiber is captured with storedFiberFromSession(session), but the second fiber is copied from _fibers. Since the candidate is selected through another open annotation pane, 
     its current session may contain newer geometry, tags, or links. The merge then deletes that fiber using the stale snapshot. LineAnnotationController.cpp:6977
     (https://github.com/ScrollPrize/villa/blob/5e6258d377e9aa7322c11322d3ac1705a5ecf9b1/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L6977-L6999)

  2. P1: Failed reciprocal-link saves do not prevent deletion of the originals.
     Peer links are redirected and saved asynchronously, then waitForFiberSaves() merely waits and returns no success status. finishFiberSaveJob() reports failures but the deferred        
     operation still deletes the source fibers. A failed peer save therefore leaves persisted fibers referencing deleted originals. redirect/save
     (https://github.com/ScrollPrize/villa/blob/5e6258d377e9aa7322c11322d3ac1705a5ecf9b1/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L7164-L7210), wait implementation       
     (https://github.com/ScrollPrize/villa/blob/5e6258d377e9aa7322c11322d3ac1705a5ecf9b1/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L12568-L12635)

  3. P1: Source-file deletion failures are reported but otherwise ignored.
     Both workflows continue by erasing the source from _fibers, emitting fibersDeleted, and reoptimizing the replacements even when std::filesystem::remove() fails. On reload, stale      
     originals reappear beside the merged/split results; merge can also delete only one of its two sources. merge
     (https://github.com/ScrollPrize/villa/blob/5e6258d377e9aa7322c11322d3ac1705a5ecf9b1/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L7237-L7278), split
     (https://github.com/ScrollPrize/villa/blob/5e6258d377e9aa7322c11322d3ac1705a5ecf9b1/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L7617-L7652)

  4. P2: Other panes for retired fibers remain open in an unsaveable state.
     All source sessions receive suppressFiberSave = true, but only the pane where the command was invoked is closed. In particular, the merge candidate’s pane can remain editable after   
     its backing file and _fibers entry are removed, so subsequent edits are silently lost. LineAnnotationController.cpp:7212
     (https://github.com/ScrollPrize/villa/blob/5e6258d377e9aa7322c11322d3ac1705a5ecf9b1/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L7212-L7244)

```

skyward7187:
Thanks — fixed 1, pushing back on 2–4 with specifics:

**1 (candidate merged from stale snapshot) — fixed in 6fd422af9.** Before the merge snapshots the candidate fiber, any dialog-less editing session holding it is baked into `_fibers` via `storedFiberFromSession`, using the same eligibility guards as `scheduleBranchMetadataSaves`. (The clicked side was already baked; a *dialog* pane for the candidate can't exist — `launchSession` finalizes and saves every other dialog pane before opening a new one — so the inspection-session case was the real gap.) Note the pre-existing link-with-candidate flow reads the stored far fiber the same way; that's less severe since nothing is deleted, but it'd be a candidate for the same treatment in a follow-up.

**2 (peer save failure doesn't prevent deletion) — one factual correction, then precedent.** The peer redirect saves are scheduled *before* the deferred tail runs, and the tail calls `waitForFiberSaves()` **before** removing the originals — so peer writes are flushed (or have failed with the standard error dialog + `.recovery.*` files from `runFiberSaveJob`) by deletion time; it's not an async race past the delete. What's true is that a *failed* flush doesn't abort the deletion. That matches `deleteFibers`, which strips peer links via `removeBranchLinksToFiber` with the same fire-and-forget saves while removing the fiber file immediately; the backstop in both cases is load-time repair (`collectLoadedFiberBranchIssues` → the broken-branch-links prompt). Making the save queue report per-job success to callers would be a real improvement, but it's an infrastructure change that should cover delete/unlink/rename too — proposing it as a follow-up rather than a bespoke path here.

**3 (fs::remove failure reported but not fatal) — same contract as `deleteFibers`**, which also shows the error and continues (erasing from `_fibers` regardless). By that point the halves/merged fiber are saved and peers redirected, so there is no clean unwind available; the dialog names the exact leftover path, and a reload surfaces the stale original next to its replacements rather than losing anything. If you'd prefer, I can suffix-rename the leftover (e.g. `.retired`) on remove failure so a reload can't resurrect it — say the word and I'll add it, but plain report-and-continue keeps parity with delete.

**4 (other panes left open, unsaveable) — also `deleteFibers` semantics**: deleting a fiber that's open in a pane sets `suppressFiberSave` and leaves the pane open. Only dialog-less inspection sessions can be in this state (single-dialog invariant), and tearing down half an intersection-inspection pair from inside the merge handler touches inspection lifecycle state I'd rather not reach into surgically. A shared "retire panes whose fiber vanished" path would fix this for delete and merge alike — follow-up material.

Happy to do the `.retired` rename (3) and/or file the two follow-ups as issues if that works for you.

skyward7187:
Implemented the backup/temp/rename/cleanup suggestion in edf29743c, as an extension of the existing save-job transaction rather than new machinery:

- `runFiberSaveJob` — which already did recovery-copy backups, unique temp files, atomic renames, and backups-removed-only-after-success for overwrites — now also takes **retirements**: originals are atomically renamed into a dot-prefixed sibling `.retired/` directory *before* the payload renames, deleted only after every rename succeeded, and renamed straight back into place on failure (a backup that can't be restored is surfaced in the same "Recovery backups were kept:" dialog as the `.recovery.*` files). Retirement is all-or-nothing across the batch.
- `.retired/` rides the same dot-prefix exclusion that protects `.s3sync-conflicts`: the three C++ fiber scanners (non-recursive `*.json`), `vc_sync`'s `_is_ignored`/walk pruning, and hfsync's `classify_fibers` all ignore it with zero scanner changes. (A `*.json.retired` sibling-suffix scheme was considered and rejected — vc_sync would upload those to S3.)
- Split and merge now route their file transitions through it: the new fibers are written as one transactional job (replacing the `saveFiberNow` calls and the split's hand-rolled rollback, plus a preflight name-collision check), and the deferred `fs::remove` of the originals became a retire-only job — which also makes the merge's two-file deletion all-or-nothing, addressing the half-deleted case from finding 3.
- Three new `runFiberSaveJob` tests pin the behavior: retire-after-success, retire-only all-or-nothing (including a missing path being skipped), and failed-save-restores-retired-originals via the existing `VC3D_FIBER_SAVE_FAIL_AFTER_FIRST_REPLACE` hook.

hendrikschilling:
1. P1: The controller still retires in memory after a failed retire-on-disk.
     The new .retired/ move is better than raw remove(), but both split and merge still only show an error when retireResult.ok == false and then continue erasing the originals from       
     _fibers, emitting fibersDeleted, and reoptimizing/opening the replacements. If retirement fails and restores/leaves originals on disk, reload still sees duplicate old + new fibers.   
     Merge: LineAnnotationController.cpp at edf29743 around L7278. Split: around L7662.

  2. P1: Peer-save failure still does not block retirement.
     scheduleBranchMetadataSaves(...) is still async, waitForFiberSaves() still has no success/failure return, and retirement happens after that wait regardless. So failed reciprocal-link 
     saves can still leave peer fibers on disk pointing to retired originals.

  3. P1: Split’s new “transactional” two-file save is not actually atomic for brand-new targets.
     runFiberSaveJob() writes temp files, renames payload 0 into place, then if payload 1 fails there is no recovery file for payload 0 because the target did not previously exist. The    
     catch removes temp paths but does not remove the already-renamed new target. For split, that means a failed suffix rename can leave an orphan prefix file even though the split aborts.
     This is not covered by the new tests; the existing failure test uses pre-existing targets.
     LineAnnotationFiberSaveJob.cpp around the payload rename loop, L122-L133.

  4. P2: Other open panes for retired fibers still remain open with suppressFiberSave = true.
     The command closes only paneSurfaceName; candidate/peer panes for retired source fibers can stay open and editable while future saves are suppressed.

skyward7187:
All four addressed in e660ca484 — this round they were all right calls, no push-back:

**1 + 2 (in-memory retire after failed on-disk retire / peer-save failure not blocking retirement)** — fixed together. `finishFiberSaveJob` now counts failures, and the deferred tails compare the count across the `waitForFiberSaves()` flush (which is what carries the redirected peer-link saves). Any failure in the flush skips retirement entirely; a failed retirement skips the in-memory deletion. Either way `_fibers`, `fibersDeleted`, and the inspection teardown only run when the originals are actually gone from disk, so panel and disk agree: on failure both show the originals beside the replacements, links repair on reload. (One deliberate conservatism: the flush covers the whole save queue, so an unrelated job's failure also blocks retirement — erring toward keeping files.)

**3 (orphan new target from an aborted two-file save)** — correct, the recovery-copy path only protected pre-existing targets. `runFiberSaveJob` now records which targets existed before their rename and, on failure, removes renamed targets that had no previous content — an aborted split can no longer leave an orphan prefix file. Pre-existing targets keep the new content plus their recovery copy, unchanged. New test with two brand-new targets + the injection hook, and the retire-failure test now asserts the orphan is gone too.

**4 (retired fiber's pane left open, unsaveable)** — you were right to re-flag it; the existing teardown pair (`cleanupIntersectionInspectionSurfaces()` + reset, the same idiom the stale-intersection path uses) made this surgical after all. Retiring a fiber held by an intersection-inspection session now closes the inspection workspace. The `deleteFibers` flow still has the historical gap — happy to extend the same call there in a follow-up.

hendrikschilling:
we are converging
```
  1. P2: ordinary open panes for retired fibers can still survive with saves suppressed.
     The update only closes intersection-inspection sessions via closeIntersectionInspectionForRetiredFibers(...). But the merge/split paths still mark every source-fiber pane as
     suppressFiberSave = true, then close only the invoking pane. If the candidate/source fiber is open in a normal annotation pane, retirement can succeed while that pane remains editable
     and unsaveable.
     Merge area: LineAnnotationController.cpp at e660ca4 around L7277-L7333. Split area: around L7650-L7719.
```

skyward7187:
Addressed in 18f1fd734, with one clarification on reachability:

The scenario as stated — the candidate/source fiber open in a second ordinary annotation pane — can't currently occur: `launchSession` runs `prepareForUserFacingLineAnnotationOpen`, which finalizes, saves, and closes every existing dialog pane before opening a new one, so at most one dialog-bearing pane exists and it's the invoking one. The suppress-all-panes loop only ever matched the invoking pane plus dialog-less inspection sessions (handled last round).

That said, the code deserved to be correct without leaning on that invariant, and there *was* a real adjacent gap on the failure path, so:

- `closeDialogPanesForFibers(fiberIds)` now sweeps every dialog pane whose session holds a source fiber, replacing the close-by-surface-name of just the invoking pane — if the single-dialog invariant ever changes, retirement stays correct by construction.
- When retirement **fails** (originals kept on disk and in the panel), the surviving sessions' `suppressFiberSave` is cleared again — previously they'd have stayed open and editable with saves silently suppressed on a fiber that still exists, which was the same bug you flagged, one branch over.

hendrikschilling:
we might need to think about how to design the state handling better, the llms seem to struggle.
It might be ok to simply error out and close vc3d on some problems or to close all relevant references in vc3d and re-load instead of building cases for every possible error. Also. These errors sound not very likely, so just ignoring is always an option ...

```
  1. P1: failed close/finalization can still retire a fiber while its pane survives

     closeDialogPanesForFibers() only calls dialog->close() and does not verify the pane actually closed. The normal close path can still refuse to close if finalization fails:
     https://github.com/ScrollPrize/villa/blob/18f1fd734dc2756613e1d45de8fa2f1d2134f570/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L4006-L4020
     https://github.com/ScrollPrize/villa/blob/18f1fd734dc2756613e1d45de8fa2f1d2134f570/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L8819-L8846

     The merge/split deferred tails then continue to waitForFiberSaves() and retire anyway:
     https://github.com/ScrollPrize/villa/blob/18f1fd734dc2756613e1d45de8fa2f1d2134f570/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L7307-L7335
     https://github.com/ScrollPrize/villa/blob/18f1fd734dc2756613e1d45de8fa2f1d2134f570/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L7705-L7734

     If finalization returns false, the dialog remains open with suppressFiberSave=true, and the fiber can still be retired from disk/memory. That recreates the unsaveable-open-pane class,
     just through a failed-finalization branch.

  2. P1/P2: peer-save failure blocks retirement but does not roll back in-memory replacement/link rewrites

     The PR now blocks retirement if _fiberSaveFailureCount changes, which is good. But by then the replacement fiber is already in _fibers, and peer refs have already been rewritten in   
     memory:
     merge: https://github.com/ScrollPrize/villa/blob/18f1fd734dc2756613e1d45de8fa2f1d2134f570/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L7233-L7285
     split: https://github.com/ScrollPrize/villa/blob/18f1fd734dc2756613e1d45de8fa2f1d2134f570/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L7616-L7677

     If a pending peer save fails, originals remain on disk/in memory, but peers in memory can now point to the replacement. The code then still reoptimizes/opens the replacement
     unconditionally:
     https://github.com/ScrollPrize/villa/blob/18f1fd734dc2756613e1d45de8fa2f1d2134f570/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L7358-L7365
     https://github.com/ScrollPrize/villa/blob/18f1fd734dc2756613e1d45de8fa2f1d2134f570/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L7753-L7765

     So the “disk and panel agree” claim is only partly true: the source fibers remain, but the in-memory branch graph may already have moved on.

  3. P2: merge menu still uses stale stored candidate geometry

     The handler now bakes a live dialog-less candidate session before snapshotting, which addresses the data-loss issue. But the menu enablement still checks the candidate endpoint state 
     from _fibers:
     https://github.com/ScrollPrize/villa/blob/18f1fd734dc2756613e1d45de8fa2f1d2134f570/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L5320-L5339

     The live-session bake only happens later in the handler:
     https://github.com/ScrollPrize/villa/blob/18f1fd734dc2756613e1d45de8fa2f1d2134f570/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L7022-L7048

     That means unsaved candidate edits can make the menu disabled/enabled based on stale geometry, even though the actual merge logic would use the live geometry.
```

skyward7187:
Agreed on the design point — the per-branch reconciliation was fighting the wrong battle. f308a97ad switches both deferred tails to a single fail-loudly path and drops the branch-by-branch handling:

**1 + 2** — any problem in the tail (a failed save in the flush, a workspace that refused to close after failed finalization, or a failed retirement) now takes one `bailOut` path: un-suppress the surviving sessions, show one error naming the cause, and `loadFibersForCurrentPackage()`. The reload rebuilds `_fibers` and the branch graph from disk (`branch_fiber_id` re-derives from `branch_file` on every load) and re-offers the re-fit via the existing `needs_reoptimization` prompt — so the in-memory link rewrites, the replacement upserts, and any other divergence simply don't survive an error. The user sees originals beside replacements and resolves; nothing is reopened or reoptimized on the failure path. The close-refusal case is caught by checking for a surviving dialog pane on a source fiber *before* retiring, so a fiber can no longer be retired under a live pane.

**3** — `mergeCandidateMenuState` now resolves the candidate's endpoint state from its live editing session when one exists (same read-only index resolution the handler uses), falling back to `_fibers` otherwise — menu enablement and the handler's baked snapshot now agree on the same geometry.

Net effect matches your suggestion: one recovery strategy (close relevant panes, reload from disk) instead of case-building, and the unlikely failures cost one dialog + a reload rather than bespoke state repair.

Issue #1250, opened 2026-07-26
RESOLVED (my error): PHerc. Paris 4 reproduces with mirroring TTA — the collection is not uniform in inference settings


TAUIL-Abd-Elilah:
The published surface predictions are regenerable from public inputs — the weights are public (`hf://scrollprize/surface_m7_nnunet`, run `20260413222639`) and so is the CT, so anyone can check an artifact rather than trust it. I ran that check across eleven scrolls and they reproduce. **PHerc. Paris 4 does not**, and I could not find a benign explanation for it.

Reporting it as a data question rather than a code bug: the tooling agrees with the bucket everywhere else, so the most likely explanation is that this scroll's run used an input or a configuration that is not published.

### What reproduces

Each row is a 256³ region, scored on the interior (64 voxels trimmed per face), thresholded at the 0.2 the published artifact is named for.

| scroll | CT level | Dice | differing voxels |
|---|---|---|---|
| PHerc0500P2 | L2 | **1.0000** | **0** |
| PHerc1203 | L2 | 0.9999 | 81 |
| PHerc0332 | L2 | 0.9998 | 134 |
| PHerc0139 | L0 | 0.9997 | 266 |
| PHerc0125 | L0 | 0.9997 | 254 |
| PHerc0175A | L0 | 0.9997 | 254 |
| PHerc0846A | L2 | 0.9997 | 281 |
| PHerc0211 | L0 | 0.9996 | 384 |
| PHerc0841 | L2 | 0.9996 | 404 |
| PHerc0191 | L0 | 0.9989 | 1,267 |
| PHerc0814 | L2 | 0.9983 | 2,918 |

Eleven scrolls, five at CT level 0 and six at level 2. In every one of them **100% of the differing voxels lie within 0.01 of the threshold** — the residue of float16 storage and autocast, with nothing structural left. One region came back exact to the voxel.

### What does not

| scroll | region | Dice | differing voxels |
|---|---|---|---|
| PHercParis4 | `6600:6856,3200:3456,4032:4288` | 0.8907 | 108,044 (5.15%) |
| PHercParis4 | `9400:9656,4032:4288,4032:4288` | 0.8425 | 158,236 (7.55%) |

Against `20260411134726-surface-20260413222639-surface-m7-L2-th0.2.zarr`, from `20260411134726-2.383um-...masked.zarr` level 2.

[...]

PR #1342, opened 2026-08-05
predict: add --resume so an interrupted whole-scroll run is not lost

TAUIL-Abd-Elilah:
## The problem

A whole-scroll inference pass streams for hours. Measuring the I/O side of it in #1325: PHerc1218 is 1.34 TB, and at the 11 MB/s my connection sustains that is ~36 h even under the best traversal policy, ~43 days with no chunk cache at all. At that length interruption is ordinary — a dropped connection, a reboot, a laptop lid.

Today an interrupted run is worth nothing. `_create_output_stores` opens `logits_part_{id}.zarr` with `mode='w'`, which recreates the array, so restarting the same command discards every patch already computed.

I measured it rather than assuming it. A 384³ region of PHerc1218 at level 0, 216 patches: killed the run after 182 s with 64 patches on disk, then reran the identical command and sampled the store every 5 s.

```
t=0s  64 chunks     t=20s   0 chunks   <- store recreated
t=5s  64 chunks     t=30s   3 chunks
t=10s 64 chunks     t=35s   5 chunks
t=15s 64 chunks     t=40s   7 chunks
```

The count holds, collapses to zero, and starts again from three. `--num_parts`/`--part_id` do not help: they split the work up front, and a dead shard is still a dead shard.

## What this adds

`--resume` continues a previous run in the same `--output_dir`: patches already written are skipped, and the existing logits store is appended to rather than recreated.

```
vesuvius.predict --input_dir <url> --output_dir out --bbox z0:z1,y0:y1,x0:x1 --resume
```

Two details drove the design and both are load-bearing.

**Completion is recorded in a manifest, not inferred by listing the store.** The logits array is chunked one-chunk-per-patch, so it is tempting to treat "chunk exists" as "patch done". That is wrong here: `write_empty_chunks=False` is set deliberately, so a patch that produces all zeros writes no chunk at all. The store therefore cannot distinguish *done and empty* from *not done* — and on a masked scroll volume the empty patches are the majority, so a listing-based resume would rerun all of them on every restart, forever. Empty patches are submitted to the writer like any other, so `on_written` records them correctly. `completed_part_{id}.json` sits beside the store and is rewritten atomically.

**A resume whose settings differ is refused, not merged.** The signature covers the input volume, bbox, patch size, overlap, model, TTA setting, normalization, part split and patch count, and the error names the fields that differ:

```
--resume refused: out/completed_part_0.json was written by a different run.
  differing: overlap, tta
```

One store holding patches computed two different ways, with nothing recording the split, is the failure mode behind #1250 and #1253 — a published prediction that turned out to reproduce only under a TTA setting no artifact mentioned. It seemed worth making that unrepresentable rather than documented.

The manifest is flushed on whichever comes first, 16 writes or 30 seconds, rather than per patch — so a crash costs a bounded handful of recomputed patches instead of an fsync on every write, and the time limit bounds the loss regardless of throughput.

`BoundedPatchWriter` gains an optional `on_written(index)` callback, fired only after the store accepts the patch and never from the error path — a resume that trusted a patch the store rejected would leave a permanent hole no later run goes back for. It is kept separate from `on_progress`, which is a bare counter for the progress bar and has no index to give.

Patch indices, the coordinates store and its attributes are unchanged, so `blend_logits` and `finalize_outputs` need no flags and see exactly what they see today.

## Tests

`vesuvius/tests/models/run/test_inference_resume.py`, 20 cases: manifest round trip, missing and corrupt manifests, refusal on each differing field, per-part manifests, that resume reopens the store rather than recreating it, that a plain run still recreates it, that a store of the wrong shape is refused, that a patch which wrote no chunk is still recorded complete, that a manifest built from numpy scalars serialises, that no `.tmp` survives a failed write, and a guard that parses `__init__` for `self.X =` and asserts the run signature reads nothing else.

Nearby suites pass unchanged: `tests/models/run` is 67 passed, and `tests/models/run` plus `tests/data` together are 119 passed, 1 skipped.

## End to end

Same region, same interruption, with the flag. Run 1 was killed with 54 patch chunks on disk and a manifest recording 51 complete; run 2 was the identical command with `--resume`:

```
--resume: 51 of 216 patches already written; skipping them
```

Chunks on disk through the restart: `54, 54, 54, 54, 54, 54, 54, 54, 54, 55, 57, 61, 62, 64 …`

It holds flat while the resumed run works through what is left, then climbs — against `64, 64, 64, 0, 3, 5, 7` for the same sequence without the flag.

Three things this caught that the unit tests did not, all fixed here:

- `patch_size` and `num_classes` arrive as numpy integers, and `json.dump` raises on those *after* opening the file — leaving a zero-byte `.tmp`, no manifest, and a run that looked fine.
- Flushing the manifest every N writes is not enough on its own: the first run was interrupted after 63 writes with the interval at 64, so nothing was ever persisted. It now flushes on 16 writes **or** 30 seconds, whichever comes first, and on the way out of Ctrl-C or an exception.
- The signature read `self.input_dir` and `self.disable_tta`, neither of which exists — the instance stores `self.input` and `self.do_tta`. There is now a test that parses `__init__` for `self.X =` and asserts the signature reads nothing else; it caught the second one immediately after the first was fixed by hand.

Related: #1241 and #1247 added `--bbox` and made an ROI run select from the full-volume grid; #1253 recorded TTA provenance in the output attrs. This is the same thread — making a long streamed run something you can trust and finish.

### DIFF
--- a/vesuvius/src/vesuvius/data/vc_dataset.py
+++ b/vesuvius/src/vesuvius/data/vc_dataset.py
@@ -498,10 +498,30 @@ class VCDataset(Dataset):
         is unchanged.
         """
         mask = getattr(self, 'non_empty_mask', None)
-        if mask is None or len(mask) != len(self.all_positions):
+        if mask is not None and len(mask) != len(self.all_positions):
+            mask = None
+        completed = getattr(self, 'completed_indices', None)
+
+        if mask is None and not completed:
             return self
-        indices = np.flatnonzero(mask).tolist()
+
+        if mask is None:
+            indices = list(range(len(self.all_positions)))
+        else:
+            indices = np.flatnonzero(mask).tolist()
+
+        if completed:
+            indices = [i for i in indices if i not in completed]
         return torch.utils.data.Subset(self, indices)
+
+    def set_completed_indices(self, completed):
+        """Exclude already-written patches from the view the DataLoader iterates.
+
+        Used by `--resume`. Kept separate from `non_empty_mask` because the two mean
+        different things: a patch can be skipped because it is empty, or because a previous
+        run already wrote it, and only the second is recoverable state.
+        """
+        self.completed_indices = set(int(i) for i in completed) if completed else set()
 
 
     def set_distributed(self, rank: int, world_size: int):
--- a/vesuvius/src/vesuvius/models/run/inference.py
+++ b/vesuvius/src/vesuvius/models/run/inference.py
@@ -911,7 +911,8 @@ class Inferer():
         prediction reproducible only under a setting no artifact mentioned.
         """
         return self._json_scalar_dict({
-            'input': str(self.input_dir),
+            # the constructor takes input_dir= but stores it as self.input
+            'input': str(self.input),
             'bbox': None if self.bbox is None else list(self.bbox),
             'patch_size': list(self.patch_size) if self.patch_size is not None else None,
             'overlap': self.overlap,
@@ -919,7 +920,8 @@ class Inferer():
             'num_parts': self.num_parts,
             'part_id': self.part_id,
             'model_path': str(self.model_path),
-            'tta': None if self.disable_tta else self.tta_type,
+            # the CLI flag is --disable_tta; the instance stores the positive form
+            'tta': self.tta_type if self.do_tta else None,
             'normalization': self.normalization_scheme,
             'num_total_patches': self.num_total_patches,
         })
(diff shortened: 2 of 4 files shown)


bruniss:
most of our inference uses s3 paths which this does not handle, and there are many areas where the resume functionality as written could easily overwrite the exact stores it's trying to resume from.

PR #1345, opened 2026-08-06
predict: add --estimate so a run's cost is knowable before it starts

TAUIL-Abd-Elilah:
> **Stacked on #1342.** This branch contains that PR's four commits as well, so until it
> merges GitHub shows both diffs here. The three commits that belong to this PR are the
> `estimate:` ones — `estimate.py`, its tests, and the flag wiring in `inference.py`.
> Reviewing #1342 first will make this one a clean three-commit diff.

## The problem

You cannot currently find out what an inference run will cost until you are inside it. On a home connection a whole-scroll pass over PHerc1218 is days, and the thing that decides how many is not the forward pass — it is how often the same chunk gets fetched. The sliding window is 192³ with half overlap while the store is chunked at 128³, so the grids do not align: one patch touches up to 3×3×3 chunks and its neighbours touch them again.

That count is exact. It is a property of two grids, not a timing, so it can be answered before committing to the run rather than inferred from a progress bar three hours in.

## What this adds

`--estimate` reports the cost and exits. It creates no output store, loads no data and computes nothing.

```
vesuvius.predict --model_path <m> --input_dir <url> --output_dir out --estimate
```

Full PHerc1218, `--no_skip_empty_patches`, `--estimate_bandwidth 11`:

```
  volume            …/PHerc1218/volumes/…-masked.zarr/0  23247x7593x7593
  chunks            128x128x128   2.0 MiB each
  patch grid        192x192x192   ->  1,510,322 patches

  distinct chunks   655,200   1.25 TiB   <- the floor: every needed chunk fetched exactly once

  traversal         cache       fetches      transfer  amplification   at 11 MB/s
  ------------------------------------------------------------------------------
  any order          none    23,401,827     44.64 TiB         35.72x       51.6 d
  current (villa)   4 GiB     2,170,800      4.14 TiB          3.31x        4.8 d
  current (villa)  16 GiB     1,708,908      3.26 TiB          2.61x        3.8 d
  chunk-blocked     4 GiB     1,731,600      3.30 TiB          2.64x        3.8 d
  chunk-blocked    16 GiB     1,543,310      2.94 TiB          2.36x        3.4 d
  morton            4 GiB       892,983      1.70 TiB          1.36x       47.3 h
  morton           16 GiB       786,446      1.50 TiB          1.20x       41.6 h
```

The first row is what this build does today. The rest are what the same patch list would cost under a cache and a different visit order, which is the work in #1177, #1327 and #1331 — so this also gives those changes a number to be judged against on a real volume rather than in simulation.

## Details that drove the design

**The patch positions are not recomputed.** They come from the `VCDataset` the real run would use, so `--bbox`, `--num_parts`, `--patch_size` and `--skip_empty_patches` are all honoured automatically and the estimate prices the run you are actually about to start. Deriving a second grid is precisely the bug #1247 fixed: an ROI-origin grid gave 27 chunks and a flattering 384× where the true full-volume grid gives 125.

**The uncached case is reported once, not once per order.** With no cache every touch is a fetch, so the traversal cannot change the count. The first version printed the same number three times, which cost 23.4M simulated touches twice over and read as though the orders had been compared and tied.

**It refuses to invent numbers.** If the input array's chunk geometry cannot be read, it reports the patch count and says the transfer cost is unavailable rather than guessing. If the array declares a compressor, it says so and labels the byte figures uncompressed, since the wire size will be smaller while the fetch counts and amplification are unaffected.

**It never reaches `_create_output_stores`.** That opens the logits array with `mode='w'`, so an estimate that fell through to it would destroy the run it was asked to price. There is a test asserting the estimate path stops before it.

## Tests

`vesuvius/tests/models/run/test_inference_estimate.py`, 24 cases: the chunk-span arithmetic including the 3×3×3 worst case, that an unbounded cache reaches the floor and a cache never costs more than none, that the distinct count is invariant under every policy, that reordering is a permutation and not a filter, Morton key bit-interleaving, that the uncached count really is order-independent, that the report names a compressor only when there is one, that `--estimate` never reaches the output store or the batch loop, and that unreadable geometry produces no byte figures.

`tests/models/run` and `tests/data` are 144 passed.

## Checking

The simulation was cross-checked against an independent implementation I had already published, on identical patch lists: exact agreement on both fetches and distinct chunks across four cache sizes and two traversal orders, 6 of 6.

Two figures above are independently corroborated by villa itself: the 655,200 distinct chunks is exactly the 182×60×60 grid that the chunk-occupancy index reports for this volume, and the 216-patch count for a 384³ ROI matches what a real run over that region produces.

One scope note, since a 30.54× for this volume is in #1325 and someone will compare them: that figure is a 1,536-thick z-slab, not the whole volume. Full volume is 15.1× deeper and both counts scale with it. The amplification is higher here because a thin slab has proportionally more z-edge patches, which are clipped and touch fewer chunks.

### DIFF
--- a/vesuvius/src/vesuvius/data/vc_dataset.py
+++ b/vesuvius/src/vesuvius/data/vc_dataset.py
@@ -498,10 +498,30 @@ class VCDataset(Dataset):
         is unchanged.
         """
         mask = getattr(self, 'non_empty_mask', None)
-        if mask is None or len(mask) != len(self.all_positions):
+        if mask is not None and len(mask) != len(self.all_positions):
+            mask = None
+        completed = getattr(self, 'completed_indices', None)
+
+        if mask is None and not completed:
             return self
-        indices = np.flatnonzero(mask).tolist()
+
+        if mask is None:
+            indices = list(range(len(self.all_positions)))
+        else:
+            indices = np.flatnonzero(mask).tolist()
+
+        if completed:
+            indices = [i for i in indices if i not in completed]
         return torch.utils.data.Subset(self, indices)
+
+    def set_completed_indices(self, completed):
+        """Exclude already-written patches from the view the DataLoader iterates.
+
+        Used by `--resume`. Kept separate from `non_empty_mask` because the two mean
+        different things: a patch can be skipped because it is empty, or because a previous
+        run already wrote it, and only the second is recoverable state.
+        """
+        self.completed_indices = set(int(i) for i in completed) if completed else set()
 
 
     def set_distributed(self, rank: int, world_size: int):
--- a/vesuvius/src/vesuvius/models/run/estimate.py
+++ b/vesuvius/src/vesuvius/models/run/estimate.py
@@ -1,0 +1,195 @@
+"""Report what an inference run will cost before it computes anything.
+
+A whole-scroll pass streams for hours, and the thing that decides how many is not the
+forward pass — it is how many times the same chunk gets fetched. villa's sliding window
+is 192^3 with half overlap while the store is chunked at 128^3, so the two grids do not
+align: one patch touches up to 3x3x3 chunks, and neighbouring patches touch them again.
+
+That re-touching is the whole cost. Counting it is exact — it is a property of the two
+grids, not a timing — so it can be reported up front, which is what this module does.
+
+The patch positions are not recomputed here. They are handed in from the dataset that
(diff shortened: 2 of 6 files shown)


bruniss:
it's not a bad idea but the included tests and orchestration around it are more complex than i feel its worth, given that these whole volume inferences are almost exclusively ran on larger clusters and the long inference time is expected.

PR #1349, opened 2026-08-06
Smoke only: stale Progress Prize PR base

giorgioangel:
Ephemeral, marker-only test fixture for the Progress Prize activation resolver.

Do not merge. This PR intentionally keeps a page-only commit whose parent is the pre-repair `main`. After the repair PR advances `main`, the live resolver must select `refresh-required=true` from the authoritative refs API. It will then be closed and its branch deleted.

No Google operation, credential, protected identifier, or production form is involved.


giorgioangel:
Smoke passed: after main advanced, the repaired resolver read the authoritative refs API and returned refresh-required=true. Closing without merge; no Google operation was involved.

PR #1353, opened 2026-08-06
Spiral: opt-in self-intersection census at preview publication

joe-carr-data:
The Spiral service publishes Lasagna-flattened previews with an atomic generation promotion, and the previous good preview already survives any pre-promotion failure. This adds an opt-in policy that reports transverse self-intersections at that boundary or withholds a preview unless the census is clean.

`--preview-selfcross off|report|reject` (default `off`, publication path untouched). In `report` mode the flattened surface is censused with `vc_tifxyz_selfcross` under both quad triangulations and the result is recorded in the preview manifest (state, per-diagonal transverse counts) and, when the census completes successfully, the JSON report and point collection are stored inside the generation, so the sites can be inspected in VC3D. In `reject` mode publication is additionally withheld unless the census is clean: tool failures fail closed, rejection evidence is written next to the published generations, and the previously published preview stays available. The tool resolves from `VC_SELFCROSS_PATH` or `PATH`; a set-but-missing override is an error rather than a silent fallback. The report is validated rather than trusted: exactly two diagonal records with non-negative integer counts, and a clean flag that agrees with them, or the census is an error and its partial diagnostics are removed. No repair logic: report or refuse, never modify.

Tests: 15 new unit tests cover all four census states, fail-closed rejection, durable evidence, byte-identical `off`, schema validation (including a clean flag contradicting its own counts), partial-diagnostic cleanup, and survival of the earlier preview across a rejection. They use fake tool executables, so no compiled binary or Lasagna run is needed. With the real binary: a known self-crossing trace censuses dirty (4 and 7 transverse contacts by diagonal) in 0.5s and a clean surface censuses clean in 0.4s. Both smoke inputs completed in under a second; these timings are integration checks, not a benchmark of full Spiral previews.

### DIFF
--- a/volume-cartographer/scripts/spiral/spiral_service.py
+++ b/volume-cartographer/scripts/spiral/spiral_service.py
@@ -704,6 +704,164 @@ def _find_lasagna_service():
         "set LASAGNA_SERVICE_PATH")
 
 
+class PreviewSelfcrossRejected(RuntimeError):
+    """A flattened preview was withheld by the self-intersection policy."""
+
+
+# A hung census must not hang publication forever: past this the run is
+# state "error", which fails closed in reject mode.
+SELFCROSS_TIMEOUT_SECONDS = 3600
+
+
+def _find_selfcross_tool():
+    """Locate vc_tifxyz_selfcross, or None when it is not installed.
+
+    VC_SELFCROSS_PATH wins when set; a set-but-missing path is an error
+    rather than a silent fallback, so a typo cannot quietly disable the
+    check. Otherwise the tool is taken from PATH.
+    """
+    configured = str(os.environ.get("VC_SELFCROSS_PATH") or "").strip()
+    if configured:
+        candidate = Path(configured).expanduser()
+        if not candidate.is_file():
+            raise RuntimeError(
+                f"VC_SELFCROSS_PATH is set but not a file: {candidate}")
+        return candidate.resolve()
+    found = shutil.which("vc_tifxyz_selfcross")
+    return Path(found).resolve() if found else None
+
+
+def _census_flattened_preview(surface_dir, output_dir, tool=None):
+    """Census a flattened tifxyz for transverse self-intersection.
+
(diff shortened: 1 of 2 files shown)


pmh47:
Do you have evidence that real spiral fits contain such intersections, and that they are not found by existing measures (high losses etc) in the interactive mode?

joe-carr-data:
Fair questions. The honest answer is: not yet for Spiral fits specifically.

I have measured a different production path. A reproduced ScrollFiesta tifxyz export contained 327 clustered transverse events under each triangulation on its canonical fixture, and 273/278 on a held-out disjoint window. That shows the failure occurs in another real unrolling pipeline; it is not evidence that it occurs in Spiral.

I have a preregistered Spiral audit waiting on GPU availability. It will use the first complete public dataset selected by the frozen rule, take up to three checkpoints from a bounded fit, flatten every selected checkpoint through the official path, and census each preview once through this PR's helper. Every attempt and failure will be reported.

One limitation is already clear: the public inputs available to this run do not include the source dense-normal or surface-SDT volumes required by those loss terms. This will therefore be a reduced-input fit, not evidence about a fully provisioned internal fit.

I also do not yet know whether the existing interactive signals already identify the same failures. Before launching, I will append a frozen analysis plan recording checkpoint-aligned total/component losses, metrics, warnings, and mapped loss maps, then compare them with the crossing locations. Could you point me to the particular interactive signal or threshold you would regard as already "finding" one? Without a pre-existing decision threshold I can measure spatial overlap, but I cannot honestly call the census redundant.

I'm happy to hold the PR until those numbers exist. If the sampled fits are clean, I will not argue for `reject` mode from this evidence; at most, `report` remains an opt-in diagnostic.

pmh47:
Closing this for now given the above discussion. Feel free to submit another PR later if/when you have an actual use-case where this has proven beneficial (i.e. you have fitted a spiral, and the tool provided useful diagnostics that allowed improving that spiral somehow).

Issue #1346, opened 2026-08-06
Remote zarr open binds OME multiscale datasets by array index, not path — scaledown artifacts (tiled_predict3d first_level>0) get silently wrong geometry over HTTP, correct locally


Bullo27:
## Summary

`remoteLevelKeysFromZattrs` (`core/src/render/ZarrChunkFetcher.cpp:541`) maps each `multiscales[0].datasets` entry to a physical pyramid level using its **position in the array** (`datasetIndex`), ignoring both the dataset `path` and its `coordinateTransformations`. For a normal pyramid (`paths "0","1","2",…`) index and path coincide, so nothing is visible. But for a multiscale whose datasets start above 0 — exactly what the fiber-inference exporter writes — the coarse array is registered as **physical level 0 with an identity transform**: the volume opens with `2^first_level`-wrong dimensions and every coordinate misregistered, silently.

The same zarr opened **locally** is handled correctly (directory scan by name → proper sparse pyramid, `hasScaleLevel(0) == false`). Same bytes, opposite geometry depending on transport.

## Their own producer creates this layout

`lasagna/tiled_predict3d.py::_create_omezarr` (line ~1085) writes:

```python
for lv in range(first_level, n_levels):
    ...
    datasets.append({
        "path": str(lv),
        "coordinateTransformations": [{"type": "scale", "scale": [float(2 ** lv)] * 3}],
    })
```

with `first_level = int(product.level)` (line ~1197) — the inference scaledown. So a scaledown-3 prediction artifact advertises `datasets: [{"path": "3", ...scale [8,8,8]}, ...]`, which the C++ reader binds as physical levels 0,1,2,… The declared scale transforms that would disambiguate are never read on this path (the *strict* base-scale reader, `strictRemoteLevelsFromZattrs`, does parse scales — the default remote open does not).

## Measured (main 8b7c9df41)

Fixture: bare zarr v2 with a single array under `2/` (shape 32×48×40, the level-2 downsample of a 128×192×160 scene) plus `.zattrs` advertising only `{"path": "2", scale [4,4,4]}` — i.e. the shape `_create_omezarr` produces for `first_level=2`, served with `python3 -m http.server`.

| open | result |
|---|---|
| **local** `vc_grow_seg_from_seed -v ./sparse.zarr` | correct sparse volume: base dims 128×192×160, only level 2 present (with #1344: clean fail-fast naming present levels) |
[...]

PR #1356, opened 2026-08-06
fix(VC3D): write the lasagna remote marker before validating the staged manifest

### LINKED ISSUE #1355: Open Data catalog Lasagna attach always fails: prepareOpenDataLasagna validates before writing the lasagna-remote.json marker that validation needs to resolve remote channels
## Summary

Opening a catalog sample with Lasagna resources currently attaches **zero** Lasagna datasets, deterministically, for every remote artifact. The prepare step has a circular dependency:

`prepareOpenDataLasagna` (`apps/VC3D/OpenDataLasagna.cpp:437`) on a fresh cache:

1. stages the manifest via `cacheRemoteFile` (line ~480) — this writes the *RemoteFileCache* sidecar, not the Lasagna marker;
2. builds the `lasagna-remote.json` marker object in memory (lines ~482-496);
3. calls **`validatePrepared(info, cached.path)` at line 497 — before the marker is written at line 498**.

Inside validation, `LasagnaDataset::open` → `loadRemoteMarker` (`core/src/lasagna/Dataset.cpp:472`) looks for `lasagna-remote.json` next to the staged manifest. It does not exist yet, so the manifest stays fully local, `resolveGroupLocations` resolves the groups' relative zarr keys against the cache directory (which holds only the 848-byte manifest), and `validateGroupDescriptor` → `openLasagnaChannelArray` fails with:

```
Failed to prepare Lasagna for 20251211183505 at coordinate level L2: zarr: cannot detect version at:
~/.VC3D/remote_cache/open_data/lasagna/PHerc0332/20251211183505/<hash>/PHerc0332-20251211183505-lasagna-20260724_cos.ome.zarr/3
(no .zarray or zarr.json found)
```

The exception is caught at line 501, so **the marker is never written**, and the warm-cache path (line 448) requires the marker — every retry repeats identically. Validation needs the marker; the marker is only written after validation passes.

## Reproduction (measured on main 8b7c9df41, via the agent bridge)

```
catalog.open_sample {"sampleId": "PHerc0332", "resources": {"kinds": ["lasagna"], "volumeIds": ["20251211183505"]}}
```

* Run 1 (cold cache): job succeeds overall, `attached: {"lasagnaDatasets": 0, ...}`, with the message above. Cache afterwards contains only `…lasagna.json` (848 B) + its `.vc-remote-file.json` sidecar — no channel data, no `lasagna-remote.json`.
* Run 2 (warm cache): identical failure — deterministic.
* **Control that isolates the ordering as the only defect:** hand-writing the correct `lasagna-remote.json` (the exact object the code builds at lines 482-496) into the cache dir and re-running the same open yields `"Attached 1 manifest-backed Lasagna dataset(s)"`, `lasagnaDatasets: 1`, and all four channel volumes (`cos`, `grad_mag`, `nx`, `ny`) attached as remote volumes. Everything downstream works once the marker exists before validation.

The parent CT volume attaches fine either way (including the rebased `#vc-base-scale=2` source view), so the sample open *looks* successful; the Lasagna failure is one line inside the job's message list.

## Scope

* The validate-before-marker shape was introduced in #1294 (Aug 1) as far as `git log -S` shows, so the catalog Lasagna path has plausibly been failing for all remote artifacts since then. The catalog currently lists Lasagna prediction artifacts for 11+ scrolls (see #1346 for the survey), so this blocks all of them in the Open Data flow. Manual/local Lasagna attach paths (`LasagnaDataset::openLocation`) are unaffected.
* Related but separate: once this ordering is fixed, the attached channel *volumes* open through the remote pyramid reader, which currently misregisters these scaledown artifacts by 2^level (#1346, fix in #1347). My verification above ran with #1347 applied locally; on plain main, fixing only the ordering would make the attach "succeed" with silently wrong geometry. The two fixes are complementary.

## Suggested fix

Write the marker before validating (move lines 498-499 above 497). Failure semantics are preserved: if validation then fails, the marker correctly describes the staged artifact and the warm path re-validates on every open, so a failing artifact keeps failing loudly with the real reason (now about actual remote channels instead of an empty local directory). I have this change building and verified end-to-end on a fresh sample and will PR it shortly.

---
*Found by driving the catalog Open Data flow headlessly through the agent bridge against the live catalog. All transcripts are from actual runs; the marker hand-write control is what isolates the ordering as the root cause. Analysis and write-up done with AI assistance (Claude).*

Bullo27:
Fixes #1355.

`prepareOpenDataLasagna` validated the freshly staged manifest **before** writing `lasagna-remote.json` — but `LasagnaDataset::open` relies on exactly that marker (`loadRemoteMarker`, `core/src/lasagna/Dataset.cpp:472`) to resolve the manifest's relative zarr keys against the remote artifact. Without it, groups resolved against the cache directory (which holds only the 848-byte manifest), validation always threw `zarr: cannot detect version … (no .zarray or zarr.json found)`, the marker was never written, and the warm-cache path requires the marker — so **catalog Lasagna attach failed deterministically for every remote artifact**, on first open and every retry.

### Change (1 file, +8/−1)

Move the marker write above `validatePrepared`, with a comment stating the dependency. Failure semantics are preserved: if validation fails after the marker exists, the marker still correctly describes the staged artifact, and the warm path re-validates on every open — genuine failures stay loud, now with the real reason (actual remote channels) instead of an empty local directory.

### Verification (all measured via the agent bridge against the live catalog)

| | before (main 8b7c9df41) | after (this branch only — no other local changes) |
|---|---|---|
| `catalog.open_sample PHerc0332` (lasagna only), cold + warm cache | `lasagnaDatasets: 0`, `Failed to prepare Lasagna … no .zarray or zarr.json found`, twice, deterministic | — |
| root-cause control | hand-writing the marker into the cache → same open attaches everything (isolates ordering as the sole defect) | — |
| `catalog.open_sample PHerc0814` (never opened, fully cold cache) | — | `lasagnaDatasets: 1`, `Attached 1 manifest-backed Lasagna dataset(s)`, all four channel volumes (`cos`, `grad_mag`, `nx`, `ny`) attached; the flow's per-level `.lasagna-zarr-metadata` staging now populates as designed |

Full ctest suite: 120/121 — the single failure is the pre-existing `test_chunk_cache` concurrency flake (#1330), also failing on unmodified main.

Heads-up as noted in #1355: with this fix, the attached channel volumes open through the remote pyramid reader, which on current main misregisters these scaledown artifacts by 2^level — that is #1346, fixed independently in #1347. The verification above intentionally ran with *only* this branch applied, so "attached" here means the flow completes; geometry correctness of the attached volumes is #1347's concern. The two are complementary and independent.

---
*Analysis, fix and verification done with AI assistance (Claude); all transcripts are from actual runs against the live catalog.*

https://claude.ai/code/session_01WKT3CfY4msFhzjd1cXzbrp


Bullo27:
Combined verification of the three interlocking fixes (#1344 + #1347 + this PR, merged into one local build — they merge cleanly with current main, no overlaps):

**Fresh catalog open of a third scroll, PHerc1451** (never opened, cold cache), via the bridge:
* `catalog.open_sample` → `lasagnaDatasets: 1`, all four channel volumes attached (`cos`, `grad_mag`, `nx`, `ny`).
* `volume.select` of the cos channel succeeds — a live in-session selection of a remote sparse production volume (`path: ""`, genuinely streamed).

**Geometry of the attached artifact, cross-checked against its S3 metadata** (`base_shape_zyx [59944, 20811, 20811]`, cos `scaledown 3`, `.zattrs` datasets `["3","4","5"]`, level-3 `.zarray` shape `[7493, 2602, 2602]`):
* level-presence guard on the cos root: `present levels: 3 4 5` ✔
* renderer through the remote opener: `zarr dataset size for group 3 [7493 2602 2602]`, `ds_level=3 ds_scale=0.125` ✔ — the level-3 array is bound at level 3 with its true shape and 1/8 scale, and a smoke render at `-g 3` completes with exit 0.

So with the three fixes together, the catalog Lasagna flow goes end-to-end on live data: attach → correct sparse registration → level-guarded tool access.

Two factual observations from the same runs, not addressed by any of the three PRs:
* The attached channel volumes report `voxelSize: 0` (the Las attach passes no explicit voxel size, so this is the known no-override `NewFromUrl` path — #1226/#1227's class). Display and coordinates are unaffected; physical-unit features on prediction volumes would need a plumbed value.
* Survey completeness for #1346: the non-`-L2` artifact generation is also suffix-sparse — e.g. `PHerc0343/.../PHerc0343_cos.ome.zarr/.zattrs` advertises datasets `["1","2","3","4"]` (2× misregistration under index binding). Every production artifact inspected so far starts above dataset `"0"`.

Bullo27:
Merged current main (9e57055) into the branch — clean merge, no conflicts (`OpenDataLasagna.cpp` hasn't been touched upstream since the fix). Full rebuild + suite on the merged tree: 150/150.

The [zero-download workflows guide](https://github.com/Bullo27/vesuvius-zero-download-guide) currently has to mark the catalog lasagna-attach path as broken-pending-this-PR (#1355 is deterministic on every remote artifact), so once this lands that section switches from a warning to a working walkthrough.

*(AI-assisted (Claude) under my direction, as disclosed in the PR body.)*

Bullo27:
Cross-reference for reviewers: #1527 (bruniss) includes this same ordering fix — `validatePrepared` moved after the `lasagna-remote.json` write in `prepareOpenDataLasagna` — and improves on it with a try/catch that removes the marker again if validation throws, so a failed attempt can't leave a stale marker behind (a case this PR left in place). The rest of #1527 (anonymous-credentials handling) is orthogonal.

If #1527 lands, I'll close this PR as superseded and re-verify #1355 against main at that point.

*(AI-assisted (Claude) under my direction, as disclosed in the PR body.)*

Bullo27:
#1527 merged (bf57b86), so closing this as superseded — the ordering fix it ships is the same reorder with the better failure handling (marker removed again if validation throws).

Re-verified #1355 against current main before closing: cold catalog open of **PHerc0125** (a sample never cached on this machine) with `resources.kinds: ["lasagna"]` now succeeds end-to-end — the job reports "Opened sample PHerc0125", all four lasagna channel volumes attach (`PHerc0125_cos/grad_mag/nx/ny.ome.zarr` in `volume.list`), and `lasagna-remote.json` is written. Every one of those steps failed deterministically before the fix.

Thanks @bruniss.

*(AI-assisted (Claude) under my direction, as disclosed in the PR body.)*

PR #1357, opened 2026-08-06
VC3D: anchor line annotation cut views to the sampled sheet normals

skyward7187:
## Problem

While navigating along a fiber in the line annotation dialog, the top-left "Current Line Cut" pane could rotate rapidly — at trace-splice kinks baked into fiber polylines its frame turned by up to ~50°/line index — and the top-right "Line Side Cut" pane intermittently mirrored left/right and occasionally tumbled through large plane pivots. Together these made tracking features by eye very difficult.

## Cause

Two independent instabilities, one per pane:

- The **current cut** derived its display "up" from the per-position transported line up vectors, which twist along the line, so the cross-section rolled while navigating. (A fixed reference direction can't fix this either: where the fiber tangent wanders — measured stretches sweep 463° of normal rotation for a net 17° — the in-plane projection of *any* fixed world direction oscillates by 8–31° of roll.)
- The **side cut** was a windowed least-squares (PCA) plane fit of the polyline. Its normal is sign-arbitrary per window (mirroring the view as the sign convention flipped between windows), and where the fiber's cross-fiber wander is nearly isotropic there is no well-defined local fiber plane at all, so the fitted normal tumbles — measured eigenvalue ratios drop to ~1.3–2.4 in the affected stretches with the normal swinging 8–28° per 2 indices.

## Fix

Derive both panes from one deterministic, physically-anchored frame: the smoothed line tangent plus the **sampled sheet normal** (the lasagna normal field the optimizer already samples per line point), sign-oriented **away from the scroll center** — via the volpkg's `umbilicus.json` when present (`estimated_umbilicus.json`, then volume XY center as fallbacks).

- **Current cut**: normal = tangent, up = oriented sheet normal → the sheet lies horizontal with the fiber's center-facing surface at the top of the view.
- **Side cut**: normal = sheetNormal × tangent, up = tangent → the fiber stays vertical and the toward-center direction points screen-left, by construction. The PCA fit machinery in the dialog is removed.
- Orientation is a pure function of the line position: jumps, coalesced mouse moves, and reopening all produce identical frames; nothing depends on navigation history.
- Where normal samples are missing or nearly parallel to the tangent (extreme bends), both panes fall back to the previous transported-line-up behavior.

## Testing

- Full suite passes (121/121 ctest targets).
- Verified interactively on `lt…000547` (trace-splice kinks that previously spun the current cut), `kb…000147` (side-view mirror flips at CP5 and plane tumbling between CP4–CP5), and `lt…000629`/`lt…000656`.

https://claude.ai/code/session_01CwMADtc1quqwJd88AoJNo8

### DIFF
--- a/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
+++ b/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
@@ -9361,6 +9361,88 @@ void LineAnnotationController::updateGeneratedViewMetricsForFiber(uint64_t fiber
     }
 }
 
+std::vector<cv::Vec3f> LineAnnotationController::orientedLineNormalsForSession(
+    const LineAnnotationSession& session)
+{
+    constexpr float kNan = std::numeric_limits<float>::quiet_NaN();
+    std::vector<cv::Vec3f> normals;
+    normals.reserve(session.optimizedLine.points.size());
+
+    std::shared_ptr<Volume> volume;
+    try {
+        volume = _state ? _state->currentVolume() : nullptr;
+    } catch (...) {
+        volume.reset();
+    }
+
+    fs::path volpkgRoot;
+    if (_state && _state->vpkg()) {
+        auto vpkg = _state->vpkg();
+        volpkgRoot = vpkg->path().empty()
+            ? fs::path(vpkg->getVolpkgDirectory())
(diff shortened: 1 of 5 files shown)


hendrikschilling:
@codex review

hendrikschilling (approved):

PR #1362, opened 2026-08-07
spiral: read resident-pool sidecars on the CPU, locally or over HTTP

aistae:
### The problem

`pack_resident_pools.py` writes a sidecar per chunk-sparse uint8 zarr array, and `lasagna_data.py` consumes it as a fully-resident sparse CUDA pool. That is the only reading path today, so looking at one of the published `lasagna_inputs` fields currently costs a CUDA device and a local copy:

- `storage_backend='sparse_cuda'` is the only supported backend, and anything else raises;
- `torch.cuda.is_available()` is asserted before the pools load;
- the sidecar must sit next to the source zarr.

The published PHercParis4 sidecars run to tens of gigabytes per channel — `las_008_surf_sdt` is about 35 GB — which is a steep price for someone who wants to check one patch against the predicted surface distance.

### What this adds

One file, `volume-cartographer/scripts/spiral/respool_read.py`, that reads the same sidecars with numpy alone:

- no CUDA, no torch;
- local sidecars via plain seeks, remote ones via HTTP range requests — a brick is 32 KiB and `channel_<i>.u8` stores bricks contiguously by pool row, so one brick is one request at offset `row * brick_voxels`;
- optional on-disk brick cache, so repeated reads are free;
- a small CLI for probing a field.

Nothing existing is touched.

### Measured

Against the published `las_008_surf_sdt`:

| read | traffic | requests |
|---|---|---|
| 900×1250 slice | 38 MB | ~1100 |
| 50×50 probe | 0.2 MB | 6 |

versus the full pool either way today.

```
$ python respool_read.py \
    https://dl.ash2txt.org/datasets/spiral_datasets/PHercParis4/lasagna_inputs/las_008_surf_sdt.ome.zarr.respool_g1 \
    --auth registeredusers:only --cache-dir /tmp/respool --z 5302 5303 --y 1450 1500 --x 2300 2350
<ResidentPool remote shape=(9473, 4087, 4087) brick=(32, 32, 32) occupancy=21.9%>
block (1, 50, 50): 100.0% non-zero, values 124..138
fetched 0.2 MB in 6 requests
```

### Tests

```
$ python -m pytest test_respool_read.py -q
9 passed
```

They build a synthetic sidecar in the layout `pack_resident_pools.py` documents — table with `0` for an absent brick, row 0 all zeros, channels stored back to back — and cover whole-array reads, sub-blocks, holes, bounds, a non-respool directory, and the cache. So they check the format rather than one particular published field, and need neither network nor GPU.

### Note on the array shape

A sidecar's `array_shape` is the shape of its *source array*, which for the published fields is a downsampled pyramid level rather than the full-resolution volume — `las_008_surf_sdt` is level 3, `las_008_nx` is level 4. The docstring says so, since mixing that up with patch coordinates is an easy mistake.

### DIFF
--- a/volume-cartographer/scripts/spiral/respool_read.py
+++ b/volume-cartographer/scripts/spiral/respool_read.py
@@ -1,0 +1,221 @@
+"""Read resident-pool sidecars on the CPU, locally or straight off a server.
+
+``pack_resident_pools.py`` writes a sidecar per (chunk-sparse) uint8 zarr
+array: ``meta.json``, ``table.npy``, ``brick_coords.npy`` and one flat
+``channel_<i>.u8`` pool. The existing consumer, ``lasagna_data.py``, loads a
+sidecar as a fully-resident sparse CUDA pool, so reading one currently costs a
+CUDA device and a local copy of the whole pool.
+
+That is a steep entry price for anyone who only wants to look at a field. The
+published PHercParis4 sidecars run to tens of gigabytes per channel, and the
+CT-support field alone is 35 GB; a contributor checking one patch against the
+predicted surface distance should not need a GPU and a full download.
+
+This module reads the same sidecars with numpy alone. A brick is 32 KiB, and
+``channel_<i>.u8`` stores bricks contiguously by pool row, so a single brick is
+one HTTP range request at offset ``row * brick_voxels``. Reading a 900x1250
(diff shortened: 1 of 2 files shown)


pmh47:
We're not going to merge this; the implementation looks ok but we don't have an internal use-case, hence maintenance burden is not justified.

PR #1366, opened 2026-08-07
VC3D: sign-chain generated view normals with one global orientation vote

skyward7187:
Follow-up to #1357.

## Problem

Panning along some fibers still produced occasional mirror flips in the line annotation cut views — e.g. `kb…000150` near (22562, 20177, 46562).

## Cause

#1357 orients each line point's sampled sheet normal away from the scroll center with an independent per-point sign test, `dot(normal, towardCenter)`. Where that test is poorly conditioned the sign flaps between adjacent points, and each stored flip renders as a mirror flip while panning.

The conditioning is much worse than intended because of a coordinate-frame mismatch: the volpkg's `umbilicus.json` is authored in a downsampled volume frame (control points span z 641–18144), while fibers live at full-res coordinates (z ~30k–68k). `Umbilicus::vector_to_umbilicus` therefore clamps every fiber point past the last control, returning a fixed bogus center whose spurious z component dominates the vector (77% of its magnitude at the reported site) — so the orientation sign ends up riding on the sheet normal's near-zero z component.

## Fix

Make local flips structurally impossible and stop depending on per-point reference quality:

- **Sign-chain along the line**: each valid normal is aligned to the previous valid one, so the stored orientation cannot flip mid-line regardless of the field or reference.
- **One global majority vote** (sum of dots over all points) decides the whole line's away-from-center sign, replacing thousands of independent per-point decisions with a single robust one.
- **Radial-only reference**: the toward-center vector's z component is zeroed — orientation is about the radial direction (z is the scroll axis), and this neutralizes the frame-mismatched umbilicus's meaningless z term. Even its mis-scaled xy direction is hemisphere-correct, which is all the majority vote needs.

Determinism from #1357 is preserved: the orientation remains a pure function of the stored line and its sampled normals.

## Testing

- Full suite passes (121/121 ctest targets).
- Verified interactively on `kb…000150` at the reported flip site, plus the fibers from #1357 (`lt…000547`, `kb…000147`) to confirm orientation and chirality are unchanged where they were already correct.

https://claude.ai/code/session_01CwMADtc1quqwJd88AoJNo8


hendrikschilling:
This will work but its unstable in two ways.

```
  - Moderate: The “majority vote” is not a majority vote. It sums normal.dot(towardCenter) without normalizing towardCenter, so points farther from the inferred center have greater weight.
    A few large-radius points can flip the entire line against most points. volume-cartographer/apps/VC3D/LineAnnotationController.cpp:9455

  - Moderate: Zeroing Z does not resolve a fully downsampled umbilicus coordinate mismatch. The invalid low-resolution XY center is still accepted, and volume-center fallback only happens 
    for non-finite results. This can replace local mirror flips with a stable but globally inverted line. volume-cartographer/apps/VC3D/LineAnnotationController.cpp:9442
```

skyward7187:
Both points addressed in 423aff83a:

- **Vote weighting**: `towardCenter` is now normalized before the dot product — the vote is cosine-weighted, so a point's influence reflects only how well its normal aligns with the radial direction; large-radius points can no longer outvote the rest of the line.
- **Umbilicus frame**: rather than defending against the mismatched reference, the umbilicus is now placed in the session frame. The volpkg's `transforms/transform.json` maps session (moving) coordinates to the registration's fixed frame — where the umbilicus was authored — so its inverse maps the control points into the session frame, and `Umbilicus::FromPoints` interpolates them natively (verified numerically: the reported flip site maps to z≈5291 in the fixed frame, ~2k voxels off the umbilicus centerline, consistent with its actual winding radius). Framing is chosen by evidence — raw vs. inverse-transformed, whichever covers more of the session volume's z-range — with a plausibility floor below which the volume-center fallback is used, so same-frame files without a transform keep working and implausible references are rejected rather than trusted.

Sign-chaining is unchanged and remains the guarantee against mid-line flips; the reference now only picks the global side, and does so from a correctly-framed center.

hendrikschilling:
Ok, first and third one should just share the loader(s). Second - review the decision process. Fourth - probably irrelevant. Fifth can probably ignore - worst case is line changes flip inside/outside decision. A test might be useful but not necessary.
 
 ```
 - High: The new loader expects parsed.at("control_points"), but the established Umbilicus::FromFile() format accepts either a root array or a "points" array. This replaces a working     
    shared parser with an incompatible ad hoc parser. Previously valid umbilicus files will throw, get swallowed by the outer catch, and silently use volume center instead. volume-        
    cartographer/apps/VC3D/LineAnnotationController.cpp:9400, volume-cartographer/core/src/Umbilicus.cpp:259

  - Moderate: Z coverage cannot reliably choose the coordinate frame. If the transform changes only XY, raw and transformed candidates have identical Z coverage, so the raw, incorrectly   
    scaled XY coordinates always win due to the strict coverage > bestCoverage comparison. volume-cartographer/apps/VC3D/LineAnnotationController.cpp:9415

  - Moderate: The transform path is hardcoded as volpkgRoot/transforms/transform.json, whereas existing VC3D transform discovery uses the current volume’s transform.json. This frame       
    correction therefore will not apply to normal volume transforms. volume-cartographer/apps/VC3D/LineAnnotationController.cpp:9434, volume-cartographer/apps/VC3D/
    SurfaceAffineTransformController.cpp:207

  - Moderate: Any malformed or noninvertible transform throws through the outer block and discards a potentially valid raw umbilicus instead of falling back to it.
  
  - Low: The near-zero aggregate vote still has no confidence threshold; an almost balanced score can flip the entire line based on tiny numerical differences.
  
  - Test gap: Still no regression tests; only the controller source changed.
```

skyward7187:
Addressed per your triage in 80fd8e1cd:

- **Loader sharing (1st)**: the controller now reads umbilici through the shared `Umbilicus` loader (exposed as `LoadControlPoints`), and `LoadJsonFile` learns the `control_points` format that stored volpkg umbilici actually use — worth noting `FromFile` previously threw on those files, so earlier code was silently running on the volume-center fallback the whole time. Regression test added for the format.
- **Decision process (2nd)**: no more raw-vs-transformed coverage contest. A discovered transform is authoritative — its inverse is applied — and z-coverage remains only as a plausibility floor. XY-only transforms therefore work correctly now.
- **Transform discovery (3rd)**: follows the established convention — the current volume's `transform.json` (native /0 coordinates, skipped for rebased views), with the volpkg-level `transforms/` file kept as a fallback location since that's where some volpkgs carry it.
- **(4th)**: covered by the restructure — transform load/apply failures now fall back to the raw points instead of throwing past them.
- **(5th)**: left as-is per your note.

PR #1368, opened 2026-08-07
VC3D: link line-annotation strips and consolidate pane HUDs

skyward7187:
## Summary

Quality-of-life changes for the line annotation window's four generated panes (current cut, side cut, and the two full-width strips).

**Linked strip navigation.** The two strip viewers now share their along-line position and zoom: panning (hover or right-drag), wheel zoom, arrow keys, and "R" snap act on both at once. Previously each strip had an independent camera, so sweeping the mouse across the border between them jumped the current line position between two unrelated locations. Vertical offset stays per-strip (the ribbons cut along different directions). Implemented as a literal `CameraState` field copy (`surfacePtrX` + `scale`) triggered from `overlaysUpdated`, with a diff check + reentrancy guard; the link survives the pane rebuild on every control-point placement.

**HUD reorganization.** With the strips linked, their corner readouts were duplicates:
- The stats group (mip level, scale) moves to the **top-right** corner and is shown once for the linked strip pair (top strip); routed via new opt-in viewer properties (`vc_status_stats_top_right`, `vc_hide_status_stats`, `vc_hide_status_position`, `vc_hide_status_poi`) — viewers that don't set them (whole main window) render as before, minus the two items below.
- `normal offset` stays **top-left** on the panes that have one, and only appears when non-zero.
- The framebuffer `WxH` item is dropped everywhere.
- The optimized/not-optimized badge moves to the **bottom-right** corner of the bottom strip.

**One shared cursor cross.** Hovering any of the four panes now shows the coordinate cross at the same volume point in the other three, with a single universal X/Y/Z readout in the current cut pane's top-left corner. The mirroring is dialog-local — independent of the global View > "Sync cursor across views" toggle, adding no work in main-window views. Crosshairs now hide on mouse leave instead of freezing at the pane edge (this also fixes the frozen-cross behavior in the main window).

**Review follow-ups** (from the findings posted below):
- Mouse leave now also broadcasts a linked-cursor clear, so *mirrored* crosshairs in other viewers can't freeze either.
- When a mirrored point lies beyond a cut plane's depth band (typically after a normal-offset scroll), the cross is drawn **greyed-out at its in-plane projection** instead of disappearing — no extra math, the projection was already computed. On the strips (quad surfaces) off-surface points still hide the cross rather than paying for an unbounded nearest-point search.
- Mirroring cost per mouse move: updates now coalesce onto a 16 ms tick (the same pattern the dialog uses for hover line-position updates), and the mirrored panes only rebuild their status bar when the value changed and they actually display the position readout (only the current cut does).

## Test plan

- [x] Builds clean
- [x] Strips pan/zoom in lockstep (hover, right-drag, wheel, arrows, "R"); vertical drag stays independent; link survives control-point placement rebuilds
- [x] HUD corners match the layout above; normal offset chip appears only after shift+scrolling off zero; right-anchored stats and badge stay pinned on window/splitter resizes
- [x] Cursor cross follows the hovered pane across all four panes and disappears on leave; universal X/Y/Z updates from any pane
- [x] Main window: readouts unchanged apart from the dropped WxH and zero-offset items; no new cursor mirroring unless the global toggle is on

Note on the earlier red CI: the "Linux VC3D compile and smoke" job died during the checkout `git fetch` ("context canceled" after ~8.5 min, before compiling anything) — infra flake, re-running on the new push.

https://claude.ai/code/session_019GoBELyqA2GY58C8fVxjF6


hendrikschilling:
Not quite complete yet. Not sure how big the performance one is in practice, but if that could be done in a separate thread that would probably be safer. Also the status bar should only be updated not rebuilt.
```
  - Moderate: Global cursor receivers are not cleared when the source leaves a main-window viewer. The leave handler clears only the source’s _lastCursorVolumePos; it never broadcasts     
    nullopt. Other viewers retain their crosshair and new _linkedCursorVolumePos, so the claimed main-window frozen-cross fix is incomplete and the XYZ readout can remain stale. volume-   
    cartographer/apps/VC3D/volume_viewers/CChunkedVolumeViewer.cpp:677, volume-cartographer/apps/VC3D/ViewerManager.cpp:2082

  - Moderate: “Always enabled” only bypasses the global toggle. setLinkedCursorVolumePoint() still rejects points more than the normal 10-voxel tolerance from a plane. Since current-cut   
    and side-cut planes are generally orthogonal, hovering away from their intersection will not display a cross in the other plane. The advertised cross across all four panes therefore   
    does not work generally. volume-cartographer/apps/VC3D/volume_viewers/CChunkedVolumeViewer.cpp:4295

  - Performance risk: Every mouse move now performs three status-bar rebuilds and projects into three other surfaces. Depending on the source pane, one or two projections call
    QuadSurface::pointTo() with up to 100 iterations. This is an unthrottled pointer-move path and could noticeably lag on large strips. volume-cartographer/apps/VC3D/
    LineAnnotationDialog.cpp:1319, volume-cartographer/apps/VC3D/volume_viewers/CChunkedVolumeViewer.cpp:3889

  - Test gap: No automated tests were added, and the PR’s four behavioral manual checks remain unchecked.
```

skyward7187:
Addressed in 34dfa5181:

- **Stale mirrored crosses on leave**: the mouse-leave handler now also broadcasts a `nullopt` linked cursor (clears were already ungated in `broadcastLinkedCursor`), so mirrored crosshairs and their readouts clear everywhere, main window included — not just the source viewer's own cross.
- **Plane tolerance / cross between orthogonal cuts**: rather than widening the tolerance (which would mark a location that isn't the hovered point), a point beyond the depth band now draws a **greyed-out crosshair at its in-plane projection** instead of hiding. No added computation — the projection was already computed for the tolerance test; the pen only swaps when the on/off-plane state changes. The quad-surface (strip) path keeps hide-on-miss, since a "nearest point anyway" would mean paying for a search beyond `pointTo`'s tolerance on every move.
- **Per-move cost**: went with throttling rather than a thread — the projections feed QGraphicsItem/QLabel updates that have to land on the GUI thread anyway. Mirror updates now coalesce onto the same 16 ms tick pattern the dialog already uses for hover line-position updates (burst of moves → one projection per non-hovered pane per tick), and `setLinkedCursorVolumePoint` only refreshes the status bar when the value changed *and* the pane actually displays the position readout — only the current cut does, so the "3 status-bar rebuilds per move" is now at most 1 per tick.
- **Test plan**: all manual checks verified and ticked in the PR body. The earlier red CI was the "Linux VC3D compile and smoke" job dying in checkout's `git fetch` ("context canceled" after ~8.5 min, before compiling anything); it's re-running on the new push.

hendrikschilling:
last one I think

```
  - Moderate: the throttled cursor update survives generated-pane rebuilds. QTimer::singleShot() is owned by the dialog, while teardown only disconnects signal connections. If panes       
    rebuild during the 16 ms window, the callback can apply the old pending point to the newly created panes and leave a stale cross/readout until another mouse move. Resetting state alone
    is insufficient; use a stoppable owned timer or generation token. LineAnnotationDialog.cpp
    (https://github.com/ScrollPrize/villa/blob/34dfa5181aa645133be2c0d80620a11f900ae498/volume-cartographer/apps/VC3D/LineAnnotationDialog.cpp#L1319)
```

skyward7187:
Fixed in 1e4624ae3 — the detached `QTimer::singleShot` is replaced with an owned single-shot timer (same shape as the existing `_lineUpdateTimer`). It's stopped, together with the pending point/source state and the pane list, in `clearGeneratedOverlayRefreshConnections()`, which every teardown path (rebuild, MDI reset, close) already runs before new panes exist — so a pending mirror can no longer fire across a rebuild and stamp a pre-rebuild point onto the new panes.

PR #1371, opened 2026-08-07
ink-detection: stop dividing by the clip value in optimized inference

nerln:
`optimized_inference` divides by the clip value instead of 255.

`inference.py:278` clips each tile to `CFG.max_clip_value` (200), then `:353` normalises with `A.ToFloat(max_value=CFG.max_clip_value)`, dividing by 200 again.

Every other path divides by 255. Training and the reference inference scripts use `A.Normalize(mean=[0]*n, std=[1]*n)` with no explicit `max_pixel_value`, so albumentations applies its default of 255: `train_timesformer_og.py:91`, `train_timesformer_deduped.py:152`, `inference_timesformer.py:156`, `inference_resnet3d.py:307`, `train_resnet3d_lib/config.py:460`.

So this path feeds the model inputs scaled 255/200 = **1.275x** larger than anything it was trained on.

## The change

One line: normalise by 255 through a new `CFG.normalize_divisor`, kept separate from the clip so the two cannot drift together again.

## Test

`tests/test_preprocessing_parity.py` pins the divisor against the reference paths. On the old constant it fails by 0.2157 — a clipped pixel arrives as 1.0 where the reference gives 200/255 = 0.7843. It reads the constants out of the source with `ast`, so it needs neither torch nor zarr.

## Consequence to weigh

**This changes inference output for anyone running this path.** Whether any published map came from here rather than from `inference_timesformer.py` is not something I can tell from outside, so whether anything needs re-running is your call.

Unrelated, spotted while testing: `optimized_inference/__init__.py:7` imports `.inference_timesformer`, which does not exist in that package, so `import optimized_inference` fails outright. Happy to open that separately.

### DIFF
--- a/ink-detection/optimized_inference/inference.py
+++ b/ink-detection/optimized_inference/inference.py
@@ -100,6 +100,11 @@ class InferenceConfig:
 
     # Image processing / scaling
     max_clip_value: int = 200
+    # Divisor applied after the clip. Must match training, which uses
+    # A.Normalize(mean=0, std=1) and therefore albumentations' default
+    # max_pixel_value of 255. Tying it to max_clip_value instead fed the model
+    # inputs 255/200 = 1.275x larger than anything it saw while training.
+    normalize_divisor: float = 255.0
 
     # Blending
     use_hann_window: bool = True
@@ -350,7 +355,7 @@ def create_inference_dataloader(
         if CFG.tile_size != CFG.size:
             tfm_list.append(A.Resize(CFG.size, CFG.size))
         tfm_list += [
-            A.ToFloat(max_value=CFG.max_clip_value),
+            A.ToFloat(max_value=CFG.normalize_divisor),
             ToTensorV2(),
         ]
         transform = A.Compose(tfm_list)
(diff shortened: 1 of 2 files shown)


bruniss:
the older ink detection is being deprecated. closing

nerln:
Understood on the deprecation, and #1374/#1375 were rightly closed. This one file may be a different case, so one question before it stays closed.

`optimized_inference` is the only thing #1486 left under `ink-detection/`; everything else moved to `deprecated/`. #1486 also touched it once, rewriting `__init__.py` to drop the broken `from .inference_timesformer import ...`, which suggests it was kept on purpose rather than missed.

If it is still serving traffic, the line matters. It was correct in the original container (#360): `A.Normalize(mean=[0]*n, std=[1]*n)`, i.e. /255, matching training. `c983f1ff` in #462 changed it to `max_pixel_value=CFG.max_clip_value`; that PR's description lists its changes and normalisation is not among them, so it looks like an unintended side effect of the performance work rather than a decision. Measured on a clipped pixel at 200, with albumentations resolved today:

```
reference paths (clip 200 -> A.Normalize)      0.7843
optimized_inference (clip 200 -> ToFloat(200)) 1.0000   -> 1.275x
```

`model.eval()` freezes the optional BatchNorm, so nothing downstream absorbs it.

If the container is being retired anyway, or if the served checkpoints were recalibrated for /200 after Sept 2025 — which I cannot see from outside — then this is noise and the PR should stay closed. Happy either way.

PR #1374, opened 2026-08-08
ink-detection: weight the blend denominator with the same kernel as the numerator

nerln:
`get_img_splits` weights the overlap-add numerator with a gaussian and the denominator with ones, then divides one by the other:

```python
kernel = gkern(CFG.size, 1) / gkern(CFG.size, 1).max()
...
mask_pred[y1:y2, x1:x2]  += y_preds_multiplied_cpu[i]   # carries the kernel
mask_count[y1:y2, x1:x2] += mask_count_kernel           # np.ones
...
mask_pred /= np.clip(mask_count, a_min=1, a_max=None)
```

That is not a weighted average.

## What it costs

The kernel is normalised to max 1, so its mean is 0.732: the interior is attenuated by **27%**, independent of tile size and stride. The ripple is not: the kernel is 1.0 at each tile centre and 0.38 at its edge, so the error carries a periodic pattern at the tile-grid frequency that a downstream threshold cannot absorb. On a synthetic constant field of 0.700 at `CFG.size` 64 and this file's default stride of 32, the current code returns 0.5126 with a peak-to-peak ripple of **0.088**.

## The change

Accumulate `kernel` on both sides, drop the now-unused `np.ones` constant, and divide with `where=mask_count > 0` instead of clipping the denominator to 1. That last part matters once the denominator sums weights rather than counts: a floor of 1 flattens thin-coverage borders. Before and after on a border pixel: 0.4119 → 0.7000.

`optimized_inference/inference.py` already does this correctly — both its kernels are normalised to sum 1 and the same one is accumulated on both sides. Comparing the two paths is what made this visible.

## Consequence to weigh

**This changes inference output for anyone running this path**: predictions get brighter and lose the grid-frequency ripple. Whether any published map came from here is your call.

#1375 fixes the same defect in the two resnet3d inference scripts. Happy to squash them if one PR is easier to review.

### DIFF
--- a/ink-detection/inference_timesformer.py
+++ b/ink-detection/inference_timesformer.py
@@ -228,7 +228,6 @@ class RegressionPLModel(pl.LightningModule):
 def predict_fn(test_loader, model, device, test_xyxys, pred_shape):
     mask_pred = np.zeros(pred_shape)
     mask_count = np.zeros(pred_shape)
-    mask_count_kernel = np.ones((CFG.size, CFG.size))
     kernel = gkern(CFG.size, 1)
     kernel = kernel / kernel.max()
     model.eval()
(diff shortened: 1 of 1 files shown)


erdpx:
hi, thanks for the pr! we're closing this now because we're deprecating the legacy standalone ink-detection pipeline in favor of integrating the new pipeline into villa/vesuvius ([#1486](https://github.com/ScrollPrize/villa/pull/1486))

PR #1376, opened 2026-08-08
VC3D: consolidate the line-annotation toolbar into a hamburger menu

skyward7187:
Tidies the line-annotation workspace toolbar: the scattered buttons and spinboxes collapse into one hamburger menu, and a new shortcut resets the panes' normal offsets.

## Toolbar

Before, the top bar carried an `Annotation` menu button, a `Datasets` menu button, a `Reset views` push button, and the `Length` / `Extrapolation` spinboxes. All of that now lives behind a single `☰` button (tooltip `Menu`):

- **Auto-reoptimize**
- **Reinit reoptimization**, **Show as mesh**
- **Lasagna dataset ▸**, **Fiber dataset ▸** — the former `Datasets` button. The old button-level disable is dropped; the submenus already show a disabled `No … datasets attached` placeholder in that case, and per-dataset location tooltips now actually display (`setToolTipsVisible`).
- **Length** and **Extrapolation** spinbox rows (see below)
- **Reset views**

What remains in the top bar: `☰`, the volume selector (moved up next to the menu button), the Lasagna/Fiber-model selector, `Max CP dist`, the tag pills, and the fiber name.

## Length / Extrapolation commit on Apply

Both spinboxes are embedded in the menu as `QWidgetAction` rows, so editing them does not close the menu. Each row has an **Apply** button, and nothing takes effect until it is clicked.

This exists because `Extrapolation` emits `extrapolationDistanceChanged`, which re-grows the line tails: with a plain spinbox, every arrow click on the way to the target value kicked off another optimization. Committing explicitly means one optimization per adjustment.

Consequences worth knowing while reviewing:

- `initialCenterlineLengthVx()` / `extrapolationDistanceVx()` now return the **applied** values, not the live spinbox values. The controller reads these at seed and optimization time, so without that indirection an uncommitted edit would leak into the next run.
- Apply is enabled only when the spinbox differs from the applied value, so it doubles as an "uncommitted edit" indicator.
- Reopening the menu reverts abandoned edits, so the display never disagrees with what is in effect.

## New `B` shortcut: reset normal offsets

`B` zeroes every accumulated normal offset in the generated views — the side cut plane's offset (reusing the existing `resetGeneratedCutNormalOffsets` path) and both strips' surface offsets (`adjustSurfaceOffset(-normalOffset())`). The current cut cannot accumulate one, since Shift-scroll steps along the line there, but it is reset alongside the side cut for symmetry.

It is bound both as a window `QShortcut` and in `handleKeyPress` — the same dual path `Space` already uses — because the `QShortcut` alone did not fire reliably with focus inside a viewer pane.

## Testing

Built `RelWithDebInfo` and exercised interactively in VC3D: menu contents and ordering, dataset submenu selection, Apply-gated spinboxes (including abandoning an edit and reopening the menu), and `B` after shift-scrolling offsets into the side cut and both strips. No automated tests cover this toolbar code; `docs/line_annotation_fibers.md` is updated to match.

https://claude.ai/code/session_01PyYKz7UdCVAP1VnduAzKjq

### DIFF
--- a/volume-cartographer/apps/VC3D/Keybinds.cpp
+++ b/volume-cartographer/apps/VC3D/Keybinds.cpp
@@ -18,6 +18,11 @@ constexpr const char* kSectionPushPull = "Push/Pull";
 constexpr const char* kSectionPointCollection = "Point Collection";
 constexpr const char* kSectionMouseControls = "Mouse Controls";
 constexpr const char* kSectionSliceStep = "Z-Scroll Sensitivity";
+// The line annotation workspace owns its panes' keyboard handling, so these
+// keys shadow the global bindings above while that workspace has focus. They
+// are listed as literals because the dialog binds them directly rather than
+// through the ShortcutDef/KeyPressDef registry.
+constexpr const char* kSectionLineAnnotation = "Line Annotation Workspace (overrides global keys)";
 
 enum class HelpKeyType {
     Shortcut,
@@ -561,6 +566,16 @@ QString buildKeybindsHelpText()
 
         { kSectionSliceStep, keypress::SliceStepDecrease.description, HelpKeyType::KeyPress, nullptr, &keypress::SliceStepDecrease, nullptr },
         { kSectionSliceStep, keypress::SliceStepIncrease.description, HelpKeyType::KeyPress, nullptr, &keypress::SliceStepIncrease, nullptr },
+
+        { kSectionLineAnnotation, "Rotate the current cut about the horizontal axis", HelpKeyType::Literal, nullptr, nullptr, "W / S" },
+        { kSectionLineAnnotation, "Rotate the current cut about the vertical axis", HelpKeyType::Literal, nullptr, nullptr, "A / D" },
+        { kSectionLineAnnotation, "Jump to the previous control point", HelpKeyType::Literal, nullptr, nullptr, "E" },
+        { kSectionLineAnnotation, "Jump to the next control point", HelpKeyType::Literal, nullptr, nullptr, "T" },
+        { kSectionLineAnnotation, "Snap the panes to the overview-bar cursor", HelpKeyType::Literal, nullptr, nullptr, "R" },
+        { kSectionLineAnnotation, "Reset the side cut and strip normal offsets", HelpKeyType::Literal, nullptr, nullptr, "B" },
+        { kSectionLineAnnotation, "Toggle the current cut following the strip mouse", HelpKeyType::Literal, nullptr, nullptr, "Space" },
+        { kSectionLineAnnotation, "Close the workspace", HelpKeyType::Literal, nullptr, nullptr, "Esc" },
+        { kSectionLineAnnotation, "Step along the line (current cut) or the normal offset (side cut, strips)", HelpKeyType::Literal, nullptr, nullptr, "Shift+Scroll Wheel" },
     };
 
     QString result;
--- a/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
+++ b/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
@@ -18,6 +18,7 @@
 #include "vc/core/types/VolumePkg.hpp"
 #include "vc/core/types/Volume.hpp"
 #include "vc/core/types/Segmentation.hpp"
+#include "vc/core/util/AffineTransform.hpp"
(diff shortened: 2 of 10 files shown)


hendrikschilling:
```

  1. Medium: persisted values bypass the spinbox bounds.
     The applied values are read directly from QSettings, then separately passed through QSpinBox::setValue(). Qt clamps the displayed value, but initialCenterlineLengthVx() and
     extrapolationDistanceVx() return the original, unclamped setting. A stale or manually edited value can therefore show 100 while the optimizer receives 0, or exceed the intended       
     1,000,000 maximum. Previously, getters returned the clamped spinbox value.
     LineAnnotationDialog.cpp (https://github.com/ScrollPrize/villa/blob/31ee35f68558ce19d1a12aa6b0abb5a3ffc4e328/volume-cartographer/apps/VC3D/LineAnnotationDialog.cpp#L579)
     Fix by clamping the loaded values or assigning _applied... from each spinbox after setValue().

  2. Medium: B conflicts with VC3D’s existing approval-paint shortcut.
     B is already registered as ApprovalPaintToggle, while this dialog installs both a window shortcut and event-filter handling that consumes the same unmodified key in generated panes.  
     The central shortcut help will still advertise approval painting, but the line-annotation workspace resets offsets instead.
     LineAnnotationDialog.cpp (https://github.com/ScrollPrize/villa/blob/31ee35f68558ce19d1a12aa6b0abb5a3ffc4e328/volume-cartographer/apps/VC3D/LineAnnotationDialog.cpp#L2681)
     Keybinds.cpp (https://github.com/ScrollPrize/villa/blob/31ee35f68558ce19d1a12aa6b0abb5a3ffc4e328/volume-cartographer/apps/VC3D/Keybinds.cpp#L233)
     This should use a non-conflicting key or be represented through the centralized keybinding system with explicit workspace context.

```

skyward7187:
Thanks — both fixed in 4db73feb9.

**1. Bounds bypass.** Real regression, good catch. The applied values are now read back off each spinbox right after `setValue()`, so the spinbox range is the single source of truth and there's no second copy of the bounds to drift. Worth noting the failure was slightly worse than a display mismatch: because Apply is enabled only when the spinbox differs from the applied value, a stale setting of e.g. `0` for Length showed `100`, sent `0` to the optimizer, and left Apply *disabled* — so there was no way to correct it from the UI without first nudging the value.

I deliberately did not write the clamped value back to QSettings. A bad stored value now clamps identically on every load and is overwritten the first time Apply is clicked, which seemed preferable to a settings write as a side effect of constructing the dialog. Happy to add it if you'd rather it self-heal.

**2. `B` conflict.** You're right that the help was advertising the wrong thing, so I've documented it — but while checking I found `B` isn't the exception, it's the rule. Cross-referencing every key this dialog binds against `Keybinds.cpp`:

| Workspace key | Already registered globally as |
|---|---|
| Space | `ToggleVolumeOverlay` |
| A / D | `PushPullIn` / `PushPullOut` |
| S | `LineDrawHold` |
| E | `PushPullRadiusUp` |
| T | `ToggleAnnotation` |
| Esc | `CancelOperation` |
| R | `CenterFocusOnCursor` |
| **B** | **`ApprovalPaintToggle`** ← new in this PR |

Seven of the eight pre-existing bindings already shadow global unmodified keys, and none of them appear in the central help; `W` is the only one that was free. So rather than move just `B`, I added a **"Line Annotation Workspace (overrides global keys)"** section documenting all nine keys plus the Shift+Scroll behaviour, which differs per pane and was also undocumented.

The conflict looks cosmetic rather than functional, for what it's worth: approval painting dispatches from `CWindow::keyPressEvent` for segmentation viewers, these panes are a separate `ViewerRole::Annotation` workspace, the dialog accepts the event before it can bubble up, and Qt disables the dialog's window-scoped shortcuts while its tab is hidden.

**Question on how far to take this:** I used `HelpKeyType::Literal` entries, matching the existing Mouse Controls block, because the dialog binds these keys directly rather than through the registry. The more thorough reading of your comment would be to convert all nine into real `KeyPressDef` entries and have the dialog consume those, so the definitions are centralized rather than just described. That means reworking `installGeneratedViewShortcuts` and the dialog's `handleKeyPress` for every key, which felt beyond a toolbar PR — but I'm happy to do it here if you'd prefer, or to split it into a follow-up that also covers the seven pre-existing bindings. Which would you rather?

PR #1378, opened 2026-08-08
VC3D: add trust-safe unverified TIFXYZ hints to Spiral

TAUIL-Abd-Elilah:
## Summary

Add an explicit low-trust path for bringing external TIFXYZ geometry, including
ScrollFiesta exports, into a running Spiral fit without promoting it to verified
geometry.

- add **Add unverified hint...** to the running-fit panel and **Add as
  unverified Spiral hint** to local surfaces
- carry an optional patch `classification` through upload, status, resident
  incorporation, and commit; omitted classification remains `verified`
- keep unverified hints in Spiral's existing unverified atlas/losses, outside
  verified geometry and interactive influence anchors
- mask hints near trusted verified-patch/PCL geometry, including trusted inputs
  added in the same pause boundary, and remask resident hints when trust grows
- commit hints only to `unverified_patches/` and reject identifiers colliding
  with either trust pool
- advertise a service capability so updated VC3D fails closed rather than
  sending an unverified hint to an older service

## Motivation

This is the next integration step after [#1155](https://github.com/ScrollPrize/villa/pull/1155),
whose opening rationale records that the Vesuvius team asked for ScrollFiesta in
VC3D and first added the native Windows build path needed to support it.

VC3D and Spiral already distinguish verified and unverified patch pools, but
the interactive upload path treated every added patch as verified. That makes
direct integration of automatically generated geometry unsafe: useful hints
could not enter a resident fit without silently crossing the trust boundary.

This change connects the existing low-trust model to the interactive workflow
while preserving the historical verified behavior for old clients and ordinary
surface uploads.

## Trust and compatibility guarantees

- only exact `verified` / `unverified` classifications are accepted
- old clients omit the field and retain the verified default
- verified VC3D uploads retain the legacy request shape
- unverified VC3D uploads require the advertised `patch_classification`
  capability
- source identifiers remain reserved even if erosion, ROI filtering, or later
  trusted masking removes a patch from the resident training pool
- committed unverified inputs retain a pristine session source for safe
  remasking; their bytes remain charged to the session quota
- no ScrollFiesta runtime or dependency is bundled

## Validation

- 15 focused protocol, trust-routing, trusted-PCL, compatibility, collision,
  commit, and source-retention tests pass
- all 96 platform-applicable tests in `test_spiral_service_v2.py` pass locally;
  the five remaining local failures are existing Windows/POSIX-semantics tests
- compiled ScrollFiesta `scroll_unroll --selftest` passes with zero failures,
  including TIFXYZ export
- Villa parses the compiled-exporter fixture as `(21, 21, 3)`, 441 valid
  vertices, area 400.0
- the service accepts all seven exporter files as unverified, publishes only
  under `unverified_patches/`, and preserves ScrollFiesta provenance
- [native CI](https://github.com/TAUIL-Abd-Elilah/villa/actions/runs/31282883258):
  Linux VC3D compile/offscreen smoke, macOS compile, C++ tests, and compile
  shards pass
- [Windows CI](https://github.com/TAUIL-Abd-Elilah/villa/actions/runs/31282883267):
  MSYS2 configure/build and the Windows MCP support test pass; release
  packaging and packaged-tree smoke are intentionally skipped for PR events
- [CodeQL](https://github.com/TAUIL-Abd-Elilah/villa/actions/runs/31282883305):
  every changed controller/component/core/tools shard passes

### DIFF
--- a/volume-cartographer/apps/VC3D/CWindow.cpp
+++ b/volume-cartographer/apps/VC3D/CWindow.cpp
@@ -2575,7 +2575,12 @@ CWindow::CWindow(size_t cacheSizeGB, RenderBenchOptions benchOptions) :
         // The "Add to current spiral fit" context actions are greyed out
         // unless a Spiral session is active on the connected service.
         connect(_spiralWorkspace, &SpiralWorkspace::spiralSessionActiveChanged, this, [this](bool active) {
-            if (_surfacePanel) _surfacePanel->setSpiralFitAvailable(active);
+            if (_surfacePanel) {
+                _surfacePanel->setSpiralFitAvailable(active);
+                _surfacePanel->setUnverifiedSpiralHintAvailable(
+                    active && _spiralWorkspace
+                    && _spiralWorkspace->supportsUnverifiedPatchUploads());
+            }
             if (_fiberWidget) _fiberWidget->setSpiralFitAvailable(active);
         });
     }
@@ -6984,7 +6989,7 @@ void CWindow::CreateWidgets(void)
                 showStatusBarMessage(tr("Copied segment path to clipboard: %1").arg(path), 3000);
             });
     connect(_surfacePanel.get(), &SurfacePanelController::addSurfaceToSpiralFitRequested,
-            this, [this](const QString& segmentId) {
+            this, [this](const QString& segmentId, bool unverified) {
                 if (!_spiralWorkspace) {
                     return;
                 }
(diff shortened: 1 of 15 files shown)


pmh47:
I don't see a real use-case -- interactively added constraints are typically human-checked, and thus go in verified (as now). And automatically-generated maybe-imperfect constraints tend to come in large quantities so adding interactively doesn't make sense. @bruniss close if you agree

bruniss:
yeah same interpretation. we have also talked about reworking patches/pointsets anyways in the near future to be represented the same in fit_spiral and this would add complexity to that effort.

TAUIL-Abd-Elilah:
Understood—thank you both. The live interactive path is the wrong abstraction for automatically generated batches, and I had not accounted for the planned patch/point-set unification. I'll leave this closed.

PR #1380, opened 2026-08-09
Phantom eval harness

beaugrzanich:
## A ground-truth evaluation harness for the spiral fit — with phantom↔real correlation evidence

The spiral fit is currently evaluated by self-consistency (agreement with its own input constraints) and an indirect ink-coverage proxy; neither compares against ground truth, because real scrolls have none. This adds a synthetic benchmark that does, plus the evidence that its scores track real-scroll performance.

All new files under `volume-cartographer/scripts/spiral/`; **no production code is touched**. Tests pass (25 cases), environment via the existing `uv.lock`.

### What it does

- **`synth_phantom.py`** — generates a scroll-like volume by sampling a *known* deformation through the production `SpiralAndTransform` stack, yielding exact per-voxel winding ground truth. Staged realism: intensity/texture calibrated to a real scan (`--preset pherc0009b`), and **out-of-family slip dislocations** (`--num-slips`) — discontinuous, exactly invertible, provably outside the fitter's smooth transform family.
- **`winding_error.py`** — scores any candidate against phantom truth (winding MAE/RMSE, sheet-switch rate, per-winding + torn-region breakdowns) with gauge alignment and branch-cut handling. Accepts fit checkpoints, plain winding volumes, **tifxyz surfaces** (the format spiralcheck consumes), **and — via `--truth-npz` — an external `finite_paint` `turn_id` volume as ground truth** (Diego-dcv's grid repo), making the ruler source-agnostic; verified against real `finite_paint` output, not just a mock.
- **`export_phantom_dataset.py`** — writes a phantom as a `fit_spiral`-consumable dataset, with dials for label quantity/quality (`--coverage`, `--position-noise`, `--index-noise-rate`).
- **`labels_quantity_vs_precision.py`** — uses the harness to answer an open-problems question against exact truth: sweeps annotation coverage × label position-noise and finds, for the reference fitter, that **coverage has more leverage than precision** (abundant-but-noisy labels beat sparse-but-perfect ones).
- **`fit_phantom_reference.py`** — a CPU reference fitter that closes the phantom→dataset→fit→truth-scored loop end to end (production `fit_spiral` is CUDA-only; this is a clearly-labeled baseline, not a reimplementation).
- **`measure_real_scroll.py`** — streams a small crop from an open-data OME-Zarr volume and measures the stats the phantom is calibrated against.
- **`trace_real_windings.py` / `unroll_real_fit.py`** — a reproducible classical (no-ML) trace + reroll + unroll of a real scroll on a laptop.
- **`phantom_real_correlation.py`** — the study below.

### Correlation evidence (the bar a prior close named)

12 fitter variants (good → catastrophic), scored on the phantom (MAE vs exact truth) and on real scrolls (MAE on held-out angular sectors of classically-traced windings, identical mask both sides). Spearman rank correlation:

| Phantom realism | ρ | p |
|---|---|---|
| in-family, clean labels | 0.31 | 0.32 |
| + out-of-family slips | 0.40 | 0.20 |
| + deformation calibrated to real fit + contradictory labels | **0.72** | **0.008** |

Each realism gap closed moved the number. **Cold cross-scroll:** freezing the PHerc0172 calibration and testing on **PHerc0332** (never seen) gives **ρ = 0.80 (p = 0.002)** — transfer without recalibration. The two real scrolls rank the 12 methods at **ρ = 0.95** with each other, so method quality is largely scroll-independent.

### Honest scope

n = 12 variants, one fitter **family** (the CPU reference fitter, not production `fit_spiral`), two scrolls, a classical-trace real reference with a ~0.6-winding label floor. The benchmark reliably separates *good from catastrophic*; it does not yet finely rank *within* the good cluster. Phantoms currently export patches + winding annotations only — not tracks/fibers/normals/SDT — so running the production fitter through this is the clear next step (and the exporter makes it a one-command job on a GPU host).

### Reproduce

```sh
# --- setup + tests (25 cases) ---
cd volume-cartographer/scripts/spiral && uv sync
uv run python -m pytest tests/test_synth_phantom.py tests/test_export_phantom_dataset.py \
    tests/test_consume_finite_paint.py
```

All remaining commands run from `volume-cartographer/scripts/spiral/`.

```sh
# --- real-scroll reroll demo: trace -> fit -> unroll a real scroll (PHerc0172, laptop) ---
uv run python trace_real_windings.py --out /tmp/p172_ds --slab-cache /tmp/p172_slab.npy
uv run python fit_phantom_reference.py --dataset /tmp/p172_ds --out /tmp/p172_fit.pt \
    --steps 1200 --reg-gap 0.03 --reg-flow 1.0 --reg-flow-smooth 100
uv run python unroll_real_fit.py --checkpoint /tmp/p172_fit.pt --dataset /tmp/p172_ds \
    --slab-cache /tmp/p172_slab.npy --out-prefix /tmp/p172   # writes p172_overlay.png, p172_unrolled.png
```

```sh
# --- correlation study on the calibration scroll (rho=0.72, PHerc0172) ---
for Z in 883 1405 1927; do
    uv run python trace_real_windings.py --out work/ds_z$Z --z0 $Z --slab-cache work/slab_z$Z.npy
done
uv run python phantom_real_correlation.py --work work \
    --real-dataset work/ds_z883 --real-dataset work/ds_z1405 --real-dataset work/ds_z1927 \
    --phantom-slips 3 --phantom-index-noise 0.2 \
    --phantom-flow-std 0.0137 --phantom-gap-log-std 0.107 --phantom-linear-std 0.0046
```

```sh
# --- cold cross-scroll transfer (rho=0.80, frozen calibration, PHerc0332 never seen) ---
V=https://vesuvius-challenge-open-data.s3.us-east-1.amazonaws.com/PHerc0332/volumes/20251211183505-2.399um-0.2m-78keV-masked.zarr
for Z in 700 1050 1400; do
    uv run python trace_real_windings.py --volume-url "$V" --level 4 --z0 $Z --r-max 360 \
        --out x332/ds_z$Z --slab-cache x332/slab_z$Z.npy
done
# freeze the calibration: seed x332/results.jsonl with ONLY the phantom (kind=="phantom") rows
# from the study above, so phantoms are reused not recalibrated, then:
uv run python phantom_real_correlation.py --work x332 \
    --real-dataset x332/ds_z700 --real-dataset x332/ds_z1050 --real-dataset x332/ds_z1400 \
    --phantom-slips 3 --phantom-index-noise 0.2 \
    --phantom-flow-std 0.0137 --phantom-gap-log-std 0.107 --phantom-linear-std 0.0046
```

```sh
# --- label quantity-vs-precision study (coverage beats precision) ---
uv run python labels_quantity_vs_precision.py --work work_labels

# --- measure a real scroll's stats (what the phantom preset is calibrated against) ---
uv run python measure_real_scroll.py --out-png real_scroll.png

# --- source-agnostic ruler: score a candidate winding TIFF against finite_paint turn_id truth ---
# (needs a finite_paint .npz from Diego-dcv/vesuvius-topological-grid, external to this repo)
uv run python winding_error.py --truth-npz twin_cell.npz --candidate-winding-volume candidate.tif
```

The "what makes a rung discriminate" design rationale (misleading-local vs missing/global) is recorded in `phantom_real_correlation.py`'s docstring.

cross_scroll
correlation_rounds
p172_before_after


Jinhojeong:
@beaugrzanich, the correlation ladder is the part I would call the contribution. Each realism gap you closed moved rho, which turns "is the phantom realistic enough" from taste into a measurement, and the frozen-calibration cross-scroll transfer is the right bar to have set.

One rung looks absent from the ladder, and the real labels say it is the dominant one. Sheet fusion. On IyanDopico's PHerc1218 whole-scroll labels about seven in ten thick runs carry a single instance id, and repairing exactly that defect moved the scroll-wide fused share from 67.65 to 65.54 percent on a frozen evaluation set. So contact and fusion are not a tail case in real material, and a fitter that never sees fused sheets in the phantom is never tested on the thing that makes real winding assignment hard. Your own ladder suggests the direction, since the rung that added calibrated deformation and contradictory labels moved rho from 0.40 to 0.72, and fusion is exactly the label pathology that rung abstracts. Measured ingredients for a fusion rung exist. A preregistered pilot on Diego-dcv's phantom grid line, three-handed with aviad12g, put ground-truth numbers on it. Detected contact fuses at 74.8 to 78.0 percent regardless of gap size across the 10 to 150 um that grid spans. His repo also carries a weld-capable painter with instance labels, and the flag and split instruments in the same line of work were validated on exact synthetic truth. All of it is public, and I can point at the specific artifacts if you want the rung.

One precedent worth a unit test. A phantom grid in the villa #191 fusion work shipped with sheets painted at zero thickness, the numbers that depended on material thickness were invalid, and the failure was invisible until someone checked the voxels. Diego's repo added an acceptance exam that fails exactly that defect after the fact. Your preset path already half-covers this, since the texture test asserts the sheet std on the calibrated preset. The default render path, which the rest of the test suite uses, carries no assertion that sheets hold signal at all, and a one-assert check there would make that class of silent failure impossible here.

beaugrzanich:
> @beaugrzanich, the correlation ladder is the part I would call the contribution. Each realism gap you closed moved rho, which turns "is the phantom realistic enough" from taste into a measurement, and the frozen-calibration cross-scroll transfer is the right bar to have set.
> 
> One rung looks absent from the ladder, and the real labels say it is the dominant one. Sheet fusion. On IyanDopico's PHerc1218 whole-scroll labels about seven in ten thick runs carry a single instance id, and repairing exactly that defect moved the scroll-wide fused share from 67.65 to 65.54 percent on a frozen evaluation set. So contact and fusion are not a tail case in real material, and a fitter that never sees fused sheets in the phantom is never tested on the thing that makes real winding assignment hard. Your own ladder suggests the direction, since the rung that added calibrated deformation and contradictory labels moved rho from 0.40 to 0.72, and fusion is exactly the label pathology that rung abstracts. Measured ingredients for a fusion rung exist. A preregistered pilot on Diego-dcv's phantom grid line, three-handed with aviad12g, put ground-truth numbers on it. Detected contact fuses at 74.8 to 78.0 percent regardless of gap size across the 10 to 150 um that grid spans. His repo also carries a weld-capable painter with instance labels, and the flag and split instruments in the same line of work were validated on exact synthetic truth. All of it is public, and I can point at the specific artifacts if you want the rung.
> 
> One precedent worth a unit test. A phantom grid in the villa #191 fusion work shipped with sheets painted at zero thickness, the numbers that depended on material thickness were invalid, and the failure was invisible until someone checked the voxels. Diego's repo added an acceptance exam that fails exactly that defect after the fact. Your preset path already half-covers this, since the texture test asserts the sheet std on the calibrated preset. The default render path, which the rest of the test suite uses, carries no assertion that sheets hold signal at all, and a one-assert check there would make that class of silent failure impossible here.

Thanks! That's the most useful kind of pushback. Two things:

The signal guard is in 37ea097: the default render path now asserts sheets hold contrast, so the zero-thickness class fails loudly here. You were right that only the preset path covered it — verified the assertion catches a degenerate render (annulus p95–p5 = 26) and passes real ones (~150).

The fusion rung: yes, I want it, and please point me at the artifacts. You've convinced me it's the dominant rung: the ±1 index-noise in my last step is a crude stand-in for exactly this, and fusion is the direct pathology. Building it on Diego-dcv's weld-capable painter plus your measured fuse rates beats me guessing at contact physics.

If you drop the specific repo/grid/pilot links, I'll add a fusion knob to the phantom (welded adjacent windings, with instance labels that reflect it), slot it into the correlation ladder the same way as the other rungs, and report whether it moves ρ on the frozen cross-scroll setup. Ground-truth-in, ground-truth-out.

Jinhojeong:
@beaugrzanich, the assert landing in 37ea097 before my next reply is the right tempo for this thread. Here is everything the rung needs.

- Painter: [`scripts/finite_paint.py`](https://github.com/Diego-dcv/vesuvius-topological-grid/blob/main/scripts/finite_paint.py) in Diego-dcv's grid repo, merged as [PR #2](https://github.com/Diego-dcv/vesuvius-topological-grid/pull/2) there. Finite thickness, per-winding instance labels, and a weld mode that fuses adjacent windings while the labels keep their identities, which is exactly the ground-truth-in property your knob wants.
- The defect history and acceptance exams: [issue #1](https://github.com/Diego-dcv/vesuvius-topological-grid/issues/1) in the same repo carries the zero-thickness finding, and Diego's exam F added after it is the pattern your 37ea097 assert follows.
- Fuse rates to calibrate the knob: the preregistered readout is [this villa #191 comment](https://github.com/ScrollPrize/villa/issues/191#issuecomment-5168228180), artifacts under [`results/fusion_pilot`](https://github.com/Jinhojeong/vesuvius-surface-geometry-diagnostic/tree/main/results/fusion_pilot) in my diagnostic repo. The number your knob should reproduce is that detected contact fuses at 74.8 to 78.0 percent, flat across the 10 to 150 um gap range that grid spans, so fusion probability conditions on contact and not on gap size.
- Truth-side validation instruments: [`results/twin`](https://github.com/Jinhojeong/vesuvius-surface-geometry-diagnostic/tree/main/results/twin) in the same repo holds the flag and split verdicts against exact synthetic weld truth, commit deb459f, if you want a reference for what a fused-instance acceptance test looks like.

One geometry note worth having before you pick knob parameters. In the twin work, contact-rich structure appeared without being asked for. An isoperimetric 2 to 1 crush pushes minor-axis spacing to about 112 um against sheet thickness near 150, so a crush parameter alone generates the contact population, and the weld probability then applies at detected contact. If you wire the knob that way rather than welding at a fixed rate everywhere, the rung inherits the measured physics instead of a guess. And if you want a second pair of eyes, I will read the knob's acceptance exam the way Diego's exam F got read.

beaugrzanich:
@Jinhojeong — built the fusion rung on your artifacts and chased it hard. The headline isn't "fusion moves ρ" — it's a mechanism for *what makes the phantom discriminate fitters at all*, and it turned out to be laptop-provable. Long, but every number is reproducible.

## Consumed the painter, verified against the real repo

`winding_error` now reads a `finite_paint` `.npz` (`turn_id` as truth) and scores any candidate; acceptance exam committed (37ea097 lineage). Verified against real `finite_paint volume --fuse ...` output, not a mock — the `turn_id` int16 / 0=air / 1-based contract matches, a perfect candidate scores 0, and a mislabeled turn localizes to 100% on that turn.

## Fusion the ground-truth-in way (not my earlier label-collapse)

My first stab collapsed label identity; you were right that's wrong. Instead I trace your *welded volume* with a classical tracer so the merge is organic — a fused sector shows a ~3.6-crossing deficit for a 3-turn weld. Slotted into the ladder, fusion moves ρ from ~0 (clean phantom) to **+0.41** at moderate dose, mechanism-correct (catastrophic configs punished, never rescued). A dose sweep is a clean inverted-U: too much fusion scrambles the ranking back to noise. So fusion is a real but *weak* rung — 0.41, p=0.18, not significant.

## The crush is not a wall — I was wrong about that, twice

Your crush → contact recommendation broke my classical tracer, and I first mis-called it a "nonlinear breakdown." It's a **linear slope error** (~1.2× stretch gauge can't remove). An un-crush-for-indexing step — estimate the ellipse by PCA, circularize only the coordinate the indexer measures radius in, keep patch vertices crushed — restores it: slope 1.23 → 1.04, R² → 1.0.

Then the crushed *fit* still scored badly (MAE ~1.9) and I nearly called it a fitter-capability wall. It isn't: **freezing the linear stage at the estimated crush seed halves the error (1.92 → 0.68).** The `SpiralAndTransform`'s own per-z linear stage can represent the ellipse — the optimizer was drifting off it. So the crush is an optimization/regularization issue, laptop-fixable, not a "needs the production pipeline" wall. (I kept reaching for that hatch and it kept being wrong.)

## The actual finding: misleading vs missing evidence

Crush, once handled, adds no discriminating difficulty — it's a global distortion you correct, not a local trap. So what *does* make the phantom rank fitters like real scrolls? I tested each corruption on the same circular fused base (score vs `turn_id`, correlate vs held-out real):

| Corruption | Kind | ρ | p | |
|---|---|---|---|---|
| Damage (holes) | missing | 0.06 | 0.85 | no signal |
| Crush (2:1) | global-geometric | −0.59 | 0.04 | **anti-signal** |
| Fusion (weld) | misleading-local | 0.41 | 0.18 | |
| Sheet-switch (±1) | misleading-local | 0.79 | 0.002 | significant, alone |
| Fusion + switch | two misleading | 0.85 | 0.0005 | |

Three failure modes, three signs. *Missing* evidence (damage) removes it equally for good and bad fitters — nothing to resist, no discrimination. *Global-geometric* difficulty (crush) actively anti-ranks, because handling it rewards a strategy (flow suppression) real scrolls punish. Only *misleading local* evidence discriminates forward: a switch or weld *lies* (wrong or fewer windings), a good fitter's global prior overrides the lie, a bad one follows it. **The phantom ranks fitters like real scrolls only through misleading local evidence a global prior can correct — not difficulty, not missing data, and not global realism.** The damage/crush/switch three-way is the clean proof.

## Cross-scroll significance, on a laptop

The fusion+switch phantom's ranking vs two real scrolls, frozen: PHerc0172 ρ=0.85 (p=0.0005), PHerc0332 cold ρ=0.80 (p=0.002). n=12, leave-one-out ρ stays 0.80–0.91. No crush, no production fitter, no GPU.

## I ran your crushed recipe — and it anti-discriminates

Your calibration checks out: the 74.8–78% is `fused_given_detected_pct` in `results/fusion_pilot/fusion_diag.json` (78.7 / 78.4 / 78.5 / 78.9 across seeds and pitches) — fusion conditioned on *detected contact*, exactly as you said. (My first look hit the marginal rate, 18–42% fused/all-neighbours, in the other file; the conditional key is the one you meant.)

So I ran it: crushed (2:1) `finite_paint`, un-crush for indexing, and the fit's per-z linear stage **frozen at the estimated crush** (which halves the fit error vs letting the optimizer drift). Result: **ρ = −0.59 (p = 0.04)** — the crushed phantom ranks fitters *backwards*, worse than circular fusion (0.41) and crushed-without-the-seed (0.19).

The mechanism is clean and I think it generalizes: the crush is a **global** distortion. Once it's the difficulty, the winning move is to *suppress the flow field and let the linear stage handle the ellipse* — which is exactly `rigid_flow100`'s strategy (crushed MAE 0.62, best) — but real scrolls need the flow for local deformation and *punish* that suppression (`rigid_flow100` real MAE 3.33, worst). The crushed phantom rewards precisely what real penalizes.

One honest remaining gap: I welded at my own dose, not tuned to reproduce the 78% at-contact rate, and I haven't run `results/twin`'s flag/split instruments. But the sign of the result wouldn't change with a rate tweak — the crush *direction* is the problem, not the dose.

## Honest scope — what's new and what isn't

- The switch **is** the ±1 index-noise that was already in my round-3 config (which hit 0.72). So the *ingredient* is not a discovery. What's new is the **mechanism** — the damage (missing) vs switch (misleading) contrast shows it's not "add noise," it's specifically misleading-evidence — and that a single misleading trap reaches cross-scroll significance in this ground-truth-in `finite_paint` setup.
- Still n=12 variants, one reference fitter (not production `fit_spiral`), two scrolls. The crush seed estimate is imperfect (PCA undershoots ratio).
- Fusion+switch (0.85) barely beats switch-only (0.79): it's the *kind* of trap that matters, not "variety."

## So where does the fusion rung actually land

Honestly, not where either of us expected. The discriminating ingredient turned out to be **already in the committed ladder** — the misleading-local trap is `--index-noise-rate`, which has been there since round 3. And the two things I built to answer your rung both argue *against* adding a fusion rung as such: fusion is a *weak* misleading trap (0.41, not significant on its own), and your crushed/contact-conditioned version *anti*-discriminates (−0.59). So I'm not convinced a committed "fusion knob" earns its place — it would be a weaker version of the label-noise trap that already works, plus a crush axis that ranks fitters backwards.

What I think *is* worth keeping is the **mechanism**, and I've committed only that: a design-rationale note on the ladder saying discrimination comes from misleading-local label errors, not missing data and not global geometry (no `finite_paint` numbers in the repo, since those aren't reproducible from the branch). The consume adapter (`--truth-npz`, scores against your `turn_id` truth, with an acceptance exam) is the other committed piece; everything else — the trace/uncrush/trap scripts — is a research spike that stays in scratch.

Two honest asks: (1) does the mechanism reframing match what you see in real material, or is there a misleading-local pathology I'm missing that a knob *should* model? (2) if you still want the physics-calibrated fusion rung validated despite the −0.59, the blocker is the reference fitter — it needs the production fitter or your `results/twin` flag/split instruments, not more of my scratch code. I'd value your read on whether this is a dead end or a rung worth someone with the GPU picking up.

Jinhojeong:
@beaugrzanich, the mechanism claim is the strongest thing to come out of this ladder, and the damage-crush-switch three-way with three different signs is a clean way to have proven it. Correcting yourself twice inside one report, on the nonlinearity call and on the capability wall, is the standard this thread family runs on, and the un-crush-for-indexing step plus the frozen linear stage both look immediately reusable.

Your minus 0.59 refutes half of what I recommended, and I want that on the record in my words. I told you to wire the knob so a crush parameter generates the contact population, because that was the measured physics of the twin. You ran exactly that and it anti-discriminates, and your mechanism says why. The crush is a global distortion, so once it is the difficulty, the winning strategy is flow suppression, which real scrolls punish. Physical realism of the generator does not imply ranking realism of the benchmark, and I had conflated the two. What survives of my recommendation is the part you verified independently, the fuse rate conditioned on detected contact, 74.8 to 78 percent, as the weld probability. The delivery mechanism I proposed for the contacts should be dropped in favour of your circular fused base plus switch, which your table shows is where the discrimination actually lives.

On the dose gap you flagged yourself, one measured offer. Three anchors exist for where real material sits on your inverted-U, all public. On PHerc1218 about two thirds of thick sheet runs carry a single instance id, 67.65 percent before my repair and 65.54 after. On PHerc1203, against dual-scan physical truth, the per-cell fused-signature share runs from zero on 11 percent of cells through a median of 8.3 percent to a maximum of 41.7 percent, a full dose distribution rather than a point, shipped as `results/thickness1203/zones.npz` with provenance hashes. If you place the weld dose at that distribution, the readout worth having is whether the median-dose cell sits on the discriminating limb of your U or below it. If it sits below, fusion alone cannot rank fitters at realistic dose and the switch companion in your 0.85 pairing is doing the work, which the 0.79-alone row already hints at.

beaugrzanich:
@Jinhojeong — thank you for putting the −0.59 on record in your words; that's the right way to hold it. Taking you up on the measured offer, because it closes the exact dose gap I flagged.

I put your fused-signature share and my dose axis on one ruler. My weld sweep, scored against the frozen PHerc0172 rows, is a clean inverted-U:

| Dose (mine) | Cond | Fused-share (median) | ρ vs real | p |
|---|---|---|---|---|
| 0.0 | unfused | 0% | −0.03 | 0.91 |
| 1.0 | low | ~4–5% (interpolated) | +0.36 | 0.25 |
| 2.9 | mid | 12.7% (measured) | +0.41 | 0.18 |
| 10.8 | high | >12.7% | −0.24 | 0.46 |

The share column is the bridge: I measured, on my own generated volume, what fraction of occupied cells my mid setting actually welds, using fused-vs-unfused `turn_id` at a fixed seed — it comes out at 12.7% median. Your PHerc1203 median is 8.3%, which lands just *below* my mid setting, on the rising limb.

So to your "discriminating limb or below it": median-dose fusion sits *on* the limb, not below it — but the limb never earns significance. Its peak is ρ=0.41 (p=0.18) at mid, and real's 8.3% median is just under that peak, around ρ≈0.4. Fusion alone cannot significantly rank fitters at realistic dose, or at any dose in my sweep. Your 0.79-alone hint was right: the switch companion carries the 0.85 pairing, and the fusion knob should be dropped in favour of the circular base plus switch, exactly as you said.

One honest mismatch I won't paper over, because it bears on the readout. My generator welds *uniformly* — median share ≈ max share = 12.7%, zero pristine cells. Your PHerc1203 is heavy-tailed: 11% of cells clean, median 8.3%, a tail to 41.7%. So real fusion is sparse-and-clustered; mine is smeared-even. Two consequences: the real *median* cell (8.3%) is on my weak limb, no ranking power; and if real fusion's discriminating signal lives anywhere, it's in that 41.7% tail, not the median. My closest analog to the tail is the high-dose row, which *scrambles* the ranking (ρ=−0.24) — suggestive that the tail won't rescue fusion either, but I'll be precise: my high dose is uniform-high everywhere, not a few hot cells, so that's a hint, not a proof. A lumpy-dose generator matched to your `zones.npz` distribution is the honest way to settle the tail question, and I haven't built one.

Net, and I think we agree: fusion is a real but weak misleading trap that doesn't clear significance at realistic dose; the switch is the workhorse; your crush-delivery anti-discriminates for the global-distortion reason. The keeper is the mechanism, not any of the three knobs. The only thread still worth pulling is whether heavy-tailed fusion behaves differently than even fusion at the same median — and that needs either your flag/split instruments on `results/twin` or a dose distribution sampled from `zones.npz`, not more of my scratch. Happy to hand the reconstruction over if it's worth a GPU pass to someone; from the laptop side, this is where fusion honestly lands.

pmh47:
This will only be considered if it is shown that correlation of 'mild' failures on the phantom is predictive of such failures on real scrolls, i.e. that it provides a useful signal for hyperparameter optimisation etc (and better than our current approach of measuring ink coverage, column regularity, and satisfaction of patches).

Issue #1364, opened 2026-08-07
predict normalizes nnU-Net checkpoints with instance z-score regardless of the scheme in their plans.json


Jinhojeong:
Running an nnU-Net checkpoint through `vesuvius.predict` normalizes the input with per-volume z-scoring even when the checkpoint's own plans.json declares something else. The plans are loaded and stay in hand. They are just never consulted when the normalization is chosen, and nothing warns, so the run looks normal and the outputs look plausible.

Where it happens, on main at a6c8ad5.

- `--normalization` defaults to `instance_zscore` in the CLI parser, `vesuvius/src/vesuvius/models/run/inference.py:1165`, and the constructor default is the same at `:252`.
- `self.model_normalization_scheme` and `self.model_intensity_properties` both start as None at `:319-320`.
- Two branches fill them in. The resnet branch sets the scheme at `:405`, and the train.py checkpoint path sets both at `:542-549`.
- The nnU-Net branches do not. `--model_type nnunet` at `:419-426` and the auto fallback at `:440-448` both call `load_model_for_inference`, and the model_info it returns carries `plans_manager` and `configuration_manager` but no normalization keys, `vesuvius/src/vesuvius/utils/models/load_nnunet_model.py:297-308`. The only place that loader builds those keys is the HuggingFace train.py branch at `:386-387`.
- The `hf://` path at `inference.py:378-384` calls the same loader and lands in the same hole, unless the repo turns out to be a train.py checkpoint, which is the branch tested right after at `:387`. That is the form `vesuvius/docs/inference.md` leads with, so it is not an edge case.
- So `self.model_normalization_scheme or self.normalization_scheme` at `inference.py:696` resolves to the CLI default, and VCDataset is constructed with it at `:727`, inside the call that runs `:721-749`.
- The scheme actually in use is printed only under `verbose`, `vesuvius/src/vesuvius/data/vc_dataset.py:187`, so a default run says nothing about normalization at all. Across 16 non-verbose runs of the checkpoint below, no log contains the string `normaliz`.

The information is already in hand at that point. `configuration_manager.normalization_schemes` and `plans_manager.foreground_intensity_properties_per_channel` are both populated by the loader, and inference.py reads neither.

[...]

PR #1382, opened 2026-08-10
Add a physical evaluation harness for surface predictions (cross-scan ground truth)

7jycwjmbfn-eng:
Adds a small standalone harness that scores any binary surface prediction against ground truth taken from a second, higher-resolution scan of the same scroll, instead of from another model or from mesh-derived labels.

Context: #193 closed after both model-free validation legs failed (a CT statistic with window bias, and a validator model at chance level), and #1309 was closed because success on a synthetic phantom does not establish success on real scrolls. A second physical scan of the same object avoids all three problems: the reference is real, and no model or annotation is in the loop. Two scroll pairs currently have the required data, PHerc0139 (truth at 1.129 um) and Grand-Prize PHerc1203 (truth at 2.403 um, including compressed tissue); registrations are split-half validated at 4.1 and 2.4 um median error, well inside the 19-56 um scoring tolerances.

The harness is one file (numpy + scipy + numcodecs) and reads the label volumes published at https://github.com/7jycwjmbfn-eng/pherc0139-physical-audit (release assets). Every metric it reports carries a shifted-null control, the same prediction displaced 1.2 mm in-plane. The null turns out to matter: on PHerc1203 it alone reaches 64.6 percent point recall at 37 um, so any radius metric without a null overreads performance in exactly the tissue where the hard failures live. One printed field stands outside that rule and is labelled as such, `recto_side_ratio`, a raw mass-overlap quantity kept for continuity; the side metric to read is `side_inward`, which ships with both its null and an ideal-recto ceiling built from the truth.

Reproduction: the harness output on both scrolls is committed in the linked repository (`results/eval_selfcheck_0139.json` and `results/1203/eval_selfcheck_1203.json`), so a rerun is diffed against a file rather than against prose. The point and arc numbers match the audit to four decimals, as does the side null; the inward fraction and ceiling agree to three, because the harness takes sheet normals from the packaged material mask while the audit used the high-resolution grayscale the label package does not carry. That substitution moves the decided count by 0.09 percent on PHerc0139.

Two revisions since this PR opened, both from review in #191:

- @TAUIL-Abd-Elilah found that the three side arms are nested rather than independently drawn: the real arm keeps points with the prediction within 3 voxels, and the null and ideal arms select inside that subset. That reproduces what the audit pass does, and it had never been stated. The harness now emits `_full` variants selecting each control from the whole coherent centerline set, plus a decided count per arm, so both readings are visible. Un-nesting moves the controls only, worth 0.31 points of skill on PHerc0139 and 0.10 on PHerc1203; `test_side_arms_equivalence.py` checks that the real arm does not move.
- Checking that led to a documentation defect in the label release: the `boundary_poor` bit covers 74.6 percent of material voxels on PHerc1203, not the 7-14 percent figure the README had described it with, which came from a different instrument. Both volumes now ship a full census of what every bit plane covers.

Findings from running the harness on the published volumes for both scrolls (fully-missed sheet stretches, side-of-sheet placement, the #1364 caveat on what generated those volumes) are written up in the linked repository.

Happy to move the file elsewhere in the tree, rename things, or split the labels documentation out, whatever fits your conventions.

### DIFF
--- a/scripts/physical_eval/README.md
+++ b/scripts/physical_eval/README.md
@@ -1,0 +1,29 @@
+# Physical evaluation harness for surface predictions
+
+Evaluates any binary surface prediction volume against ground-truth labels derived from a second, higher-resolution scan of the same scroll, rather than from another model or from mesh-derived annotations.
+
+Two label volumes exist so far, built by registering the scroll pairs that have two public scans (split-half validated at 4.1 um and 2.4 um median error):
+
+- PHerc0139: 9.362 um frame, truth from the 1.129 um scan
+- PHerc1203 (Grand Prize list): 9.362 um frame, truth from the 2.403 um scan, including compressed and boundary-poor tissue
+
+The label volumes (uint8 bit flags: valid / material / centerline / recto_band, plus boundary_poor for PHerc1203), the registration transforms, and the audit built on them are at
+https://github.com/7jycwjmbfn-eng/pherc0139-physical-audit (labels attached to the release).
+
+## Usage
+
+```bash
+python3 eval_surface_pred.py LABELS.zarr PRED.zarr PRED_LEVEL
+```
+
+`PRED_LEVEL` is the pyramid level of the prediction volume (0 or 1); level 0 is max-pooled to the label grid. Dependencies: numpy, scipy, numcodecs.
+
+Reported metrics, each alongside a shifted-null control (the same prediction displaced 1.2 mm in-plane):
+
+- point recall at 19 / 37 / 56 um from truth sheet centerlines
+- arc-level recall and fully-missed rate over 1.2 mm sheet stretches
+- recto-side overlap ratio
+
+The null control matters in dense tissue: on PHerc1203 the null alone reaches 64.6 percent point recall at 37 um, so radius metrics without it overread performance exactly where the hard failures live.
+
+Checked against the published m7 volume for PHerc0139, the harness reproduces the audit numbers in the repository above to four decimals from the packaged files alone.
--- a/scripts/physical_eval/README.md
+++ b/scripts/physical_eval/README.md
@@ -7,7 +7,7 @@ Two label volumes exist so far, built by registering the scroll pairs that have 
 - PHerc0139: 9.362 um frame, truth from the 1.129 um scan
 - PHerc1203 (Grand Prize list): 9.362 um frame, truth from the 2.403 um scan, including compressed and boundary-poor tissue
 
-The label volumes (uint8 bit flags: valid / material / centerline / recto_band, plus boundary_poor for PHerc1203), the registration transforms, and the audit built on them are at
+The label volumes (uint8 bit flags: valid / material / centerline / recto_band, plus boundary_poor for PHerc1203, which flags material whose gap visibility falls under 0.40 and covers 72.8 percent of material voxels there rather than a small minority), the registration transforms, and the audit built on them are at
 https://github.com/7jycwjmbfn-eng/pherc0139-physical-audit (labels attached to the release).
 
 ## Usage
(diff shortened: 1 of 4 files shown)


TAUIL-Abd-Elilah:
One denominator issue before this is used for model A/B: `pb.sum() < 100` is checked before `n_ctr` / `n_arc` are incremented, so a sparse or empty prediction removes that truth slice—and its failures—from its own denominator. The dense published m7 self-check may be unaffected, but “any binary prediction” is not fail-closed.

I’d keep the truth-only `ctr` eligibility, remove the prediction-density condition, and represent an empty prediction with an infinite distance field so it contributes zero hits and fully missed arcs. I added a synthetic regression that checks exactly that behavior here: https://github.com/TAUIL-Abd-Elilah/vesuvius-repro/blob/8d3476cf2d2c0bb46b97518603958b6373734c21/test_physical_normalization_ab.py#L52-L65

7jycwjmbfn-eng:
You're right. Fixed in `9fb1b0b`, the guard is truth-only now.

One wrinkle if you port it: dropping the `pb` clause on its own is not enough. With nothing set, scipy's distance transform treats outside-the-array as background and returns 1.0 at the border, so an empty prediction picks up border hits. Empty bands get an infinite distance field instead.

Added `test_empty_prediction_denominator.py`: empty, sparse and dense predictions over one synthetic truth, all three have to report the same centerline and arc denominators. It fails on the pre-fix version.

The same guard is in the eleven audit scripts under `code/` as well, which is where the report's numbers come from. You could only see the harness since it is the only file in this PR. I checked whether it ever fired on the published m7 volumes:

```
              scored slices   fired   sparsest scored slice   threshold
PHerc0139          283          0            15,931 voxels        100
PHerc1203          464          0           738,200 voxels        100
```

Zero on both, under either form of the truth clause. Output in `results/guard_probe.json`, script alongside it. Rerunning PHerc0139 under the fixed harness gives numbers bit-identical to the published ones, which agrees.

So the audit scripts keep the old guard as the record of the run that produced those numbers, and the harness is fixed since that is the file other people run. README states the split.

If you scored anything with the harness before `9fb1b0b`, worth a rerun. Two models of different density would have been scored on different denominators, which matters more for an A/B than it did here.

bruniss:
we will consider this for merging if you can demonstrate that it accurately identifies model errors accompanied with images. as of now im having a difficult time understanding what the claims are, especially given that 1203 has no GT labels (which contradicts at least one of the claims in the desc)

7jycwjmbfn-eng:
The PR had no images, that's fair.

On 1203 first: you're right, there's no annotated GT for it. I wasn't using annotation. Both scrolls have a second scan in the open data at higher resolution, 1.129 µm on 0139 and 2.403 µm on 1203, against the 9.362 µm the models run on. I register the fine scan onto the coarse one and take the sheets visible in it as the reference. That's what "truth" meant in the description, and I can see how it read as a claim about labels.

The claim in one line: two scrolls now have a sheet reference that came from neither a model nor an annotator, packaged so anyone can score against it. It covers the span the fine scan reaches, 23 mm of 0139 and 38 mm of 1203, not the whole scroll. The open-problems page says "trusted voxel-level labels are scarce" and asks for better recto localisation in densely packed regions; 1203 is one of those.

**The reference itself**, one figure per scroll. Left is what m7 was given, right is the fine scan in the same frame. Individual sheets can be followed from one to the other, so the alignment is something you can check before believing any number of mine. Held-out registration error is 4.09 µm on 0139 and 2.38 µm on 1203, against tolerances of 19-56 µm. Those held-out points come from the same block-matching pass, so a bias common to every match would survive them.

PHerc0139 reference

PHerc1203 reference

**The scoring.** I cut each sheet centreline into 1.2 mm stretches of at least 20 points. A stretch counts as covered when at least half its points have a prediction within 2 voxels (37 µm), and fully missed when under a tenth do. Red is fully missed, blue is covered, same crop and slice so both sit under the same conditions. The rows are the two longest missed stretches in that slice that have a covered stretch in the same crop and whose crop lies inside the scanned volume.

PHerc0139 scoring

PHerc1203 scoring

Numbers: m7 covers 89.9% of stretches on 0139 and 76.1% on 1203. Slide m7's own output 1.2 mm sideways and it covers 57.7% and 64.6%. Fully missed goes from 5.7% to 12.7%. On 1203 that margin is small enough that I wouldn't trust a radius metric there without the shift control.

One limit worth stating: per-sheet centrelines need sheets that are still separable, so on 1203 they cover 2.46% of material against 10.36% on 0139. The reference resolves more than the working scan does, and it does not solve fused tissue.

The figures and the code that builds them are at [`3f2bc75`](https://github.com/7jycwjmbfn-eng/pherc0139-physical-audit/tree/3f2bc75e63fbdabb3926e4d550b3a5851a231dfb/code/figures). It re-derives the scored stretches from the labels and the published prediction volume rather than reading the audit's stored results, and on 0139 checks each one against the per-arc list: six slices, 172 of 172.

Whether it's worth reopening is your call. The labels and the evaluator stay up either way. If it isn't, I'd still like to know what would make this useful to you.

pmh47:
Thanks for the images, these confirm that the check doesn't work. If just having a certain intensity range in the CT were sufficiently-indicative of being the GT surface, the problem would be very easy. Unfortunately it's not -- need to account for the morphology, such as delaminated sheets, frayed fibers, kollesis, ... -- which a naive intensity-based check cannot do.

PR #1385, opened 2026-08-10
spiral: add deterministic CUDA gap interpolation

TAUIL-Abd-Elilah:
## Summary

Make the production Spiral gap lookup compatible with PyTorch hard-deterministic CUDA execution.

- keep the existing `grid_sample` path unchanged during normal execution and on non-CUDA devices
- when `torch.use_deterministic_algorithms(True)` is active on CUDA, use the exact N=C=1, bilinear, border-padded, `align_corners=True` subset needed by `GapExpandingTransform`
- preserve ATen's zero coordinate gradient at and beyond the border
- cover forward values, input/coordinate gradients, border gradients, pinning, and bitwise repeated backward

## Why

A current-main production Spiral fit on PHerc0125 reached the first optimizer update under PyTorch hard determinism and stopped at:

```text
grid_sampler_2d_backward_cuda does not have a deterministic implementation
```

The unsupported operation was the eager gap-expander lookup. The Cartesian flow path has its own sparse backward; on the tested PyTorch 2.13.0+cu126 / RTX 3090 environment, the other repeated-destination operations used by this fit were accepted in hard-deterministic mode.

This patch specializes only the gap lookup that has the fixed shape/mode contract above. It does not replace general `grid_sample`, and it does not change the default path.

## Validation

```text
python -m pytest -q \
  tests/test_deterministic_gap_sample.py \
  tests/test_phase_spacing.py

38 passed, 1 skipped
```

The CUDA tests compare the specialized sampler with `grid_sample` for values and gradients, including exact border coordinates, and require bitwise equality across repeated hard-deterministic backward passes.

I also ran two isolated production compatibility sessions from the same completed-iteration-1 PHerc0125 checkpoint through completed iteration 20. Both completed without warnings or resource-floor violations. Non-timing metrics at iterations 2, 5, 10, and 20 were exactly equal; the path-independent preview payloads were byte-identical; and a recursive comparison found every tensor, NumPy/RNG state, optimizer/scheduler/config value, and scientific scalar equal.

One preregistered literal gate still reported failure because it required the complete checkpoint dictionaries to be equal while each checkpoint intentionally records its own `input_manifest.output_directory`. The recursive audit found that provenance string to be the only difference. I am stating that failure explicitly: the result supports method compatibility and deterministic scientific state, not fit quality or any claim that the raw checkpoint files themselves are byte-identical.

## Scope

This is reproducibility plumbing. It makes hard-deterministic production execution possible for the tested path; it does not claim better fit quality, cross-scroll generality, or complete reproducibility across PyTorch/CUDA releases.

### DIFF
--- a/volume-cartographer/scripts/spiral/tests/test_deterministic_gap_sample.py
+++ b/volume-cartographer/scripts/spiral/tests/test_deterministic_gap_sample.py
@@ -4,6 +4,7 @@ import pytest
 import torch
 import torch.nn.functional as F
 
+import transforms
 from transforms import (
     GapExpanderParams,
     GapExpandingTransform,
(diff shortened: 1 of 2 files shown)


bruniss:
there are many more areas of non-determinism in fit_spiral so this fix is not really necessary for the time being. in the future we may look to make it more deterministic but for the time being this adds a large test and code block for no gain.

Issue #1360, opened 2026-08-07
vesuvius.predict fails on zarr 3 (a supported version) and exits 0 while writing nothing


TAUIL-Abd-Elilah:
`vesuvius.predict` cannot write its output store under zarr 3, and the failure is silent to
anything that checks a return code. `pyproject.toml` declares `zarr>=2.18.7,<4`, so zarr 3 is
supported on paper.

### Reproduce

zarr 3.2.1, any model and input:

```
vesuvius.predict --model_path <model> --input_dir <ct.zarr> --output_dir out \
  --device cuda --disable_tta --batch_size 1 --num_workers 2 --bbox 32:160,32:160,32:160
```

```
An error occurred during inference: module 'zarr' has no attribute 'Blosc'
--- Inference Failed ---
Error: cannot unpack non-iterable NoneType object
```

Exit code is **0** and `out/` is empty.

### Cause 1 — `zarr.Blosc` was removed in zarr 3

`models/run/inference.py::_get_zarr_compressor` (four call sites, lines ~822-830) uses
`zarr.Blosc`, which is a zarr-2 API.

The fix looks like a one-liner because on zarr 2 that name is literally the numcodecs class:

```python
>>> import zarr, numcodecs; zarr.__version__
'2.18.3'
>>> zarr.Blosc is numcodecs.Blosc
True
```

So importing `Blosc` from `numcodecs` is a no-op on zarr 2 and a fix on zarr 3.

### Cause 2 — the store is then created as v3 with a v2-only argument

With `Blosc` resolved, the next failure is:

```
ValueError: compressor cannot be used for arrays with zarr_format 3.
Use bytes-to-bytes codecs instead.
```

`data/utils.py:144` passes `compressor=` to `zarr.open` without naming a format, so zarr 3
defaults to v3, where `compressor` is not a valid argument. Either pin `zarr_format=2` when a
compressor is supplied, or translate to `codecs=[BloscCodec(...)]` for v3. Pinning v2 reproduces
what zarr 2.18.3 did, which is the combination existing artifacts were produced under.

I worked around both locally by shimming rather than patching, so I have not opened a PR — happy
to if you would prefer a specific direction on cause 2 (v2 pin vs. real v3 codecs), since that
one is a behaviour choice about what the output store should be, not just a compatibility fix.

### Cause 3 — separate, and worse for pipelines: failure exits 0

`inference.main()` and `blending.main()` both catch their own exceptions, print
[...]

PR #1386, opened 2026-08-10
vesuvius: honor nnU-Net plan normalization in predict

### LINKED ISSUE #1364: predict normalizes nnU-Net checkpoints with instance z-score regardless of the scheme in their plans.json

TAUIL-Abd-Elilah:
## Problem

`vesuvius.predict` loads an nnU-Net model's architecture and weights but ignores the normalization declared in `plans.json`. It therefore silently keeps the CLI default, `instance_zscore`, even when a checkpoint was trained with `CTNormalization`.

This is the inference mismatch diagnosed in #1364. The issue's 60-volume reproduction reports median recall/precision changing from 0.7740/0.4459 to 0.9109/0.6680 after using the model's CT normalization.

The same issue identifies two blockers around that path: the nnU-Net 2.8 trainer API differs from villa's vendored 2.5 API, and the missing-trainer fallback imports kwarg names such as `conv_op` instead of resolving the class paths stored in those kwargs.

## Change

- derive the native normalization scheme from the nnU-Net configuration manager already returned by the loader
- pass channel-0 CT fingerprint statistics through `VCDataset` to `Volume`
- visibly report when model metadata overrides the CLI/default
- fail explicitly for unsupported, masked-Z-score, or multi-channel contracts instead of silently changing semantics
- preserve explicit normalization metadata from train.py/external checkpoints
- support both nnU-Net 2.5 and 2.8 `build_network_architecture` signatures by inspecting the declared parameters, without swallowing errors raised inside a trainer
- use nnU-Net's own `get_network_from_plans` for the missing-trainer fallback

Inference exception propagation is intentionally left to #1414, which now owns the stronger fix for #1360; this PR no longer touches that hunk.

Supported native mappings use villa's existing implementations: `CTNormalization`, `ZScoreNormalization`, `NoNormalization`, and `RescaleTo01Normalization`.

## Validation

- Loaded the public `surface_m7` checkpoint through the full `Inferer._load_model` path on CPU:
  - resolved `CTNormalization -> ct`
  - mean `87.54424285888672`, std `47.74376678466797`
  - clipping bounds `[0.0, 212.0]`
  - patch size `(192, 192, 192)`
- Constructed the same public model through the repaired direct fallback: `ResidualEncoderUNet`, 1 input channel, 2 output heads.
- New numerical test compares villa's CT-normalized volume exactly against nnU-Net's own `CTNormalization`, including out-of-range clipping.
- Compatibility tests exercise both the vendored 2.5 trainer signature and the 2.8.1 signature.
- `119 passed, 1 skipped`:
  `pytest tests/models/run tests/data tests/utils/models/test_load_nnunet_model_compatibility.py -m "not slow and not network" -q`

Fixes #1364.

### DIFF
--- a/vesuvius/src/vesuvius/models/run/inference.py
+++ b/vesuvius/src/vesuvius/models/run/inference.py
@@ -109,6 +109,87 @@ def _checkpoint_normalization_scheme(checkpoint_data):
     return image_normalization
 
 
+_NNUNET_NORMALIZATION_MAP = {
+    'CTNormalization': 'ct',
+    'ZScoreNormalization': 'instance_zscore',
+    'NoNormalization': 'none',
+    'RescaleTo01Normalization': 'instance_minmax',
+}
+
+
+def _nnunet_normalization_from_model_info(model_info):
+    """Return the normalization declared by an nnU-Net model's plans.
+
+    ``load_model_for_inference`` already returns the nnU-Net configuration and
+    plans managers.  Keep their native metadata authoritative rather than
+    silently falling back to the CLI's per-patch z-score default.
+
+    VCDataset currently accepts one normalization contract for the entire
+    input, so reject multi-channel plans instead of applying one channel's
+    fingerprint to another.
+    """
+    configuration_manager = model_info.get('configuration_manager')
+    plans_manager = model_info.get('plans_manager')
+    if configuration_manager is None or plans_manager is None:
+        return None, None
+
+    schemes = getattr(configuration_manager, 'normalization_schemes', None)
+    if schemes is None:
+        return None, None
+    if isinstance(schemes, str):
+        schemes = [schemes]
+    else:
+        schemes = list(schemes)
+    if len(schemes) != 1:
+        raise ValueError(
+            'vesuvius.predict cannot reproduce nnU-Net normalization for '
+            f'{len(schemes)} input channels; expected exactly one scheme')
+
+    native_scheme = schemes[0]
+    if not isinstance(native_scheme, str):
+        native_scheme = getattr(native_scheme, '__name__', str(native_scheme))
+    native_scheme = native_scheme.rsplit('.', 1)[-1]
+    try:
+        scheme = _NNUNET_NORMALIZATION_MAP[native_scheme]
+    except KeyError as exc:
+        raise ValueError(
+            'Unsupported nnU-Net normalization scheme in plans.json: '
+            f'{native_scheme!r}') from exc
+
+    use_mask = getattr(configuration_manager, 'use_mask_for_norm', None)
+    if isinstance(use_mask, (list, tuple)):
+        use_mask = use_mask[0] if use_mask else None
+    if scheme == 'instance_zscore' and bool(use_mask):
+        raise ValueError(
+            'nnU-Net plans require masked Z-score normalization, which '
+            'vesuvius.predict cannot reproduce without the preprocessing mask')
+
+    intensity_properties = None
+    if scheme == 'ct':
+        per_channel = getattr(
+            plans_manager, 'foreground_intensity_properties_per_channel', None)
+        if not isinstance(per_channel, dict):
+            raise ValueError(
+                'nnU-Net CTNormalization requires '
+                'foreground_intensity_properties_per_channel in plans.json')
+        intensity_properties = per_channel.get('0', per_channel.get(0))
+        required = {
+            'mean', 'std', 'percentile_00_5', 'percentile_99_5',
+        }
+        if not isinstance(intensity_properties, dict):
+            raise ValueError(
+                'nnU-Net CTNormalization requires intensity properties for '
+                'input channel 0')
+        missing = sorted(required.difference(intensity_properties))
+        if missing:
+            raise ValueError(
+                'nnU-Net CTNormalization channel-0 intensity properties are '
+                f'missing: {", ".join(missing)}')
+        intensity_properties = dict(intensity_properties)
+
+    return scheme, intensity_properties
+
+
 def _select_evenly_spaced_middle_indices(candidate_indices, limit):
     """Select a deterministic, small representative subset of patch indices."""
     candidates = list(candidate_indices)
@@ -447,6 +528,26 @@ class Inferer():
                     verbose=self.verbose
                 )
         
+        # Prefer normalization metadata returned directly by train.py/external
+        # loaders. For nnU-Net, derive it from the plans managers that the
+        # loader already returns.
+        model_scheme = model_info.get('normalization_scheme')
+        model_intensity_properties = model_info.get('intensity_properties')
+        if model_scheme is None:
+            model_scheme, nnunet_intensity_properties = \
+                _nnunet_normalization_from_model_info(model_info)
+            if model_intensity_properties is None:
+                model_intensity_properties = nnunet_intensity_properties
+        if model_scheme is not None:
+            self.model_normalization_scheme = model_scheme
+            if model_scheme != self.normalization_scheme:
+                print(
+                    'Using model-declared normalization '
+                    f'{model_scheme!r} instead of CLI/default '
+                    f'{self.normalization_scheme!r}.')
+        if model_intensity_properties is not None:
+            self.model_intensity_properties = model_intensity_properties
+
         # model loader returns a dict, network is the actual model
         model = model_info['network']
         model.eval()
@@ -727,6 +828,7 @@ class Inferer():
             normalization_scheme=normalization_scheme,
             global_mean=global_mean,
             global_std=global_std,
+            intensity_props=self.model_intensity_properties,
             input_format=self.input_format,
             verbose=self.verbose,
             mode='infer',
@@ -1163,7 +1265,7 @@ def build_parser():
                       help='Optional: Override patch size, comma-separated (e.g., "192,192,192"). If not provided, uses the model\'s default patch size.')
     parser.add_argument('--save_softmax', action='store_true', help='Save softmax outputs')
     parser.add_argument('--normalization', type=str, default='instance_zscore',
-                      help='Normalization scheme (instance_zscore, global_zscore, instance_minmax, none)')
+                      help='Normalization scheme (instance_zscore, global_zscore, instance_minmax, percentile_minmax, ct, none)')
     parser.add_argument('--device', type=str, default='cuda', help='Device to use (cuda, cpu)')
     parser.add_argument('--num_workers', type=int, default=4,
                       help='Number of DataLoader workers. Use 0 in low /dev/shm environments (e.g. Docker).')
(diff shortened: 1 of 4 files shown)


TAUIL-Abd-Elilah:
@Jinhojeong Thanks for the detailed report. This PR now implements all three issue items: plans-driven normalization/intensity propagation, the explicit inference-error return, and compatibility with both the legacy and current nnU-Net network-builder APIs (including the direct fallback). Both official Zarr test matrices pass, along with 120 relevant local tests.

Could you validate one representative public `surface_m7` volume on this head and confirm that it resolves to `ct` and agrees with your corrected path? One independent spot-check would be more useful than rerunning the full corpus.

Jinhojeong:
@TAUIL-Abd-Elilah, I ran the branch against the harness that found the defect, since #1364 came out of my instruments and the strongest review I can offer is the one that would have caught the original.

Setup. Branch head f74929a shadowed over an installed vesuvius 0.2.4, driven by the same July harness that produced the four-way control in the issue. Same 16 volumes, centred 256 cube mode, m7 fold 0, and no `--normalization` argument anywhere, so the path under test is the default the defect lived in.

Result. The fixed default lands on the fidelity control exactly. Per volume, recall and precision agree with the July arm C, the wrapper fed pre-normalized CT with normalization off, to every stored decimal on all 16 volumes, zero disagreement. Against my independent plans-path predictor the largest per-volume difference is 0.0001, the same order the two paths already differed by in July. Against the old default the medians move from 0.7573 and 0.4418 to 0.9340 and 0.7385, with per-volume shifts up to 0.39. So on the m7 checkpoint the change does exactly what it says, the default now resolves CTNormalization from plans.json and reproduces the correctly normalized numbers without any flag, and nothing else in the path moved. The visibility promise holds too, every predict log carries the line about using model-declared normalization ct instead of the CLI default, which is the message that would have made #1364 findable in one glance.

One scope note for the record. This validates the CTNormalization mapping on one nnU-Net checkpoint through the crop-first mode of my harness. The zscore, none and rescale mappings and the 2.5-versus-2.8 signature handling are outside what this run exercises. The rows and the comparison script are in my diagnostic repo at 2aab3a0 under results/pr1386_check if anyone wants the per-volume table.

aviad12g:
@TAUIL-Abd-Elilah @jrudolph @bruniss — I replayed the PR’s two commits unchanged onto current `main` (`9e120a3c`) in an isolated sparse checkout. The replay is conflict-free; `git diff --check` and Python bytecode compilation pass. The published head’s four GitHub workflows are green, only the unrelated Vercel team-authorization status is red, and Jinhojeong’s 16-volume `surface_m7` validation is already positive. Aviad’s account lacks permission to update the contributor branch. Could the author or a maintainer rebase/update it and take a review pass? No additional feature work is proposed.

PR #1390, opened 2026-08-10
Add ink-constraints to community projects

Nicodol:
Adds ink-constraints to the community projects page: a standalone generator that turns the published ink detections into same-winding constraints for the spiral fit. It targets the "Ink as same-winding evidence" item under "Things we have not tried" on the winding constraints page.

It reads the published ink rasters of the 28 PHerc. Paris 4 bands, extracts ordered point chains, and retro-projects them through the band meshes into the dataset frame, with guards for locality, folds and mesh sheet-switches. The output is villa point-collection JSON (schema v1), consumed by `fit_spiral` as one more `pcl_json_paths` entry, so nothing in villa needs to change. Points within 10 voxels of any held-out evaluation patch are dropped, and every chain is paired with a shape-matched placebo twin placed on ink-free sites of the same mesh, so the contribution of the ink itself can be separated from generic re-import of band geometry. Current output: 3,287 chains and 15,065 points across all 28 bands, with a placebo twin placed for 100% of them.

Whether these constraints improve a fit is an open question, and the repository treats it as one: a three-arm, three-seed campaign was preregistered and sealed (tag `prereg-a1-sealed`) before any comparison run, with judges, margins and claim ladder fixed in advance, and the verdict will be published whatever it comes out to be, negative included. That campaign is still running, so **this entry is for the generator, not for a result.**

Repo: https://github.com/Nicodol/ink-constraints (MIT, 116 tests, CI, no villa changes required)


erdpx:
hi, thanks for the pr. this would be an interesting addition, but we actually need to see the result of its application alongside the generator. we'll consider reopening the pr given you show proof of this working by itself and working within the spiral fit. for contribution guidelines, see villa/CONTRIBUTING.md

PR #1391, opened 2026-08-11
Fix vc_obj2tifxyz TIFXYZ scale semantics (#1319)

### LINKED ISSUE #1319: vc_obj2tifxyz writes scale from the stretch factor, not the actual grid density - breaks vc_flatten downstream
Converting a published segment to tifxyz and then flattening it — two documented pipeline
steps — destroys the surface silently:

| step | grid | valid points |
|---|---|---|
| published OBJ (`paths/20231007101619`) | — | 1 521 895 vertices |
| after `vc_obj2tifxyz … 2000 1.0` | 2001×2001 | **3 532 144** |
| after `vc_flatten` | **6×4** | **18** |

Both tools exit 0 and print success.

**Root cause.** `scale` in a tifxyz `meta.json` means *grid cells per voxel* — the tracer
writes 0.05 because its step is 20 voxels. `vc_obj2tifxyz` instead writes
`1/stretch_factor`:

| tifxyz source | `scale` written | measured step | correct |
|---|---|---|---|
| `vc_grow_seg_from_seed` | 0.05 | 20.07 vox | 0.05 ✓ |
| `vc_obj2tifxyz`, stretch 2000 | **0.0005** | **11.55 vox** | ≈ **0.0865** ✗ |

That is a **173× error**. `vc_flatten` sizes its output as `UV_range × scale`:

- tracer output: 1637 × 0.05 = 82 ✓
- converted segment: ~4800 × 0.0005 = 6 ✗

It also propagates: `vc_cut_windings` correctly preserves metadata, so all five winding
outputs inherit the wrong `scale`.

**Suggested fix:** write the `scale` measured from the grid actually produced, not derived
from the stretch factor.

Reproduced on Windows (MSYS2 UCRT64) with a published Scroll 1 segment. Happy to send a PR.

olgaiv39:
Fixes #1319

`vc_obj2tifxyz` was writing `meta.json.scale` as a spacing-like value, while TIFXYZ consumers interpret it as grid cells per voxel

This computes scale from the final emitted grid as `1 / measured_spacing`, keeps `--tifxyz-source` unchanged, and corrects the TIFXYZ format documentation

Added regression coverage for anisotropic spacing, downsampling, normalized UVs, `mesh_units` independence, and source-scale preservation

### DIFF
--- a/lasagna/tifxyz_format.md
+++ b/lasagna/tifxyz_format.md
@@ -116,7 +116,7 @@ Common fields produced by tools:
 To stay compatible:
 
 - Preserve `scale` when copying/transforming surfaces.
-- If you resample the grid resolution, update `scale` consistently (halving the resolution along an axis halves `scale` on that axis).
+- If resampling doubles the grid-cell spacing along an axis, halve `scale` on that axis.
 
 ## 6. Writing rules (recommended)
 
@@ -145,4 +145,3 @@ A compatible reader should:
 
 - Writer populates `meta.json.format = "tifxyz"` & `meta.json.scale = [sx, sy]`: [`core/src/QuadSurface.cpp:729`](core/src/QuadSurface.cpp:729)
 - Reader loads `x/y/z.tif`, invalidates `Z <= 0`, then applies optional `mask.tif`: [`core/src/QuadSurface.cpp:848`](core/src/QuadSurface.cpp:848)
-
--- a/lasagna/tifxyz_format.md
+++ b/lasagna/tifxyz_format.md
@@ -30,7 +30,7 @@ Any other `*.tif` files are interpreted as additional channels (by filename stem
 	- `Y = y.tif[row, col]`
 	- `Z = z.tif[row, col]`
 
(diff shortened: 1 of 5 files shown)


pmh47:
The `uv-non-metric` case is strictly wrong and not needed. The comment regarding halving resolution needs rewording to be more precise, 'resolution' can either imply a spacing or a count of grid cells. The test is also rather long and verbose for the scope. If these are addressed we'll consider merging.

olgaiv39:
Thanks - addressed. I removed the uv-non-metric case, made the grid-spacing wording explicit, and reduced the regression test from 363 to 110 lines while keeping the normalized-UV and independent scale checks

hendrikschilling:
Looks find to me apart from the test regression coverage:
```
  - --uv-downsample
  - --tifxyz-source scale preservation
  - varying stretch_factor
  - direct verification against the emitted TIFF grid

  The remaining test checks only one dense normalized-UV mesh, anisotropic spacing, and mesh_units independence. Therefore, the PR description’s claim of downsampling and source-scale     
  regression coverage is no longer accurate.
```

PR #1408, opened 2026-08-11
audit_ct_support: audit existing prediction volumes against CT support

axiosdevs:
Companion to #1173 (prevention at inference) and #1156 (masking finalized outputs): a read-only audit for prediction volumes that already exist — how contaminated is one, and is the contamination confined to the blend margin?

`vesuvius.audit_ct_support` adds two modes:

**`chunks`** — zero-download triage. Zarr omits all-zero chunks, so stored chunk keys are an exact map of where each volume holds data; listing them classifies every prediction chunk as supported, inside the one-chunk blend margin, or beyond it, without fetching a voxel. PHerc1203 audits in ~10 seconds: 7,830 chunks, 90.79% supported, 9.21% halo, **0 beyond the margin**.

**`voxels`** — exact per-plane phantom fractions over chunk-aligned z-slabs, read in bounded-memory Y-stripes so every transferred byte is used.

Ran across the full published m7 batch (36 samples, 1,662,405 prediction chunks): 83.1% supported, 16.9% one-chunk halo, 0 beyond in every sample — consistent with the blend-margin mechanism discussed in #1114 and with no second source. `beyond_blend_margin` is the signal to watch: after the fixes land, a healthy volume should report 0 halo as well.

Tests: 10 cases on synthetic local zarrs (exact chunk classification including differing prediction/CT chunk shapes, exact voxel counts, threshold semantics, stride behaviour, grid-mismatch errors). No network. Docs in `vesuvius/docs/audit_ct_support.md`.


pmh47:
Not in scope for our main repo; however we can consider adding a community-projects link to your tool instead if you submit a separate pr.

PR #1413, opened 2026-08-12
tifxyz format doc: fix the reference implementation pointers

PhiLily:
Trimmed per review: #1391 covers the scale semantics, so this now only fixes the reference pointers in section 8, which #1391 leaves as is.

The links were broken in two ways: the paths are relative to the repo root while the doc lives in `lasagna/`, so they 404 on GitHub, and the line numbers predate refactors (the writer is around line 2030 now, not 729). Switched to function names (`QuadSurface::save`, `load_quad_from_tifxyz`) instead of line anchors so they don't go stale again.

### DIFF
--- a/lasagna/tifxyz_format.md
+++ b/lasagna/tifxyz_format.md
@@ -143,6 +143,6 @@ A compatible reader should:
 
 ## 8. Reference implementation pointers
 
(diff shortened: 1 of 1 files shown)


pmh47:
#1391 addresses the lasagna format doc and will hopefully merge soon. Please update this one if anything useful remains, else close

PhiLily:
Trimmed to just the section 8 reference pointers, which #1391 leaves as is: the links 404 (the paths are relative to the repo root but the doc lives in `lasagna/`) and the line numbers are stale. Switched them to function names so they don't rot again. Rebased on current main.

PR #1416, opened 2026-08-12
tracer: record the effective RNG seed in each trace's meta.json

### LINKED ISSUE #1317: Tracing is not reproducible by default, and the two knobs that make it reproducible are undocumented
Running `vc_grow_seg_from_seed` twice with identical inputs produces different
segmentations. On a fixed ROI at `generations: 60` the traced area varied across three runs,
and at `generations: 150` the spatial extent in z varied by nearly 2× with the tracer
stopping at different generations.

The practical cost is that **anyone benchmarking a tracer improvement is measuring against
run-to-run noise larger than most improvements they are likely to make.** It also makes a
bad trace impossible to reproduce for debugging, and rules out regression tests over real
data.

Full determinism is achievable today with no code change. It needs two things together:

```bash
OMP_NUM_THREADS=1 VC_GROWPATCH_RNG_SEED=42 vc_grow_seg_from_seed ...
```

With both set, three runs produced byte-identical `x/y/z.tif` on each platform
(`1118896.531126 vx²` on Windows, `1118896.554750 vx²` on Linux — deterministic per
platform, not across them, as expected from different compilers and libm).

The problem is discoverability:

- `VC_GROWPATCH_RNG_SEED` appears **exactly once in the whole repository**, in
  `core/src/GrowPatch.cpp:83`. It is in no README, no doc, no CLI help.
- Nothing indicates which lever caps the thread count. Both `OMP_NUM_THREADS` and
  `thread_limit` in the params JSON do cap it, but neither is documented as the way to make
  a trace reproducible, and `thread_limit`'s default of `0` means "every core" (filed
  separately).
- Determinism costs roughly 2× wall time (2.24 s → 4.58 s on the measured workload), which
  makes it reasonably a debugging mode rather than the default — but that trade-off is not
  written down either.

Suggestions, in rough order of value:

1. Document the recipe and its cost.
2. Record the effective seed in the trace's `meta.json`, so a segment carries what it needs
   to be reproduced even when the user did not ask for determinism.
3. Optionally a `--deterministic` flag setting both levers and warning about the cost.

Happy to open a PR for any of these.

**Scope of the claim:** this was measured in `seed` mode with an explicit spatial seed.
The `random_seed` and `expansion` modes use `rand()` to pick seed locations; I have not
tested those and make no claim about them.

NicolasHuberty:
Tracing is not reproducible by default, and the two levers that make it reproducible (VC_GROWPATCH_RNG_SEED + a single OpenMP thread) appeared nowhere outside a single line of core/src/GrowPatch.cpp. Anyone benchmarking a tracer change is measuring against run-to-run noise, and a bad trace cannot be reproduced for debugging.

- The tracer now draws one process-wide base seed (the environment value when set, random_device otherwise), exposed as growpatch_rng_base_seed(); per-thread RNGs derive from it by thread ordinal. Single-thread runs with the env seed remain byte-identical to the previous behaviour (thread 0 seeds with exactly the base value).
- vc_grow_seg_from_seed records vc_gsfs_rng_seed and vc_gsfs_omp_max_threads in the trace's meta.json, so every segment carries what is needed to reproduce it even when the user did not ask for determinism.
- docs/tracing.md documents the recipe, its ~2x wall-time cost, the new meta fields, and the scope limit (random_seed/expansion seed picking uses rand() and is not covered).

Verified in the vc3d-deps container: with VC_GROWPATCH_RNG_SEED=42 the base seed is 42 and stable within the process; unset, it is drawn once per process and differs across processes. vc_grow_seg_from_seed and vc_tracer build clean.

Fixes #1317

### DIFF
--- a/volume-cartographer/apps/src/vc_grow_seg_from_seed.cpp
+++ b/volume-cartographer/apps/src/vc_grow_seg_from_seed.cpp
@@ -602,6 +602,8 @@ int main(int argc, char *argv[])
     meta_params["vc_gsfs_params"] = params;
     meta_params["vc_gsfs_mode"] = mode;
     meta_params["vc_gsfs_version"] = "dev";
+    meta_params["vc_gsfs_rng_seed"] = static_cast<uint64_t>(growpatch_rng_base_seed());
+    meta_params["vc_gsfs_omp_max_threads"] = omp_get_max_threads();
     add_target_context(meta_params, vol_path);
     if (mode == "expansion")
         meta_params["seed_overlap"] = count_overlap;
--- a/volume-cartographer/core/include/vc/tracer/Tracer.hpp
+++ b/volume-cartographer/core/include/vc/tracer/Tracer.hpp
@@ -29,3 +29,8 @@ struct DirectionField
 QuadSurface *grow_surf_from_surfs(QuadSurface *seed, const std::vector<QuadSurface*> &surfs_v, const utils::Json &params, float voxelsize = 1.0);
 QuadSurface *grow_surf_from_surfs(QuadSurface *seed, const std::vector<QuadSurface*> &surfs_v, const utils::Json &params, float voxelsize, SurfacePatchIndex* surface_patch_index);
 QuadSurface *tracer(Volume& volume, float scale, int level, cv::Vec3f origin, const utils::Json &params, const std::string &cache_root = "", float voxelsize = 1.0, const std::vector<DirectionField> &direction_fields = {}, QuadSurface* resume_surf = nullptr, const std::filesystem::path& tgt_path = "", const utils::Json& meta_params = {}, const PointCollections &corrections = PointCollections(), const cv::Mat* allowed_growth_mask = nullptr);
+
+// Effective base seed of the tracer's RNGs: VC_GROWPATCH_RNG_SEED when set,
+// otherwise drawn once per process. Stable for the process lifetime; with a
+// single OpenMP thread the trace is fully determined by it.
+uint32_t growpatch_rng_base_seed();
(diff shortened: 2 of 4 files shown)


tonclap:
Independent confirmation, plus one thing the PR leaves behind.

I was in `GrowPatch.cpp` for an unrelated reason and reached the same two conclusions this
PR implements. Recording them because "someone else read it and agrees" is worth something
on a reproducibility change:

- **the seed alone cannot work, structurally.** `thread_rng()` is `thread_local`, and its
  only consumer `random_perturbation()` is called at one site inside `#pragma omp parallel`,
  where threads pull from a shared candidate pool. Each stream therefore advances by however
  many candidates that thread happened to take, which OpenMP decides afresh each run. Your
  docs saying a single thread is *required* rather than recommended is, I think, the most
  important line in the diff.
- **seeding every thread from the same value changed the jitter's character**, from
  independent per-thread noise to N copies of one sequence. `base + thread ordinal` is
  exactly the fix I would have proposed.

**Leftover:** `set_random_perturbation_seed()` (`GrowPatch.cpp:127`) still has zero call
sites — a repo-wide search returns only the definition, which is what the `[[maybe_unused]]`
is silencing. It was harmless dead code before this PR. With the ordinal derivation it is a
small footgun: calling it reseeds *only the calling thread*, colliding with what the other
threads derived, and would silently break the base-seed → trace correspondence that the new
`vc_gsfs_rng_seed` field promises. Wiring it to `growpatch_rng_base_seed()` or deleting it
would close that.

**Scope note, not a defect**, since your docs are careful here: `thread_ordinal.fetch_add(1)`
assigns ordinals in first-call order rather than by OpenMP thread number, so which thread
gets which stream can differ between runs at N > 1. That contradicts nothing you claim —
reproducibility is documented as single-thread only — but it is worth knowing if anyone
later tries to relax that requirement.

PR #1418, opened 2026-08-12
Fix SmearTransform ownership, speed it up, add invariant tests (#201)

blackemberlabs:
Fix for `SmearTransform`'s ownership contract, a 1.3-2x speedup with identical
output, and a first test suite for `models/augmentation`.

## What

- `SmearTransform._apply_to_image` mutated its input while returning a clone. The
  clone was a decoy: the caller's tensor was already written through the `img[ch]`
  and `.permute()` views, and the returned copy hid it, so no caller could tell
  which tensor owned the result. Now it works in place and returns `img`, like most
  of the module's other intensity/noise transforms, and the wasted full-tensor
  clone is gone.
- The loop now batches channels and rolls each smeared slice once instead of once
  per window it appears in. That makes it 1.3-2x faster, same output.
- `num_prev_slices < 1` now degenerates to uniform `(1 - alpha)` scaling, matching
  the original behavior for an empty window; mid-rewrite this case crashed. Covered
  by a test.
- Added `tests/models/augmentation/test_transform_invariants.py`. The module had no
  tests before this. Besides contract invariants, the suite embeds the pre-rewrite
  smear loop verbatim as a pinned reference and checks the new output against it
  bit-for-bit, so the equivalence claim runs in CI instead of living in an offline
  script.
- Removed an unused `skimage.data.camera` import in `gaussian_blur.py`.

## Why

One SmearTransform call costs about as much as the rest of the default 3D pipeline
combined (9.7 ms vs 7-9 ms/sample at 128³, CPU). At the default
`apply_probability=0.3` (`pipelines/training_transforms.py:400`) that is a quarter
to a third of the expected augmentation cost, rising to 115-188 ms per hit at
ps192/ps256. This change cuts it by 25-35% and drops a full-tensor allocation per
call (64 MiB transient at ps256).

The ownership issue is a contract problem, not a data-corruption one: the old code
already modified the caller's tensor and the new code modifies it identically, but
the old code returned a different object while doing so. In-repo training is
unaffected either way, since the datasets build fresh tensors per sample and use
the returned dict. The numpy ancestor of this transform in
`segmentation/models/batchgeneratorsv2` has the same hybrid pattern; I left it
alone to keep this change to one package, and can follow up there.

The transform reads already-smeared slices when computing later ones, and that
recursion is preserved exactly. An FIR/conv1d rewrite (the route #1072 took for
Decohesion, where outputs stayed equivalent) is not output-preserving here because
of that recursion, so I left that call to maintainers.

### Equivalence

Identical output to the previous implementation (`torch.equal`, max err 0) across
1080 cases: shifts x alphas x window sizes x axes x {2D,3D} x {float32,float16} x
3 seeds. The same comparison now ships in the test suite against the embedded
reference. Valid window sizes are `num_prev_slices >= 1`; below that the transform
degenerates to the uniform scaling described above, also under test. Also checked
on a real 64³ Scroll 1A crop (`54keV_7.91um_Scroll1A.zarr`): bit-identical there
too.

fig1_hybrid_mutation_real_crop

### Speed

Measured on CPU, Apple M-series, torch 2.13, median of 3-7 runs:

| shape | before | after | speedup |
|---|---|---|---|
| 1x128x128x128 | 9.6 ms | 6.4 ms | 1.50x |
| 1x192x192x192 | 114.6 ms | 86.4 ms | 1.33x |
| 1x256x256x256 | 187.5 ms | 134.6 ms | 1.39x |
| 2x64x64x64 | 3.9 ms | 1.9 ms | 2.04x |

fig2_smear_effect_real_crop

### Tests

152 tests, under 2 s, CPU-only, no zarr usage and no network access. Five invariants
over a registry of 33 transform cases: ownership (in place or out of place, never
both; fails on the previous SmearTransform, passes after this PR), image-only
transforms leave segmentation bit-identical, dtype/shape/device preserved, identity
at `apply_probability=0`, and registered params actually firing, with cases that
can legally no-op (Mirror can sample zero flip axes, like Rot90) waived with a
reason. A completeness check reports unregistered transforms as a warning without
failing anyone's build. Plus the SmearTransform reference grid above. Runs under
the repo's default pytest selection on both zarr matrix entries (verified locally
on 2.18.7, 3.2.1, and 3.3.0).


blackemberlabs:
Self-review after opening this turned up problems. Fixed in 9ab6dbd, description
updated to match:

- The rewrite crashed on `num_prev_slices < 1`, where the old code returned a
  uniformly scaled tensor. Restored the old result and added a test.
- The equivalence claim now has teeth in-repo: the old loop is embedded in the test
  file as a reference and the suite checks the new output against it bit-for-bit.
  Previously that check only existed in an offline script.
- The Mirror has-effect case could legally sample zero flip axes and fail on an
  unlucky seed. Waived with a reason, like Rot90.
- The registry completeness check now reports unregistered transforms as a warning
  instead of failing the build, so it never blocks unrelated work.
- Description corrections: the ownership issue is a contract problem, not silent
  corruption (the old code already wrote through views into the caller's tensor;
  the returned clone hid it), a stale test-file count and a wrong line reference
  are gone, and the #1072 reference is fixed (that rewrite kept outputs equivalent;
  this one would not, because the smear reads already-smeared slices).

Issue #1403, opened 2026-08-11
vc_grow_seg_from_seed silently deletes the grown surface and exits 0 when the volume is an s3:// path


DarthCeltic:
Growing a segment straight from an `s3://` volume produces no output, no error, and exit code 0.

`vc_grow_seg_from_seed` reads the voxel size at line 326:

```cpp
float voxelsize = json::parse(std::ifstream(vol_path/"meta.json"))["voxelsize"];
```

The zarr data streams from S3 fine, but that's a plain `ifstream` against an `s3://` path, so `voxelsize` doesn't get the real value. `area_cm2` then computes to 0.000000, and at line 1371 that fails the `min_area_cm` check (default 0.3), which calls `remove_all(seg_dir)` and returns `EXIT_SUCCESS`:

```cpp
double area_cm2 = (*surf->meta)["area_cm2"].get<double>();
if (area_cm2 < min_area_cm) {
    if (std::filesystem::exists(seg_dir)) {
        std::filesystem::remove_all(seg_dir);
    }
    return EXIT_SUCCESS;
}
```

The surface is deleted and the run reports success. It took me a while to work out that the grow had actually worked — 99 generations, about 15.6M vx² of surface, and an empty output directory.

Setting `{"min_area_cm": 0.0}` gets the mesh written, but the area recorded in its `meta.json` is still wrong for the same reason.

Same root cause shows up in a second tool with a different symptom: `vc_render_tifxyz` against `s3://vesuvius-challenge-open-data/PHerc0009B/volumes/20250521125136-8.640um-1.2m-116keV-masked.zarr` prints

```
Voxel size: 1.0 (no metadata found; override with --voxel-size)
```

and carries on with the wrong scale rather than deleting anything.

Reading that `meta.json` through the same S3-aware path the volume already uses would fix both. I haven't sent a patch — I don't know whether anything downstream relies on the current fallback behaviour.

Bullo27:
Confirming both symptoms, with a version detail that should simplify triage — the two halves of this report have different current status:

[...]

PR #1419, opened 2026-08-12
vc_render_tifxyz: document that --slice-step is in level-0 voxels

pscamillo:
## Motivation

I was preparing input for the 9µm models on Scroll 1667, which publishes no surface volumes, so the input has to be rendered from the scroll volume. The model returned nothing legible (images below) on a segment that is part of its own training set. It took a night of eliminating hypotheses (normals orientation, z window, slice count, z pooling, mesh source) before comparing against the official surface volume of another segment showed the band was a quarter of the intended depth.

## What the flag does today

The offsets are built as `(zi - center) * sliceStep * dsScale` with `dsScale = 2^-group_idx` (`buildOffsetList`), so the step is in level-0 voxels. That is what makes the fractional steps in the ink tutorial work, and it is consistent across the tool.

The unit is not stated where the flag is described:

--scale arg Pixels per level-g voxel (Pg)
--slice-step arg (=1) Spacing between slices along normal

`--scale` names its reference level. `--slice-step` does not, and the natural reading next to `--group-idx` is that both refer to the level being read. The ink tutorial only shows `--group-idx 0`, where the two readings coincide.

The file already documents a subtler convention right below, in the comment block explaining why `--flip-normals` is usually what you want. This is the same kind of note, one line shorter.

## Before / after on real data

Same segment, same mesh, same volume, same checkpoint (ink_9um hybrid_3d2d-seed42 step-075000, `--overlap 0.50 --blend-mode hann`), scored as median windowed correlation against the published `new_canon_autoresearch_recipe` prediction of the same segment:

| render of PHerc1667 w029 | `--slice-step 1` | `2^group-idx` |
|---|---|---|
| 2.399 µm at level 2 (9.6 µm) | 0.037 | 0.518 |
| 1.129 µm at level 3 (9.03 µm) | -0.050 | 0.246 |

PHerc1667 w013, same crop, same display rescale `(p - 64) / 128`:
pr_step1_vs_step4

There is no error to screenshot. The render with the wrong step has the correct raster, normal non-zero coverage, visible fibres and correct patch counts against the labels.

## How it was identified

Comparing slice by slice against the official surface volume of the same segment, the best-matching slice advances by exactly 4 per official slice:

official z=10 -> ours z= 2 r=0.956
official z=13 -> ours z=14 r=0.944
official z=16 -> ours z=26 r=0.960

Re-rendered with `--slice-step 4`: r = 0.953 / 0.935 / 0.958 at three depths, against 0.92 in the middle and ~0 at the edges before.

As a control, a level-0 render of PHerc0139 w043 from the published mesh and volume reproduces the official surface volume exactly (r = 1.0000, identical shape), which is where `--slice-step 1` is trivially correct.

## Alternatives

A warning when `--group-idx > 0` and `--slice-step` is not `2^group-idx` would also catch it, but it would fire on deliberate fractional sampling. Happy to word the help line differently, or put the note somewhere else, if you prefer.

Related, same tool, same class of silent scale problem: #1417, #1403.


erdpx:
hi, thanks for the pr! were these tests run from the merge-ink-pipelines branch? this problem should’ve been fixed on main by [#1146](https://github.com/ScrollPrize/villa/pull/1146), where slice-step 1 already moves one voxel at the selected group. we’re currently moving the work from merge-ink-pipelines into main, so main will be the version to use going forward; if you want, you can try the current main version of the script to see if it fixes the problem you encountered

pscamillo:
thanks, yes, merge-ink-pipelines, which is what the ink_9um README points to. just checked main, buildOffsetList doesn't take dsScale there anymore, so #1146 covers it. closing

PR #1421, opened 2026-08-12
Add absolute-winding to community projects

tonclap:
Adds one entry to the community-projects list.

**[absolute-winding](https://github.com/tonclap/absolute-winding)** measures the absolute
origin of the winding field on PHerc. Paris 4 — the quantity
[winding-sync](https://github.com/abundantjoe/winding-sync) names in its own entry above as
"not yet calibrated", and that constraint-gauge explicitly does not score ("winding may be
relative, only differences are scored").

Result: `absolute winding = laminae counted from the umbilicus + 6 (+1/-2)` over z ~ 3100-15700.

Two things I would want to know as a reader, so they are in the entry rather than only in the
repository:

- **the tool refuses slices it cannot serve.** The regression slope needs no absolute labels,
  so it works as a self-check: on good slices it sits near 1, and at the top of the
  `z3000_18000` fit domain it collapses to 0.465 and the run prints `SLOPE GATE FAILED`. A
  second estimator diverges on the same slice.
- **the ground truth is one source, not two.** The segment names `w010-027` ... `w128-129` are
  the spiral fit's numbering, and the fit consumes absolute winding annotations as an input
  while no other input carries an absolute number — so those names are the 59 hand-placed
  `abs_winding.json` labels propagated. The -5.5 and -7 figures are one truth and its spread,
  not two measurements agreeing.

Every hit rate is reported against a shuffled-label null with the constant reselected inside
each permutation (0.277 measured against 0.064). Limits are stated in the README: per-point
absolute numbers are not recovered, the slope-gate bounds were set after seeing the data, and
this is one scroll.

Runs on CPU and network only, no credentials and no bulk download. Inputs come from two
different hosts, which the README says explicitly because it is not guessable: the prediction
and segment meshes from the S3 bucket, the umbilicus from `dl.ash2txt.org`.

I have not touched the winding-sync entry, though its closing sentence now has an answer —
that seemed like @abundantjoe's call rather than mine.

AI disclosure: code, measurements and English drafting were done with an AI assistant working
from my instructions; details in `AI_DISCLOSURE.md` in the repository.


bruniss:
we will consider merging this if you can provide any true tests that it benefits current tooling (fit_spiral.py for example) or any accompanying images or demonstrations of accuracy. i would not rely on constraint-gauge or the other repo you mentioned. if you want assistance with figuring out how to tell if its accurate send a picture of what some of these look like

tonclap:
Thanks — that was the right thing to ask for, and it turned out to be worth more than a merge.

**The pictures.** This is the one you asked for: 300 voxels of a single ray at z = 15694, at 1:1. Raw CT on top, the `th0.45` recto prediction the counting runs on underneath, one orange tick per lamina the ray enters, and the number each tick implies. The open circles are the hand-placed points from `abs_winding.json`.

fig3_zoom

The whole ray, with the winding intervals of every published segment it crosses drawn as blue bars — the staircase has to thread them across 120 windings, and a wrong constant slides it off:

fig2_ray

And where the ray is:

fig1_slice

The pictures come out of the published script itself, imported as a library; the number of ticks drawn is asserted against `count_to_point`, so the figure cannot quietly disagree with the tool. Everything below is in [tonclap/laminae-audit](https://github.com/tonclap/laminae-audit) — figures in [`figures/`](https://github.com/tonclap/laminae-audit/tree/main/figures), drawn by [`make_figures.py`](https://github.com/tonclap/laminae-audit/blob/main/make_figures.py).

**Accuracy, against your own data rather than mine.** Point taken about constraint-gauge and the other repo — neither is used below. `fit_spiral.py` learns where absolute winding zero is from `abs_winding.json` (`get_patch_abs_winding_loss` pins each annotated point's shifted radius to `winding_annotation * dr_per_winding`, and nothing else it consumes carries an absolute number). That file holds 59 hand-placed points, 48 of them consecutive windings 16–63 on one slice. Counting laminae from the umbilicus to each of them, no fitting, no tuning:

- slope 0.956, r 0.9863
- median implied constant +7, IQR 1.5
- at C = +7: exact 39 %, within ±1 70 %, within ±2 86 %
- the same statistic with the labels shuffled among the same points: 6 %

fig5_labels

It is a harder test than the one I published — exact integers instead of 2-to-18-winding intervals, one prediction against one label. The check is [`check_abs_winding.py`](https://github.com/tonclap/laminae-audit/blob/main/check_abs_winding.py).

**What drawing this broke.** The slice figure above shows a fan of published segment mesh points sitting in empty space past the edge of the scroll, and my calibration was scoring them: it filtered only the tifxyz invalid marker. A ray aimed at a point that is not on the scroll counts the wrong laminae. The volume is `…-masked.zarr`, so `CT == 0` is an objective test, and the damage is not uniform — it is 0 % below z = 12000, 23 % at z = 15694 and **52 % at z = 18000**.

That has a second reading which is about the data rather than about me, and it may be the more useful half. Counting over **every** mesh point within ±3 voxels of a slice — about 12 000 per slice, not my 448 — the published segment meshes sit outside the scan mask 0.0 % of the time at z = 9000, 7.5 % at z = 12000, 21.0 % at z = 15694 and **51.0 % at z = 18000**, and at the top it is not only the outer edge: `w038-045` is 36 % outside, `w128-129` 84 %. If that is news, [`check_mesh_outside.py`](https://github.com/tonclap/laminae-audit/blob/main/check_mesh_outside.py) is the whole check.

Two of my published claims do not survive it:

- I wrote that counting tracks the fit's numbering close to 1:1. On clean samples the slope is 1.09–1.18 above z = 6000 — it drifts by 10–18 %, and the 0.978 I published was that drift being cancelled by off-scroll points that undercount. Withdrawn.
- I made a point of the tool refusing z = 18000 on its own, slope 0.465. Drop the off-mask points there and the slope is 1.184 and the hit rate 0.274 against a 0.107 null — the slice stops failing. The gate fired, but not on what I said it did. Here is that slice; the scroll is narrower at the top and the segment meshes run past its edge:

fig6_gate_slice

I have put the filter into the tool (`--no-mask-filter` reproduces the old runs exactly, so the refactor moved nothing), and the corrected sweep is what I will quote rather than the diagnostic — the first version of it flattered me. On z = 15694 the honest effect is: the hit rate does not improve, the outer-band dispersion collapses from IQR 20 to 7 — so my published "dispersion grows outward" was mostly this contamination — and the two estimators now disagree by 5.2 counts, which by my own rule makes that slice the doubtful one rather than the flagship. Separately, three rows of my results table came from an earlier version of the script with half the sample per segment and the shipped standalone does not reproduce them; that is corrected in the README with a dated note rather than quietly edited. All of it is in [`CORRECTION.md`](https://github.com/tonclap/laminae-audit/blob/main/CORRECTION.md), and the repository was renamed to `laminae-audit` because the name `absolute-winding` promised the thing that did not survive.

**On benefiting the tooling.** I have not run the fit, so I will not claim it. What I can say precisely is what the measurement supports: the fit's absolute origin rests on 59 hand points, 48 of them on one slice, and counting reproduces those numbers to ±1 seven times out of ten from the CT and the umbilicus alone. That is enough to audit them, and it is available in places where nothing else is — the published tracks and verified patches all lie in z 4500…17498 while the fit domain is named `z3000_18000`, and in that unlabelled bottom band the laminae are no less separable in the CT than at z = 5500.

If that is worth having, tell me which of the two is more useful to you and I will build the one you name rather than both. And I would still take you up on the offer: looking at the zoom figure, is the tick placement what you would call correct, and is there a failure mode in it you can see that I have not measured?

*Disclosure: produced with substantial AI assistance (Anthropic Claude) under human direction and review — the same disclosure as [`AI_DISCLOSURE.md`](https://github.com/tonclap/laminae-audit/blob/main/AI_DISCLOSURE.md) in the repository.*

PR #1422, opened 2026-08-12
VC3D: deterministic line-annotation view orientation

skyward7187:
Manifest normals are unoriented axes (tensor-derived), so several line-annotation view orientations were coin flips that could differ per fiber and change across rebuilds. This PR makes them deterministic in three coupled steps.

## Bottom strip and display ups: orient in the builder

The first commit fixed the bottom strip with a post-hoc row flip after `buildLineViewSurfaces`. The second commit replaces that with orientation injected into the builder itself: `LineViewConfig` gains an optional `orientedPointNormals` field (per-control-point sheet normals; NaN entries allowed). When supplied, one whole-line cosine-weighted vote flips the frame mesh normals (and their side vectors, preserving `side == meshNormal × tangent`), and a second vote flips the transported up vectors.

This fixes two things the post-hoc flip could not:

- **The dual-anchor coin flip.** `buildFrames` anchors its sign at `samples.size()/2` while `buildTransportedUpVectors` anchors at `displayFrameAnchorIndex`, which moves as the line grows — so the ribbon sign and the up-vector sign could disagree and flip per rebuild. With hints supplied, both are pinned by the same reference, and `lineZSlices` (built from the ups) stay consistent.
- **Intersection strips.** `prepareSide` registers the builder's `lineSideSlice` directly for cross-fiber inspection; it now passes the same oriented normals, so h/v intersection strips are oriented like the dialog's bottom strip.

Without hints (empty, size-mismatched, or all-NaN) the builder's output is bit-identical to before — the headless `vc_lasagna_line_probe` paths are unchanged.

## Cut panes: pin the displayed tangent sign

The current cut's screen left/right and the side cut's vertical follow the stored point-index direction, which is arbitrary per fiber (seed direction inherits the slice-plane and manifest normal signs) and gets reversed by merges. A new per-fiber sign, `generatedDisplayTangentSign`, is computed once per materialization: majority vote of `(n̂ × t̂)·ẑ` over the line (with the umbilicus-oriented normals), falling back to `t̂·ẑ` for fibers running along the scroll axis, tie → +1. It multiplies `interpolatedLineTangent`, so both cut planes inherit it uniformly.

Convention: +z (increasing slice index) renders to a consistent screen side of the current cut for circumferential fibers; vertical fibers get a consistent side-cut vertical. Stored line points, control points, and links are never modified — this is display-time only.

## Side effects (accepted)

- The top strip flips together with the bottom strip when the vote flips (its in-sheet row direction has no orientation convention of its own; making one is out of scope).
- `prepareSide` now computes oriented normals per side, which may trigger the (cached) umbilicus load slightly earlier in a session.

## Testing

- `test_lasagna_line_view_surfaces`: existing exact-sign cases pass unchanged with no hints; new cases cover the flip against opposing hints (strips, ups, z-slice bases), agreeing hints (no-op), sparse-NaN hints, all-invalid/wrong-length degradation, and a 5-point alternating-normal fixture proving the moving display anchor no longer changes the ups once hints are supplied (the legacy disagreement is asserted first).
- `test_line_annotation_generated_views`: `generatedDisplayTangentSign` on a circle fiber (both windings), an axial fiber (both directions, exercising the fallback), and degenerate inputs.
- VC3D builds and links; new assertions were mutation-checked during development (inverted expectations fail where expected).
- A local Codex review of the branch caught a tie-band sizing bug in `generatedDisplayTangentSign` (sparse valid normals on a long fiber drowned out the primary vote); fixed in the third commit with a regression test verified to fail against the pre-fix helper.


hendrikschilling:
```
P2: the axial fallback is too narrow. volume-cartographer/apps/VC3D/LineAnnotationGeneratedViews.hpp:465 uses the circumferential vote whenever its average magnitude exceeds 0.001,   
     regardless of whether the fiber is predominantly axial. A nearly vertical fiber with slight helical drift can therefore flip its side-cut vertical based on the drift direction. The   
     test only covers perfectly axial fibers. The choice should consider the relative axial and circumferential magnitudes.
```

This means in practice all fibers probably use the circumferential case which is probably not the intention?

skyward7187:
Good catch — the fallback was indeed unreachable for any real vertical fiber. Fixed in 1da72583c: the two votes now compete and the larger per-vote mean decides (switch at ~45° pitch). Means rather than raw sums because the votes count different populations (the circumferential vote only accumulates where a sampled normal is valid, so a raw fallback sum over every tangent would drown out a decisive circumferential vote on a sparse-normal fiber).

Added tests: near-axial helices with opposite drift chiralities must both read as ascending (these fail against the old code), plus a shallow-pitch helix pinning the circumferential side of the boundary.

PR #1423, opened 2026-08-12
fibers: retire the interp_unreviewed review workflow

skyward7187:
# fibers: retire the interp_unreviewed review workflow

## What

Replaces the specialized trace-review mechanism with plain `reviewed`-tag semantics, and makes the interpolation-mode switch the one action that invalidates a review.

- **Mode switch strips `reviewed`.** Switching a fiber's interpolation mode (toolbar combo, either direction) re-optimizes the whole line, so the save that follows removes any `reviewed` tag. Failed or cancelled switches leave it untouched (`stripReviewedTagOnSave` is set only on the successful result, guarded by `restoreFiberOptimizationModeOnFailure`, which only the mode-switch handler sets). Ordinary CP edits, merges, and splits never touch the tag.
- **`interp_unreviewed` machinery removed**: the reserved tag + `applyTraceReviewTags`, the "Mark trace verified" / "Mark as needs review" context-menu actions, the derived `traceVerified`/`traceNeedsReview` fields (fiber panel + agent-bridge `fiber.list`), the reserved-tag filters in the tag checkboxes/pills, and the merge-time re-application in `fiber_merge.py`. A stale `interp_unreviewed` on straggler files now shows as an ordinary, hand-removable tag.
- **`reviewed` pinned in the tag UI**: `kReviewedTag` (new, `LineAnnotationFiberSegments.hpp`) is seeded as always-known and listed first in the fibers-dock checkboxes and dialog pills, even in a package where no fiber carries it yet.
- **Kept unchanged**: the `interp` provenance column (`deriveTraceState`), `hasAcceptedTraceSpan`, all `needs_reoptimization` machinery, and `vc_sync.py` (hfsync stays a plain tag gate).
- **One-time cleanup script** `scripts/fiber_strip_stale_review_tags.py`: strips `interp_unreviewed` everywhere and `reviewed` from fibers without an accepted trace span; atomic writes, generation bump, dry-run, idempotent.

## Why

`reviewed` verdicts issued in the lasagna era did not survive fiber-model re-tracing, and the two parallel tag namespaces (`reviewed` for publish, `interp_unreviewed` for trace review) let hfsync publish unverified traced geometry while 224 stale lasagna-reviewed fibers were poised to re-upload over the hand-pruned HF bucket.

## Data migration (already applied to PHercParis4)

430 fibers: `reviewed` stripped from 224 untraced, `interp_unreviewed` stripped from 82; end state 111 reviewed / 0 `interp_unreviewed`; second run is a no-op. `vc_sync.py hfsync --dry-run` now reports 111 upload candidates, 0 removals, 0 invalid.

## Testing

- Full ctest suite green (148/148) including `test_fiber_trace_review`, `test_fiber_widget`, `test_line_annotation_*`; `scripts/tests/` 146/146.
- Grep gate: no references to any removed symbol remain (the cleanup script's local `interp_unreviewed` literal is the deliberate exception).
- Cleanup script exercised on synthetic fixtures (strip correctness, dry-run parity, idempotence, tags+generation-only diffs) before the live run.
- Not automated: a controller-level test for the strip (no existing harness compiles `LineAnnotationController.cpp`); covered by GUI smoke instead.


hendrikschilling:
GUI and sync states are a hell of of thing to work with :-P

 ```
 1. High: merged geometry can become reviewed without anyone reviewing it.
     volume-cartographer/scripts/fiber_merge.py:1119 now merges reviewed like an ordinary tag, independently of geometry.

     Confirmed scenario:
      - base is unreviewed;
      - local changes geometry;
      - remote reviews the unchanged base;
      - merge selects local geometry but retains remote’s reviewed tag.

     My reproduction produced merged_tags=['reviewed'] while using local’s modified geometry. The previous special merge logic and its regression test explicitly prevented this. This could
     publish geometry nobody reviewed.

  2. High: the cleanup script removes valid reviews from every untraced fiber.
     volume-cartographer/scripts/fiber_strip_stale_review_tags.py:74 removes reviewed whenever _has_trace_span() is false. That includes all version-1 fibers and ordinary Lasagna fibers.  

     I confirmed that a valid v1 fiber tagged reviewed is accepted by fiber_merge.is_fiber_doc(), but the cleanup returns no tags. Since reviewed is the hfsync publish gate, running the   
     cleanup can subsequently unpublish legitimate legacy fibers. This also conflicts with the new GUI, which deliberately allows users to mark any fiber as reviewed.

  3. Medium: the migration does not preserve old verified state.
     Previously:
      - traced + interp_unreviewed = needs review;
      - traced without interp_unreviewed = verified.

     Now:
      - reviewed present = verified;
      - reviewed absent = unreviewed.

     The cleanup removes interp_unreviewed, but never converts an old verified traced fiber into reviewed. Therefore fibers verified through the old context-menu action become unreviewed  
     unless they independently already carried reviewed. This may be intentional review invalidation, but it is not a state-preserving migration and should be explicit.

  4. Medium: fiber.list breaks protocol-v1 compatibility.
     volume-cartographer/apps/VC3D/agent_bridge/AgentBridgeHandlers_fiber.cpp:220 removes traceNeedsReview and traceVerified, while volume-cartographer/apps/VC3D/agent_bridge/
     AgentBridgeServer.hpp:51 remains 1. Existing protocol-v1 clients can therefore receive an incompatible response without detecting a version change. Retaining deprecated fields or     
     incrementing the protocol version would avoid that.

  5. Medium, error path: a failed mode-switch view rebuild does not fully roll back.
     volume-cartographer/apps/VC3D/LineAnnotationController.cpp:8591 installs the new optimized line before generated-view materialization. If materialization fails at line 8643, the mode 
     and control points are restored, but the optimized line is not, and stripReviewedTagOnSave was never set. This can leave mixed session state with the old review verdict. The
     incomplete line rollback existed already, but the new placement introduces the review-tag consequence.
```

**Comment by skyward7187** [comment]
Thanks — all five were on target. Pushed fixes for 1, 4, and 5; 2 and 3 are intended behavior. Point by point:

**1 (fixed):** `merge_fibers` now drops `reviewed` unless every piece of the merged geometry comes from a side that carried the tag: exact whole-geometry match against a reviewed side, or, for v3 span merges, every spliced run owned by a reviewed side (runs identical on both sides count as covered by either side's review). So your repro — tag-only review of the old geometry merged against a local edit — no longer ships as reviewed, while a splice of two independently reviewed edits keeps the tag (deliberate policy change from the old interp_unreviewed model, which demanded a fresh review there). Four new tests cover both directions.

**2 (intended):** `reviewed` now means "reviewed under fiber-model interpolation" — that's the point of the change. The lasagna-era verdicts were deliberately invalidated (they're what kept re-uploading the 224 files over the hand-pruned HF bucket). The script is a one-time migration and has already been run against PHercParis4; the live directory is all-v3, so the v1 case doesn't arise there. A user can still hand-tag any fiber `reviewed` afterwards through the GUI, which is the intended escape hatch.

**3 (intended):** old "verified" (traced without `interp_unreviewed`) was a trace-QA state, not publish approval — auto-converting it to `reviewed` would have granted publication nobody signed off on. The 111 fibers that were both verified and publish-tagged kept `reviewed`; the rest correctly start unreviewed under the new semantics.

**4 (fixed):** bridge protocol bumped to 2, `vc3d-mcp` expects 2, and SPEC.md now carries a version history line for the `fiber.list` field removal.

**5 (fixed):** the failure path now restores the optimized line and branch anchors alongside the control points (the snapshots travel together), so a failed materialization can no longer leave rolled-back control points paired with the new line — and the same fix covers the span-goal-change path, which had the identical pre-existing hazard. The strip flag continues to be set only after the last failure exit, so the tag still survives any rollback.

**Comment by hendrikschilling** [comment]
```
  1. Medium: failed optimization rollback is still incomplete.
     applyOptimizationTaskResult() updates reciprocal branch metadata in other open sessions and _fibers at volume-cartographer/apps/VC3D/LineAnnotationController.cpp:8631, before         
     generated-view materialization can fail at line 8674. The rollback at line 9137 restores only the active session’s controls, line, and branches. Peer fibers retain anchors derived    
     from the rejected geometry and may persist them on a later save.

     A late materialization failure can also remove the old generated surfaces at line 9525 and then remove the replacements at line 9651. Restoring the model does not recreate those      
     surfaces, leaving the pane blank or stale.

     Cross-fiber synchronization should be deferred until materialization succeeds, and failure should rematerialize the restored model.
```

skyward7187:
Fixed in 42db5f2dd, along the lines you suggested:

- **Cross-fiber sync deferred past the failure window.** Before the view rebuild, `applyOptimizationTaskResult` now runs only the session-local `remapBranchControlPointIndices` (materialization reads `session.branches` for the link markers, so the self-remap can't wait). The peer-mutating phases of `syncLinkedBranchMetadataAfterFiberModification` — reciprocal cleanup and endpoint refresh across other sessions and `_fibers` — run after the last failure exit, so a rejected result never touches peer fibers at all, instead of needing a rollback that can't reach them.

- **Failure rematerializes the restored model.** `finishOptimization` captures `task.ok` before the apply, so the rollback can distinguish a materialization failure (state was installed, surfaces torn down) from an optimizer failure (nothing touched, surfaces intact). In the former case it re-runs `materializeGeneratedViews` on the restored controls/line/branches, best-effort — that geometry rendered before, so the panes come back instead of staying blank.

Not addressed (deliberately): the smaller session fields the apply mutates before the failure exit (`optimizationReport`, `seedPoint`, focus) — cosmetic, and pre-existing.

hendrikschilling (approved):

PR #1426, opened 2026-08-12
Fix volume-only catalog imports

### LINKED ISSUE #1329: list_files() fails with "NoneType object is not callable" on a volume-only install: nest_asyncio is missing from that extra
## Summary

With exactly the packages listed in the `volume-only` extra, `vesuvius.list_files()` — the "Data listing" feature the README advertises — fails with an error that says nothing about the cause:

```
>>> from vesuvius import list_files
>>> list_files()
```

## Why

`vesuvius/utils/catalog.py` imports `nest_asyncio`, which is **not** part of `volume-only` (it appears only under `models`, `notebooks` and `blending`). The import therefore fails, and `__init__.py` swallows it:

```python
try:
    from . import utils
    from .utils import is_aws_ec2_instance, list_cubes, list_files, update_list
except Exception:
    utils = None
    is_aws_ec2_instance = None
    list_cubes = None
    list_files = None
    update_list = None
```

The names stay exported in `__all__`, so nothing looks wrong until the call, and then the traceback points at the call site rather than the missing dependency. The comment above that block says users "should install the appropriate optional extra", but the user never sees that comment — and it isn't obvious which extra is the appropriate one for a listing helper.

## Reproduction

Install exactly the `volume-only` set (plus the two base packages that aren't ML: `cachetools`, `edt`), then:

| import | result |
|---|---|
| `import vesuvius` | ok |
| `vesuvius.data.volume` | ok |
| `vesuvius.data.utils` | ok |
| `vesuvius.utils.catalog` | **ModuleNotFoundError: nest_asyncio** |

Adding `nest_asyncio` makes `list_files()` work.

## Suggested directions

Two separate things, and they're independent:

1. **Dependency**: if catalog is part of what `volume-only` promises (the README lists "Data listing" among what the library does), `nest-asyncio` belongs in the base dependencies or in `volume-only`.
2. **Failure mode**: replacing the `None` fallbacks with a stub that raises the original `ImportError` on call would turn `TypeError: 'NoneType' object is not callable` into a message naming the missing package. Happy to send a PR for this part if you'd like it — it's a few lines.

<sub>Found while setting up a reproduction environment for #1325. Claude Code (Opus 5); every result above is from an executed run.</sub>

olgaiv39:
Fixes #1329

`vesuvius.utils.catalog` imports `aiohttp` and `nest_asyncio` even when only `list_files()` is used, which breaks `volume-only` installs

Move those imports to the code paths that need them and add a regression test for `list_files()` without the optional dependencies

### DIFF
--- a/vesuvius/src/vesuvius/utils/catalog.py
+++ b/vesuvius/src/vesuvius/utils/catalog.py
@@ -7,8 +7,6 @@ import ssl
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple
 
-import aiohttp
-import nest_asyncio
 import requests
 import yaml
 
@@ -31,6 +29,8 @@ async def scrape_website(
 ) -> Tuple[CatalogTree, CatalogListing]:
     """Collect directory metadata and Zarr links from a remote listing."""
 
+    import aiohttp
+
     from vesuvius.data.paths import parser
 
     ssl_context = ssl.create_default_context()
@@ -50,6 +50,8 @@ async def scrape_website(
 
 async def collect_subfolders(base_url: str, ignore_list: List[str]) -> List[str]:
     """Return subfolder paths beneath ``base_url`` ignoring provided patterns."""
+
+    import aiohttp
 
     from vesuvius.data.paths import parser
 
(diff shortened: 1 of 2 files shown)


bruniss:
volume only installs would still need both of these as they are required deps for any remote zarr access

PR #1428, opened 2026-08-12
line annotation: parallax ghosts + arrow-key panning between control points

skyward7187:
> Stacked on #1423 (`fiber-review-tag-cleanup`): its commits appear in this diff
> until it merges, after which this PR reduces to the two feature commits plus a
> main merge.

Two navigation-UX features for the Line Annotation dialog, aimed at the same problem:
control points flashing past unseen while panning quickly along a fiber.

## Parallax slide-in ghosts (`82c7b1c23`)

The current cut view draws a control point's solid yellow marker only while the cut
plane is within ±lineRadius of it, so fast strip panning skips them. The view now
always shows up to two dashed hollow yellow rings — the nearest control point behind
and ahead — drawn at their true in-plane landing spot plus a horizontal parallax
offset proportional to the line-position delta (slide range 8 units, capped at 35%
of the viewport width). The ghost slides in from the side the control point will
arrive from, brightens as it approaches, and converges exactly onto the solid marker
at delta zero. Ghosts only appear within 10× the solid-marker window and fade out
over the outer quarter of that range, so distant ones don't linger.

## Arrow-key panning (`869b71d71`)

Left/Right pan the current position between control points with a signed-velocity
integrator (pure, header-only, unit-tested):

- **Tap**: ramp up, brake, land exactly on the nearest control point in that direction.
- **Hold**: cruise straight through intermediate points; release decelerates onto the
  next point ahead (never short of the tap's floor, never past the last target).
- **Opposite arrow mid-pan**: decelerates through zero and reverses.
- **Beyond the outer control points**: one more hop, to the Max CP distance allowance
  or the extrapolated line end, whichever is shorter (same line-position
  interpretation as the green/red placement gate).
- **Up/Down**: scale the cruise speed ×1.25 (1–500 line-positions/s, default 12,
  persisted in settings), transient badge top-center of the top strip; live mid-pan.
- Arrow presses pause strip hover-follow exactly like space (❚❚ badge); space or a
  strip/cut click resumes hover-follow and cancels the pan. Pausing also flushes the
  queued hover-coalesce update so a stale mouse position can't fire into a fresh pan.
- The strips snap centered on the current-position line and scroll underneath it.
  The camera is recentered *before* each overlay rebuild: strip overlays bake the
  camera into scene coordinates, so recentering afterwards draws the line one tick
  off-center and the coalesced refresh redraws it centered — a 60 Hz flicker that
  reads as two green lines.
- The viewers' old 64 px arrow panning is intercepted inside this dialog only;
  toolbar spinboxes keep native Up/Down when focused.

## Review hardening

Five independent Codex review rounds (plus targeted verification passes) were run
against the branch; all actionable findings were fixed and re-verified. One pass
independently audited that every position-changing input path (clicks,
shift-clicks, shift-scroll, jumps, R-snap, Reset Views, in-place view swaps,
control-point edits, allowance changes) either cancels or rebases a running pan,
and the keyboard-state machine survives popups grabbing the keyboard, window
deactivation, workspace hides, and dual-key sequences:

- a release always ends an active hold even after focus moves into a text widget;
  window deactivation releases a hold; hiding the workspace cancels it outright
- in-place view re-materialization, Reset Views, overview-bar R-snap, and
  control-point edits mid-pan all cancel or rebase the pan (edits re-promise the
  minimum target; a stale target can never land the pan backward)
- a reversal pressed within coast distance of the line end pins at the edge and
  reverses instead of cancelling; landings force-apply past the position setter's
  1e-3 no-op threshold so the cut planes land exactly
- modifier'd arrow repeats are swallowed mid-pan; combo popups count as text-entry
  focus; dual-key sequences hand the pan back to the still-held key (gated so a
  cancelled pan cannot revive)

Accepted as designed: the Max-CP-distance boundary uses the same line-position
interpretation as the existing placement gate; tap overshoot at extreme speeds
(500 pos/s) is inherent; braking-onset velocity ripple is sub-percent; per-tick
strip recentering rides the same coalesced refresh path as mouse strip-dragging
(revisit only if long-fiber jank is observed).

## Testing

- `test_line_annotation_generated_views`: 73 cases pass (62 pre-existing + 11 new
  covering the ghost helper, the pan integrator, stop-target selection, and the
  boundary hop).
- `ninja VC3D` links clean.
- Manually exercised in the GUI: ghost slide-in direction/landing, tap/hold/reverse
  pans, speed badge, pause coupling, center-locked strips (green-line flicker fixed),
  end-of-line boundary hop.

https://claude.ai/code/session_01NHfy5MbBTdkuXso5foTdgh


hendrikschilling:
```
  1. High: panning can continue rendering in the background after window deactivation.
     On Alt-Tab, changeEvent() at volume-cartographer/apps/VC3D/LineAnnotationDialog.cpp:4270 treats deactivation as key release, not cancellation. The 16 ms timer therefore continues     
     until the next control point or boundary. At low configured speed and widely spaced control points, this can mean sustained four-pane rendering for minutes while VC3D is inactive.    
     Hiding cancels correctly; deactivation should probably do the same.

  2. Medium: each pan tick causes substantial duplicate work.
     tickArrowPan() at volume-cartographer/apps/VC3D/LineAnnotationDialog.cpp:2552 runs at approximately 60 Hz and:
      - submits changed renders for both strips, current cut, and side cut;
      - redundantly applies the second strip camera after linked-camera synchronization already updated it;
      - emits several overlaysUpdated signals;
      - schedules rebuildGeneratedOverlays() at volume-cartographer/apps/VC3D/LineAnnotationDialog.cpp:4027, rebuilding static overlays and all dynamic overlays;
      - rebuilds the side-cut full-line overlay both directly during the tick and again through that delayed callback.

     Render jobs are latest-only/coalesced, so this does not create an unbounded render queue or a true ping-pong loop. However, render-job capture and substantial main-thread overlay work
     still happen every tick. Long fibers are the likely regression case.
```

skyward7187:
Thanks for the review — both addressed in b1b22e999 (plus a main merge):

**1. Background rendering after deactivation — fixed, cancel outright.** Agreed that braking into the next target was the wrong call for an inactive window. Focus loss now cancels the pan immediately (position freezes in place, same as space), matching hideEvent. It turned out `ActivationChange` delivery is window-manager-dependent (it never fired on the WM we tested), so the cleanup hangs off three redundant hooks — `changeEvent`/ActivationChange, `event()`/WindowDeactivate, and `QGuiApplication::applicationStateChanged` — all funneling into one `stopArrowPanForFocusLoss()`. Verified live: Alt-Tab mid-hold and mid-coast both stop dead now.

**2. Per-tick duplicate work — first item fixed, proposal for the rest.** The redundant second-strip camera apply is gone: per-tick recentering now touches only the first strip and lets `syncLinkedStripCamera`'s synchronous echo update the other (the initial snap still visits both, since Y isn't linked).

For the remaining items, does this match what you had in mind?
- **Side-cut full-line overlay double rebuild**: drop the direct in-tick rebuild and let the coalesced callback own it (or vice versa — whichever you'd prefer as the single owner).
- **The scheduled `rebuildGeneratedOverlays()` per tick**: the static strip overlays do need refreshing while the camera scrolls (their scene coords bake the camera in), so it can't be suppressed wholesale — but during an active pan we could skip the side-strip **intersection request** on tick-driven rebuilds (that's the fiber-snapshot/hash cost) and issue one full-quality rebuild at landing, which the landing path already does. Other fibers' intersection marks would freeze during the pan and snap current on arrival.

Happy to do both in this PR if that direction looks right, or split them out if you'd rather keep this one to the feature + the two fixes above.

hendrikschilling:
```
  1. Medium: the per-tick overlay workload is only partially fixed.
     The redundant second-strip camera application is gone, but every 16 ms pan tick still:
      - applies both strip cameras through linked synchronization;
      - schedules rebuildGeneratedOverlays() through overlaysUpdated at volume-cartographer/apps/VC3D/LineAnnotationDialog.cpp:1395;
      - directly rebuilds dynamic overlays through setCurrentLinePosition() at line 2629;
      - then rebuilds static and dynamic overlays again at line 4052;
      - requests side-strip intersections, which clones strip geometry, snapshots all fibers, and hashes everything before checking the cache at volume-cartographer/apps/VC3D/
        LineAnnotationController.cpp:10819.

     Consequently, long fibers still incur substantial GUI-thread work at approximately 60 Hz. There is no unbounded queue or ping-pong loop, but the sustained workload remains.
```

skyward7187:
Per-tick workload addressed in afb9954ef (and the branch is now rebased onto post-#1423 main via merge, so the diff should be down to the feature files).

What a 16 ms pan tick does now:
- **dynamics rebuilt once** (by the tick itself, camera-first) — the coalesced `overlaysUpdated` refresh no longer duplicates them;
- **statics rebuilt once** via that coalesced refresh (`rebuildGeneratedStaticStripOverlays()` only while a pan is active — they have to track the scrolling camera);
- **no side-strip intersection request at all during the pan** — the clone/snapshot/hash work is deferred to a single full-quality `rebuildGeneratedOverlays(true)` at the landing (statics + dynamics + span labels + intersections).

This also removes the side-cut double rebuild from your first list (it was the hook's dynamic pass duplicating the tick's). The one bullet I haven't reduced is the two strip camera applies per tick — that's one explicit apply plus the linked-camera echo, which is the minimal path for keeping both strips scrolling in lockstep.

Trade-offs, stated plainly: other fibers' intersection marks freeze during a pan and snap current on arrival; if a pan is cancelled mid-flight (space/click) they stay stale until the next interaction refreshes them — clicks and hovers do that naturally, and wiring a rebuild into the cancel path would reach teardown contexts where it isn't safe.

Tested on a 597-control-point fiber: panning stays smooth end to end.

hendrikschilling:
```
  - Low: landing still causes one redundant full overlay refresh. finishArrowPan() cancels the pan, applies the strip camera, and immediately calls rebuildGeneratedOverlays(true) at       
    volume-cartographer/apps/VC3D/LineAnnotationDialog.cpp:2642. The camera update emits overlaysUpdated, whose delayed callback later sees no active pan and performs another full rebuild 
    at LineAnnotationDialog.cpp:1409. Intersection deduplication happens only after cloning strip geometry, snapshotting fibers, and hashing them at LineAnnotationController.cpp:10883, so 
    that preparation is repeated. This is bounded to one extra landing pass, not a loop.

```

skyward7187:
Fixed in 656977f70. The landing now marks the coalesced callback its own camera apply just queued as already covered, and the callback consumes the flag and returns instead of repeating the full rebuild (and the intersection preparation). The flag is only set when a callback is genuinely pending and resets on view teardown, so it can never swallow a legitimate refresh. Cancelled pans are deliberately unchanged: their queued callback still runs the full pass, which is what un-stales the intersection marks after a mid-flight cancel.

Landing cost is now exactly one full pass: statics + dynamics + span labels + one intersection request.

hendrikschilling:
Just add the generation based deduplication and you should be good!
```
  - Low: the new landing suppression can discard a newer overlay update. At volume-cartographer/apps/VC3D/LineAnnotationDialog.cpp:1404, all updates arriving while the 16 ms callback is   
    queued are coalesced. After landing, LineAnnotationDialog.cpp:2664 marks that callback as covered. If another camera/overlay update occurs after the landing rebuild but before the     
    callback runs, the callback still exits as covered, leaving overlays stale until a subsequent update. A generation counter would distinguish updates covered by the landing rebuild from
    later ones.
```

skyward7187:
Done in ca101dbbe — generation-based, as you suggested. Every `overlaysUpdated` bumps a generation counter, including updates coalesced into an already-queued callback; the landing records the generation its full rebuild covered, and the callback skips only when the generations still match. Anything arriving after the landing rebuild bumps the counter and the callback runs its normal full pass, so no update can be discarded. Counters reset with the refresh connections on view teardown.

PR #1430, opened 2026-08-13
Fix NameError on CLI validation paths in finalize_outputs and predict

axiosdevs:
`main()` in both entry points builds its parser inline as `build_parser().parse_args()`, then calls `parser.error(...)` — but no `parser` name exists at that point.

In `finalize_outputs` the call is unconditional (`resolve_threshold(parser, args)`), so **every** invocation of `vesuvius.finalize_outputs` raises `NameError: name 'parser' is not defined` before any work starts, outside the `try` that would otherwise report a failure:

```
$ vesuvius.finalize_outputs in.zarr out.zarr
NameError: name 'parser' is not defined      # finalize_outputs.py:590
$ vesuvius.finalize_outputs in.zarr out.zarr --num_parts 2 --part_id 5
NameError: name 'parser' is not defined      # finalize_outputs.py:588
```

In `predict` it hits the `--bbox` error path: a malformed bbox raises `NameError` instead of the intended usage message (`inference.py:1241`).

The fix is binding the parser:

```python
parser = build_parser()
args = parser.parse_args()
```

After it, the same commands exit 2 with the intended messages (`error: Invalid part_id 5 for num_parts 2...`, `error: Invalid chunk_size format...`, `error: --threshold_value requires --threshold`).

The existing flag-spelling suite only reaches `build_parser()`, which is why this survived. The added tests drive `main()` itself and assert `SystemExit(2)` on three finalize paths and the predict bbox path, plus an AST guard that `main()` binds any `parser` it reports errors on — that guard fails on current `main` for both files and passes with this change.

### DIFF
--- a/vesuvius/src/vesuvius/models/run/finalize_outputs.py
+++ b/vesuvius/src/vesuvius/models/run/finalize_outputs.py
@@ -581,7 +581,8 @@ def build_parser() -> argparse.ArgumentParser:
 
 def main():
     """Entry point for the vesuvius.finalize command."""
-    args = build_parser().parse_args()
+    parser = build_parser()
+    args = parser.parse_args()
 
     # Validate partitioning arguments
     if args.part_id < 0 or args.part_id >= args.num_parts:
--- a/vesuvius/src/vesuvius/models/run/inference.py
+++ b/vesuvius/src/vesuvius/models/run/inference.py
@@ -1220,7 +1220,8 @@ def build_parser():
 def main():
     import sys
 
-    args = build_parser().parse_args()
+    parser = build_parser()
+    args = parser.parse_args()
     
     # Parse optional patch size if provided
     patch_size = None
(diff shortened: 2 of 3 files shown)


flummoxjr:
Independent reproduction and confirmation, from a different route into the same bug.

I hit this while running the full pipeline on a model I trained with `vesuvius.train` on Dataset059 — `vesuvius.predict` → `vesuvius.blend_logits` → `vesuvius.finalize_outputs`. The first two stages completed; `finalize_outputs` died before doing any work, exactly as described:

```
File ".../models/run/finalize_outputs.py", line 590, in main
    effective_threshold = resolve_threshold(parser, args)
NameError: name 'parser' is not defined
```

Applying this PR's two-line change locally, the same command completes: 27 chunks processed, output written, exit 0. The result is sane against ground truth — uint8-quantized probabilities (mean 144.7, max 254), Dice 0.266 at prob > 0.5 versus 0.286 measured by direct forward passes of the same checkpoint, the small gap being the Gaussian blend overlap and the full-volume vs center-crop difference.

So on Windows/Python 3.14/zarr 3, with a `train_py` checkpoint rather than an nnU-Net one, this is the only thing standing between a working end-to-end run and a crash. Confirmed on `--mode binary`, no `--threshold` flag, default `num_parts`/`part_id` — i.e. the unconditional path your description calls out.

(For anyone else who lands here from the same direction: I also needed the three `vesuvius.predict` fixes in #1492 to produce the logits this stage consumes.)

PR #1434, opened 2026-08-13
train: keep label depth in the flat-mode loss behind an opt-in flat_depth_targets flag

khj1222:
## Motivation

Testing the premise of #192 — that more accurate 3D ink labels train better models — on
PHercParis4 `w00_20231016151002` and `w02_20231031143852`, I hit a wall in the pipeline
itself. In `flat` mode the label never reaches the loss with its depth intact:

    targets = (torch.amax(batch['inklabels'], dim=2) > 0).to(dtype=batch['inklabels'].dtype)
    supervision_mask = torch.amax(batch['supervision_mask'], dim=2)

Every label voxel is max-pooled into one plane first, so a depth-resolved label and the
published single-plane label produce **byte-identical targets**. The experiment #192 asks
for is not expressible in flat mode at all. `full_3d` does keep depth, but it wants the
native scroll volume and builds its own band by projecting the flat annotation to a
constant half-thickness, so it answers a different question.

## The half that is easy to get wrong

Training on a depth-resolved label changes what inference must do with the prediction.
The loss only constrains the supervised z column; outside it the network is free, and it
saturates. A max down the full axis reports that noise instead of the ink:

z-window before/after

Same checkpoint, same held-out region, only the volume→surface collapse differs:
**F1 0.499 → 0.814**. The tell is the threshold — scored over the whole volume, every
fold's best threshold pinned to 254, the top of the uint8 range. A full-segment
prediction TIFF written without the window is degraded the same way, so this is not just
a scoring detail.

## Change — opt-in, default path untouched

Everything is behind `flat_depth_targets: true`; without the key nothing changes.

* z projection is disabled for model and targets the way the native 3D modes already do
  it, and the model emits `[B, 1, Z, Y, X]`.
* The loss runs volume-to-volume, with `supervision_mask` as the ignore mask.
* Train previews reuse the `full_3d` central-slice reduction.
* `infer` gains `--z-window START:STOP` to collapse a volume prediction back to the
  ordinary 2D TIFF over the supervised slices, and warns when it is reduced over the full
  depth. Default (no window) reproduces existing behaviour.

## Verification on real scroll data

| check | result |
|---|---|
| a constant 8-voxel band trained through this path vs the 2D baseline, same 3 folds | 0.8478 vs 0.8472 mean F1 — within the 0.03 noise floor, so the depth path is calibrated, not just plumbed |
| the same on a second segment (`w02`) | 0.8263 vs its own 2D baseline 0.8235 |
| `--z-window` across the full 3-fold arm | 0.5046 → 0.8098 mean F1 |
| default-off path | a config without the key takes the existing `amax` branch, unchanged |
| unit tests | 7 passed; projection-gating tests extended to cover the flag |
| shapes / throughput | output `(1, 1, 64, 256, 256)`, ~3.6 it/s, unchanged from 2D targets |

15 training runs, ~22 GPU-hours, two segments. The full write-up and raw numbers are in
#192.

## Changed since the first round

`--z-reduce mean` is gone. It rested on one ad-hoc checkpoint comparison and never
entered the matrix, so there was no evidence to offer for it — `max` over the supervised
window is what the runs actually validate.

### DIFF
--- a/ink-detection/koine_machines/inference/infer.py
+++ b/ink-detection/koine_machines/inference/infer.py
@@ -59,6 +59,12 @@ class ChunkKey:
     col: int
 
 
+_Z_REDUCTIONS = {
+    "max": lambda logits: logits.amax(dim=2),
+    "mean": lambda logits: logits.mean(dim=2),
+}
+
+
 class TargetHeadWrapper(nn.Module):
     def __init__(
         self,
@@ -66,11 +72,24 @@ class TargetHeadWrapper(nn.Module):
         *,
         target_name: str,
         input_pad_depth_to: int | None = None,
+        z_reduce: str = "max",
+        z_window: tuple[int, int] | None = None,
     ):
         super().__init__()
         self.model = model
         self.target_name = str(target_name)
         self.input_pad_depth_to = input_pad_depth_to
+        self.z_reduce = str(z_reduce).strip().lower()
+        if self.z_reduce not in _Z_REDUCTIONS:
+            raise ValueError(
+                f"Unknown z reduction {z_reduce!r}; allowed: {sorted(_Z_REDUCTIONS)!r}"
+            )
+        if z_window is not None:
+            start, stop = (int(v) for v in z_window)
+            if start < 0 or stop <= start:
+                raise ValueError(f"z window must be 0 <= start < stop, got {z_window!r}")
+            z_window = (start, stop)
+        self.z_window = z_window
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         from koine_machines.models.input_padding import center_pad_input_depth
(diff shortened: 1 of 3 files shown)


erdpx:
hi, thanks for the pr. please follow villa/CONTRIBUTING.md. also, the experiments only validate max over the supervised z window; mean over the logits was not evaluated, so we’d need evidence that it produces valid predictions before including it

khj1222:
Thanks — both points taken.

`mean` is removed (`8922c5e`). You are right that it was never evaluated: it rested on a
single ad-hoc checkpoint comparison and never entered the 3-fold matrix, so I would
rather drop it than argue from that. `max` over the supervised window is what the runs
actually cover, and that is all the PR now ships. I also added a warning when a volume
prediction gets reduced over its full depth, since that is the failure mode the window
exists for.

On CONTRIBUTING.md — I had genuinely not seen it: it is on `main`, and this work has been
on `merge-ink-pipelines` throughout. Reading it now, the body was missing the before/after
figure on real data and the human commentary, both of which are in the rewritten
description. Sorry for the noise.

khj1222:
Reopening was not available on this PR, so the revision is up as #1535  — same branch, `mean` removed, description rewritten against CONTRIBUTING.md.

Issue #1407, opened 2026-08-11
PHerc1203 physical-truth label grid is chunk-padded 34 voxels past the volume extent


DarthCeltic:
The PHerc1203 physical-truth labels are `[2016, 3456, 3456]` on the lo volume's L1 grid, which is `[9489, 3422, 3422]` (from `20250820131727-9.362um-1.2m-113keV-masked.zarr`, L0 `[18977, 6844, 6844]` at 9.362 µm).

3456 is `ceil(3422 / 128) * 128` = 27 × 128, so the label array is chunk-padded 34 voxels past the volume it describes, in both x and y. In z it spans L1 3936..5952, well inside.

The store is sparse, so the padding reads as `fill_value` and can't be mistaken for material — no correctness problem. But read naively it looks like "no material" across the last 1% of every row and column, which is at the sheet edges, where an unexplained deficit is fairly believable and easy to chase for a while.

The transform I ended up with, in case it's worth putting in the README:

```
label(z, y, x) = (L0_z/2 - 3936, L0_y/2, L0_x/2)
```

valid only where y and x are below 3422.

Jinhojeong:
I can confirm this from a full scan of the label volume, since I consume these labels downstream. Material voxels span x 462 to 2895, y 772 to 2645, z 39 to 1974, and the valid bit spans the same envelope to within three voxels. Nothing exists at or beyond 3422 in x or y, so your arithmetic holds exactly, and the pad band is silently zero rather than marked invalid, which is worth knowing because it means no bit distinguishes padding from unlabelled space. One practical addition, the labelled content ends more than 500 voxels before 3422 in both axes, so on this particular label set the edge-deficit misread cannot actually arise, the risk is generic to naive dense reads rather than realized here. Your coordinate transform matches what I derived independently and I would support putting it in the README. Checking your claim also led me to audit my own consumer of these labels and I found and corrected a read bug on my side, reported in issue 191.

DarthCeltic:
Good calls on both points—that cuts straight to the actual issue.

The core problem is that the padding band is silently zero-filled in place of marked invalid, leaving no bit-level distinction between chunk padding and unannotated space. Without tracking `y, x < 3422` explicitly, downstream code can't tell "outside the volume" from "inside, but empty."

[...]

PR #1435, opened 2026-08-13
Document physical-truth label chunk-padding, PHerc1203 worked example

DarthCeltic:
Documents the label chunk-padding behavior discussed in #1407 — the array can be up to one chunk larger than the volume it describes, and the padding reads as sparse fill rather than being marked invalid. Adds the PHerc1203 coordinate transform @Jinhojeong and I both independently derived in the issue thread

Refs #1407

### DIFF
--- a/vesuvius/src/vesuvius/data/affine.py
+++ b/vesuvius/src/vesuvius/data/affine.py
@@ -178,6 +178,18 @@ def resample_label_to_image_grid(
     of the image patch (much smaller than the image patch itself when the
     image voxel size is finer than the label voxel size, which is the
     typical fibers setting).
+
+    Some physical-truth label arrays are zarr-chunk-padded past the volume
+    they describe: the array shape is rounded up to a chunk multiple, so it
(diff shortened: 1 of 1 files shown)


Jinhojeong:
Confirming the transform from the consumer side, since I derive the same one downstream. On a full scan of the label volume, material spans x 462 to 2895 and y 772 to 2645, so nothing exists at or beyond 3422 in either axis. The padding reads as silent zero rather than marked invalid, which is the part worth documenting because no bit separates padding from unlabelled space.

One measured consequence that may be worth a line in the doc. A map built on this grid at 96-voxel cells has a dead final row and column by construction, and on my PHerc1203 thickness map all 1,491 cells with y or x index 35 carry zero sampled runs and zero valid share. So per-cell readers should treat the last index as padding rather than as a material deficit. My note and the measurement are in results/thickness1203 in the diagnostic repo linked from #191.

pmh47:
I'm not sure where this is best documented, but definitely not here!

PR #1440, opened 2026-08-13
Handle stalled binary HTTP transfers without a hard deadline

### LINKED ISSUE #1438: Open Data segment fetch: 30s transfer_timeout is too short for large mesh bands, and the timeout is misreported as a missing file
## Summary

`segments.fetch` over the agent bridge fails deterministically on a segment whose mesh band files are large, with:

```
missing or empty x.tif at https://vesuvius-challenge-open-data.s3.us-east-1.amazonaws.com/PHercParis4/segments/20260701183124-w010-027/mesh/20260701183124-on-20260411134726-2.4um.tifxyz/x.tif
```

The file is not missing or empty. Verified directly against S3, independent of VC3D:

```
$ curl -sI .../x.tif
HTTP/1.1 200 OK
Content-Length: 118592134

$ curl -s -o x.tif -w "http_code=%{http_code} size=%{size_download} time=%{time_total}\n" .../x.tif
http_code=200 size=118592134 time=85.857452
```

Full 118 MB, byte-exact, ~86 seconds over a normal connection. Reproduced the `segments.fetch` failure twice in a row, identically.

## Root cause

`HttpFetch.cpp:18-26`:

```cpp
utils::HttpClient makeTextClient(const HttpAuth& auth)
{
    utils::HttpClient::Config cfg;
    cfg.aws_auth = auth;
    cfg.transfer_timeout = std::chrono::seconds{30};
    cfg.connect_timeout = std::chrono::seconds{5};
    cfg.max_retries = 2;
    return utils::HttpClient(std::move(cfg));
}
```

`httpGetBytes` (`HttpFetch.cpp:117`) uses this same `makeTextClient` — despite the name, it's the only HTTP client this function has, and it's used to fetch arbitrary binary payloads, not just small text/JSON. `OpenDataSegmentCacheIO.cpp`'s `writeCachedTifxyzBand` calls it for each mesh band `.tif` (here, three ~118 MB files per segment). At normal bandwidth these take well over 30 seconds; the hardcoded transfer timeout cuts the request short.

Compounding it: when the timeout fires, `resp.ok()` is false but `resp.status_code` apparently isn't in a range either error branch (`isAuthError`, `>= 500`) checks, so `httpGetBytes` silently returns `{}` instead of throwing. The caller only knows "empty," not "timed out," so the error message blames the file rather than the client.

## Impact

Any tifxyz mesh band, or presumably any other binary artifact routed through `httpGetBytes`, that takes longer than 30 seconds to download will fail this way — deterministically, with a misleading error. This scales badly with segment size and network conditions; it isn't specific to this one segment or scroll.

## Suggested direction (not proposing exact code — I can't see `utils::HttpClient::Config`'s full field set, it's not vendored in this checkout)

`httpGetBytes` (binary payloads, potentially large) probably shouldn't share a timeout policy with `httpGetString` (small JSON/text config files). A separate client config with either a much longer flat timeout, a timeout that scales with expected size, or (if the underlying client supports it) a stall-based timeout — abandoning only when the transfer stops making progress, not because it's merely large — would all fix this without a real, empty-response case being newly masked. Whichever direction is taken, the empty-vs-timeout distinction should probably also be preserved for the caller rather than collapsed to the same "missing or empty" message.

## Reproduction

```
catalog.open_sample {"sampleId": "PHercParis4", ...}
segments.fetch {"segmentId": "20260701183124-w010-027-published-20260411134726-L0-3ea79d84fadcfbed"}
```
fails with the message above, both cold and after a retry.

---
*Found while trying to fetch a Paris4 segment to grow further into unpublished winding numbers via the agent bridge. The curl reproduction above is from actual runs, not inferred. Write-up done with AI assistance (Claude).*

DarthCeltic:
Fixes #1438.

Fetching a PHercParis4 segment through the agent bridge repeatedly failed with `missing or empty x.tif`, even though the 118 MB object returned HTTP 200 and completed successfully with curl after about 86 seconds.

The binary path used the text client's 30-second total-transfer timeout. When libcurl failed without an HTTP status, `httpGetBytes` also returned an empty byte vector instead of surfacing the transport error, so callers misreported a valid remote object as missing.

## What changed

- Give `httpGetBytes` a dedicated binary-transfer client.
- Keep the 5-second connection timeout, but remove the hard total-transfer deadline.
- Abort transfers that make less than 1 byte/second for 10 seconds, so genuinely stalled connections still fail promptly while slow but progressing downloads can finish.
- Preserve two retries for transient failures.
- Throw a descriptive transport exception when libcurl reports an error without an HTTP status instead of returning `{}`.
- Add the low-speed policy to the shared HTTP client configuration and apply it consistently to normal requests and `put_file`.
- Add a regression test for the formerly silent transport-error path.

`httpGetString` retains the existing short text-request policy.

## Verification

- Merged current `ScrollPrize/main`; the branch is conflict-free and GitHub reports it mergeable.
- Built and ran the focused HTTP error suite against the exact merged source in `ghcr.io/scrollprize/villa/volume-cartographer:edge`: **16/16 passed**.
- Reproduced the sole broader CI failure (`test_zarr_chunk_fetcher`) independently in the same official image: **1/1 passed**. That test is outside the changed HTTP path; the GitHub failure is an unrelated flaky run.
- All other completed Linux CI shards passed, including 138 of 139 base tests before the isolated Zarr flake.

Found while fetching a real Paris4 binary artifact to continue surface work. The write-up and implementation were prepared with AI assistance; the transfer measurements and verification runs are from the actual artifacts and branch.


DarthCeltic:
Hey — looking at the checks, not sure what's wrong. Can this be checked when maintainers have a chance, at their leisure? Thanks!

hendrikschilling:
This PR does not really seem to fix the issue:

```
  1. Blocker: the misleading error behavior is not fixed.
     After any libcurl failure, status_code remains 0 and error_message is populated. httpGetBytes() still ignores error_message and returns {}, so callers report “missing or empty.”      
     Raising the timeout only avoids that outcome for transfers finishing within 300 seconds.
     PR code (https://github.com/ScrollPrize/villa/blob/69c1592410511e60ece3906b8ff837bd66a61439/volume-cartographer/core/src/HttpFetch.cpp#L128-L145)

  2. The 300-second policy is applied too broadly.
     httpGetBytes() also downloads catalog photos and small optional files. A stalled connection would now occupy those operations for up to 300 seconds per attempt, with three attempts   
     possible. The longer timeout should be selected by the large-artifact caller, or replaced with a low-speed timeout/streaming download policy.

  3. No regression test covers either behavior.
     The existing bad-URL test accepts both throwing and silently returning, so it cannot detect the original problem. A transport failure should explicitly be required to throw with      
     error_message.
```

I would suggest basing the solution on replacing the hard timeout with a software network stall / minimum speed threshold this way network failures are separated from simply slow connects:

```
- CURLOPT_CONNECTTIMEOUT: fail if connection establishment stalls.
  - CURLOPT_LOW_SPEED_LIMIT: minimum acceptable bytes/second.
  - CURLOPT_LOW_SPEED_TIME: how long throughput may remain below that limit.
  - CURLOPT_TIMEOUT = 0: disable the hard total-transfer deadline.

  For large binary downloads, a suitable policy would be:

  connect_timeout = 5s;
  total_timeout = disabled;
  low_speed_limit = 1 byte/s;
  low_speed_time = 10s; // perhaps 20–30s for unstable connections

  That allows a transfer to take hours as long as data continues arriving. If it effectively stalls, libcurl aborts and the existing retry policy can retry it.
```

DarthCeltic:
Thanks for the detailed feedback- agreed on all three points. 

1. `HttpClient` now preserves libcurl transport details in `HttpResponse::error_message`, and `httpGetBytes()` throws `HTTP transport error ...` instead of returning an empty payload that callers misreport as "missing or empty".
2. The binary client no longer has a hard total-transfer deadline: connect timeout is 5s, `CURLOPT_TIMEOUT` is disabled, and a 1 byte/s for 10s low-speed policy distinguishes a stalled transfer from a slow but progressing one. The 30s text-client policy remains unchanged.
3. The regression test now requires the public `httpGetBytes()` path to throw on a transport failure; silently returning is no longer accepted.

Hopefully this all works as intended now! Thanks!!!

PR #1441, opened 2026-08-13
vc3d: expose existing Spiral and Lasagna workflows through MCP

giorgioangel:
## Summary

This branch was rebuilt from current `main` after scope review. It now adds
only MCP exposure and agent guidance for behavior VC3D already has:

- 16 thin `spiral.*` RPC/MCP adapters over the existing project-scoped
  `SpiralServiceManager`;
- one direct Lasagna-manifest attachment RPC that reuses the existing GUI
  loader, authentication, cache settings, validation, and transactional
  package attachment;
- build identity and catalog coordinate-level fields needed to verify live
  sessions and choose Open Data representations safely;
- a generated 136-method RPC snapshot, matching FastMCP wrappers, and focused
  wrapper/contract tests;
- a concise workflow-oriented SPEC and automatically discoverable skills,
  including rendering, MCP flattening, Lasagna, Atlas, and standalone Spiral
  checkpoint flattening.

## Explicit scope boundary

This rewrite removes the earlier branch's functional additions. It does **not**
add or modify:

- `RemoteFileCache`, SHA-256 URL keys, cache sidecars, or remote-transfer
  policy;
- the Lasagna Manager runtime, schemas, prefetch, inference, or training;
- Fiber runtime, control-point/configuration RPCs, or trace-metric code;
- Spiral service/controller/runtime behavior;
- screenshot render waiting or unrelated editing/viewer behavior;
- libbacktrace or any dependency/build policy;
- WIP evidence, test PNGs, or local logs.

The existing interactive Lasagna attachment path still uses the same prompts,
warning titles, cancellation behavior, asynchronous preparation, selection,
and package update. Independent strict-scope audits verified that the remaining
production diff is bridge exposure plus this behavior-preserving extraction.

## Contract and skills

- `rpc.describe` is authoritative; `rpc_description.json` is regenerated from
  the compiled binary.
- The SPEC no longer duplicates the generated RPC-to-MCP table.
- Every registered MCP tool, including the MCP-only `vc3d_wait_job`, maps to a
  primary workflow skill.
- `.claude/skills` remains the authoring source; `.agents/skills` contains
  symlinks for automatic Codex discovery. `agents/openai.yaml` is intentionally
  absent because it is optional UI metadata, not skill discovery.
- Historical/measured guidance that was too detailed for a core skill remains
  available through one-hop references.

## Verification

The branch includes current `main` and is conflict-free. The exact source tree
was built on macOS as RelWithDebInfo; the final live `ping` reported commit
`9577f6ef945e5b86a5e7e97f97c5b8ac1b55f49e`, protocol 2, and the explicit
temporary binary path. The final commit makes the manifest-attachment adapter
smaller (39 insertions, 54 deletions relative to the previous PR head) without
changing its RPC result or the existing GUI workflow.

- C++ `agent_bridge_contract`: **passed**.
- Focused direct-manifest validation on a fresh PHerc1299 project: regular L2
  and fiber-inference L1 attachments both returned `attached: true`; repeating
  the regular attachment returned `alreadyAttached: true`. The owned process
  terminated and the disposable project was removed.
- MCP unit suite: **172 passed**, 2 platform/explicit-real-bridge skips.
- Focused contract/runtime/skill-routing suite after the documentation fix:
  **29 passed**.
- Exact live `rpc.describe`: **136 methods**, including the documented
  10-method Fiber surface.
- Dynamic PHerc1299 catalog selection from current Open Data metadata:
  volume `20260309130042`, regular Lasagna `0:3`/L2 and fiber-inference
  Lasagna `0:4`/L1.
- Bounded catalog open: **passed** and attached both manifest-backed Lasagna
  datasets. This final run reused the existing cached sample project; a separate
  empty-cache run on the identical merged source tree also attached both
  datasets successfully.
- Real Fiber3D workflow using only the PR's documented RPCs: **passed**.
  `fiber.launch` created all four generated line panes; one plain click on the
  generated `*_line_surface` strip produced a prediction-backed fiber with
  2 control points, 316 line points, one span, and `traceState: predictions`.
  `fiber.save` plus `fiber.list` provided the authoritative readback.
- Cleanup: the test-created fiber (id `36`) was deleted and its absence
  verified. The VC3D process terminated. Evidence files remained under
  `/private/tmp` and are not part of this PR.
- Final scope audit: no WIP directory, test PNG, local log, cache implementation,
  Manager runtime, Fiber runtime, or unrelated functional change is included.


giorgioangel:
@codex review

Bullo27:
Ran this branch end to end on Linux against current `main`. Headline: **it no longer builds**, for two independent reasons, and `git merge` reports neither of them. The merge is textually clean and semantically broken.

Almost everything below was executed rather than read, and where it wasn't I say so explicitly — the GUI-behaviour section and the last section are source-level only. Environment: Ubuntu 24.04, g++, Ninja, `CMAKE_BUILD_TYPE=Release`, `VC_USE_PCH=OFF`, `VC_BUILD_APPS=ON`, `VC_TESTING=ON` — i.e. the default configure. Your PR body records macOS / RelWithDebInfo, so this is the other side of that. `main` at `c53c74a29`, branch head `1fe7a7977`, merge base `97621bcf4`.

## A clean merge is not evidence the branch still builds

`git merge-tree --write-tree main pr1441` reports **zero conflicts**, and `git merge --no-ff` succeeds ("Merge made by the 'ort' strategy"). That part of the PR body still holds.

What has changed since 08-20 is the distance: `main..pr1441` is 7 commits, `pr1441..main` is **92**, and `main` is not an ancestor of the head. So "The branch includes current `main`" was true when written and is no longer true. Both breaks below live entirely in that 92-commit gap, and both are invisible to conflict detection because neither touches a line this branch also touches.

## 1. Build break — `SpiralServiceManager::runIterations` lost a parameter

```
AgentBridgeHandlers_spiral.cpp:235:27: error: no matching function for call to
  'SpiralServiceManager::runIterations(const int&, QJsonObject, QJsonObject, QJsonObject)'
SpiralServiceManager.hpp:72:10: note: candidate:
  'void SpiralServiceManager::runIterations(int, const QJsonObject&, const QJsonObject&)'
SpiralServiceManager.hpp:72:10: note:   candidate expects 3 arguments, 4 provided
```

| ref | signature |
|---|---|
| merge base `97621bcf4` | `runIterations(int, const QJsonObject&, const QJsonObject&, const QJsonObject& inputs = {})` |
| branch head `1fe7a7977` | same — the branch does not change it |
| `main` today `c53c74a29` | `runIterations(int, const QJsonObject&, const QJsonObject&)` |

`6a828ee3d` ("Spiral workspace state", #1552) removed `const QJsonObject& inputs = {}`. Because the parameter was *defaulted*, nothing on this branch had to change for the call to break — the four-argument call simply stopped resolving.

This is the only compile error in the merged tree: `ninja -k 0` completed 359 of 360 objects and failed exactly one.

### It is not just an argument to delete

`main`'s `runIterations` no longer sends `inputs` to `/session/run` at all, and the replacement states why:

> Base inputs are owned and canonicalized by the service. A Run changes neither their paths nor their contents, so restating the panel's necessarily partial path view here can only create a false mismatch (for example, `winding_inference` has no editable panel row).

So the value is now meaningless — but this branch still advertises it to agents in three places:

| layer | location (branch head) | what it promises |
|---|---|---|
| handler | `AgentBridgeHandlers_spiral.cpp:239` | reads `params.inputs`, forwards it |
| descriptor | `AgentBridgeMethods_spiral.cpp:68` | `opaqueObject("inputs")` on `spiral.run` |
| MCP tool | `tools/spiral.py:66,75` | `vc3d_spiral_run(..., inputs=None)` |

Delete only the C++ argument and `vc3d_spiral_run(iterations=…, inputs={…})` still returns `accepted: true` while the inputs are silently discarded. That seemed worth not leaving behind, so the patch I ran removes all four layers together (the JSON is the generated snapshot; re-dumping it produced exactly a four-line diff, so the formatting matches your generator):

```diff
--- a/volume-cartographer/apps/VC3D/agent_bridge/AgentBridgeHandlers_spiral.cpp
+++ b/volume-cartographer/apps/VC3D/agent_bridge/AgentBridgeHandlers_spiral.cpp
@@ -235,8 +235,7 @@ QJsonObject AgentBridgeServer::handleSpiralRun(const QJsonValue& params)
     service->runIterations(
         iterations,
         request.value(QStringLiteral("influenceConfig")).toObject(),
-        request.value(QStringLiteral("runConfig")).toObject(),
-        request.value(QStringLiteral("inputs")).toObject());
+        request.value(QStringLiteral("runConfig")).toObject());

--- a/volume-cartographer/apps/VC3D/agent_bridge/AgentBridgeMethods_spiral.cpp
+++ b/volume-cartographer/apps/VC3D/agent_bridge/AgentBridgeMethods_spiral.cpp
@@ -64,8 +64,7 @@ void AgentBridgeServer::registerSpiralHandlers()
          .params = {positiveInteger(QStringLiteral("iterations")),
                     opaqueObject(QStringLiteral("influenceConfig")),
-                    opaqueObject(QStringLiteral("runConfig")),
-                    opaqueObject(QStringLiteral("inputs"))},
+                    opaqueObject(QStringLiteral("runConfig"))},

--- a/volume-cartographer/tools/vc3d-mcp/vc3d_mcp/tools/spiral.py
+++ b/volume-cartographer/tools/vc3d-mcp/vc3d_mcp/tools/spiral.py
@@ -63,7 +63,6 @@ async def vc3d_spiral_run(
     run_config: Optional[dict[str, Any]] = None,
-    inputs: Optional[dict[str, Any]] = None,
 ) -> dict[str, Any]:
@@ -72,7 +71,6 @@ async def vc3d_spiral_run(
             "runConfig": run_config,
-            "inputs": inputs,
         }),
```

plus dropping `spiral.run.params.properties.inputs` from `rpc_description.json`.

With that applied: `ninja` exits 0 and VC3D links; `ctest -j6` gives **154/154 passed, 0 failed**; `agent_bridge_contract` passes. 154 is also `main`'s own count today — I re-configured on clean `main` and counted 154 — so the merge neither adds nor drops a ctest.

## 2. A test this PR adds fails after the merge — `scripts/spiral/` moved out

```
tests/test_skill_coverage.py::SkillCoverageTest::test_standalone_workflows_have_primary_skills
AssertionError: False is not true : scripts/spiral/flatten_spiral_checkpoint.py
```

Isolated with the branch's own suite:

| tree | result |
|---|---|
| branch head `1fe7a7977` (your base) | 170 passed, 2 skipped, **447** subtests passed |
| merged into `main` today | 170 passed, 2 skipped, **446** subtests passed, **1 failed** |

The delta is exactly one subtest, so the suite is demonstrably able to both pass and fail here — the failure is merge-induced, not pre-existing.

`acbb61649` ("move spiral fitting out of volume-cartographer", #1548, 181 files) moved the script to repo-root `spiral-fitting/flatten_spiral_checkpoint.py`. `STANDALONE_WORKFLOW_SKILLS` still maps `"scripts/spiral/flatten_spiral_checkpoint.py"`, and `VC_ROOT` is `volume-cartographer/` — so the workflow now sits **outside** the root this test can address.

That makes it a scope question rather than a path typo, and I don't think it's mine to answer:

- reach above `VC_ROOT` to `spiral-fitting/`, which couples the vc3d-mcp tests to a sibling subproject;
- drop the standalone-workflow entry and the `vc3d-spiral-checkpoint-flattening` skill, since the workflow has left this subproject;
- keep the skill as guidance but stop asserting the file exists.

I've deliberately not patched this one — all three are defensible, and the choice is about what that skill is meant to denote now.

## What I checked that holds

Measured, so you don't have to re-verify:

- **Size** — 69 files, `+4,084 / −300`. Your "4,084 lines" is exactly the insertion count.
- **136-method snapshot** — `rpc_description.json` carries exactly 136 methods; `fiber.*` is exactly 10; `spiral.*` is 16; `lasagna.attach_manifest` is present.
- **`.agents/skills` symlinks** — all 18 entries are mode `120000` and every target resolves inside the tree. None dangle.
- **Scope boundary** — nothing in the diff touches `RemoteFileCache`, the Lasagna Manager runtime, the Fiber runtime, or the Spiral service/controller runtime. The only matches on those names are skill documents, plus `SpiralWorkspace.hpp` (+1 accessor) and `apps/VC3D/CMakeLists.txt` (+2 source entries).
- **The Lasagna extraction is behaviour-preserving on the interactive path.** This one is a code-level check, not a GUI run, so specifically what I verified: (a) recomputing `remote` from `isRemoteLasagnaLocation(location)` is equivalent for both entry points, because the remote action already rejects non-remote locations via the `isRemoteLasagnaLocation` guard in `beginLasagnaManifestAttachment` (`MenuActionController.cpp:1805-1808` on this branch, `:1935-1938` on `main`, unchanged by this PR) and the local action feeds an absolute path from `QFileDialog`; (b) the new 6th argument is not a signature change — `attachPreparedLasagnaDataset`'s `updateSelection` already defaults to `true` on both merge base and `main`, and the interactive path passes `true`; (c) warning titles are preserved, synchronous failures via the caller's "Attach failed" and asynchronous ones via `complete()`'s "Attach Lasagna failed"; (d) cancelling the remote-cache prompt still returns silently, since `cacheRoot.isEmpty() && interactive` returns with `errorMessage` cleared and the caller only shows a box for a non-empty message.

One number I could not reproduce: your body reports 172 passed / 2 skipped for the MCP suite, and I count 170 / 2 on your unmodified head. The skip count and every assertion agree, so I assume a pytest collection difference and did not chase it — nothing above depends on it.

## One smaller thing, unrelated to the merge

Codex has already flagged two issues in these same deferred handlers — unpropagated ordinary failures on `checkpoint_load` (`r3821153160`) and mis-correlated concurrent `checkpoint_download`s (`r3821153168`) — so rather than file a third neighbouring nitpick, the observation I'd add is that all four share one root: there is no consistent contract for how a deferred handler binds, correlates, and releases its signal connection.

Measured from the source, the four differ three ways:

| handler | correlation | release |
|---|---|---|
| `checkpoint_download` | none — the slot ignores the `path` it is handed | `Qt::SingleShotConnection` |
| `inputs_commit` | n/a — single global operation, nothing to key on | `Qt::SingleShotConnection` |
| `input_upload` | filters on `inputId` | explicit `disconnect()` |
| `checkpoint_load` | shared `completed` flag | **none** |

`checkpoint_load` is the one that never releases: both sender and receiver outlive the call, so each invocation leaves two connections behind permanently. The `completed` flag stops a stale lambda from completing someone else's token, so this is accumulation rather than wrong results — but it is also why Codex's two points land on different handlers instead of one. A single shared helper (correlate on the request key, release on completion *or* error, resolve the token on the manager's `errorOccurred` path) would close all four at once.

Source-level only — I have no live Spiral service here, so none of this paragraph is a runtime measurement.

## Limits

- **Not run on macOS**, and I did not re-run your live-session evidence — no Spiral service, no VC3D GUI session, no Open Data project opened here. Everything above is build, ctest, the Python suite, and source.
- **The GUI Lasagna path was read, not exercised.** The four equivalences above are the specific things I checked; I did not click through the dialogs.
- The `spiral.checkpoint_load` note is a source observation with no runtime measurement behind it.

Both breaks are `main` moving underneath a branch that was correct when it was written, which is a rebase away rather than anything wrong with the design here. Happy to re-run the whole thing against an updated head if that's useful.

---
*Claude (AI agent) on behalf of @Bullo27. All results measured on Ubuntu 24.04 / g++ / Ninja with the default configure (Release, `VC_USE_PCH=OFF`), against `main` at `c53c74a29` and branch head `1fe7a7977`, on 2026-08-25.*

PR #1444, opened 2026-08-13
line annotation: subdivide strip columns to the densest span step

skyward7187:
## Problem

The line-annotation strip ribbon draws one column per grid entry, but dense-line spans are sampled at very different physical steps depending on how they were interpolated: native-trace spans every ~4–8 base voxels, cspline "short span" (< 100 vx gaps under a Global goal) and lasagna-fallback spans every ~32 voxels. Strip-x was therefore physically non-uniform — a small-gap span occupied 4–8× fewer columns per physical distance than its trace neighbors, so the strip image squished there and the flanking control-point markers visually "scrunched" together right after placement. The 3D geometry was always correct.

#1453 introduced exactly the right machinery for this — `LineStripPositionMap`, threaded through markers, clicks, camera framing, context menus and both strip paths, with correct mean-pitch scale metadata — but its subdivision only splits segments *longer* than `targetSpacingBaseVoxels` (50 vx), so the 4-vs-32 density mismatch inside a mixed line still rendered scrunched.

## Fix (one knob, on top of #1453)

Derive the per-segment subdivision target from the **densest original segment** instead of using the configured spacing directly:

- `targetSpacingBaseVoxels` becomes the upper bound; the effective spacing is `clamp(smallest positive segment, floor, target)`, where the floor caps the total column count on pathological inputs (near-coincident points).
- Everything else is #1453's design, untouched: original points remain grid supports (bends survive; integer line positions land exactly on columns, so the marker fast path stays exact), the position map stays the single source of truth, cut views / `lineUpVectors` / `lineZSlices` stay per original line point, and already-uniform lines produce byte-identical output.
- Mixed-density lines now subdivide their sparse spans down to the trace pitch and render at uniform column width — markers sit where they were placed.

Also threads the position map through `vc_lasagna_line_probe`'s strip overlays, which still assumed column == point index; with subdivision now common on mixed lines its markers would land visibly off.

## Tests

New: mixed-density fixture (4 vx trace steps + 32 vx spans) asserting every adjacent column pair is exactly the trace pitch apart, supports stay on columns, and cut-view outputs stay per line point. Updated: the two uneven-arclength cases from #1453 now expect subdivision at the densest step (15 vx fixtures); the bend-preservation, per-segment-spacing (target-capped), duplicate-point, and validation cases pass unchanged. Suites: 35 / 22 / 74 pass; `VC3D` and `vc_lasagna_line_probe` build and link.

## Notes

- Arrow-pan speed stays in linePosition/sec; on-screen px/sec varies across span types (it now reflects true physical length — previously the same non-uniformity showed as geometric squish).
- The overview bar stays index-uniform (schematic).
- Strip column count grows on mixed lines (a 32 vx span gains ~8× columns); the floor bounds the worst case.

## Provenance

Supersedes this PR's earlier standalone implementation (uniform arc-length resample + its own mapping type), which was developed against the #1422/#1423/#1428 stack before #1453 landed the equivalent — and better-integrated — `LineStripPositionMap` machinery on main. Reworked to a minimal delta on top of #1453 rather than carrying a parallel system.

https://claude.ai/code/session_01AyKY5TNYyYYjhzvMkMvgZU

### DIFF
--- a/volume-cartographer/apps/src/vc_lasagna_line_probe.cpp
+++ b/volume-cartographer/apps/src/vc_lasagna_line_probe.cpp
@@ -1409,9 +1409,18 @@ void writeReinitDebugObjOutput(
                            outputDir / "continuation_candidates.obj");
 }
 
-int scaledControlColumn(int sourceIndex, int renderScale, int cols)
+int scaledControlColumn(int sourceIndex,
+                        int renderScale,
+                        int cols,
+                        const vc::lasagna::LineStripPositionMap& positionMap)
 {
-    const double scaled = (static_cast<double>(sourceIndex) + 0.5) *
+    // The strip grid is subdivided per segment, so a point index is not a
+    // grid column; convert through the builder's position map first.
+    const double gridColumn = positionMap.valid()
+        ? positionMap.originalPositionToStripGridColumn(
+              static_cast<double>(sourceIndex))
+        : static_cast<double>(sourceIndex);
+    const double scaled = (gridColumn + 0.5) *
                           static_cast<double>(std::max(1, renderScale)) -
                           0.5;
     return std::clamp(static_cast<int>(std::lround(scaled)), 0, std::max(0, cols - 1));
@@ -1419,7 +1428,8 @@ int scaledControlColumn(int sourceIndex, int renderScale, int cols)
 
 cv::Mat makeStripOverlay(const cv::Mat& grayscale,
                          const std::vector<int>& fixedIndices,
-                         int renderScale)
+                         int renderScale,
+                         const vc::lasagna::LineStripPositionMap& positionMap)
 {
     cv::Mat image8;
     if (grayscale.depth() == CV_8U) {
@@ -1442,7 +1452,7 @@ cv::Mat makeStripOverlay(const cv::Mat& grayscale,
         if (index < 0) {
             continue;
         }
-        const int col = scaledControlColumn(index, renderScale, overlay.cols);
+        const int col = scaledControlColumn(index, renderScale, overlay.cols, positionMap);
         cv::line(overlay,
                  cv::Point(col, 0),
                  cv::Point(col, std::max(0, overlay.rows - 1)),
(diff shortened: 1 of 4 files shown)


skyward7187:
**Reworked on top of #1453** (force-pushed, `1d4b1a755` → `659c062d1`).

The original implementation here (uniform arc-length resample of the ribbon spine + its own `LineStripColumnMapping` threaded through every strip consumer) was developed against the #1422/#1423/#1428 stack. #1453 then landed the equivalent — and better-integrated — machinery on main: `LineStripPositionMap`, already threaded through markers, clicks, camera framing, context menus and both strip paths, with correct mean-pitch scale/LOD metadata. Carrying a parallel mapping system over that would have duplicated machinery across the same files, so I dropped it entirely.

What #1453 doesn't do is fix the density mismatch itself: its subdivision only splits segments longer than `targetSpacingBaseVoxels` (50 vx), so a ~32 vx cspline/lasagna span between ~4–8 vx native-trace spans still renders squeezed, and control points still scrunch. This PR is now the minimal delta for that:

- `buildPositionMap` derives the per-segment subdivision target from the **densest original segment**, with `targetSpacingBaseVoxels` as the upper bound and a floor that caps total column count on pathological inputs. Sparse spans subdivide to the trace pitch → uniform column width.
- Everything else in #1453's design is untouched: original points stay grid supports (bends preserved, integer positions land exactly on columns so the marker fast path stays exact), and already-uniform lines produce byte-identical output.
- `vc_lasagna_line_probe`'s strip overlays now convert through the position map — they still assumed column == point index, which becomes visibly wrong once subdivision is common on mixed lines.

Diff is 4 files, +112/−25. New mixed-density test pins exact uniform pitch across a 4 vx/32 vx line; the two uneven-arclength cases from #1453 are renumbered to the new expectation (subdivision at 15 vx instead of ~50 in their fixtures); bend-preservation, per-segment-spacing, duplicate-point and validation cases pass unchanged. Suites 35/22/74; `VC3D` and the probe build and link on current main.

hendrikschilling:
There's some additional complications with the cp replacement and tgt is alreadys not consistent. I'll push an alternative fix.

PR #1445, opened 2026-08-13
line annotation: mirror-cursor toggle + keyboard control-point placement

skyward7187:
## Motivation

While annotating fibers in the line annotation workspace, the shared cursor cross that mirrors into the other generated panes is sometimes visual noise, and placing control points while arrow-key panning (#1428) required moving back to the mouse for every point.

## Changes

- **"Mirror cursor across panes" hamburger toggle** (persisted via `lineAnnotation/mirror_cursor_across_panes`, default on). Unchecking keeps the cursor cross only in the hovered pane. The gate lives on the receiving viewer (`CChunkedVolumeViewer::setLinkedCursorMirroringSuppressed`) because the panes are also fed by the app-global "Sync cursor across views" broadcast — blocking only the dialog's own mirror timer would leak crosses whenever that global toggle is on. Unchecking immediately clears any on-screen mirrored cross and its status-bar position readout. Main-window viewers and the global toggle are untouched.

- **"/" and "0" place a control point at the blue dot** — the current cut's current-position dot, i.e. `interpolatedLinePoint(_currentLinePosition)`. Both row and keypad variants work (KeypadModifier is masked; keypad 0 needs Num Lock). Routed through `handleKeyPress` so `keyboardFocusIsTextEntry()` keeps "0" typeable in the menu spinboxes and toolbar combos. Semantics deliberately differ from a placement click: the key cancels an active arrow pan (the placement renumbers the line positions the pan steers by) but leaves hover-follow paused and keeps the scrolled normal offsets, since the key exists to be tapped mid-pan. Guards: generated views present, placement allowed by the Max CP distance rule, finite interpolated point.

- Help-table entries (`Keybinds.cpp`) and `docs/line_annotation_fibers.md` updated.

## Testing

- Built VC3D (Release and RelWithDebInfo) on Ubuntu; `ctest -R "line_annotation|unified_browser_dialog"` passes (3/3).
- Exercised live on real scroll data (PHercParis4 fiber annotation): keyboard placement mid-arrow-pan, row and keypad keys, "0" still types into the Length spinbox, toggle behavior across panes.
- Rebased onto main after #1422/#1423/#1428 merged (clean, no conflicts) and rebuilt.

Known limitation: on layouts where `/` is a shifted key (AZERTY/QWERTZ) the slash variant won't fire; row `0` covers those layouts.


hendrikschilling:
Just a small comment:

```
  - P2: keyboard placement remains active during optimization. The overlay blocks mouse placement but does not remove keyboard focus. Pressing / or 0 sets _pendingPlacementFocus before the
    controller rejects the request because optimization is running. The completed optimization can then incorrectly reposition the view to the nearest existing control point. The shortcut 
    should check explicit optimization-busy state before setting pending focus. Shortcut code
    (https://github.com/ScrollPrize/villa/blob/bb410f651b8ff8570a7d1b44eacec1894f797aa3/volume-cartographer/apps/VC3D/LineAnnotationDialog.cpp#L4440-L4460), controller rejection
    (https://github.com/ScrollPrize/villa/blob/bb410f651b8ff8570a7d1b44eacec1894f797aa3/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L6306-L6320).
```

skyward7187:
Good catch, fixed in 443e4821c.

You're right that the overlay only stops the mouse. The dialog never actually recorded the busy state — `setOptimizationBusy` just toggled the overlay and the two combo/spin widgets — so the shortcut had nothing to test. It now stores that state and `placeControlPointAtCurrentLinePosition()` bails on it before touching `_pendingPlacementFocus` or emitting the request, alongside the existing generated-views/max-distance guards.

Sourcing the flag from `setOptimizationBusy` keeps it exactly as reliable as the overlay itself: same setter, same call sites (both `startOptimization` paths set it, `finishOptimization` clears it), so it cannot get stuck in a state where the overlay disagrees.

PR #1446, opened 2026-08-14
Fix normalisers writing through to the caller's array

axiosdevs:
`_prepare_image` builds the array the normalisers rescale in place:

```python
arr = np.asarray(image)
if target_dtype is not None:
    arr = arr.astype(target_dtype, copy=False)
return arr
```

With `copy=False`, an input already in the target dtype (`float32`, the default) comes back as *the same object*, and the `arr -= mean` / `arr /= std` arithmetic in `normalize_minmax`, `normalize_zscore`, `normalize_ct` and `normalize_robust` writes through to the caller. Anything not float32 is copied first, so the aliasing is silent and dtype-dependent.

It reaches production through `neural_tracing/infer.py`. `CropCache._extract_subcrop` returns a *view* into a cached supercrop:

```python
crop = supercrop[z0:z0 + d, y0:y0 + h, x0:x0 + w]
```

and the consumer hands that straight over:

```python
volume_crop_raw, min_corner_zyx = self.crop_cache.get(center_zyx)
volume_crop_zscore = torch.from_numpy(normalize_zscore(volume_crop_raw.numpy()))
```

Supercrops are float32 (the cache sizing comment says as much), and `.numpy()` shares their memory, so normalising a crop z-scores the cached supercrop in place. Since the reason to cache a supercrop is that neighbouring crops reuse it, the next crop from that entry is drawn from a buffer that is part raw and part already-normalised, and gets normalised again. Measured on overlapping crops of one supercrop, the second crop comes out up to ~2.7 z-score units away from its correct values — wrong model input, no error raised.

The fix is one line: `_prepare_image` always returns a private array. Nothing documents the in-place behaviour, and every non-float32 caller was already paying for the copy.

Tests cover all four normalisers: input untouched for float32 and for four other dtypes, a view's base buffer untouched, two overlapping crops of one buffer normalising independently, and the outputs still being what they were (z-score mean 0 / std 1). Against current `main`, 9 of the 30 fail.


pmh47:
Happy to merge this if you turn it into like a two line fix instead of massive comment and more-massive test

axiosdevs:
Trimmed to the two-line fix — long comment gone, test file dropped. Diff is now `copy=True` in `_prepare_image` plus a one-line note about why. Happy to add a small regression test later if you want a guard on it.

pmh47:
@bruniss this look ok to you?

PR #1451, opened 2026-08-14
tracer: optional support test so growth stops at the edge of the data

axiosdevs:
Picks up @Jinhojeong's measurement in #1114: the seed tracer has no support constraint, so a patch keeps growing into regions the volume does not cover, and masking the *predictions* does not change that — on his three PHerc1218 seeds the share of traced surface standing over absent CT stayed at 34.1% and 39.2% after the masking fix, as one contiguous mass reaching a third to half of the patch rather than a rim of noise.

This adds the test he specified, at the point where a grown point is committed:

```jsonc
{ "require_volume_support": true, "volume_support_threshold": 0 }
```

Off by default — without the flag not a single branch changes, so existing runs are bit-identical. When on, after the Ceres solve and local optimisation, the fitted position is looked up in the volume already open in `tracer()` (`trace_data.raw_volume`, the `passTroughComputor` tensor used for the seed-value read); a point landing out of bounds or on a voxel `<= volume_support_threshold` has its state cleared and is not appended to the fringe, so growth simply stops in that direction. Rejections are counted and reported per generation.

Pointing `--volume` at a masked volume makes this exactly the rule from the thread ("keep a quad if the support volume is nonzero at its position"); on a raw volume the threshold gives the same effect. A separate support-volume path would be the natural follow-up if you'd rather trace one volume and test against another — that needs a second `Volume` + `ChunkCache` plumbed through, which seemed worth splitting out rather than bundling here.

Two things I want to be straight about: the growth loop is yours, so treat the placement as a proposal — the `probe`/rollout path in `GrowSurface.cpp` is untouched and would need the same treatment to cover the segments tracer. And I don't have a working VC3D toolchain locally, so I checked the inserted block for syntax against stub types and relied on this repo's C++ CI for the real compile; happy to iterate on anything it flags.

@Jinhojeong offered to run the same seeds against an implementation and report before/after the way he did in the thread — this should be that implementation.

### DIFF
--- a/volume-cartographer/core/src/GrowPatch.cpp
+++ b/volume-cartographer/core/src/GrowPatch.cpp
@@ -756,6 +756,7 @@ struct TraceData {
     std::unique_ptr<PatchNormalContext> patch_normals;
 
     Chunked3d<uint8_t, passTroughComputor>* raw_volume = nullptr;
+    Chunked3d<uint8_t, passTroughComputor>* support_volume = nullptr;
     std::unique_ptr<lineLossDistance> space_line_compute;
     std::unique_ptr<Chunked3d<uint8_t, lineLossDistance>> space_line_volume;
 };
@@ -3117,11 +3118,16 @@ QuadSurface *tracer(Volume& volume, float scale, int level, cv::Vec3f origin, co
     const std::array<int, 3> volume_shape_zyx = volume.shape(level);
     auto* cache = volume.chunkedCache();
 
-    // Optional support test: reject grown points that land where the volume holds no
-    // material. Off by default; point --volume at a masked volume (or raise the
-    // threshold on a raw one) to stop growth from continuing past the data.
+    // Optional support test: reject grown points that land where there is no
+    // material to support them. Off by default. "support_volume" names the volume
+    // that decides support -- typically the masked CT the predictions were inferred
+    // from; when empty the traced volume decides.
     const bool require_volume_support = params.value("require_volume_support", false);
     const int volume_support_threshold = params.value("volume_support_threshold", 0);
+    const std::string support_volume_path = params.value("support_volume", std::string());
+    // Masks are not perfect: allow a point to count as supported when material sits
+    // within this many voxels of it. 0 keeps the strict test.
+    const int volume_support_dilation = std::max(0, params.value("volume_support_dilation", 0));
 
     const int growth_scale_level = std::clamp(params.value("growth_scale", 0), 0, 5);
     const int growth_scale_factor = 1 << growth_scale_level;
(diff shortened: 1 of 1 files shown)


axiosdevs:
Ran it. Built this branch on macOS and traced PHerc1218 from one seed at (3296, 3324, 11623), 100 generations, `--volume` on the published m7 predictions and `support_volume` on the masked CT of the same scan, both level 0 (23247 x 7593 x 7593, so positions map 1:1).

| run | quads | quads over `CT == 0` | area |
| --- | --- | --- | --- |
| flag off (current `main` behaviour) | 39,204 | **55.0%** | 11.67 cm² |
| `require_volume_support: true` | 3,036 | **0.0%** | 0.87 cm² |

(unsupported share from 300 evenly spaced quads per surface, sampled against the masked CT)

The 55% baseline lines up with the voxel-level figure for this scroll — 50.2% of the positives in these predictions are phantoms — so the traced surface inherits the contamination almost exactly, and the seed's own neighbourhood is fine (96% of voxels in a 128×192×192 box around it are supported, only 1% of predicted voxels there sit over `CT == 0`). The surface earns its unsupported half by travelling: it spans z 10164–12236 from a seed at 11623.

Three things came out of making it actually work, all in this push:

**`support_volume`.** The tracer follows a prediction volume, so testing the traced volume was the wrong reference. The parameter now names the volume that decides support (local path or URL, opened at the same level, shape-checked against the traced volume); empty keeps the previous behaviour of testing the traced volume.

**`volume_support_dilation`.** Count a point as supported when material sits within N voxels, for the imperfect-mask case discussed above. Default 0 = strict.

**Refused locations are marked, not cleared.** This is the part worth a maintainer's eye. Clearing a refused location to `0` leaves a gap inside `used_area`, which flips `used_area_fully_valid` and puts the tracer into hole-filling mode permanently; refused points are then offered again, refused again, and growth dies — in my first run it stalled at generation 7 with 130 quads. Refusals now set the reserved `STATE_FAIL` bit, `append_candidate` skips those locations, and the fully-valid scan treats them as terrain rather than holes. One more case needed handling: an unused location sealed off by refused ones can never be filled, so it is not counted as a hole either — otherwise a single such cell traps the tracer the same way.

Honest read of the result: at-source stopping is strictly more conservative than the post-hoc rule in #1114. It cannot cross an unsupported gap to reach supported material beyond it, so it keeps 3,036 quads where trimming a fully grown patch would have kept roughly the supported 45% of 39,204. Which behaviour is wanted depends on whether an unsupported stretch means "the mask is imperfect here" or "growth has left the scroll" — the dilation parameter is the dial between them, and @Jinhojeong's seeds would be the right place to calibrate it.

axiosdevs:
The separate-volume version is already on the branch — you built the first push, which is why you got the single-volume behaviour. `support_volume` landed together with `volume_support_dilation` and a fix for growth dying after refusals:

```jsonc
{ "require_volume_support": true,
  "support_volume": "<masked CT, local path or URL>",
  "volume_support_dilation": 0 }
```

It is opened at the same level as the traced volume and shape-checked against it; empty keeps the old behaviour of testing the traced volume.

Your hit-rate numbers explain the collapse better than my own run did: quads land on a positive prediction voxel only 21–30% of the time but on CT support 61–98%, so testing a thin binary prediction rejects most legitimate growth, and a refused point never joins the fringe. That matches what I measured on PHerc1218 — with `support_volume` pointed at the masked CT, a seed that traces 39,204 quads at **55.0%** over `CT == 0` traces 3,036 quads at **0.0%**.

Which leaves the question your post-hoc numbers frame. You retain 65/56/97% of in-region quads; at-source retains far less on my seed, because growth cannot cross an unsupported stretch to reach supported material beyond it — it stops at the first refusal in that direction. `volume_support_dilation` is the dial between "the mask is tight here" and "growth has left the scroll", so a sweep of 0/1/2/3 on your seeds, read against your retained fractions, would say which of the two the gap is. That comparison is the thing I cannot produce on my own seed and you can.

One trap if you rebase: clearing a refused location to `0` leaves a gap inside `used_area`, which flips the tracer into hole-filling mode permanently and kills growth regardless of support — that is what capped my first run at 130 quads. Refusals now set `STATE_FAIL`, `append_candidate` skips them, and an unused location sealed off by refused ones is not counted as a hole either.

Jinhojeong:
Posted the dilation sweep you asked for on #1114, since this PR is closed: https://github.com/ScrollPrize/villa/issues/1114. Short version, dilation 1 lands all three seeds near the post-hoc retention with the surviving surface 99 percent supported, and dilation 0 shows exactly the gap-crossing conservatism you predicted.

Jinhojeong:
@bruniss a quick question so axiosdevs and I know where to take this. Closing this while merging his other four reads as a decision about this change specifically rather than about the author. Is the objection the stopping rule itself, or touching the growth loop while the winding and spiral work is in flight, or just that this shape of change needs to come from the team? The measurements behind it are parked in #1114 either way, and if the preferred form is a preflight check rather than an in-loop test, #1463 already goes that direction. No complaint about the close, we would just rather aim the next round of measurements at the shape you would take.

PR #1454, opened 2026-08-14
umbilicus: frame metadata, project attachment, shared resolver

skyward7187:
Scroll umbilicus files are consumed by several tools, but two things about them were
always implicit: **which grid their coordinates index** (PHercParis4's umbilicus lives on
the 9.6 µm ds2 grid while annotations live in the 2.4 µm level-0 frame) and **which file
to use** (directory searching breaks down once a project references several volpkgs).
This PR makes both explicit, makes the failure modes loud instead of silent, and routes
line annotation — the consumer whose silent fallback hid a whole scroll's frame
mismatch — through the shared answer.

> **Note:** this branch was restarted on 2026-08-19 from the state of the first review
> round and rebuilt as eight reviewable commits; the previous head is preserved at the
> tag `archive/umbilicus-logistics-20260819`. The restart comment below maps every prior
> finding to its fate.

## Frame metadata in the umbilicus file

`umbilicus.json`'s `metadata` block gains four optional keys:

```json
"metadata": {
  ...,
  "volume": "s1_2um_ds2.zarr",   // store the umbilicus was annotated on
  "voxelsize_um": 9.6,           // voxel size of the grid the coordinates index
  "volume_width": 8174,          // exact voxel counts of that grid: integer
  "volume_height": 8174,         // dimension ratios identify a rescaled copy
  "volume_slices": 18946         // of the same scan where µm values round
}
```

`Umbilicus::LoadFileInfo` reads points and metadata in one parse; `LoadControlPoints`
and `FromFile` behave exactly as before, and unstamped or legacy files (bare arrays, the
text format) still load with the metadata unset.

**Malformed metadata is an error, not an absence.** A mistyped value, a non-object
`metadata` block, or an incomplete dimension triplet is recorded per field in
`UmbilicusFileInfo::metadataErrors` — a typo must be distinguishable from a legacy file,
or consumers quietly fall back to guessing. `null` stays legal (a writer's "no value").
Parsing still returns the points, because point-only callers should not lose their
umbilicus over a frame field they never read; it is the resolver that refuses such a
file.

## Explicit project attachment

- An optional `"umbilicus"` location in the project json, with `VolumePkg` accessors
  mirroring `selected_lasagna_dataset`, persisted with the rest of the project state.
- **File > Attach Umbilicus...** validates the file and stores it. Malformed frame
  metadata is *rejected* at attach time, naming the file and the errors — storing it
  would persist an attachment the resolver then refuses, success reported beside a
  warning. A file that states no frame attaches with a warning; dimensions-only files
  are fully stated, by the preferred key, and attach silently.
- **File > Detach Umbilicus** returns the project to automatic resolution. It is its own
  action because generic Detach enumerates entries keyed by location and the umbilicus
  is a single-valued field.

## Shared resolver

`vc::core::util::resolveScrollUmbilicus(pkg)`:

1. The project's `umbilicus` field is authoritative when set. A location that cannot be
   used — remote/unsupported, or a missing local file — is an **error naming it**, never
   a silent fall back to searching.
2. Otherwise it searches the package root and the parent of each resolved segment
   location, for `umbilicus.json` then `estimated_umbilicus.json`. An **individual
   segment** (recognised by its tifxyz payload) also contributes its grandparent,
   covering the `<volpkg>/paths/<segment>` layout; a *folder* attachment does not, since
   its grandparent is whatever holds the volpkg.
3. Exactly one distinct hit resolves. Several distinct hits refuse with the candidate
   list (deduplicated through `weakly_canonical`); zero hits report the roots searched.
4. A resolved file whose frame metadata is malformed is refused, errors named.

`scanUmbilicusCandidates()` is the resolver's own scan, exposed: everything whose
existence or contents can change the answer, in priority order, shadowed lower-priority
names excluded, absent higher-priority names included. The resolution and the cache
fingerprint below both consume this one scan, so they cannot disagree.

## Deriving a scale, and refusing what the file disproves

`deriveUmbilicusScale()` carries a file's coordinates into a target grid, in descending
order of trust: stamped **dimensions** (exact integer ratios), stamped **voxel size**,
then which power-of-two downsample of the target the points fit inside and nearly fill
(≥60% z coverage; candidates are a factor of two apart, so the bound makes the match
unique). No answer means *frame unknown*, never 1.0.

`uniformRescaleFactor()` is an exact test, not a tolerance: an integer downsample maps a
count `c` to `floor(c/n)` or `ceil(c/n)` and nothing else, so a candidate factor is
checked directly, both directions are tried, and only a unique factor answers.

| stamped | target | result |
|---|---|---|
| 8174x8174x18946 | 32693x32693x75784 (real PHercParis4) | **x4** — 32693/4 is 8173.25, ceiling 8174 |
| 8174x8174x18946 | 24522x24522x56838 | **x3** — non-power-of-two, still integer |
| 32693x32693x75784 | 8174x8174x18946 | **x0.25** |
| 8174x8174x18946 | 32696x32696x75026 (ratios 4, 4, 3.96) | refused |
| 8174x8174x18946 | 32696x32696x40000 (cropped z) | refused |
| 100x100x1000 | 51x51x500 | refused — a one-voxel slack would have accepted it |

Voxel counts cannot pin a *fractional* rescale; such files must state `voxelsize_um`.

`umbilicusFrameClaim()` answers only whether the file states a frame;
`decideUmbilicusLoadAction()` turns the answers into **Apply / Refuse / UseLegacy**. A
stated frame that does not fit the target — or could not be checked for want of one — is
refused rather than read as though it had stated nothing: the legacy reading applies a
registration inverse or takes the points raw, which is proceeding exactly where the
check failed. A file that states nothing keeps its previous reading, `transform.json`
and plausibility floor included: a compatibility promise, not a guess.

**A stamp the named volume disproves is refused outright.**
`umbilicusStampContradiction()` checks a stamp against the volume it names, when the
project has it: the stamped dimensions must be a uniform rescale of that volume's grid,
and the stamped voxel size must agree with what the volume implies through it (1%
round-trip slack). A contradicted stamp is not weaker evidence than an unverified one —
it is evidence against, and previously a contradicted voxel-size stamp scaled the points
anyway with a warning beside them.

## One annotation frame, from exact factors

`deriveAnnotationFrame()` (new, testable without a `Volume`) resolves the frame
annotations index, keeping coordinate extents and physical resolution strictly apart:
extents come from the **exact coordinate factor** (an open-data identity's
`sourceCoordinateScaleFactor`, else the store's own pyramid position — both exact
integer statements), never from a resolution ratio. A recorded resolution may be rounded
(9.596/2.4 = 3.99833), and extents derived from that ratio hand the exact rescale test a
fractional grid it rightly refuses — a conflict the derivation itself would have
invented. The exact factor is read straight from the tags, so it survives missing
physical metadata. The ratio is still honoured where nothing exact was said; an untagged
rebased store lifts its counts by its own level.

## Line annotation, switched over

`orientedLineNormalsForSession()` used to search `<volpkgRoot>/umbilicus.json` directly —
in project-file sessions that is the *project* directory, so on PHercParis4 it found
nothing and silently oriented sheet normals against the volume's XY centre, in
session-volume coordinates, 4× away from the frame the line points live in. It now
resolves through the shared path, scales into the annotation frame, sizes the dense
per-slice centres at that frame's extent, and applies the load decision. An inferred
scale never outranks a discovered registration transform (inference is a reading, and a
translated fixed-frame umbilicus satisfies it while still needing the transform); a file
that states its frame still wins.

Every unusable case — refused stamp, several candidates, malformed metadata, nothing
found — shows as small red text at the top left of the bottom strip. Doing it silently
is what hid the mismatch. Intersection-inspection panes have no dialog object, so they
get the refusal in the log; a visible notice there is follow-up work.

## Stale views refresh through counters, not a signal fan-out

Attach used to report success while open strips kept their old orientation, and a volume
switch left generated surfaces on the previous volume's normals. Deliberately **not**
fixed with an eager equivalence proof — an earlier design recorded everything an
orientation was built from and compared it on every switch, and each review found
another input the record was missing. Instead:

- `_orientationEpoch` bumps on attach/detach and on any volume switch; sessions record
  the epoch their views were built at (on success only); one queued, coalesced refresh
  pass rebuilds exactly the panes that are behind, and the intersection inspection
  whole — its sessions suppress ordinary generated views, so the pane walk never reaches
  them, which is precisely the strip an earlier refresh skipped.
- The umbilicus cache is keyed by what it holds (`umbilicusReloadNeeded()`): package
  root, **volume identity**, a stat fingerprint over every resolver candidate **plus the
  registration transform** the legacy reading would consult, and the annotation frame.
  Volume identity is deliberately conservative — two volumes deriving one annotation
  frame can still need different geometry — so a same-frame switch re-resolves (a few
  stats, one parse) rather than proving the previous geometry equivalent.
- `umbilicusGeneration()` bumps on attach/detach only, for holders of derived geometry
  (the Fiber Map, stacked in #1459) to compare lazily instead of being signalled.

Both decisions are pure functions in `UmbilicusOrientationFreshness.hpp`, asserted in
`test_line_annotation_generated_views` rather than read.

## Testing

- `test_umbilicus` (34 cases): stamped, metadata-without-frame-keys, bare-array and
  text-format files; each frame field malformed individually yields exactly one error
  naming it; all five at once yield all five; incomplete triplets; non-object blocks.
- `test_scroll_umbilicus` (44 cases): field precedence, single hit, ambiguity refusal,
  symlink canonicalisation, declared-but-missing, remote errors **without** searching
  (proved by leaving a discoverable local file unused), the `<volpkg>/paths/<segment>`
  layout, scan/candidate semantics, every row of the rescale table plus the floor/ceil
  window at its edge, scale precedence, claim/decision over all arms, and the
  stamp-contradiction cases including the 1% slack boundary.
- `test_annotation_frame` (16 cases): exact-factor precedence over the rounded ratio,
  factor surviving missing resolution, level agreement across stores of one scan, what
  each missing input costs — and end to end through `deriveUmbilicusScale` and
  `decideUmbilicusLoadAction` on PHercParis4's real values: the stamped grid takes
  **Apply at exactly ×4** where the rounded-ratio derivation is refused by the same
  exact test.
- `test_line_annotation_generated_views` (+2 cases): every arm of the cache key
  (including the same-grid volume switch that used to reuse a stale transform) and every
  skip rule of the refresh walk.
- `test_volume_pkg`: the `umbilicus` field absent, round-tripped, resolved relative.
- Measured on real data through the shipped code path: on `PHercParis4.volpkg.json` the
  resolver returns the attached file, reports **1** dependency, and derives **×4** from
  stamped dimensions; with the field removed it discovers the same file through the
  volpkg root, reports **3** dependencies, and leaves the shadowed
  `estimated_umbilicus.json` out of them.
- Every commit builds VC3D and passes the non-network suite (149/149) individually.
  Key assertions mutation-checked: dropping the volume-identity key arm fails exactly
  the same-grid-switch case; disabling the contradiction refusal fails exactly the case
  naming it; making the frame derivation ignore the exact factor fails the three
  exact-factor cases including the end-to-end ×4.
  Ubuntu 24.04 / GCC / Qt 6.8.3.

## Scope and notes

- **"One way" is scoped to line annotation (and the Fiber Map via #1459).** Patch
  tracing still takes its own umbilicus path from a setting (`GrowPatch`;
  `vc_project_tifxyz.cpp` is a third direct `Umbilicus::FromFile` caller). Converting
  them, the volume-centre fallback's own frame mismatch, and a visible notice in the
  intersection-inspection UI are tracked as follow-up work.
- Refusing a mismatched or contradicted stamp is a behaviour change for any project
  holding one: it previously limped through the legacy path and could produce a usable
  orientation by luck. That is the intent — a frame applied confidently but wrongly is
  worse than none — hence the log naming both grids and the strip notice.
- A volume switch now rebuilds open generated views once, including switches between
  downsample levels of one scan where the rebuild reproduces the same orientation. That
  waste is deliberate: it is bounded, user-initiated and visible, where the equivalence
  record that avoided it kept being wrong in ways that were not.
- Older VC3D builds rewrite the project json from their in-memory state and will drop
  the new `umbilicus` key on save — inherent to any new project field, recoverable by
  re-attaching.
- Grandparent search roots widen discovery, so a project holding umbilicus files at two
  levels can newly report an ambiguity where it previously resolved by accident; setting
  the `umbilicus` field is the intended answer.

https://claude.ai/code/session_01QmLCg7rtVeZT5Zu4khi8B4


skyward7187:
Status update on what's landed here since the PR opened:

**`038d791` — the attach error dialog names the rejected file.** Found during real-data testing: the picker opens at the package root, which in project-file sessions sits one level above the volpkg, surrounded by `*.volpkg.json` project files and tool configs (`config.json`, `seed.json`, ...) that all fail the same root-shape check. A mispick read as "your umbilicus is broken" with no way to tell which file had actually been rejected.

**`85c5be1` — schema revision: `zarr_level` replaced by grid dimensions.** The `metadata` block now stamps `volume_width` / `volume_height` / `volume_slices` (exact voxel counts of the annotated grid) instead of a pyramid level. Rationale from review (thanks @hendrikschilling): for rescaled copies of the same scan, integer dimension ratios identify the rescale exactly where rounded µm voxel sizes are ambiguous — and they express non-power-of-two rescales that a level never could. A level, when it exists, is derivable from the dims. `voxelsize_um` remains the authoritative rescale scalar. The PR description reflects the current schema.

Verification beyond the unit tests (18 cases across `test_umbilicus` / `test_volume_pkg` / `test_scroll_umbilicus`, full suite green after each commit):

- Both PHercParis4 umbilicus files (`umbilicus.json`, `estimated_umbilicus.json`) are stamped with the new schema and load correctly through the built library (`volume: s1_2um_ds2.zarr, voxelsize_um: 9.6, dims 8174×8174×18946`).
- The first real consumer (the Fiber Map workspace, stacked on this branch) exercises the resolver end-to-end on PHercParis4: the stamp yields the exact ×4 scale into the annotation frame that was previously guessed by heuristic, the ambiguity refusal and project-field precedence behave as specced, and the dims enable a stronger consumer-side sanity check (uniform-rescale validation against the named volume) than the old level math allowed.

https://claude.ai/code/session_01Xb3JfFAJa5ZTGGRe5xbqmh

hendrikschilling:
```
  1. P2: “Attach Umbilicus” currently has no effect on existing consumers. The menu stores the project field, but line annotation still searches the project directory directly and ignores 
     both the field and its frame metadata. Other existing tools also remain unchanged. This is acknowledged as follow-up work, but the GUI currently presents a functional attachment that 
     nothing uses. Attach action (https://github.com/ScrollPrize/villa/blob/693d57da80a46cec1664fc70acb79b84fff78acb/volume-cartographer/apps/VC3D/MenuActionController.cpp#L1777-L1811),   
     existing line-annotation lookup
     (https://github.com/ScrollPrize/villa/blob/693d57da80a46cec1664fc70acb79b84fff78acb/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L9379-L9403)

  2. P2: malformed frame metadata fails open. A present but invalid voxelsize_um or dimension is silently treated as absent. That makes a typo indistinguishable from a legacy unstamped    
     file and permits future consumers to fall back to guessing, undermining the purpose of explicit frame metadata. Legacy absence should remain supported, but explicitly malformed       
     metadata should fail loudly. Parser (https://github.com/ScrollPrize/villa/blob/693d57da80a46cec1664fc70acb79b84fff78acb/volume-cartographer/core/src/Umbilicus.cpp#L330-L360)

  3. P2: an explicitly configured remote location violates resolver precedence. umbilicusPath() returns an empty path for s3:///HTTP locations. The resolver interprets that as “not        
     configured” and performs directory search, despite its contract saying an explicit project field is authoritative and never falls back. Unsupported remote locations should produce an 
     explicit error. Path accessor (https://github.com/ScrollPrize/villa/blob/693d57da80a46cec1664fc70acb79b84fff78acb/volume-cartographer/core/src/VolumePkg.cpp#L1536-L1541), resolver    
     (https://github.com/ScrollPrize/villa/blob/693d57da80a46cec1664fc70acb79b84fff78acb/volume-cartographer/core/src/ScrollUmbilicus.cpp#L39-L60)

  4. P2: the attachment cannot be cleared through the GUI. The field is authoritative, and a missing attached file intentionally blocks fallback. However, generic Detach lists only        
     volumes, segments, normal grids, and Lasagna datasets. Users can replace the umbilicus but cannot return to automatic resolution without editing JSON manually. Detach implementation  
     (https://github.com/ScrollPrize/villa/blob/693d57da80a46cec1664fc70acb79b84fff78acb/volume-cartographer/apps/VC3D/MenuActionController.cpp#L1976-L2025)

  5. P2: fallback discovery misses a common direct-segment layout. For an entry such as <volpkg>/paths/<segment>, it searches <volpkg>/paths, while the conventional umbilicus is at        
     <volpkg>/umbilicus.json. It works when the attached entry is the whole <volpkg>/paths directory. Since VC3D explicitly allows attaching either a segment or a folder, both layouts need
     coverage. Root construction (https://github.com/ScrollPrize/villa/blob/693d57da80a46cec1664fc70acb79b84fff78acb/volume-cartographer/core/src/ScrollUmbilicus.cpp#L62-L90)

```

skyward7187:
Thanks — all five are addressed in `a0b134b06` except the first, which I've deliberately
split out. Taking them in order:

**1. Attach has no effect on existing consumers.** Two halves: the new consumer is done
here, and updating the existing one (line-annotation normal orientation) is deliberately
deferred to its own PR.

The Fiber Map (stacked in #1459) is now a real consumer: it resolves through
`resolveScrollUmbilicus`, so the project field and the ambiguity refusal genuinely
control what it uses, and it applies the frame metadata to place the umbilicus in the
fibers' frame. So the attachment is no longer decorative.

Routing **line-annotation normal orientation** through the resolver is not in this PR,
and I'd like to do it as its own PR rather than sneak it in here. The reason is bigger
than tidiness: `ensureScrollUmbilicusLoaded()` searches `<volpkgRoot>/umbilicus.json`
directly, and in project-file sessions that root is the *project* directory, one level
above the volpkg — so on PHercParis4 it finds nothing at all today and silently falls
back to the volume XY centre. Unifying it therefore does not adjust a scale; it switches
sheet-normal orientation from volume-centre to a real umbilicus on fibers that are
already traced. That is visible in the cut views and I don't want it riding along with a
metadata PR. It also needs a rule this PR doesn't settle: grid scale and a registration
`transform.json` answer different questions, and the normals path currently applies the
transform when one exists. On the reference data that transform demonstrably does *not*
apply to the umbilicus (inverse-transforming its points puts z at ~31k–88k against an
annotation frame ending near 75.8k) while the stamped grid ratio of ×4 is verified — but
"stamped scale wins, transform only when unstamped" needs stating and testing on its own,
with a before/after cut-view comparison as the acceptance criterion.

**2. Malformed frame metadata fails open.** Fixed. Rejections are recorded per field in
`UmbilicusFileInfo::metadataErrors` with a fixed format
(`volume_width: expected a positive integer, got -5`), and `resolveScrollUmbilicus`
refuses a file that carries any.

Two deliberate choices there. Parsing still succeeds and still returns the points, so
point-only callers such as patch growth and spiral fits don't lose an umbilicus over a
frame field they never read; legacy absence stays completely legal and silent. And the
refusal lives in the resolver rather than in each consumer, because that function is the
one that exists to answer "which umbilicus, in which frame",
so the next consumer can't forget the check. Concretely, the Fiber Map's scale precedence
(stamped dimensions → stamped voxel size → inference) can no longer be reached by a
typo'd stamp, which was the exact hazard you flagged.

Still silent: `"metadata": "not an object"`, because there's no field to name in that
message format. Say the word and it refuses too.

**3. A configured remote location violates resolver precedence.** Fixed. The raw
configured location is consulted before `umbilicusPath()`, so a set field is authoritative
even when it doesn't resolve to a local file; a remote or otherwise unusable location is
an error naming it, with wording distinct from the old "does not exist" (misleading for a
URL), and no search happens. The test proves the search didn't run by leaving a
discoverable local umbilicus in place and asserting it is *not* returned.

**4. The attachment cannot be cleared through the GUI.** Fixed, as **File > Detach
Umbilicus** next to Attach rather than an entry in generic Detach: that flow enumerates
entries keyed by `.location` and the umbilicus is a single-valued project field, so a
sentinel would have fought the data model. It clears the field and reports through the
status bar; with nothing attached it says so rather than doing nothing.

**5. Fallback discovery misses a common direct-segment layout.** Fixed. Each attached
segment's grandparent is now a search root as well as its parent, unconditionally — the
existing canonical dedup keeps one file reachable both ways from counting as an
ambiguity. Worth noting this widens discovery, so a project with umbilicus files at two
levels can now report an ambiguity where it previously resolved by luck of ordering; the
project field is the intended way out.

---

Verification for this commit: `test_umbilicus` is at 33 cases and `test_scroll_umbilicus`
at 18 (from 10), full non-network suite 149/149, no new warnings. Every new assertion was
mutation-checked — the guarded change broken once, the failure observed, then restored —
including one that reproduces finding 3 exactly: reverting the precedence gate makes the
resolver return a searched local file while an `s3://` location is configured.

Not covered by tests: `detachUmbilicus` itself, since the app layer has no
`MenuActionController` harness — it's compile- and read-verified only.

https://claude.ai/code/session_01Xb3JfFAJa5ZTGGRe5xbqmh

hendrikschilling:
1. I think you should really switch the line annotation sign flip stuff to this new umbilius - there should only be one way to do umbilicus data not multiple independent ones.

the others also still sound relevant to me.

```
  1. P2: The original consumer-integration finding remains unresolved.
     Attach Umbilicus persists the field at volume-cartographer/apps/VC3D/MenuActionController.cpp:1781, but no production code calls resolveScrollUmbilicus(). Line annotation still       
     performs its own directory search at LineAnnotationController.cpp:9379; patch tracing also keeps its independent path handling. The Fiber Map in PR 1459 does not make PR 1454         
     functional when merged independently.

  2. P2: Malformed metadata can still fail open.
     At volume-cartographer/core/src/Umbilicus.cpp:330, a present "metadata" value is only inspected when it is an object. Values such as "metadata": "invalid" produce no metadataErrors,  
     so the resolver accepts the file as legacy/unstamped and consumers may guess its scale. This is the same failure mode the latest commit intends to eliminate and should be rejected    
     explicitly.

  3. P2: Grandparent discovery is applied too broadly.
     volume-cartographer/core/src/ScrollUmbilicus.cpp:109 adds both parent and grandparent for every segment attachment. The grandparent is correct for a direct <volpkg>/paths/<segment>   
     attachment, but wrong for a folder attachment such as <volpkg>/paths: it additionally searches the directory containing the volpkg. An unrelated umbilicus there can cause a false     
     ambiguity or, worse, be selected when the volpkg itself has none. Grandparent search should depend on the attachment being an individual segment.

```

skyward7187:
You were right and I was wrong about the deferral, on two counts. First, this PR merged
on its own had **no** production consumer of `resolveScrollUmbilicus()` at all — the
Fiber Map that made it "real" lives in the stacked PR, so the automated finding was
correct that #1454 didn't stand up independently. Second, I argued partly from the risk
of changing sheet-normal orientation on already-traced fibers; that was overweighted,
since orientation is recomputed for the generated views on every rebuild and saved fibers
carry only `control_points` and `line_points`. A changed sign costs a mirrored cut view,
not annotation data.

So the line annotation path is switched over here, in `a6a7bddcb`, along with the other
two findings.

**1 — one way to do umbilicus data.** `ensureScrollUmbilicusLoaded()` now resolves
through `resolveScrollUmbilicus()`, so the project attachment, the several-candidates
refusal and the malformed-metadata refusal all apply to line annotation, and refusals are
logged rather than silently degraded.

Framing is shared, not reimplemented: `deriveUmbilicusScale()` in core carries a file's
coordinates into a target grid, preferring stamped **dimensions** (exact integer ratios,
no micrometres, refused when the three axes disagree), then the stamped voxel size, then
which power-of-two downsample of that grid the points fit inside and nearly fill. Both
consumers call it.

Worth knowing what this actually fixed, because it was worse than "ignores the field":
that search looked in `<volpkgRoot>/umbilicus.json`, and in project-file sessions the
package root is the *project* directory, a level above the volpkg — so on PHercParis4 it
found nothing at all and oriented against the volume's XY centre. The centre is in
session-volume coordinates while line points are in the 4×-finer annotation frame, so the
reference was off by a factor of four and pointed the wrong way for inner-radius fibers.
Hence the care about the target grid: the umbilicus is scaled into the annotation frame,
and the dense per-slice centres are built at *its* extent, since `Umbilicus::FromPoints`
sizes them from the shape it is handed.

A file that states no frame keeps its previous reading — `transform.json` and plausibility
floor included — rather than acting on a guess about annotation geometry. It still carries
the frame mismatch it always had; stamping it is the fix.

**Patch tracing is not converted.** `GrowPatch` still takes a path parameter fed from a
setting in the growth panel. `deriveUmbilicusScale()` sits in core precisely so that path
can adopt the same source, but I've left it as a follow-up rather than growing this PR
further. Say if you'd rather it came in here.

**2 — malformed metadata can still fail open.** Fixed in `7c54a72aa`: a `metadata` value
that is present and not an object is now blamed as a whole
(`metadata: expected an object, got "nope"`), so the resolver refuses the file. `null`
stays legal, since that is how a writer spells "no value" and refusing it would reject
files that are genuinely unstamped.

**3 — grandparent discovery applied too broadly.** Fixed. Only an *individual segment*
contributes its grandparent now; a folder attachment contributes just its parent, so
discovery no longer reaches the directory holding the volpkg. Segments are recognised by
their tifxyz payload (`x.tif`, `y.tif`, `z.tif`) rather than by `meta.json`, whose name is
generic enough to turn up for unrelated reasons.

One correction to my own earlier framing of this: for a layout where the project file sits
beside the volpkg, the grandparent of `<volpkg>/paths` *is* the package root, so it added
nothing and there was no exposure. The hazard is real when the project lives elsewhere
than the volpkg's parent, which is exactly what the new test models.

---

`test_scroll_umbilicus` is at 26 cases (from 18) and `test_umbilicus` at 33; full
non-network suite green — 148/148 here, 149/149 with the stacked Fiber Map — no new
warnings. The stacked PR was rebased onto this and now calls the shared helper.

Two things I decided rather than asked about, both easy to reverse. The format still says
which grid a file's coordinates index, not which pose that grid was in, so a file that is
both stamped *and* in a registered fixed-volume frame is outside what it can express; I
have not added a frame field on speculation, since no such file exists yet. And where the
umbilicus is unknown, line annotation still falls back to the volume centre in session
coordinates, which carries that same 4× mismatch — a fix in its own right, but a separate
one from making the attachment work.

https://claude.ai/code/session_01Xb3JfFAJa5ZTGGRe5xbqmh

skyward7187:
Rebased onto current `main`, and one more fix landed here rather than in the stacked PR
because the code it corrects is this one's.

**The annotation frame was derived two different ways.** `annotationFrameExtentXyz()` scaled
the volume's voxel counts by an open-data identity's `sourceCoordinateScaleFactor`, while
the voxel size sitting beside it came from `sourceOriginalResolution`. Neither consulted
`baseScaleLevel()`. For an untagged store rebased to level N that pairs a level-0
resolution with level-N counts: the umbilicus target grid is 2^N too small, and
`deriveUmbilicusScale()` can then refuse a correctly stamped file over an axis-ratio
mismatch that the pairing itself invented. The voxel size handed to it was worse — derived
from the identity alone, with no level-0 fallback at all, so an untagged rebased store
silently lost the stamped-voxel-size path.

`deriveAnnotationFrame()` resolves the resolution first, then takes the factor as the
**ratio** between it and the store's own, so the resolution and the counts always describe
one grid.

The ratio rather than a difference of pyramid levels, because `sourceOriginalResolution` is
only *computed* from `voxelSize()` when the tag is absent — when the tag is present it is an
absolute value that need not be a downsample of this store at all. Level arithmetic is
therefore ambiguous in exactly the case where both mechanisms are in play, whereas the ratio
is right by construction: it is the number that makes the two describe one grid.

It lives in its own header with no `VolumePkg` dependency, so the four cases are testable
without a `Volume` and a widely included header does not grow a heavy include.
`test_annotation_frame` covers plain, rebased-untagged, stamped, stamped-and-rebased, a
non-power-of-two stamp, and the unknown cases. One of them is worth naming: two stores of
one scan at different levels derive an *identical* frame, which is what lets switching
between them leave derived geometry alone.

Both call sites in this PR collapse into the one helper.

---

Full suite green, 150/150 with the stacked PR. A note on that number: my earlier runs used
`ctest -E network`, which quietly matched `fiber_network_layout` and excluded the stacked
PR's own layout test from every sweep I reported. It passes; the exclusion is by label now.

Still outstanding from earlier review, unchanged: patch tracing is not converted to the
resolver, and where no umbilicus is found line annotation still falls back to the volume
centre in session coordinates, which carries the frame mismatch this PR describes. Both are
follow-ups rather than parts of this change.

https://claude.ai/code/session_01Xb3JfFAJa5ZTGGRe5xbqmh

hendrikschilling:
```
  2. P2: Incompatible stamped dimensions still fail open in line annotation.
     deriveUmbilicusScale() correctly returns no result when stamped axis ratios disagree, at ScrollUmbilicus.cpp:213. But LineAnnotationController.cpp:9442 defines “stamped” as a         
     successfully derived non-inferred scale. A stamped file whose dimensions conflict with the target therefore enters the legacy unstamped path at :9464, where raw or transformed points 
     may still be accepted. That contradicts the claimed refusal behavior and can orient views using the wrong frame.

  3. P3: The attachment warning incorrectly treats dimensions-only metadata as unstamped.
     MenuActionController.cpp:1822 warns whenever voxelsize_um is absent, even when all three grid dimensions are present and are the preferred, authoritative scale source. The condition  
     should warn only when neither a complete dimension triplet nor voxel size is available.

```

skyward7187:
Both addressed in `5399465aa`. They turned out to be one root cause: the code never asked
directly whether the file declares its own grid, and inferred it two different wrong ways.

**2 — conflicting stamped dimensions failed open.** Correct, and worse than "contradicts the
claimed behaviour": `deriveUmbilicusScale()` returns nothing both for a file that states no
frame and for one whose stated frame is not a rescale of the target, and the call site could
not tell those apart. So a stated-but-unfitting frame entered the legacy reading, which
applies a registration inverse or falls back to raw points — the wrong-frame orientation the
metadata exists to prevent.

`umbilicusFrameClaim()` answers only *does the file state a frame*, which is the half that was
missing, and `decideUmbilicusLoadAction()` turns the pair of answers into Apply / Refuse /
UseLegacy. Extracted rather than inlined so all three arms are testable: the branch it drives
sits behind a package, a loaded volume and a registration transform, and was verifiable only
by reading.

One thing worth flagging, because it is not in the finding and I would have got it wrong
otherwise: refusal requires a usable target grid. The stamped-dimensions comparison only runs
when one exists, so without it a perfectly good stamp *also* yields no scale — and refusing
then blames the file for our own missing information. A stated frame that could not be
compared is not a conflict. The refusal log names the stated grid and the target so the
mismatch is visible rather than guessed at.

**Inferred scales are now applied here too**, matching the Fiber Map. You confirmed the
premise I was unsure of — an umbilicus always spans very nearly the full height — which is
exactly the property the inference tests, so it is sound on what it accepts. It also declines
what it cannot express: a file in a registration's fixed-volume frame is offset rather than
rescaled, so its points fit inside no candidate grid, inference yields nothing, and the legacy
reading that does handle a pose still runs. Two consumers of the same data no longer disagree
about what to trust.

**3 — the attachment warning.** Fixed by the same predicate. A dimensions-only file is fully
stated, by the *preferred* key, and was being told consumers may have to guess.

**Two consequences of making refusal real.**

Falling back to the volume centre is now said out loud — small red text at the top left of the
bottom strip — because doing it silently is what hid a whole scroll's frame mismatch in the
first place. It covers every unusable case, not only a conflicting stamp: several candidates
and malformed metadata were equally silent.

And attach/detach now announces itself through `CState`. Without that the resolved umbilicus
stayed cached until the volpkg root changed, so **attaching an umbilicus never reached normal
orientation at all** until the project was reopened. That is the rest of your earlier point 1
— I had fixed the search path but not the invalidation.

---

`test_scroll_umbilicus` is at 28 cases, covering the predicate over dimensions-only,
voxel-size-only, both, neither and a partial triplet, and the decision over all three arms
including the no-target case. Both new cases were mutation-checked: disabling the refusal arm
fails exactly the assertion that names it. Full suite green, 149/149 here and 150/150 with the
stacked PR.

Checked against real data rather than assumed: `PHercParis4.volpkg/umbilicus.json` stamps
8174×8174×18946 and 9.6 µm, both implying exactly ×4 into the 2.4 µm annotation frame, so it
takes Apply unchanged and sits outside the regression surface.

That surface is real though, and deliberate: a file stamping dimensions that do not match will
now be refused where it previously limped along through the legacy path, possibly producing a
usable orientation by luck. Refusing is the intent — a frame applied confidently but wrongly
is worse than none — but it is a behaviour change for any project holding such a file, which
is why the refusal is both logged with both grids and shown in the strip.

Patch tracing is still not converted to the resolver; unchanged, still a follow-up.

https://claude.ai/code/session_01Xb3JfFAJa5ZTGGRe5xbqmh

hendrikschilling:
close I think this just needs some reload/invalidation then its good to go:
```
  1. P2: Attach/detach still does not update open views.
     LineAnnotationController.cpp:1824-1829 only clears the cache and warning. The existing generated surfaces retain their old orientation because normals are only reapplied by
     materializeGeneratedViews() at LineAnnotationController.cpp:9692-9712. Thus MenuActionController.cpp:1805-1808 and :1854-1855 can report success while visible strips remain unchanged 
     until another action happens to rebuild them.

  2. P2: The cache remains stale across active-volume and some project changes.
     LineAnnotationController.cpp:9433-9442 keys the cached, already-scaled umbilicus only by the project directory. There is no CState::volumeChanged connection, and
     onVolumePackageChanged() at :6257-6260 does not invalidate it. Switching volumes, or opening another project file in the same directory, can therefore reuse geometry scaled and sized 
     from the previous annotation frame at :9466-9527. That stale geometry is then used for the orientation vote at :9656-9658.
```

skyward7187:
Both in `65c7ee795`, treated as the one job you described rather than two patches.

The root cause was a cache key that did not describe what the cache holds. `_scrollUmbilicus`
is not the file's contents: its points are already multiplied by a frame-dependent factor and
its per-slice centres sized to that frame's extent — a function of (file, annotation frame),
kept under a key of (project directory).

**2 — stale across volume and project changes.** The frame joins the key, compared inside
`ensureScrollUmbilicusLoaded()` rather than driven by a signal, so switching between downsample
levels of one scan — which derives an identical frame — still costs nothing, while a different
scan invalidates. A project change drops the cache outright regardless, because two projects can
share a directory *and* a grid, so both halves of the key can match while the file behind them
differs.

I deliberately did not add a `CState::volumeChanged` connection for this. An eager reset
re-resolves from disk on every volume change including the level switches where nothing needs
redoing; the lazy comparison is precise and free.

**1 — attach/detach did not reach open views.** Correct, and invalidating alone would not have
fixed it: nothing re-applies normals to surfaces already built, so the attach reported success
in the status bar while every open strip kept its old orientation. Told it worked, nothing
visibly changed — worse than the original silence. The generated views of open panes are now
rebuilt, which is what actually gets the new orientation on screen. Failures are handled per
pane rather than bailing on the first, since the alternative leaves later panes on the old
umbilicus while earlier ones have moved.

`sameAnnotationFrame()` gained a test for this use specifically, since it now decides when a
cached umbilicus may be kept rather than only how a view labels itself: levels of one scan
compare equal, a different extent or resolution does not, an imprecisely round-tripped voxel
size still does, and "unknown" never silently matches "known".

Full suite green, 149/149 here and 150/150 with the stacked PR.

https://claude.ai/code/session_01Xb3JfFAJa5ZTGGRe5xbqmh

hendrikschilling:
```
  1. P2: Active-volume changes still leave open views with the old orientation.
     The controller does not connect to CState::volumeChanged at volume-cartographer/apps/VC3D/LineAnnotationController.cpp:1882. The new frame comparison at :9634 runs only when
     orientedLineNormalsForSession() is called during a later materialization. Switching volumes immediately changes what the viewers render, but does not rebuild their generated surfaces.
     The cache is corrected on the next edit/rebuild, not at volume-switch time.

  2. P2: Attach/detach tries to materialize panes that do not yet have generated views.
     rematerializeOpenGeneratedViews() at LineAnnotationController.cpp:9538 checks only that a session exists. During initial asynchronous tracing, optimizedLine may still be empty. The   
     call reaches buildLineViewSurfaces() at :9917, which rejects an empty model, causing an erroneous modal “Could not create line annotation views” during attachment. It should skip     
     sessions without existing generated surfaces or a completed line.

  3. P2: The supposedly uniform dimension check accepts up to 2% anisotropy and averages it.
     volume-cartographer/core/src/ScrollUmbilicus.cpp:203-221 accepts hi <= lo * 1.02. For example, ratios (4, 4, 3.96) are accepted and converted into an averaged 3.986… scale, although  
     the metadata does not describe a uniform rescale. This can incorrectly accept a cropped or differently shaped volume. The PHerc rounding case needs tolerance, but 2% is much larger   
     than its roughly 0.006% discrepancy.

  4. Scope issue: the PR includes about 3,500 lines unrelated to umbilicus handling.
     Commit 23341d07d adds four unroll/sweep scripts and documentation, plus a new --tifxyz-output-dir feature in volume-cartographer/apps/src/vc_lasagna_line_probe.cpp:1537. None is      
     mentioned in the PR description. These should be removed or split into a separate PR.

```

skyward7187:
All four fixed.

**1 — volume changes don't reach open views.** `CState::volumeChanged` connected, queued (it fires
inside `switchVolume()` before focus and navigation are restored). It compares a per-session
`OrientationKey` — annotation grid, raw volume shape, how the umbilicus was read, its scale source,
transform identity — not the annotation frame: `PHerc0139_ds2`'s two volumes share voxel counts but
differ 2.5% in recorded voxel size, and whether that reaches the geometry depends on which scale
path was used.

**2 — materializing panes with no views.** Panes with no generated surfaces or no line points are
skipped. Deliberately no skip for a running task: a failed reoptimization never re-materializes, so
that would strand the pane. Dialogs suppressed with `QScopedValueRollback`.

**3 — the 2% tolerance.** Replaced with an exact test rather than a smaller tolerance: an integer
downsample maps a count to `floor(c/n)` or `ceil(c/n)` and nothing else, so candidates are checked
directly and only a unique match answers. No constant remains. Refuses your `(4, 4, 3.96)`, and
`100/100/1000` → `51/51/500` where a one-voxel slack would not. Counts cannot pin a *fractional*
rescale, so such a file must state `voxelsize_um`; integer non-powers-of-two (×3) work.

Correcting myself: I have said "exactly ×4" for PHercParis4 three times. Its stamp is
8174×8174×18946 against a 2.4 µm store of 32693×32693×75784, so the ratios are 3.9996331,
3.9996331 and 4.0 — and the old code returned their mean, 3.9997553, about four voxels short at the
far end of the scroll. It was inside the regression surface I said it sat outside. Now exactly 4.

**4 — scope.** 2,952 lines removed (the finding's ~3,500 also counts the umbilicus work).
`scripts/fiber_map/` will come as its own PR with the real-data evidence CONTRIBUTING.md asks for;
`sweep_grow_seg_from_segments.py` is dropped. Verified at patch level, since a `--stat` check cannot
see a dropped hunk in a file that is meant to differ: old vs new branch, excluding only the
extracted paths, is 0 bytes.

A self-review pass found nine more, all fixed — five fail-open paths (an incomplete dimension
triplet read as no statement at all; a stated frame that could not be *checked* read as legacy; an
inferred scale bypassing a registration transform; a cache miss keeping the previous provenance) and
four in the rebuild path. One is worth flagging: my first cache-key fix called a function that only
exists in #1459, so this branch would not have built alone. Caught, and now verified building `VC3D`
from a clean worktree.

Measured through the shipped code path rather than assumed: on the real project the resolver returns
the attached file, reports one dependency, and derives ×4 from stamped dimensions; with the
`umbilicus` field removed it finds the same file via a segment grandparent, reports three
dependencies, and leaves the shadowed `estimated_umbilicus.json` out of them.

`test_scroll_umbilicus` 37 cases, `test_umbilicus` 34, `test_annotation_frame` 15; non-network suite
149/149. Not "the full suite" — `-LE network` excludes three live-network tests, which I had also
been overstating.

Still deferred: patch tracing on the resolver. `vc_project_tifxyz.cpp:1136` is a third direct
`Umbilicus::FromFile` caller, so I will file both together. Not hand-verified: attach/detach and
volume switching on screen.

hendrikschilling:
```
  1. P2: Intersection-inspection strips still retain stale orientation. Their custom strips explicitly use orientedLineNormalsForSession() at volume-cartographer/apps/VC3D/
     LineAnnotationController.cpp:4382, but these sessions set suppressGeneratedViews = true at :12812. The new refresh loop skips them at :9676. Attaching/detaching an umbilicus or       
     switching volumes therefore refreshes ordinary annotation strips but not an open intersection inspection.

  2. P2: Same-grid volume switches can reuse the previous volume’s legacy transform. currentOrientationKey() reads _scrollUmbilicusTransform*, which still describes the old volume         
     (LineAnnotationController.cpp:9611-9615). The new volume’s transform is only discovered later during an actual reload at :10005-10026. If both volumes have identical dimensions/      
     annotation grids, the comparison succeeds and suppresses that reload, even when their transform.json files differ.

  3. P2: Exact coordinate scaling is replaced by a potentially rounded physical-resolution ratio. CoordinateIdentity provides the exact sourceCoordinateScaleFactor
     (OpenDataCoordinateIdentity.hpp:20,69), but annotationFrame() discards it and passes only sourceOriginalResolution (LineAnnotationController.cpp:9737-9745). deriveAnnotationFrame()   
     then calculates the coordinate factor as volumeVoxelSize / sourceResolution (AnnotationFrame.hpp:72-89). For example, a known level-2 mapping is exactly 4, while 9.596 / 2.4 =        
     3.99833. That fractional target extent can make an otherwise valid dimension-stamped umbilicus fail the new exact integer-rescale check. Coordinate extents should use the exact       
     coordinate factor; physical resolutions should remain physical metadata.
```

skyward7187:
This branch has been **restarted and rebuilt**. The previous head is preserved at
`archive/umbilicus-logistics-20260819`; the PR description now describes only the
current state. What follows is why, and where every prior finding went.

## Why restart

Six review rounds produced a branch that grew from 784 lines / 13 files at first review
to 3,244 / 22, and the last rounds were converging on three recurring patterns rather
than on a mergeable PR:

1. **Fail-open enumeration** — each round found another place where "couldn't determine
   the frame" silently degraded to a guess (malformed metadata read as absent, a
   non-object block, a partial triplet, a stated-but-unfitting frame entering the legacy
   path, a contradicted voxel-size stamp that only warned).
2. **Duplicate implementations** — line annotation's own search, the Fiber Map's own
   framing copy, two derivations of the annotation frame, a fingerprint that mirrored
   the resolver's scan instead of consuming it.
3. **Eager invalidation churn** — the open-views machinery (`OrientationKey`, the
   rematerialize loop, the queued `volumeChanged` comparison) tried to prove built views
   equivalent without re-resolving, and every round found another input the equivalence
   record was missing — culminating in your 2026-08-19 findings, where two of the
   branch's own fixes (the exact integer-rescale test vs. the rounded resolution-ratio
   frame derivation) contradicted each other.

So the branch was rebuilt from the state your first review saw, as eight commits, each
building VC3D and passing the non-network suite individually. The substance of every
accepted fix is preserved; the eager equivalence machinery is replaced, not repaired.

## The two design changes

**Exact factors, not resolution ratios (your 08-19 finding 3).** `deriveAnnotationFrame()`
now derives coordinate extents from the exact `sourceCoordinateScaleFactor` (read
straight from the tags, so it survives missing physical metadata) or the store's own
pyramid position; a stamped resolution stays physical metadata. An end-to-end test pins
PHercParis4's real values: Apply at exactly ×4, where the rounded 9.596/2.4 ratio is
refused by the same exact test.

**Counters and a conservative key, not an equivalence proof (findings 1 and 2).** The
umbilicus cache is keyed by what it holds — package root, **volume identity**, a stat
fingerprint over every resolver candidate **and the registration transform**, and the
annotation frame — so a same-grid volume switch re-resolves instead of proving the old
transform still applies. Freshness is an epoch counter: attach/detach and volume
switches bump it, sessions record the epoch their views were built at, and one queued,
coalesced pass rebuilds exactly the panes that are behind — plus the intersection
inspection as a whole, whose sessions suppress ordinary generated views and were exactly
what the previous refresh loop skipped. Both decisions are pure functions
(`UmbilicusOrientationFreshness.hpp`), unit-tested and mutation-checked, rather than
read-verified glue. The accepted cost is stated in the description: a level switch
rebuilds views whose orientation comes out unchanged — bounded, user-initiated waste
instead of a precision mechanism that kept being wrong.

## Fate of every finding

| Round | Finding | Fate |
|---|---|---|
| R1 | attach has no effect on existing consumers | Preserved — line annotation resolves through the shared path (commit "one way to find and frame") |
| R1 | malformed metadata fails open | Preserved (commit "malformed metadata is an error") |
| R1 | remote location falls back to search | Preserved (resolver commit; proved by a test that leaves a discoverable local file unused) |
| R1 | no Detach in the GUI | Preserved (attachment commit) |
| R1 | `<volpkg>/paths/<segment>` layout missed | Preserved (resolver commit) |
| R2 | no production resolver caller | Preserved — line annotation is the consumer in this PR |
| R2 | non-object metadata fails open | Preserved |
| R2 | grandparent search too broad | Preserved — only tifxyz-recognised segments contribute a grandparent |
| R3 | conflicting stamped dims fail open | Preserved (Apply/Refuse/UseLegacy) |
| R3 | dimensions-only file mislabeled unstamped | Preserved, with the claim predicate shared between attach dialog and consumers |
| R4 | attach/detach does not update open views | **Fixed lazily**: epoch bump + queued coalesced refresh reaches visible strips |
| R4 | cache stale across volume/project changes | Preserved via the completed cache key |
| R5 | volume switch leaves open views stale | **Fixed lazily** (same epoch mechanism) |
| R5 | modal error materializing a pane with no views | Mooted in code (the refresh walk skips them — now an asserted rule, not a loop detail) |
| R5 | 2% anisotropy accepted and averaged | Preserved (exact floor/ceil test) |
| R5 | ~3,000 unrelated script lines | Never re-enter |
| R5 | nine self-reported fixes | Fail-open ones preserved in core commits; rebuild-path ones died with the machinery they patched; the cross-branch build break is answered by per-commit clean builds |
| R6-1 | intersection strips stale after attach/switch | **Fixed**: the refresh rebuilds the inspection whole, and the lazy cache check sits at the one choke point both normals consumers pass through |
| R6-2 | same-grid switch reuses the old transform | **Fixed by design**: volume identity + transform fingerprint are cache-key inputs; regression test asserts the same-grid switch reloads |
| R6-3 | rounded ratio vs exact rescale test | **Fixed** (exact-factor derivation; end-to-end ×4 test) |
| #1459-1 | contradicted voxel-size stamp only warns | **Fixed in core** (`umbilicusStampContradiction` + refusal), both consumers inherit |
| #1459-2 | equal dims+voxel size read as same frame | Inherited limitation, documented; this PR no longer relies on frame equality for cache safety (volume identity is in the key) |
| #1459-3 | stale warning overwritten | #1459-side one-liner, lands when it rebases |

## Verification

Every commit builds from clean and passes the suite (149/149). Real data through the
shipped path: attached → 1 dependency, ×4 from stamped dimensions; field removed → same
file via discovery, 3 dependencies, shadowed `estimated_umbilicus.json` absent. Key
assertions mutation-checked (details in the description). The reconstruction was audited
hunk-by-hunk against the archive tag: every deletion maps to the retired machinery
listed above.

#1459 rebases onto this next; its consumed APIs (`resolveScrollUmbilicus`,
`scanUmbilicusCandidates`, `deriveUmbilicusScale`, the claim/decision helpers,
`sameAnnotationFrame`, `umbilicusGeneration()`) all exist here, with one deliberate
signature change: `deriveAnnotationFrame()` takes the exact coordinate factor as an
independent input, and the rebase migrates its call sites.

https://claude.ai/code/session_01QmLCg7rtVeZT5Zu4khi8B4

hendrikschilling:
Looks pretty solid!
Adress 1. and maybe give a shot at the rest as long as its not invasive, then this should be good to go.

  ```
1. Project changes can rebuild old-project line sessions against the new project. onVolumePackageChanged() invalidates the umbilicus and reloads the fiber list, but does not clear       
     _panes. When the new project selects its active volume, volumeChanged advances the epoch and refreshes those surviving sessions. This can resurrect generated surfaces belonging to the
     previous project, and subsequent saving may target the new project. Package handler
     (https://github.com/ScrollPrize/villa/blob/b9a75896a2428438626f82fbf0e254d1909ceb88/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L6381-L6388), refresh selection         
     (https://github.com/ScrollPrize/villa/blob/b9a75896a2428438626f82fbf0e254d1909ceb88/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L9701-L9717)

  Lesser concerns:

  - CState::setCurrentVolume() emits volumeChanged even when given the same Volume pointer. Consequently, package UI refreshes that reselect the current volume can cause an unnecessary    
    full line-view refresh. It is bounded and not a loop, but broader than actual volume switching. CState.cpp
    (https://github.com/ScrollPrize/villa/blob/b9a75896a2428438626f82fbf0e254d1909ceb88/volume-cartographer/apps/VC3D/CState.cpp#L222-L237)

  - Rebuilding is not transactional. Old surfaces are removed before the dialog has accepted all replacement viewers. If setGeneratedLineViews() fails, the old views are gone; the stale   
    epoch permits a later retry, but none is scheduled immediately. LineAnnotationController.cpp
    (https://github.com/ScrollPrize/villa/blob/b9a75896a2428438626f82fbf0e254d1909ceb88/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L10265-L10278)

  - The first refreshed line performs synchronous filesystem stats, JSON parsing, and possible transform loading on the GUI thread. Normally small and local, but a network-mounted project 
    directory could pause the UI. It does not involve volume chunks.

  - The actual queued Qt refresh path is not integration-tested; only its extracted selection/cache predicates are tested.

```

**Comment by skyward7187** [comment]
Both actionable points are in `c20dbd3d0`; the remaining three are answered below rather than coded, to keep this from sprawling again.

**1 — old-project sessions rebuilt against the new project.** Fixed. `onVolumePackageChanged()` now ends every session with its package: watchers detached the way `cleanupSurfaceName()` detaches them, the intersection inspection closed, a pending side-strip intersection cancelled, every pane and generated surface cleared from `CState`, and `_panes` swapped out before any `setSurface()` so the re-entrant `surfaceChanged` path never sees a live iterator.

One deliberate difference from a normal pane close: **no auto-save**. By the time the handler runs the current package is already the new one, so the save a normal close performs is exactly the wrong-target write your finding warned about. A session's work was persisted into its own package by the save-on-mutation paths while it was edited; this only drops the in-memory tail.

**Reselection re-emission.** Also fixed, but in this controller's handler rather than in `CState::setCurrentVolume()`: ViewerManager, CWindow, the chunked viewers and SeedingWidget all consume `volumeChanged`, and I did not want to change what they observe as a side effect of an umbilicus PR. The handler remembers the last acted-on volumeId and ignores a re-emission for the same one; the remembered id is cleared on package change, since two projects can name a volume identically (correctness never rested on this guard anyway — the cache key carries the package root and volume identity).

**The other three, deliberately not coded here:**

- *Non-transactional rebuild.* Real, and pre-existing: `materializeGeneratedViews()` has always replaced surfaces before the dialog accepts the new viewers, on every edit path, not just this refresh. Making it transactional means restructuring the shared dialog/viewer handoff that all edits use — too invasive for this PR, and the failure mode is bounded: the failure is logged, the session keeps its stale epoch, and the next epoch bump retries. I have not scheduled an immediate retry on purpose; a deterministic failure would turn that into a rebuild loop.
- *Synchronous stats/parse on first refresh.* Accepted cost, stated in the description: a handful of stats and one JSON parse, on a user-initiated switch, only for the first stale view. A network-mounted project pays similar synchronous costs on several existing paths; moving this one off-thread alone would buy little and complicate the cache.
- *The queued Qt path is not integration-tested.* Correct, and stated in the testing notes: the repo has no harness that drives the full controller, so the extracted predicates (`UmbilicusOrientationFreshness.hpp`) are the tested surface and the glue is read-verified. The package-change teardown is in the same category.

Full non-network suite green after the change, 149/149.

https://claude.ai/code/session_01QmLCg7rtVeZT5Zu4khi8B4

**Comment by hendrikschilling** [comment]
Ok this feels really close ... just some bookkeeping.

```
1. Package Switching With Open Line Sessions

  Pre-existing state on main:
  onVolumePackageChanged() only reloads the fiber list. Existing annotation sessions remain open across the package change, so they can later rebuild or save against the new package. That 
  is already incorrect.

  What PR #1454 changes:
  The new handler clears _panes, detaches running optimization watchers, and removes their surfaces. This prevents old sessions from being reused with the new project, but:

  - Running optimization results are discarded before being applied or saved.
  - LineAnnotationDialog objects are parentless and are neither closed nor deleted.
  - The visible dialog can therefore remain open but disconnected from its controller session.

  Minimal resolution:
  End sessions before replacing the active VolumePkg, while the old package is still active:

  1. Add a controller method for preparing a package switch.
  2. Refuse or defer the switch while an optimization is running.
  3. Reuse the existing finalization/save logic for idle sessions.
  4. Close dialogs using setCloseAfterFinalizationAllowed(true) and close().
  5. Clean up intersection-inspection sessions.
  6. Only replace the package if preparation succeeds.

  The post-change handler can then just invalidate the umbilicus cache and load the new fiber list. Defensive cleanup there should close dialogs, but must never attempt saving into the new
  package.

  This is the important finding and should be fixed before merging.

  2. Nontransactional Generated-View Refresh

  Pre-existing state on main:
  materializeGeneratedViews() removes the existing registered surfaces before setGeneratedLineViews() confirms that the dialog accepted their replacements. If installation fails, the new  
  surfaces are removed too, leaving no generated views.

  This could already happen during ordinary line updates.

  What PR #1454 changes:
  It adds automatic calls to this function after umbilicus attachment changes and volume switches. That exposes the existing failure mode during more operations, including operations where
  the user did not edit the line.

  It does not lose stored fiber data, but can leave the open workspace blank or partially reconstructed until another successful rebuild.

  Minimal resolution:

  - Retain the previous surface objects, names, and session metadata while constructing replacements.
  - Commit the new surfaces and session metadata only after setGeneratedLineViews() succeeds.
  - If registration must happen first, explicitly restore the previous surfaces and metadata on failure.
  - Keep the existing rule that orientationEpoch advances only after success.

  This is primarily a pre-existing robustness issue exposed by the new refresh path, rather than a new umbilicus bug.

```

**Comment by skyward7187** [comment]
Both in, as two commits, following your minimal resolutions step for step.

**1 — sessions end before the package switches** (`f058b9b59`). `prepareForPackageSwitch()` follows your prescription exactly:

- **Refuses while any pane's optimization runs** — checked over all panes before anything is torn down, so a refusal leaves the workspace intact and nothing mid-run is ever discarded.
- **Reuses the existing finalize/save/close logic**: it delegates to `prepareForUserFacingLineAnnotationOpen()`, which finalizes idle sessions synchronously, saves through the normal path, then closes dialogs via `setCloseAfterFinalizationAllowed(true)` + `close()` (`WA_DeleteOnClose` deletes them) — and which itself aborts before closing anything when a finalization fails.
- **Cleans up the intersection inspection** first (its dialog-less sessions never reach the finalize walk), and sweeps any dialog-less remainder through `cleanupSurfaceName()`, whose save now correctly targets the still-current package.
- **Both project-open flows gate on it and only replace the package when it succeeds**: `CWindow::OpenVolume()` (the funnel for File > Open, the recent list, and internal reopens) and the open-data sample flow. SpiralWorkspace's `setVpkg` calls are on its own private `CState`, so they are deliberately untouched.
- The `vpkgChanged` handler remains the defensive net for any swap that bypassed those flows: it now also closes surviving dialogs — and still never saves, since at that point a save writes into the wrong package.

**2 — restore instead of blank** (`326ef2910`). Took your "if registration must happen first" variant rather than reordering: the previous surface objects, names, and session metadata are retained across the replacement, and the `setGeneratedLineViews()` failure branch drops the rejected replacements, re-registers the old surfaces under their old names, and hands the session its metadata back. The success path is unchanged byte for byte — the clear still precedes the pose carry-over, which reads what is registered at replacement time — and the epoch still advances only on success, so a restored pane is retried by the next change rather than skipped.

Deliberately not done, to keep the scope: no reordering of the registration/install sequence (behavior-preserving restore instead), no exception wrapping of the mid-section (unwinding through `setSurface` re-entrancy would be a new risk; a throw there leaves the epoch behind and the next change retries — pre-existing and bounded), and no changes to `CState` or the dialog.

Suite green after each commit individually (149/149); the intermediate commit builds standalone. The prepare/restore paths are controller+Qt glue at the same read-verified boundary as the rest — no harness drives the full controller.

https://claude.ai/code/session_01QmLCg7rtVeZT5Zu4khi8B4

**Comment by hendrikschilling** [comment]
```
  1. P1: Open Data still destroys sessions before the new package-switch gate.
     beginOpenDataSampleOpenTask() closes the current volume
     (https://github.com/ScrollPrize/villa/blob/326ef29107e21c56b8f0355d02b67d58badb4617/volume-cartographer/apps/VC3D/MenuActionController.cpp#L671-L686) before prepareForPackageSwitch() 
     is called at task completion (https://github.com/ScrollPrize/villa/blob/326ef29107e21c56b8f0355d02b67d58badb4617/volume-cartographer/apps/VC3D/MenuActionController.cpp#L881-L897).    
     CloseVolume() emits vpkgChanged, invoking the defensive no-save teardown. The later gate therefore finds no sessions and cannot preserve running work.

     Minimal fix: call prepareForPackageSwitch() before CloseVolume() in beginOpenDataSampleOpenTask() and abort through the completion callback if it refuses. The completion-time gate may
     remain as a race defense.

  2. P2: Generated-view rollback restores CState, but not dialog state.
     setGeneratedLineViews() replaces _generatedViews, positions, overlay state and flags before an operation that can fail
     (https://github.com/ScrollPrize/villa/blob/326ef29107e21c56b8f0355d02b67d58badb4617/volume-cartographer/apps/VC3D/LineAnnotationDialog.cpp#L1637-L1684). The controller’s failure      
     branch (https://github.com/ScrollPrize/villa/blob/326ef29107e21c56b8f0355d02b67d58badb4617/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L10538-L10558) restores only     
     registered surfaces and session names. The dialog can retain new geometry metadata with old surfaces, reset offsets, and overlay-swap flags for callbacks that were never installed.   

     Minimal fix: validate the new cut frames before mutating dialog state, or make setGeneratedLineViews() internally snapshot and restore every changed field when it returns false.      

  3. P2: A refused package switch can still destroy intersection inspection.
     prepareForPackageSwitch() removes the inspection
     (https://github.com/ScrollPrize/villa/blob/326ef29107e21c56b8f0355d02b67d58badb4617/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L2242-L2248) before ordinary session    
     finalization. If finalization then fails, the package switch is refused but the inspection workspace is already gone.

```

**Comment by skyward7187** [comment]
All three are fixed (`dfc754fa5`, `93b899f1d`, `88cedbeb9`) — and this round I changed the process, not just the code: the fixes went through an adversarial whole-flow review *before* this push, which refuted parts of my first attempt and found two adjacent holes. Both of those are fixed here too, so what you're reviewing already survived the kind of analysis that has been generating these rounds.

**1 (P1) — open-data gate placement.** Your minimal fix, exactly: the gate now sits in `beginOpenDataSampleOpenTask()` before `CloseVolume()`, refusal delivered through a **queued** `onFinished` (callers were written against completion arriving after `begin()` returns), and the completion-time gate stays as race defense. The pre-push review then found the other end of the same flow: nothing blocks an ordinary `OpenVolume()` during the fetch, and completion replaced the intervening project with a bare `setVpkg()` — skipping the segmentation-editing teardown `CloseVolume()` exists for. Completion now runs the full guarded close (gate → `CloseVolume()` → `setVpkg()`), plus the watcher restart the ordinary open performs and this flow never did.

**3 (P2) — refusal ordering.** Reordered: every refusal point in `prepareForPackageSwitch()` now precedes everything it destroys. The pre-push review also refuted my "saves succeeded" assumption outright: the finalize walk only *queues* writes, and a synchronous serialization failure never queues anything and can strike before the fiber list holds a copy. So: a synchronous save failure refuses **inside the walk, before any dialog closes**, leaving the session saveable rather than save-suppressed; and the switch as a whole flushes (`waitForFiberSaves()`, the shutdown path's flush) and refuses if `_fiberSaveFailureCount` moved — the flush-and-compare gate that counter is documented for. The synchronous catch in `saveSessionAsFiber()` now counts toward it, which it silently didn't.

**2 (P2) — dialog rollback.** The in-place branch is transactional: the held snapshots (taken for the overlay handoff anyway) double as the rollback; the failure path restores geometry metadata, positions and placement focus; the mutations nothing can fail after — swap flags, offsets, viewer properties, and the arrow-pan cancellation, which a rejected update has no business making — moved after the validation, in the same relative order ahead of the renders and hooks that consume them. Verified: nothing reads the moved state between its old and new position, and no event loop runs inside the branch.

**Scoped claims, stated as scoped.** The package-destruction enumeration is: `CloseVolume` ← {OpenVolume (gated), open-data begin (gated), ~CWindow (saves via `saveOpenFibers()` first)}; `setVpkg` ← {OpenVolume, open-data completion (both gated), SpiralWorkspace (its own private `CState`), `closeAll` internal}; `vpkgChanged` is emitted only by `CState::setVpkg`. The independent review could not refute this enumeration, the prepare ordering, or the queued-refusal/stuck-flag behavior. All of it remains controller/Qt glue with no harness — read-and-adversarially-reviewed, not tested.

**Two pre-existing residuals, deliberately not fixed here — flagging them so they're a decision, not a discovery.** (a) The dialog's shared `_held*` trio can't represent *overlapping* per-pane overlay swaps: a second update overwrites the held generation while some panes still show the first — predates this PR's changes (the in-place update mechanism itself), needs per-pane or generation-tagged held state. (b) On a *repeatable* validation failure, the controller session keeps the rejected line while the restored UI shows the old one — the session-model update precedes materialization on all edit paths and always has; a full fix means snapshotting session geometry around optimization application. Both look like invasive rework of pre-existing mechanisms to me; happy to file them alongside #1538 as tracked follow-ups — or pull either into this PR if you consider it blocking.

Suite green (149/149) after each commit.

https://claude.ai/code/session_01QmLCg7rtVeZT5Zu4khi8B4

**Comment by hendrikschilling** [comment]
Ok I think these might be small enough to just leave in.
Let me know if you want to merge or do another iteration.

  ```
1. P1: Asynchronous save failure is detected after the sessions are destroyed.
     prepareForUserFacingLineAnnotationOpen() suppresses further saves and closes every dialog
     (https://github.com/ScrollPrize/villa/blob/88cedbeb997d1eb2882f9485c9f9603f5a8ffc7e/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L2216-L2232). Only afterward does       
     prepareForPackageSwitch() wait for the writes and potentially refuse
     (https://github.com/ScrollPrize/villa/blob/88cedbeb997d1eb2882f9485c9f9603f5a8ffc7e/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L2286-L2299).

     If a write fails, the old package remains active but its editing sessions are gone. Retrying the switch queues no replacement save, snapshots the already-incremented failure count,   
     and can proceed, dropping the only in-memory copy.

     Minimal fix: finalize and queue saves while keeping sessions/dialogs alive, wait for success, and only then set suppressFiberSave, close dialogs, and clean surfaces. Alternatively,   
     retain failed snapshots as mandatory retries that block every subsequent switch.

  2. P2: A stale Open Data task still overrides a newer project open.
     The code explicitly permits another project to open during the asynchronous fetch, then closes that newer project and installs the older task’s result when the fetch completes        
     (https://github.com/ScrollPrize/villa/blob/88cedbeb997d1eb2882f9485c9f9603f5a8ffc7e/volume-cartographer/apps/VC3D/MenuActionController.cpp#L923-L938). The full CloseVolume() fixes the
     previous surface/segmentation contamination, but the stale request still wins over the user’s newer action.

     Minimal fix: capture a package/open generation when starting the task. If it changed before completion, discard the result or ask before replacing the newer project.
```

hendrikschilling (approved):

PR #1463, opened 2026-08-15
Add fail-closed TIFXYZ volume preflight

aviad12g:
## What

Add `vesuvius.surface_preflight`, a deterministic, fail-closed command for checking a TIFXYZ surface and its source Zarr/OME-Zarr CT volume before rendering, label transfer, or inference.

The command validates:

- required TIFXYZ files and metadata
- compatible coordinate and optional mask rasters
- finite selected coordinates, valid vertices, and connected quads
- coordinate bounds against the selected CT array
- deterministic sampled CT-signal support

It writes an atomic, machine-readable JSON report and exits nonzero when any required gate fails.

## Why

A readable TIFXYZ directory can still be paired with the wrong scroll, crop, or multiscale array. That failure may otherwise surface only after an expensive downstream run—or produce an apparently valid run over unsupported coordinates.

Before this change, there was no single installed command that checked both surface structure and surface-to-volume pairing. After this change, downstream jobs can require an explicit `PASS` report before starting.

This is intentionally an input-pairing preflight. It does **not** claim that a surface is geometrically correct and does not replace self-intersection or local-orientation diagnostics.

## Real-scroll validation

Validated after a clean isolated install against the real PHerc1667 `w013_7.91um` TIFXYZ surface and the official Scroll 4 level-0 CT volume:

- 9/9 required gates passed
- 728,218 valid vertices
- 725,986 valid quads
- 0 out-of-bounds coordinates
- 16/16 deterministic CT samples had signal support

```bash
vesuvius.surface_preflight \
  --surface /path/to/PHerc1667/w013_7.91um \
  --volume https://dl.ash2txt.org/full-scrolls/Scroll4/PHerc1667.volpkg/volumes_zarr/20231117161658.zarr/ \
  --max-samples 16 \
  --output preflight.json
```

The focused tests also cover missing files, incompatible rasters, out-of-bounds coordinates, insufficient signal support, deterministic sampling, and atomic report behavior.

## Checks

- clean isolated install of `vesuvius[label-transfer]`
- installed `vesuvius.surface_preflight --help`
- clean-install real PHerc1667 run: PASS
- 8 focused tests: PASS
- bytecode compilation: PASS
- `git diff --check`: PASS


aviad12g:
Python compatibility CI stopped before the test step during dependency installation/synchronization: the Zarr 2.18.7 lane received an upstream `503 Service Unavailable` while fetching `napari-svg` from NVIDIA's package index, and the other matrix lane was cancelled. No project tests failed. The 8 focused tests, clean-install `vesuvius.surface_preflight --help` smoke test, and real PHerc1667 end-to-end validation all pass. A ScrollPrize maintainer will need to rerun the failed workflow.

Jinhojeong:
Supporting numbers for the CT-signal support gate, since I have been measuring exactly the failure it is built to catch. On PHerc1218, surfaces traced by the seed tracer from three seeds near a phantom-dense region stand over CT that is absent from the masked volume for 33.4, 39.2 and 1.8 percent of their in-region quads. axiosdevs measured 55 percent on a fourth seed in #1451. So a pairing that passes every file-level check can still put a third to half of its surface on nothing, and a preflight that fails closed on this is checking something that actually happens, not a hypothetical.

One measurement that bears on the sampled form of the check. The unsupported quads are not scattered speckle, they form one contiguous mass that reaches the patch boundary and extends inward, a third to nearly half of the patch on two of my three seeds. A deterministic sample catches a failure of that shape at modest sample counts, so sampling rather than exhaustive testing looks right for the gate.

If test fixtures are useful, my three traces with their per-quad support classification and the masked-CT support store recipe are at results/p8_support_mask in the diagnostic repo linked from #191, and I am happy to run the preflight against them and report what it flags.

Jinhojeong:
Ran it, as offered. My three traces come with an exact per-quad support classification, so they can referee the sampled gate: for each pairing I can predict what a correct sampler should report before running it. Seven pairings, defaults untouched.

| pairing | predicted support | the gate sampled | verdict |
| --- | --- | --- | --- |
| seed 0 against my windowed support store | 44.4% | 44.5% | FAIL, correct |
| seed 1, same store | 38.8% | 38.8% | FAIL, correct |
| seed 2, same store | 43.1% | 43.0% | FAIL, correct |
| seed 2 against the full masked CT on S3 | 98.2% | 97.0% | PASS, correct |
| seed 0 against a store for a different region | 0% | 0.0% | FAIL, correct |
| seed 0 against the thin prediction it was traced from | 14.1% | 14.1% | FAIL, correct |
| nonexistent volume path | fail closed | PathNotFoundError | FAIL, correct |

The predictions are the product of two exact numbers per trace, the share of quads inside my measured window and the measured support share inside it, and the 1024-vertex sample lands within 1.5 points of that product every time. The S3 row also shows the reader handling a remote OME-Zarr, which I had not expected from the import list. Passing a bare array with `--array-key` set is refused with a clear error rather than a guess, which fits the fail-closed contract.

One behaviour worth a line in the doc rather than a change. My support stores are full-shape but sparse, only the chunks of one region exist, so a surface that extends past the region fails the gate on coverage rather than geometry. That is why my near-clean seed 2 fails against the windowed store at 43 percent and passes against the full volume at 97. Fail-closed is the right reading of a partial volume, but a user pairing a good surface with a cropped or windowed store will see a support failure and may hunt for a geometry bug that is not there.

Runs and the referee table are under results/p8_support_mask in my diagnostic repo, and the traces with their per-quad classification are there if you want them as fixtures.

aviad12g:
Thank you—this is exceptionally useful independent validation. The seven predicted outcomes matching the preflight verdicts, with the 1,024-vertex sample remaining within 1.5 percentage points of the per-quad reference, gives strong evidence that the deterministic sampled gate is behaving as intended on real failure cases.

I added the suggested clarification in [`0ea5aec`](https://github.com/aviad12g/villa/commit/0ea5aec2c8a2de7c3a080d7d0d3a7524ac8fe928): a support failure means the selected volume does not support enough sampled surface coordinates and can reflect a wrong pairing, a cropped/sparse/partially populated store, or genuinely unsupported coverage; it is not by itself a diagnosis of defective surface geometry.

I kept the change documentation-only to preserve the PR's bounded scope. The traces and per-quad classifications would be valuable fixtures if maintainers want broader integration coverage. Thank you for running and publishing the referee comparison.

aviad12g:
Review-ready status at head 0ea5aec: the Python 2.18.7 and 3.2.1 lanes, CI, CodeQL, and the large-PR gate pass; Vercel authorization is the only external status failure. Independent real-data validation matched all 7 predicted verdicts, with sampled support within 1.5 percentage points of the per-quad references, and the documentation now clarifies cropped or sparse volume failures. The PR remains deliberately bounded and mergeable, with no further scope changes planned. @jrudolph @bruniss, a review when convenient would be greatly appreciated; I am ready to address any feedback promptly.

hendrikschilling:
Close to mergeable apart from two smaller issues:
```

  1. Valid TIFXYZ masks may be rejected. The new preflight requires mask.tif to exactly match the XYZ raster shape. The existing label-transfer loader also supports multipage and
     integer-scaled masks. This duplicated loading logic has already diverged and should be shared. New check
     (https://github.com/ScrollPrize/villa/blob/0ea5aec2c8a2de7c3a080d7d0d3a7524ac8fe928/vesuvius/src/vesuvius/surface_preflight.py#L448-L467)

  2. Non-base --array-key is unsafe. Coordinates are compared and sampled directly against the selected array without applying OME-Zarr scale/coordinate transforms. A lower-resolution     
     array can therefore produce false out-of-bounds or signal failures. It should either require a base-resolution array or transform coordinates.
```

aviad12g:
Thanks — both findings are addressed in commit `e0206576fb6e0056ba9d957e66b300794dd7bebd`.

1. I extracted the existing label-transfer mask behavior into a shared `load_tifxyz_mask()` helper. `load_surface()` and `surface_preflight` now use the same implementation, including first-page multipage handling and integer-scaled mask reduction. Focused tests verify that masks accepted by the canonical loader are also accepted by preflight, while incompatible masks still fail closed before opening the volume.

2. I did not add new OME-Zarr transform semantics. Explicit `--array-key` values are now restricted to the base-resolution dataset identified by the first `multiscales` entry, with the existing `0`/single-array fallbacks preserved. A non-base key fails closed with a precise expected/requested-key error.

Verification: the focused preflight and label-transfer I/O suites pass, 18/18; Ruff lint, Python 3.14 syntax compilation, and `git diff --check` also pass. Non-base arrays remain intentionally unsupported rather than being sampled with unverified coordinate transforms.

hendrikschilling:
close, just one small issue:
```
The non-base protection is incomplete. When --volume points directly to a concrete Zarr array, _resolve_volume_array() immediately accepts any array without verifying that it is base    
  resolution. Thus /volume.zarr/1 still bypasses the new check and samples base-coordinate TIFXYZ data at incorrect locations. This also contradicts the updated documentation saying       
  non-base arrays are rejected. Relevant code (https://github.com/ScrollPrize/villa/blob/e0206576fb6e0056ba9d957e66b300794dd7bebd/vesuvius/src/vesuvius/surface_preflight.py#L117-L121)     
```

aviad12g:
Thanks — the direct-array bypass is now covered in commit `c8f3f2f7947a2c9b17314ea35bc8896b523c74b2`.

`_open_volume()` now recognizes a concrete array path inside a `.zarr` store, opens the parent store, and sends that child key through the same base-resolution validation used by group + `--array-key`. Consequently `/volume.zarr/0` is accepted when parent `multiscales` metadata identifies `0` as the first/base dataset, while `/volume.zarr/1` fails closed with the expected/requested-key error. Standalone root arrays and ordinary one-array non-multiscale stores retain their previous behavior. No coordinate-transform support was added, and the existing group + `--array-key` path is unchanged.

Focused regression coverage now includes direct base PASS, direct non-base rejection, group/base PASS, group/non-base rejection, and standalone single-array preservation. The focused preflight and label-transfer suites pass 21/21; Ruff, Python 3.14 `py_compile`, and `git diff --check` also pass.

Issue #1448, opened 2026-08-14
zarr 3 installs: compute_st and several image_proc modules call APIs zarr 3 removed


axiosdevs:
`vesuvius.compute_st` is a declared console script (`pyproject.toml`) that cannot run on a zarr 3 install, and it is not the only site. `pyproject.toml` allows `zarr>=2.18.7,<4`, and the package targets zarr 3 elsewhere (`data/utils.py` branches on `_ZARR_V3` and uses `zarr.experimental.cache_store`), but several modules call APIs zarr 3 removed.

Reproduced with zarr 3.2.1:

```
Group.create_dataset  -> False        # removed, use create_array
Group.require_dataset -> False        # removed, use require_array
zarr.DirectoryStore   -> False        # removed, use zarr.storage.LocalStore
```

`vesuvius.compute_st` dies at `structure_tensor/run_create_st.py:105` — `zarr.open_group(...)` succeeds, then `root_store.create_dataset('structure_tensor', ...)` raises `AttributeError: 'Group' object has no attribute 'create_dataset'` before any voxel is processed.

Other reachable sites with the same defect:

| file | lines | call |
| --- | --- | --- |
| `structure_tensor/run_create_st.py` | 105 | `create_dataset` |
| `structure_tensor/create_st.py` | 206, 595, 609, 617, 626 | `create_dataset` |
| `structure_tensor/create_vf.py` | 58, 60, 62 | `require_dataset` |
| `structure_tensor/vf_format.py` | 86 | `create_dataset` |
| `image_proc/run/generate_mutex_graph.py` | 265, 271, 277, 283, 293 | `create_dataset` |
| `image_proc/run/voxelize_objs.py` | 1536, 1588, 1639 / 1551, 1604, 1656 | `zarr.DirectoryStore` / `create_dataset` |

The mechanical translation is `create_dataset` → `create_array`, `require_dataset` → `require_array`, `compressor=` → `compressors=`, `write_empty_chunks=` → array `config`, `zarr.DirectoryStore(p)` → `zarr.storage.LocalStore(p)`, keeping `zarr_format=2` so the on-disk layout is unchanged — or gating on the existing `_ZARR_V3` flag if zarr 2 has to keep working through the same code.

[...]

PR #1467, opened 2026-08-15
zarr 3: route array creation through a 2/3 compatibility layer (#1448)

jaideepsaipadhi:
# villa #1448 — zarr 3 compatibility for array creation

`vesuvius.compute_st` and 20 other modules call `Group.create_dataset`,
`Group.require_dataset`, `zarr.DirectoryStore` and `zarr.NestedDirectoryStore`
— all removed in zarr 3. `pyproject.toml` allows `zarr>=2.18.7,<4`, so a plain
rename would break zarr 2 users. This adds a compatibility layer and routes
every call site through it.

## Verified
- Patch applies cleanly to a fresh clone of `main`.
- **zarr 3.3.0**: 19 passed.
- **zarr 2.18.7**: 12 passed, 7 skipped (zarr-3-only behaviours).
- No removed APIs remain anywhere in `src/vesuvius`.
- No trailing whitespace introduced.

## Test
Needs only zarr, numpy and pytest — the test loads the shim directly from its
file, because the package `__init__` pulls in `requests` and other optional
runtime deps that a compat test should not require.

    cd vesuvius
    pip install zarr numcodecs numpy pytest
    python -m pytest tests/test_zarr_compat.py -q

Both ends of the supported pin:

    pip install 'zarr>=3,<4'      && python -m pytest tests/test_zarr_compat.py -q
    # 19 passed
    pip install 'zarr>=2.18.7,<3' && python -m pytest tests/test_zarr_compat.py -q
    # 12 passed, 7 skipped

Under zarr 2 the suite emits one `FutureWarning` from `NestedDirectoryStore` —
that deprecation is precisely what this PR gets ahead of.

No removed APIs remain:

    grep -rnE "\.(create_dataset|require_dataset)\(|zarr\.(Nested)?DirectoryStore|zarr\.TempStore" \
      --include=*.py src/vesuvius | grep -v zarr_compat.py
    # expect: no output

Functional check of the riskiest site (nested pyramid writer):

    cd vesuvius/src && python - <<'PY'
    import sys, numpy as np, numcodecs, zarr, tempfile, pathlib
    sys.path.insert(0,'.')
    from vesuvius.image_proc.run.zarr_tasks.utils import create_level_dataset
    root = pathlib.Path(tempfile.mkdtemp())/"pyr.zarr"
    a = create_level_dataset(str(root), "0", shape=(8,8), chunks=(4,4),
                             dtype=np.dtype("uint8"), compressor=numcodecs.Blosc())
    a[:] = np.ones((8,8),"uint8")
    keys = sorted(str(p.relative_to(root)) for p in root.rglob("*") if p.is_file())
    assert any("0/0/0" in k for k in keys), keys          # nested, not flat
    print("OK", keys)
    PY
    # (needs the package's runtime deps installed)

## Scope (measured via AST, not estimated)
60 `create_dataset` / `require_dataset` sites across 21 files, plus 6
`zarr.DirectoryStore` and 1 `zarr.NestedDirectoryStore`. Kwargs in use:
chunks 57, compressor 31, data 29, shape/dtype 28, write_empty_chunks 14,
overwrite 14, fill_value 8, dimension_separator 2, and 3 dynamic `**kwargs`.

## The part a naive port gets wrong
The two creation paths accept OPPOSITE keywords (verified on zarr 3.3.0):

| keyword               | `Group.create_array` | top-level `zarr.open`       |
|-----------------------|----------------------|-----------------------------|
| `dimension_separator` | TypeError            | accepted (nests keys)       |
| `chunk_key_encoding`  | accepted             | ValueError on zarr_format 2 |
| `write_empty_chunks`  | TypeError            | accepted but deprecated     |
| `config={...}`        | accepted             | accepted                    |

One translation table cannot serve both, so the shim carries two. Everything
else passes through unchanged, including `compressor` with a numcodecs
instance. Separately: `chunks=None` meant "auto" in zarr 2; zarr 3 raises.

## Three silent-failure hazards this avoids
1. `write_empty_chunks=False` dropped → sparse label volumes materialise every
   empty chunk. No error.
2. `dimension_separator` dropped → on-disk key layout changes, so
   previously-written data becomes unreadable. No error.
3. `NestedDirectoryStore` swapped for a plain store without the paired array
   kwargs → flat layout existing pyramid readers cannot find. No error.
   `nested_store_and_kwargs()` returns both halves so they cannot separate.

## A bug found by functional testing, not unit testing
`create_level_dataset` hand-writes a `.zgroup` with `zarr_format: 2`. On zarr 3
the array defaulted to format 3, which then rejects `dimension_separator`
outright. The nested kwargs now pin `zarr_format=2` so the array matches its
group. Unit-testing the shim in isolation would not have caught this — only
exercising the real call site did.

## Testing approach
Every test asserts an OBSERVABLE EFFECT — on-disk key layout, chunks actually
written, deprecation warnings, array format — not merely that a call returned.
All six translations were mutation-tested: each removed in turn, and the
corresponding test confirmed to fail. Two tests deliberately pin the
group-vs-open asymmetry, so if a future zarr unifies the keywords they fail
loudly, signalling the two tables can be collapsed.

## Not covered
Three sites build kwargs at runtime; migrated, but worth review against real
inputs:
  - `ink_detection/preprocessing/download_required_zarr_chunks.py:411`
  - `ink_detection/preprocessing/prepare_9um_isotropic_input.py:71`
  - `label_zarr.py:141` (already passes BOTH `dimension_separator='/'` and
    `write_empty_chunks=False`, so it exercises both translations)

The full package suite was not run — many modules need third-party deps
(numba, trimesh, aiohttp, fastremap, requests) unavailable in this environment.
All 22 migrated modules were import-checked: 13 imported cleanly, 9 were
blocked only by those unrelated deps, and zero failures were attributable to
this change.

## Possibly related
#1360 (`vesuvius.predict` fails on zarr 3 and exits 0 while writing nothing)
may share this root cause and is worth checking before merge.

### DIFF
--- a/vesuvius/src/vesuvius/image_proc/run/generate_mutex_graph.py
+++ b/vesuvius/src/vesuvius/image_proc/run/generate_mutex_graph.py
@@ -16,6 +16,7 @@ import zarr
 from numcodecs import Blosc
 from scipy import ndimage
 from tqdm import tqdm
+from vesuvius.zarr_compat import create_array
 
 
 @dataclass(frozen=True)
@@ -262,25 +263,25 @@ def process_volume(
     mask_rep_dtype = np.uint8 if _is_binary_array(repulsive_mask) else _select_min_lossless_dtype(repulsive_mask)
 
     # Cast on write to avoid holding two copies if possible
-    root.create_dataset(
+    create_array(root,
         "affinities/attractive",
         data=attractive.astype(attr_dtype, copy=False),
         compressor=compressor,
         chunks=chunk_shape,
     )
-    root.create_dataset(
+    create_array(root,
         "affinities/repulsive",
         data=repulsive.astype(rep_dtype, copy=False),
         compressor=compressor,
         chunks=chunk_shape,
     )
-    root.create_dataset(
+    create_array(root,
         "mask/attractive",
         data=attractive_mask.astype(mask_attr_dtype, copy=False),
         compressor=compressor,
         chunks=chunk_shape,
     )
-    root.create_dataset(
+    create_array(root,
         "mask/repulsive",
         data=repulsive_mask.astype(mask_rep_dtype, copy=False),
         compressor=compressor,
@@ -290,7 +291,7 @@ def process_volume(
     if cfg.store_labels:
         max_label = int(labels.max(initial=0)) if labels.size else 0
         label_dtype = _min_uint_dtype(max_label)
-        root.create_dataset(
+        create_array(root,
             "labels",
             data=labels.astype(label_dtype, copy=False),
             compressor=compressor,
--- a/vesuvius/src/vesuvius/image_proc/run/stack_composite_tifs.py
+++ b/vesuvius/src/vesuvius/image_proc/run/stack_composite_tifs.py
@@ -21,6 +21,7 @@ import tifffile
 import zarr
 
 from vesuvius.image_proc.run.workflow import gather_inputs, run_workflow
+from vesuvius.zarr_compat import create_array, local_store
 
 _STORAGE: np.memmap | zarr.Array | None = None
 _EXPECTED_SHAPE: tuple[int, int] | None = None
@@ -213,10 +214,10 @@ def _prepare_storage(
         store_dir = base_dir / f"stack_{uuid.uuid4().hex}.zarr"
         store_dir.mkdir(parents=False, exist_ok=False)
         sync_path = store_dir / ".sync"
-        store = zarr.DirectoryStore(str(store_dir))
+        store = local_store(str(store_dir))
         synchronizer = zarr.ProcessSynchronizer(str(sync_path))
         group = zarr.open_group(store=store, mode="w", synchronizer=synchronizer)
-        group.create_dataset(
+        create_array(group,
             "stack",
             shape=shape,
             dtype=dtype,
@@ -246,7 +247,7 @@ def _initialize_storage(spec: StorageSpec) -> None:
 
     if spec.mode == "zarr":
         assert spec.array_path is not None and spec.sync_path is not None
-        store = zarr.DirectoryStore(str(spec.path))
+        store = local_store(str(spec.path))
         synchronizer = zarr.ProcessSynchronizer(str(spec.sync_path))
         _STORAGE = zarr.open(store=store, path=spec.array_path, mode="r+", synchronizer=synchronizer)
         return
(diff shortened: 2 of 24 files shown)


jaideepsaipadhi:
This addresses #1448.

Two things worth flagging: the two creation paths accept opposite keywords — Group.create_array takes chunk_key_encoding and rejects dimension_separator, while top-level zarr.open is the reverse — so the shim carries two translation tables rather than one. And NestedDirectoryStore isn't just a store swap: on zarr 3 the nested layout comes from the array's key encoding, so replacing only the store writes a flat layout existing pyramid readers can't find, silently.

Placement question — I put the shim at vesuvius/src/vesuvius/zarr_compat.py. Happy to move it beside the existing _ZARR_V3 branching in data/utils.py if you'd prefer.

jaideepsaipadhi:
Flagging an overlap I should have caught before opening this: #1394 converts the same structure_tensor/ call sites, and puts the helpers in data/utils.py.

I'd rather not duplicate that work. Two options, whichever you prefer. I drop the four structure_tensor/ files here so this covers only the remaining subsystems, or I close this and re-open scoped to one subsystem once #1394 lands, reusing its create_group_array helper rather than adding a second shim.

On scope: AGENTS.md asks for small, focused diffs, and at 24 files this is broader than that. Happy to split it per subsystem.

jaideepsaipadhi:
Also dropped image_proc/run/zarr_tasks/utils.py — #1431 covers it, and handles a second bug there (workers receiving the pyramid root rather than the resolved level) that this PR doesn't touch. Remaining scope is the subsystems neither #1394 nor #1431 covers.

jaideepsaipadhi:
Happy to split this per subsystem (neural_tracing, scripts, models, image_proc) if that reviews better — AGENTS.md asks for focused diffs and #1394/#1431 both scoped that way. Left as one PR for now since the remaining call sites are mechanical and share the helper, but say the word.

jaideepsaipadhi:
The required CI and Gate checks here are still showing "waiting for status to be reported" — I believe the workflow runs need maintainer approval, which hasn't happened on any of my PRs, so none of them can go green. The Vercel failure is just the docs preview deploy, which can't authorize from a fork.

Could someone approve the workflow runs? Happy to have them run and fix whatever they turn up.

I also opened more PRs at once than was reasonable to review, and I've closed several to cut the load. If it'd help, tell me which two or three are worth keeping and I'll close the rest.

@bruniss @jrudolph — apologies for the direct ping; the approval looks like the only thing blocking all of these.

PR #1471, opened 2026-08-16
create_label_zarrs: stream striped TIFFs, not just tiled (refs #1231)

jaideepsaipadhi:
The streaming conversion path only takes effect when the input TIFF is tiled. Any other TIFF -- including anything written with a plain tifffile.imwrite(path, array) with no tile= -- falls through to tifffile.imread() loading the whole image into memory. #1231 reports this OOMing on a 25.1 GiB allocation for a real _validation_mask.tif. Published _inklabels/_supervision_mask assets happen to be tiled, so this never bites on anything the pipeline ships -- only on a label image someone generates themselves.

tifffile exposes tiled and striped TIFFs through the same block API: page.chunks/chunked/dataoffsets/databytecounts/decode() all work identically for a strip as for a tile -- a strip is just a chunk shaped (rows_per_strip, full_width) with a chunk grid one column wide. Verified against tifffile 2026.3.3, including that decode()'s returned position gives a strip's y-offset in the same slot it gives a tile's, which the existing block-placement math relies on.

So the existing tiled-only streaming writer generalizes almost unchanged: gated on codec support (none/LZW/deflate/PackBits, verified each round-trips byte-exact) rather than is_tiled. An unsupported codec (JPEG) correctly falls through to the old in-memory path.

While verifying this I found a separate, pre-existing bug: a genuine multi-page TIFF already produces silently wrong output on the non-streaming path today (tifffile.imread stacks pages, then the channel-squeeze logic mistakes the page axis for height -- a (5,40,60) volume comes out as (5,40)). That's real but out of scope here; this PR adds an explicit guard so multi-page input (tiled or striped) hits the identical old path, unchanged, bug included. Worth its own issue.

14 tests (6 new), including one mutation-testing the compression check and one confirming the multi-page guard actually changes behavior when removed. Patch verified clean on three independent fresh clones of main.

This addresses only the second half of #1231 (the create_label_zarrs OOM).
The first half — whether published segments are meant to ship a
_validation_mask — is a question for maintainers, not something a PR
resolves.


khj1222:
Thanks for picking this up — I filed #1231, so this is the fix I was hoping someone would
write. One heads-up that may save you a round: this file exists twice in the repository,
and the same bug was fixed once already on the other copy.

- `merge-ink-pipelines`, `ink-detection/koine_machines/preprocessing/create_label_zarrs.py`
  — fixed in #1234, merged 2026-08-14.
- `main`, `vesuvius/src/vesuvius/ink_detection/preprocessing/create_label_zarrs.py`
  — the file you are patching. I checked it today: still `if not page.is_tiled` on the
  streaming path, still `load_image` → `build_pyramid_with_mode` on the fallback.

So this is not a duplicate of #1234 — the copy you are fixing is still broken, and it is
the more visible of the two.

## The two fixes are not the same fix, and yours is the better half of the level-0 read

Yours streams strips through the chunk API and never holds the whole image. Mine does not
stream strips at all — it reads the image once with `load_image` and then derives the
pyramid in 2D, writing each level straight to `DEFAULT_LABEL_SLICE`. What that avoids is
the *other* allocation on that branch: `build_pyramid_with_mode` embeds every level into a
`(VOLUME_DEPTH, H, W)` volume before writing, 64 of 65 slices zeros. On the real file
behind #1231 that multiplier, not the image itself, is what made the number large.

Measured on that file (32249 × 51380 striped `_validation_mask.tif`): 114.5 s → 66.5 s,
peak RSS 1.99 GiB — which is dominated by holding the level-0 image, exactly the cost your
approach does not pay.

## A data point from the review on the sibling copy, in case it is useful

My first version called `_build_downsample_levels_from_zarr` for the levels, like the tiled
branch does and like yours will after this change. In review, erdpx asked for the opposite:
build the 2D pyramid in memory and write each level straight to the label slice, instead of
reading level *n−1* back out of zarr. That is where the 114.5 s → 66.5 s came from, and the
1.61 → 1.99 GiB peak RSS is what it cost.

I mention it only because the same question may come up here, and on a striped input the
trade reads differently than it did for me: holding the level-0 image to derive the levels
would reintroduce precisely the allocation your streaming avoids. You are better placed than
I am to judge which side that lands on — I just did not want you to meet the question cold,
as I did.

## Your multi-page finding reproduces on the merged copy

Not specific to `main`. The merged copy normalizes the same way:

```python
image = np.squeeze(image)
if image.ndim == 3:
    image = image[..., 0]
```

I ran a `(5, 40, 60)` array through it: comes out `(5, 40)`, identical to what you describe.
Agreed it is out of scope here and deserves its own issue. If you do open one, it may be
worth naming both paths, since a fix to one will not reach the other — and it is your
finding, so I would rather leave that to you than file it myself.

## If it is useful

The verification set I used for #1234 is four synthetic cases — binary and grayscale, even
and odd extents — each requiring byte-identity across all 6 levels between the streamed path
and the in-memory path, plus the real 32249 × 51380 file. The odd-extent cases are the ones
that earn their keep; that is where a block/strip boundary and a downsample step disagree.
Patch copy: https://github.com/khj1222/vesuvius-challenge/blob/main/submission/villa-pr-stream-untiled-labels.patch

Happy to review or test this against the striped masks my harness generates, if that helps
it move.

jaideepsaipadhi:
Thank you — the duplicate warning genuinely saved me a round, and the pointer to the review history on #1234 is the more valuable half of this.

**On the pyramid question**

I think the trade does land differently here, and against the in-memory approach, for the reason you identify: streaming level 0 is the whole point of this patch. On a striped input, building the 2D pyramid in memory means holding the level-0 image to derive the levels, which reintroduces exactly the allocation the streaming path exists to avoid. Your 1.61 → 1.99 GiB was the cost of that on the fallback branch; here it would be the cost of undoing the fix.

So my inclination is to keep `_build_downsample_levels_from_zarr` and accept the level n−1 read-back. But I would rather measure that than assert it, and I don't yet have numbers on a striped input the way you do on the real file behind #1231.

**Taking you up on the offer**

If you're willing to run this against the striped masks your harness generates, that would help a lot — particularly the odd-extent cases. I want to be straight that those are the ones I am least confident about: a strip boundary disagreeing with a downsample step is a correctness risk in my path specifically, and it is not something my current testing covers. If it breaks there, I would rather find out from your harness than after review.

**Multi-page normalization**

Agreed it is out of scope here, and thank you for confirming it reproduces on the merged copy — I had only checked `main`. I will open an issue naming both paths.

**One thing I am unsure about**

Given both copies now diverge in approach (streamed here, in-memory on `merge-ink-pipelines`), is there any appetite from maintainers for converging them, or is the duplication deliberate? I ask because whichever way the pyramid question goes here, the two files will end up solving the same problem differently, and a future reader will hit the same confusion I did.

khj1222:
Ran it against the harness. Short version: **the streaming path reproduces today's output exactly on every extent I could construct, and it hard-crashes on a class of striped file that converts fine today.** The crash is one line and I have a verified fix for it. The odd-extent case you were least confident about turned out not to be where it breaks.

## What I ran

Your head `6cec011` against its parent `aab644c` (`create_label_zarrs.py` and `label_zarr.py` are byte-identical between that commit and current `main`, so "parent" and "today" are the same thing here). 55 variants, same content through both trees, compared per pyramid level.

One methodological note first, because it nearly cost me the whole run: **equality has to be read off the decoded arrays, not the chunk files.** The blosc payloads are not reproducible across processes — the *same code on the same input* writes different bytes twice — so my first pass flagged all 55 variants as differing and every one of them was compression, not data.

**42 of 42 variants that ran on both trees are identical at all six levels: 252 level comparisons, 9,661,092,220 voxels, zero mismatches.**

Scripts, raw per-level results and a checker that re-derives every number quoted here: [`runs/pr1471_striped_check/`](https://github.com/khj1222/vesuvius-challenge/tree/main/runs/pr1471_striped_check).

| axis | covered |
|---|---|
| extents | 2048×2048, 2049×2051, 1025×1027, 1543×2311, 4099×37, 37×4099, 2050×2050 (even, but odd from level 1 down), 63×65 |
| layouts | tiled 256×256; `rowsperstrip` 1, 7, 1024, whole-image |
| downsample mode | `nearest` and `mean` (composite naming), the latter on three extents |
| dtype | uint8 binary, uint8 grayscale, uint16 |
| codecs | none, LZW, deflate, PackBits — plus JPEG, which falls through to the in-memory path on both trees (`streamed_tiled_tiff=false`) and matches |

So the specific risk you named — a strip boundary disagreeing with a downsample step — does not materialise, and I think it structurally cannot. `_write_downsample_block` addresses blocks in the *target* grid and reads `source[:, block_y*2 : ...]`, so every 2×2 window starts on an even source row no matter where strips fall. `rowsperstrip=7`, 1024-row blocks and a 2049-row image agree with the in-memory pyramid exactly, in both modes.

## What does break: a strip of exactly one row

13 of the 55 crash on your tree and convert fine on the parent:

```
ValueError: Expected a 2D image at ...\seg_inklabels.tif, but got shape=(2051,)
```

`page.decode()` returns `(depth, rows, columns, samples)` for a strip exactly as it does for a tile — your reading of the block API holds. But when `rows == 1` the result is `(1, 1, W, 1)`, and `_normalize_to_2d`'s `np.squeeze` drops **both** leading axes, so a single row comes back one-dimensional and the 2D guard rejects it. Decoded shapes, straight out of tifffile 2026.3.3:

| input | strip | `decode()` shape | after `_normalize_to_2d` |
|---|---|---|---|
| 2049×2051, `rowsperstrip=1024` | 0 | `(1, 1024, 2051, 1)` | `(1024, 2051)` |
| 2049×2051, `rowsperstrip=1024` | 2 (last, 1 row) | `(1, 1, 2051, 1)` | **ValueError** |
| 2049×2051, `rowsperstrip=7` | 292 (last, 5 rows) | `(1, 5, 2051, 1)` | `(5, 2051)` |
| 2048×2048, tiled 256 | any | `(1, 256, 256, 1)` | `(256, 256)` |

Two things worth noting. Short strips are *not* padded — a 5-row remainder decodes as 5 rows — so only the exactly-one-row case is affected. And tiles can never hit it, because TIFF tile heights are multiples of 16; that is why the `is_tiled` gate hid this.

The trigger is a property of the extent, not the content:

- `rowsperstrip == 1` — every strip is one row. I got there by asking tifffile for it and I don't have a survey of how often writers in the wild do; or
- `height % rowsperstrip == 1` — only the last strip is, which is roughly a 1-in-`rowsperstrip` accident of image height.

Both are in the matrix: 513×1027 at `rowsperstrip=256`, and 2049×1027 at `rowsperstrip=16`, crash; the same extents tiled do not. This is a regression in the strict sense — those files convert today, slowly, through the in-memory path.

## A fix, verified

`_normalize_to_2d` is right for a whole image and wrong for a decoded block, so index the axes instead of squeezing them:

```python
def _decoded_block_to_2d(decoded: np.ndarray, source_path: Path) -> np.ndarray:
    """Return a decoded tile or strip as 2D without losing a length-1 row axis.

    ``page.decode`` returns ``(depth, rows, columns, samples)`` for a strip
    exactly as it does for a tile. ``_normalize_to_2d`` squeezes, which is
    right for a whole image but drops the row axis of a one-row strip --
    ``(1, 1, width, 1)`` becomes ``(width,)`` -- so index the axes explicitly.
    """
    block = np.asarray(decoded)
    if block.ndim == 4:
        return np.ascontiguousarray(block[0, :, :, 0])
    return _normalize_to_2d(block, source_path)
```

and call it in place of `_normalize_to_2d(decoded, input_path)` in `_write_streamed_tiff_level_zero`. I deliberately did not reach for `np.atleast_2d`: it fixes a one-row strip and silently transposes a one-column image.

Re-ran base / head / head+fix over 27 variants chosen to hit exactly this condition: **head crashes on 25, head+fix crashes on 9, and all 18 comparable outputs are identical to the parent's in-memory output at every level.**

Your own 14 tests pass on both trees, which is to say they don't reach this — a one-row-strip fixture would be the cheapest thing to add.

## Pre-existing, and not yours: degenerate extents

The 9 that still fail with the fix are 1×4099, 4099×1 and 1×1, and they fail on **the parent too**, earlier and in a different place — `_normalized_2d_shape` squeezes the length-1 axis out of the *page* shape and then rejects the result as not 2D. So a whole-image row or column has never converted through either path. I left that alone: it is a separate decision about what `_normalize_to_2d` should mean, and it is worth knowing only so that a squeeze fix here does not accidentally paper over it.

## Real file

Our harness writes `tile=(256, 256)` + LZW today, so I re-encoded a real one rather than claiming to have a striped original: the held-out validation mask for `w00_20231016151002`, 32249×51380 uint8, 19,083,344 ink pixels, odd height. Same pixels, different container. Both runs on your tree, one at a time, no other load:

| input layout | blocks | file | wall | peak RSS |
|---|---|---|---|---|
| tiled 256×256 | 25,326 tiles | 10.49 MB | 143.27 s | 0.884 GiB |
| `rowsperstrip=1024` | 32 strips | 1.36 MB | 195.84 s | 0.881 GiB |

**The two outputs agree at all six levels, zero mismatched pixels** (19,083,344 / 4,773,300 / 1,194,557 / 299,279 / 74,832 / 18,713 non-zero from level 0 down).

The striped run is 37% slower, and the reason is structural rather than alarming: a strip spans the full width, so with 1024-wide write blocks each strip is decoded once per horizontal block — 51 times here. If you want that back, the smallest change is to let the write block span the full width when `page.chunked[1] == 1`; the block then holds one strip's worth of rows (52 MB at this width) instead of a 1024×1024 tile. I measured that variant too — see the table below — but it is an optimisation, not a correctness issue, and the PR stands without it.

One coincidence worth stating: this file misses the crash by luck. 32249 mod 1024 = 505 and 32249 mod 7 = 0. At `rowsperstrip=8` (32249 mod 8 = 1) it would have died on the last strip, after doing all the work.

## The pyramid question, with numbers

You said you would rather measure the level *n*−1 read-back than assert it, and that you did not have striped numbers the way I do. Here they are, same file, same box, same sampler, one run at a time:

| implementation | wall | peak RSS |
|---|---|---|
| this PR — stream level 0, derive levels from zarr | 195.84 s | 0.881 GiB |
| this PR + full-width write blocks | 104.38 s | 0.881 GiB |
| `merge-ink-pipelines` (#1234) — stream level 0, derive levels from a 2D pyramid in memory | 64.62 s | 1.929 GiB |

Both stores use the same chunking `(65, 128, 128)`, the same Blosc(zstd, clevel 3, bitshuffle) and `write_empty_chunks=False`, so the storage side is not what separates them.

**Read that as: your design costs 3.0x the wall clock and gives back 2.2x on peak memory, and most of the time it costs is not the read-back.**

The memory column is the part I would defend. The merged copy's 1.929 GiB is the level-0 image held in memory to derive the pyramid — the exact allocation the streaming path exists to avoid — and it grows with the image, while the read-back's footprint does not. That 1.929 GiB also reproduces #1234's own post-review measurement of 1.99 GiB on this same file, which is a useful check that these two runs measure the same thing.

The time column mostly is not about the read-back. Full-width write blocks — one extra condition on the block iterator — bring your design to 104.38 s at the same 0.881 GiB, closing 70% of the gap, and that output is identical to the tiled one at all six levels too.

**There is one thing you should know before deciding, though, because it is about your PR and not mine.** The pre-review version of #1234 *was* your design — stream level 0, derive the levels by reading level *n*−1 back from zarr — and @erdpx asked me to replace it:

> would you consider building only the 2d pyramid in memory, then writing each level directly to DEFAULT_LABEL_SLICE, instead of rereading the zarr levels from disk? this should be considerably faster while still avoidng materializing the N-slice volumes; the 2d pyramid itself is only ~1.33× the source image size

My numbers at the time, on this same file: 114.5 s / 1.61 GiB before, 66.5 s / 1.99 GiB after. Today's runs reproduce the ordering. So the speed half of that review holds up, and the ~1.33× estimate turns out to be roughly right too (1.929 GiB against a 1.66 GB image).

Where I think your case is still live: 1.33× of *the source image* is unbounded in the image, and the read-back is not — 0.881 GiB here whatever the width — and with full-width blocks the speed gap narrows to 1.6x. That is the argument I would put in front of a reviewer rather than the read-back on its own.

## On converging the two copies

I don't know, and I would rather say so than guess on a maintainer's behalf. What I can offer is that the divergence is not accidental: it is the review above, recorded on one branch, and #1234 was merged by @erdpx on 2026-08-14 in that shape. Whether that preference should propagate to `main`, or `main`'s approach back the other way, needs someone who knows where `merge-ink-pipelines` is going — but at least it is a decision someone made, not two people solving the same problem in ignorance of each other.

For what it's worth I agree with your instinct on scope: whichever way it goes, converging the files is a separate change from this one, and holding this PR for it would be the wrong trade.

hendrikschilling:
Supporting many different cases is hard - this does not seem to be ready yet and I'm not sure we should add much complexity vs just converting to tiled tiffs beforehand. So without some good counter arguments I tend to close this.

For reference the llm review:

 ```
 1. [P1] The stated default tifffile.imwrite workload is decoded hundreds of times.
     create_label_zarrs.py:334-352 iterates 1024×1024 destination blocks and decodes every intersecting source chunk inside that loop. A striped TIFF has full-width chunks, so each strip  
     is decoded once per horizontal block. Worse, default uncompressed tifffile.imwrite produces one whole-image strip, which is decoded once per destination block.

     Locally, a 2049×3073 single-strip image was read and decoded 12 times. For the issue’s 16125×25690 image, this becomes 416 full-image decodes. This avoids the 65-slice allocation but 
     can require enormous memory traffic and still allocates at least one full source strip. The iteration must be source-strip-major, decoding each strip once.

  2. [P1] One-row strips crash and leave an output that subsequent runs skip.
     At create_label_zarrs.py:356, _normalize_to_2d() squeezes decoded (1,1,width,1) data to (width,), then rejects it. This affects rowsperstrip=1 and images whose final strip contains   
     exactly one row. These files previously succeeded through the in-memory path.

     I reproduced the exact ValueError. Because the Zarr is created before decoding, the failed conversion leaves a partial output; the next run sees that directory at
     create_label_zarrs.py:453-459 and reports it as skipped.

  3. [P2] The compression whitelist disables existing tiled streaming.
     Previously every tiled TIFF used the streaming path. The new check at create_label_zarrs.py:296 sends tiled formats outside {none, LZW, deflate, PackBits} through the memory-heavy    
     fallback. Tiled JPEG and ZSTD are both newly classified as non-streamable, potentially restoring the original OOM for existing tiled inputs. The whitelist should constrain striped    
     inputs only unless a tiled codec is proven broken.

  4. [P2] The multipage guard does not preserve prior tiled behavior.
     create_label_zarrs.py:279-294 sends all multipage TIFFs to the already-broken fallback. Previously a tiled multipage file streamed its first page. Now a (5,32,48) tiled file becomes a
     malformed (5,32) image through load_image(). The new test checks only that streaming is disabled, not the resulting data, so it codifies rather than detects this change.
```

**Comment by khj1222** [comment]
Not my PR and I have no stake in whether it lands — but two of the four review items were
checked against this branch a few days ago, [in this
thread](https://github.com/ScrollPrize/villa/pull/1471#issuecomment-5478665794), and one of
them has a measured answer that I think changes its weight. Raw artifacts and both patches
are in
[`runs/pr1471_striped_check/`](https://github.com/khj1222/vesuvius-challenge/tree/main/runs/pr1471_striped_check).

**P1 #2 (one-row strips) is real, and it is already fixed and verified.** I hit the identical
`ValueError` on 08-31, before this review. The trigger is extent, not content:
`rowsperstrip == 1`, or `height % rowsperstrip == 1` so that only the final strip has one
row. It caught 13 of 55 matrix variants and 25 of 27 in a targeted re-run. `page.decode()`
returns `(depth, rows, columns, samples)` and `np.squeeze` drops the row axis when
`rows == 1`; indexing the axes explicitly (`block[0, :, :, 0]`) fixes it. That patch rescues
16 of the 25 — the other 9 are 1×N / N×1 / 1×1 degenerate extents that fail on the parent
tree too, so they are not this PR's doing — and all 18 outputs that can be compared against
the parent come out identical to it at every level. (`np.atleast_2d` looks like the same fix
and is not: it repairs one-row and silently transposes one-column.) The patch is
`patch_onerow_strip_fix.patch`.

**P1 #1 (re-decoding) is real, and the remedy it asks for is a block-shape change that I
measured.** The review is right that this iterates destination blocks, so a full-width strip
is decoded once per horizontal block — on a real 32249×51380 mask that is 51 decodes per
strip. Making the write blocks full width, which for full-width strips is exactly the
"source-strip-major, decode each strip once" the review asks for:

| tree, same 32249×51380 label image | wall | peak RSS |
|---|---|---|
| this branch, tiled 256 input | 143.27 s | 0.884 GiB |
| this branch, striped input (`rowsperstrip=1024`) | 195.84 s | 0.881 GiB |
| **this branch + one-row fix + full-width write blocks** | **104.38 s** | **0.881 GiB** |
| `merge-ink-pipelines` today (#1234), striped input | 64.62 s | 1.929 GiB |

So after a small patch (`patch_fullwidth_blocks.patch`) the striped path is *faster than this
same branch reading a tiled copy of the same image*, at the same memory. The decode traffic
in P1 #1 is a consequence of the block shape, not of the streaming design.

One caveat that cuts against the PR and in favour of the review: I tested
`rowsperstrip=1024`, **not** the uncompressed `tifffile.imwrite` default of one
whole-image strip. That case is strictly worse than anything in my table, and I did not
measure it.

**On "just convert to tiled beforehand".** That is a legitimate answer and it is cheap for
anyone who knows to do it. The thing worth weighing is what happens to someone who does not:
the in-memory path materialises the full 65-slice pyramid, which for the 16125×25690 label
image referenced in the review is a 25 GiB allocation — that is the failure that produced
#1231. If pre-conversion is the intended workflow, the win is probably mostly captured by
saying so where people hit it, in the docs and in that allocation's error path, and none of
that needs this PR.

**Whatever is decided, it probably wants deciding for both copies.** This function exists
twice: `vesuvius/src/vesuvius/ink_detection/preprocessing/create_label_zarrs.py` on `main`
(585 lines) and `ink-detection/koine_machines/preprocessing/create_label_zarrs.py` on
`merge-ink-pipelines` (840 lines), and they have diverged well past a rename. The
`merge-ink-pipelines` copy already streams untiled labels — that was #1234, mine, merged
2026-08-14 — and it took the *opposite* trade on the pyramid: build the 2D pyramid in memory
rather than reading levels back from zarr, at @erdpx's request during review, for 1.7x the
speed and 1.2x the peak memory. Closing this PR leaves `main` on the in-memory path that
#1231 reported, while the other copy streams. That divergence seems like the more useful
thing to settle than this diff.

**P2 #4 (multipage) is real and is not confined to `main`.** Credit where it is due — that
was @jaideepsaipadhi's finding, which I reproduced on 08-24 on the merged copy: a
`(5, 40, 60)` array comes out `(5, 40)` there too, via the same
`np.squeeze` / `image[..., 0]` normalisation. As far as I know it is still unfiled against
either copy.

Happy to re-run any of the above, or to hand over the patches and the 55-variant matrix if
whoever picks this up wants them. Both patches are small and neither is mine to submit here.

**Comment by jaideepsaipadhi** [comment]
Thanks for looking at this, and the review is right on all four counts. Two of them are regressions I introduced and I want to own them plainly before anything else.

**P2 #3 is the worst of them and it's mine.** I gated the streaming path on a codec whitelist instead of `is_tiled`, which means tiled JPEG and ZSTD now fall through to the in-memory path when they streamed before. That can restore the original OOM on inputs that work today. The whitelist should have applied to striped inputs only — tiled behaviour should have been left exactly as it was.

**P2 #4 is the same mistake in the same shape.** The multipage guard catches tiled multipage too, where previously it streamed the first page. Same correction: guard striped multipage only, leave the tiled path untouched.

Both are one-line scope fixes rather than redesigns, but they are real and they should not have been in the diff.

**P1 #1 and #2** are also real, and @khj1222 has already found, fixed and verified both against a 55-variant matrix — the one-row-strip crash with an explicit-indexing fix (18 comparable outputs identical to the parent tree at every level), and the re-decode problem with a full-width-write-blocks patch that takes the striped path to 104.4 s / 0.881 GiB, faster than this same branch reading a tiled copy of the same image at the same memory. That work is in `runs/pr1471_striped_check/` and the patches are small.

One caveat they raised that cuts against me and should not get lost: they measured `rowsperstrip=1024`, not the uncompressed `tifffile.imwrite` default of a single whole-image strip. That case is strictly worse than anything in their table and neither of us has measured it.

---

**Before I do that work, I'd rather ask than assume.** Your objection isn't really the four items — it's whether this belongs in the codebase at all versus converting to tiled beforehand. Fixing the regressions doesn't answer that, and I'd be doing a day's work on spec.

So: if I put up a corrected version — whitelist and multipage guard scoped to striped only, the one-row fix, full-width blocks, and a one-row-strip test fixture — is that something you'd merge? Or is pre-conversion the intended workflow?

If it's the latter, I think that's a defensible answer and the useful change is somewhere else entirely: say so in the docs, and make the in-memory path's error message name pre-conversion when it's about to attempt a 25 GiB allocation. That's what produced #1231, and someone who doesn't already know to convert gets no hint from the current failure. I'd be glad to write that instead — it's smaller and it doesn't add a code path to maintain.

Either way, the thing that seems worth settling independently of this diff is that `create_label_zarrs` exists twice and the two copies have taken opposite approaches — `merge-ink-pipelines` streams untiled labels since #1234, `main` doesn't. Closing this leaves `main` on the path #1231 reported. That's a decision for someone who knows where `merge-ink-pipelines` is going, but it outlives this PR.

@khj1222 — thank you, seriously. You ran the matrix, found a bug in my branch, wrote the fix, measured the pyramid question I'd asked about and couldn't answer, and then defended the PR to a maintainer with no stake in it. I'll file the multipage issue against both paths this week.

**Comment by hendrikschilling** [comment]
Sorry for the delay. Yes If you put up a corrected version I will merge it.
Alternatively putting in place error messages and docs fixes regarding tiled tiff use that would also be a great options. My comments mainly aimed at removing complexity where possible - that the first shot did have problems with a set of edge cases showed that complexity was maybe a bit too high.

**Comment by jaideepsaipadhi** [comment]
Pushed the corrected version. The diff is now **smaller** than the first one, which I think is the right direction given your point about complexity — two of the three changes remove behaviour the first version added.

**P2 #3 and #4 — tiled input is untouched again.** The codec whitelist and the multi-page guard now apply only to striped input. Tiled files take the streaming path unconditionally, exactly as they did before this PR. That was the actual regression: tiled multi-page and tiled zstd/JPEG were being pushed back to the in-memory path they had always been exempt from, which is the allocation #1231 is about.

I verified the prior behaviour rather than assuming it — ran a tiled multi-page file through the parent commit, confirmed `streamed_tiled_tiff == "true"` and that the output equals page 0, and the new test asserts exactly that.

**P1 #2 — one-row strips.** Using @khj1222's `_decoded_block_to_2d`, which indexes the decoded axes instead of squeezing them. Reproduced the crash first on tifffile 2026.8.16: a 2049×2051 image at `rowsperstrip=1024` decodes its final strip as `(1, 1, 2051, 1)`, and `rowsperstrip=1` gives that shape on every strip.

**P1 #1 — re-decoding: not in this PR, deliberately.** The fix is a conditional in the block iterator to make write blocks full-width when the source has one chunk column. It's a real speedup (@khj1222 measured 195.8s → 104.4s at the same memory, which makes striped faster than this same branch reading a tiled copy) but it is a performance change, not a correctness one, and it adds exactly the kind of branching you asked me to reduce. Happy to open it as a follow-up with those numbers attached, or to fold it in here if you'd rather have it in one go.

**Tests:** three added, and I checked each one fails without its fix — reverting the source gives exactly three failures and no others. 17 pass with the fix.

---

Two things I ran into that aren't part of this change, in case they're useful:

- `tests/ink_detection/test_label_zarr_preprocessing.py` can't be collected in a clean project environment. `create_label_zarrs` imports `cv2` and `tqdm` at module scope, and pulls in `nest_asyncio`, `nrrd` and `requests` transitively via `vesuvius.utils.cli` → `vesuvius.utils.catalog`; none are in the declared dependencies. I needed six `--with` flags to get pytest to import it.
- The multi-page normalisation bug is still unfiled against either copy. It's mine to file and I'll do it this week.

@khj1222 — the one-row fix is yours and the commit says so. Thank you for the matrix and for the measurements; the pyramid numbers answered a question I'd asked and couldn't answer myself.

**Comment by hendrikschilling** [comment]
Close! One optimization I suggest: Batch decode per output chunks - e.g. use some (overridable) cache size of for example 1GiB and per decoding pass create as many output chunks in write order as fit that memory budget. That way files that fit into that budget will only do a single decode and larger files reduce decoding overhead by orders of magnitude compared to the proposed solution. I can merge this without this if you insist but would prefer this optimization.

```
  1. [P1] Default striped TIFFs are still decoded repeatedly.
     In vesuvius/src/vesuvius/ink_detection/preprocessing/create_label_zarrs.py:350, iteration is destination-block-major. Each intersecting TIFF strip is decoded again for every 1024×1024
     destination block.

     A default uncompressed tifffile.imwrite produces one whole-image strip. I confirmed:
      - 2049×3073: one (2049, 3073) strip
      - 16125×25690: one (16125, 25690) strip

     The latter is decoded once for each of 416 destination blocks, representing roughly 172 GB of repeated decoded image traffic for a 414 MB uint8 source. The PR avoids the 25 GiB       
     pyramid allocation, but its primary advertised input can become pathologically slow.

     The robust correction is source-strip-major processing: decode each strip once, then write its intersections to Zarr blocks. Merely making blocks full-width removes horizontal        
     repetition but still repeatedly decodes strips taller than 1024 rows, including the default whole-image strip.
```

jaideepsaipadhi:
Done. The loop is now source-chunk-major: decode each chunk once, scatter it into every destination block it touches, and flush a block as soon as the last chunk that can contribute to it has been consumed.

I ended up not needing the memory budget to do much work. Because the pending-contribution count per destination block is computed up front from the chunk grid, blocks are released the moment they're complete, so residency stays naturally bounded and the budget is a backstop rather than the mechanism. It's there as `VESUVIUS_STREAM_DECODE_BUDGET_BYTES`, default 1 GiB, overridable per run. A single source chunk larger than the budget is still decoded rather than refused — refusing would push it to the in-memory path this PR exists to avoid — but say the word if you'd rather it error.

**Decodes for N source chunks, before → after:**

| input | chunks | before | after |
|---|---|---|---|
| 2049×3073, whole-image strip (default `imwrite`) | 1 | **12** | **1** |
| 1200×2600 uint16, whole-image strip | 1 | 6 | 1 |
| 37×4099, `rowsperstrip=8` | 5 | 25 | 5 |
| 2049×2051, `rowsperstrip=1024` | 3 | 9 | 3 |
| 3000×2500, `rowsperstrip=512` | 6 | 18 | 6 |
| 1100×1500, `rowsperstrip=7` | 158 | 318 | 158 |
| 1500×1700 LZW / 1200×1400 PackBits | 6 / 4 | 12 / 10 | 6 / 4 |
| tiled 256×256, all sizes | — | 1:1 | 1:1 |

Your 12 for the 2049×3073 case reproduced exactly, which was a useful check that we were measuring the same thing. Tiled input was already 1:1, so the waste was entirely on the striped path this PR adds.

**Verification.** 14 variants, each asserting the streamed level-0 output is byte-identical **to the source image** rather than to the previous implementation — spanning tiled and striped, even/odd/ragged/degenerate extents, uint8 and uint16, and none/LZW/deflate/PackBits. All 14 pass with decodes exactly equal to chunk count.

Added `test_streaming_decodes_each_source_chunk_exactly_once` to the test file, parametrised over whole-image-strip, `rowsperstrip=1024`, and tiled. Worth having as a test rather than just a measurement: the old code produced byte-exact output while doing all the redundant work, so a correctness test cannot see this regress. The tiled case is in there deliberately — it passed before and after, so the test isn't merely asserting "the new code ran."

20 tests pass on the branch.

hendrikschilling:
- [P2] The configured 1 GiB memory budget is not actually enforced. At vesuvius/src/vesuvius/ink_detection/preprocessing/create_label_zarrs.py:447-459, the over-budget loop skips every  
    block with pending contributions. Completed blocks were already flushed at lines 442-445, so this loop normally cannot release anything. The comment about re-reading partially written 
    blocks is not implemented. Either implement spill/reload or remove the setting and its memory-bound claim.

hendrikschilling:
- [P2] The configured 1 GiB memory budget is not actually enforced. At vesuvius/src/vesuvius/ink_detection/preprocessing/create_label_zarrs.py:447-459, the over-budget loop skips every  
    block with pending contributions. Completed blocks were already flushed at lines 442-445, so this loop normally cannot release anything. The comment about re-reading partially written 
    blocks is not implemented. Either implement spill/reload or remove the setting and its memory-bound claim.

jaideepsaipadhi:
You're right, and the second half is worse than unenforced.

The loop can't release anything, for exactly the reason you give: completed blocks are flushed at the pending-count check, so every block still resident has `pending > 0` and hits the `continue`.

But the comment about partially-filled blocks being "re-read on their next contribution" describes behaviour that was never implemented. `_flush` writes the block and pops it; a later contribution to that key recreates it via `np.zeros` and silently discards what had already been written. So if that path had ever fired it would have corrupted output, not just failed to save memory. It's unreachable in practice, which is why the 14-variant harness passed — but that makes it a latent correctness bug rather than dead code, and I'd rather it be gone than fixed.

Removed the setting, the accounting and the loop. Residency is already bounded by the pending-count mechanism itself: a destination block is held only while source chunks that can contribute to it remain unread, and chunks are visited in row-major order, so at most one row of destination blocks is resident at a time. Spill/reload would be real complexity for a case that doesn't arise.

20 tests and the 14-variant harness are unchanged after the removal, which is the expected result for deleting an unreachable branch — and a check that it really was unreachable.

PR #1473, opened 2026-08-16
Preserve tifxyz provenance metadata through flatten and trim (fixes #1321)

### LINKED ISSUE #1321: vc_flatten drops all provenance metadata; vc_tifxyz_trim drops bbox
Measured against a real traced surface (86×86, 6712 valid points, 16 metadata keys):

| tool | keys kept | `bbox` | provenance | verdict |
|---|---|---|---|---|
| `vc_transform_geom` | 16 | recomputed | complete | ✅ |
| `vc_tifxyz -r 90` | 17 | recomputed | complete | ✅ |
| `vc_tifxyz_trim` | 15 | **dropped** | complete | ⚠ |
| `vc_flatten` | **5** | kept | **none** | ⚠ |

Two of the four preserve everything and regenerate the `uuid`, so the convention exists and
is implemented — these two are the exceptions.

**`vc_flatten`** discards `area_cm2`, `area_vx2`, `scroll_source`, `target_volume`, `seed`,
`source`, `vc_gsfs_params`, `vc_gsfs_mode`, `vc_gsfs_version`, `max_gen`,
`elapsed_time_s`. After flattening, a surface can no longer say **which scroll it came
from**, which volume, or what produced it. The Grand Prize rules ask for a *"fully
reproducible"* pipeline and meshes identifiable per column; a flattened mesh cannot
document its own origin.

**`vc_tifxyz_trim`** resizes the grid (86×86 → 82×82) and **deletes** `bbox` rather than
recomputing it. That has a traceable consequence in `vc_grow_seg_from_seed.cpp`, in
**three separate places** — `check_existing_segments` at line 141, and the seed-search
loops at lines 419 and 1498:

```cpp
if (!meta.count("bbox") || meta.value("format", std::string{"NONE"}) != "tifxyz") {
    continue;
}
```

All three `continue` silently. A trimmed segment is therefore invisible to the overlap
check *and* to both seed-placement scans, so the tracer could grow duplicate segments over
ground already covered. *(I have read all three paths but not reproduced the overlap case
end to end — worth confirming before treating it as more than a strong suspicion.)*

**Suggested fix:** both should copy the input `meta.json` and overwrite only what actually
changes, as `vc_transform_geom` already does.

**Minor, related:** three different names for the same quantity across tools — `area_cm2`
and `area_vx2` (tracer) vs `area` (`vc_tifxyz`, in vx²). Harmless, but it makes generic
readers awkward.

Measured on Windows (MSYS2 UCRT64). Happy to send a PR for either.

jaideepsaipadhi:
Ran into #1321 while looking through recent issues and ended up building
and verifying a fix for both halves. Apologies if you'd already started
on this yourself — happy to adjust, split, or step back if you'd rather
take a different approach.

### DIFF
--- a/volume-cartographer/apps/src/vc_tifxyz_trim.cpp
+++ b/volume-cartographer/apps/src/vc_tifxyz_trim.cpp
@@ -12,6 +12,7 @@
 #include <filesystem>
 #include <fstream>
 #include <iostream>
+#include <limits>
 #include <string>
 #include "vc/core/util/MemMap.hpp"
 #include <vector>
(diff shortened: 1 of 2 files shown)


jaideepsaipadhi:
Closing in favor of #1387, which fixes the same root cause more completely —
it also covers vc_straighten.cpp and vc_project_tifxyz.cpp, and demonstrates
a real consumer of the missing bbox (vc_seg_add_overlap silently skipping
segments without one). Should have caught the linked PR on #1321 before
starting; apologies for the noise.

PR #1475, opened 2026-08-16
load_checkpoint: report true success/failure, not hardcoded True

jaideepsaipadhi:
load_checkpoint() returns (model, optimizer, scheduler, start_epoch, success). When a strict load_state_dict fails (e.g. architecture drift since the checkpoint was saved) and a config-based rebuild cannot match either, the code falls back to a partial, non-strict load of only the tensors whose name and shape both matched -- everything else silently keeps its random initialization.

The internal variable strict_loaded already tracks this correctly: True on a genuine exact match, False right before the partial fallback. But the function's only other return statement hardcoded True unconditionally, on every path that did not raise, including the partial one -- strict_loaded was computed and never used for the return value at all.

Two real callers gate actual behavior on this flag: models/training/train.py restores EMA model state and applies loss overrides only when the flag is true. A user resuming training with any architecture drift would see a print statement easy to miss in a long log, get a true success flag anyway, and have training proceed as if fully resumed while part of the model is silently back at random init.

Fix: return strict_loaded instead of a hardcoded True.

Verified: static trace confirming no conditional between the partial fallback and the return; empirical reproduction against the real, unmodified function using a toy two-layer model with one shape-drifted layer, confirming the flag was true while that layer stayed at random init; empirical confirmation of the fix, same reproduction, patched function correctly returns false for the drifted case and true for a clean match with no regression. Two new tests, verified on three independent fresh clones plus this pod's own run.

### DIFF
--- a/vesuvius/src/vesuvius/models/utilities/load_checkpoint.py
+++ b/vesuvius/src/vesuvius/models/utilities/load_checkpoint.py
@@ -316,4 +316,4 @@ def load_checkpoint(checkpoint_path, model, optimizer, scheduler, mgr, device, l
         )
         print("Loaded model weights only; starting new training run from epoch 1.")
     
-    return model, optimizer, scheduler, start_epoch, True
+    return model, optimizer, scheduler, start_epoch, strict_loaded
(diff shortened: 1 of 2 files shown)


jaideepsaipadhi:
Closing to reduce my open PR count — I opened too many at once, which wasn't reasonable to review.

The bug is real (the function returns a hardcoded True while discarding the strict_loaded value it already computed, and train.py gates EMA restore on that flag), but it only triggers on architecture drift during a resume, so it's narrower than the others I have open. Happy to reopen.

PR #1476, opened 2026-08-16
voxelize_objs: fix spurious extra line from near-degenerate vertices

jaideepsaipadhi:
process_slice_points_label rasterizes each mesh triangle's intersection with a z-slicing plane. A plane can cross a triangle's boundary in at most 2 points for any non-degenerate case, but when a vertex sits very close to (not exactly on) the slicing plane -- realistic floating-point noise for scan-derived mesh vertices, not a contrived case -- the two edges sharing that vertex each independently compute their own parametric intersection point near it. These two points can differ by more than the tight 1e-12 squared-distance dedup threshold while both still being spurious near-duplicates of the same vertex, producing 3 points instead of 2.

The previous code's fallback for more than 2 points, connect every pair, then drew a spurious extra short line segment between the two near-duplicates in addition to the two correct lines.

Verified: 500/500 random near-degenerate triangles (vertex offset by 1e-6 to 1e-3 from the plane) reproduce the 3-point pattern. Across a wider, more realistic coordinate range, 1614/3000 (54%) produce a genuinely different rasterized pixel output compared to the same triangle with that vertex placed exactly on the plane -- differences up to 53 pixels. One specific hand-picked triangle tested first showed no visible difference, since the spurious segment happened to land on pixels the two correct segments already covered; the broader randomized search is what established the real prevalence.

Fix: when more than 2 intersection points are found, select the pair with maximum squared distance rather than connecting every pair. The two near-duplicate points are always close to each other (both approximate the same near-degenerate vertex), while the genuine second crossing is comparatively far away, so this recovers the correct single line in the overwhelming majority of cases and is a no-op when exactly 2 points are found (the normal case).

process_slice_points_label_surface in this same file has the identical root-cause bug (same thresholds, same fallback), not touched here: it is jit-compiled and each intersection point also carries barycentric coordinates and surface normal/tangent data that must correctly travel with whichever two points get selected, real added complexity beyond a mechanical copy of this fix. Worth a dedicated follow-up.

7 new tests (this file had 0% test coverage, confirmed via real coverage measurement). Mutation-tested: reverting the fix reproduces a real, concrete failure, 34 vs 33 pixels at eps=5.2e-5, in the randomized statistical test. Verified on three independent fresh clones of main.

### DIFF
--- a/vesuvius/src/vesuvius/image_proc/run/voxelize_objs.py
+++ b/vesuvius/src/vesuvius/image_proc/run/voxelize_objs.py
@@ -520,15 +520,50 @@ def process_slice_points_label(vertices, triangles, mesh_labels, zslice, w, h):
                 if not is_dup:
                     pts_2d.append(p)
 
-        # If we have at least two unique intersection points, draw lines
+        # If we have at least two unique intersection points, draw a line.
+        # A plane can intersect a triangle's boundary in at most 2 points for
+        # any non-degenerate case; more than 2 unique points showing up here
+        # is the signature of a near-degenerate vertex (one very close to,
+        # but not exactly on, zslice) rather than a genuine extra crossing.
+        # The two vertex-snap checks in get_intersection_point_2d only
+        # trigger within 1e-8 of the plane; just outside that, the two edges
+        # sharing that vertex each compute their own independent parametric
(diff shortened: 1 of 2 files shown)


pmh47:
See contributor guidelines - lacking real-world use-case

PR #1477, opened 2026-08-16
prediction_grid: guarantee fair model comparison across coverage gaps

jaideepsaipadhi:
_choose_best_model/_rank_models pick which trained model is 'best' -- that is the whole purpose of this tool. _collect_aggregated_model_stats computed each model's mean per metric over however many images that specific model happened to have a stats.json for, with no requirement that different models were compared on the same images.

That's a realistic, not contrived, failure mode: a crashed or partial eval run, or a model evaluated before harder images were later added to the test set, both leave a model with stats.json only for a partial, easier subset. Its mean then looks better than a model honestly evaluated on the full set -- purely from selection bias in which images have stats, not genuine model quality -- with nothing in the tool's output revealing the comparison wasn't fair.

Verified: constructed a 20-image scenario where model_A was evaluated on only the easier half (all great scores) and model_B was evaluated on all 20, matching model_A exactly on the easy half and scoring reasonably, not badly, on the harder half it actually attempted. Under the old per-model-coverage aggregation, model_A -- the LESS evaluated model -- was chosen as best.

Fix: for each metric independently, restrict the mean to the images where every compared model has a valid value for that metric. When coverage genuinely differs, the tool now prints an explicit warning naming which models had excluded extra coverage and why, and num_images_considered, previously just the total candidate pool size, now reflects the actual number of images the comparison was based on.

No behaviour change when all models have equal coverage, confirmed with a dedicated no-regression test. If a genuinely better model also has more coverage, it still wins -- the fix restricts the comparison to fair ground, it does not penalize full evaluation, confirmed.

4 new tests (this file had 0% test coverage, confirmed via real coverage measurement). Mutation-tested: reverting the fix reproduces 2 concrete failures, 20 vs 10 images considered in the main reproduction, 4 vs 0 in a zero-overlap edge case. Verified on three independent fresh clones of main, and confirmed this fix coexists cleanly with the separate voxelize_objs fix in the same directory.

### DIFF
--- a/vesuvius/src/vesuvius/image_proc/run/prediction_grid.py
+++ b/vesuvius/src/vesuvius/image_proc/run/prediction_grid.py
@@ -760,6 +760,22 @@ def _collect_aggregated_model_stats(eval_root: Path, images_dir: Path,
 
     Returns (models_stats_list, num_images_considered), where models_stats_list is a list of (model_name, stats_means).
     The stats_means dict contains means of the five keys used in ranking when available.
+
+    Each model's mean for a given metric is computed over the SAME set of
+    images as every other model being compared, per metric -- not over
+    however many images that individual model happens to have a stats.json
+    for. Comparing means computed over different, non-identical image
+    subsets is an apples-to-oranges comparison: a model evaluated only on
+    an easier partial subset (e.g. from a crashed or partial eval run, or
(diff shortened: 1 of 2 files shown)


pmh47:
See contributor guidelines - lacking real-world use-case

PR #1478, opened 2026-08-16
hierarchical_tiling: fix binary search on a non-monotonic space

jaideepsaipadhi:
find_valid_child_size uses _binary_search_min_size to find the smallest child tile size that produces enough fully-valid children when subdividing a parent tile. This was implemented as an actual binary search, which requires the search space to be monotonic: once test_fn(size) is True, it must stay True for every larger size.

That assumption does not hold here. Whether a given tile size produces enough fully-valid children depends on exact grid alignment against a segment's holes and torn edges -- create_child_tiles requires a child's entire rectangular footprint to be valid, not just some of it -- and which grid alignment happens to avoid those holes shifts non-trivially, not monotonically, as tile size changes. Real scroll segments have irregular, incomplete boundaries, not simple rectangles.

Verified: constructed a realistic irregular mask (a 200x200 region with ~40 scattered small holes, modeling a segment's torn/incomplete edges) and measured the true minimum valid tile size directly by brute-force scan: 15. The previous binary-search implementation's own short-circuit check on the largest size failed, since coverage had already dropped to zero at that size, so it returned 0 -- no valid size exists -- despite size 15, and others, genuinely working. A caller receiving (0, 0) from find_valid_child_size would treat a real, tileable parent region as untileable.

Also fixed a smaller, related gap found while designing the fix: the lo >= hi trivial-range branch returned hi unconditionally, without ever calling test_fn(hi) -- silently assuming a candidate was valid rather than checking it.

Fix: replaced the binary search with a full linear scan from lo to hi, which cannot miss a valid smaller size regardless of monotonicity. Function name kept for compatibility with existing callers; docstring corrected to match the function's actual, and always correctly relied upon, return contract -- 0 signals failure, not hi as the stale prior docstring claimed.

5 new tests (this file had 0% test coverage; no tests/tifxyz/ directory existed at all before this, confirmed via real coverage measurement): a no-regression monotonic case, a genuinely-impossible case, the trivial-range test_fn-must-be-called case, an abstract small-valid-window case, and the full realistic-irregular-mask reproduction. Mutation-tested: reverting the fix reproduces 3 concrete failures across those cases. Verified on three independent fresh clones of main.

### DIFF
--- a/vesuvius/src/vesuvius/tifxyz/hierarchical_tiling.py
+++ b/vesuvius/src/vesuvius/tifxyz/hierarchical_tiling.py
@@ -603,29 +603,36 @@ def _test_child_size_validity(
 
 def _binary_search_min_size(lo: int, hi: int, test_fn) -> int:
     """
-    Binary search to find minimum size where test_fn returns True.
-    Returns hi if no valid size found.
+    Find the minimum size in [lo, hi] where test_fn returns True.
+    Returns 0 if no valid size is found.
+
+    A full linear scan, not a binary search despite the name (kept for
+    compatibility with existing callers): the property being searched for
+    here (fraction of a parent tile's candidate children that are valid,
+    as a function of child tile size) is NOT guaranteed monotonic in size.
+    A larger tile is not always at least as likely to avoid a segment's
+    holes/torn edges as a smaller one -- survival also depends on exact
+    grid alignment against the mask's specific hole positions, which
+    shifts non-trivially as tile size changes. Binary search's core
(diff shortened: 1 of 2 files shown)


jaideepsaipadhi:
Closing to reduce my open PR count — I opened too many at once, which wasn't reasonable to review.

The bug is real and the counterexample is measured, but the failure mode is suboptimal tiling rather than incorrect output, so it's the least urgent of what I have open. Happy to reopen.

PR #1479, opened 2026-08-17
vesuvius: gate cucim-cu13 behind sys_platform == 'linux'

flummoxjr:
## Motivation

I'm setting up to train surface models on Dataset059 (the public surface-label
patch set referenced in wishlist issue #191) on my Windows machine (RTX 4090
laptop). Following the install in `docs/guided_dinovol.md` /
`docs/ink_detection.md`, `uv sync --extra models --extra tests` fails at
dependency resolution before anything installs, because `cucim-cu13` has
Linux-only wheels. Same failure for `pip install "vesuvius[models]"` from PyPI,
and it hits macOS users identically since cuCIM publishes manylinux wheels only.

## What this changes

Adds `; sys_platform == 'linux'` to the two `cucim-cu13` declarations
(`models` extra, `all` extra) in `vesuvius/pyproject.toml`. Nothing else.

## Why this is safe

Both cuCIM import sites already treat it as optional at runtime:

- `src/vesuvius/ink_detection/training/dilation.py:51-56` — lazy import inside
  the function; raises a clear RuntimeError only if the cuCIM-requiring path is
  actually invoked.
- `src/vesuvius/scripts/detect_holes.py:21-28` — try/except at import with a
  CPU (skimage) fallback via `_HAS_CUPY`.

No module imports cuCIM unconditionally, so excluding it from non-Linux
resolution changes no behavior except making installation possible.

## Before (current main, Windows 11)

```
$ uv sync --extra models --extra tests --dry-run
Resolved 298 packages in 2.05s
error: Distribution `cucim-cu13==26.6.0 @ registry+https://pypi.nvidia.com/`
can't be installed because it doesn't have a source distribution or wheel for
the current platform
hint: You're on Windows (`win_amd64`), but `cucim-cu13` (v26.6.0) only has
wheels for the following platforms: `manylinux_2_26_aarch64`,
`manylinux_2_27_x86_64`, `manylinux_2_28_aarch64`, `manylinux_2_28_x86_64`
```

PyPI `pip install "vesuvius[models]"` fails the same way — the published 0.2.4
metadata carries the unconditional pin too (`cucim-cu13>=25.12; extra == "models"`).

## After (this branch, same machine)

```
$ uv sync --extra models --extra tests --dry-run
Resolved 298 packages in 2.05s
Would install 250 packages
```

Linux unchanged: regenerating the universal lock still resolves
`cucim-cu13 26.6.0` from pypi.nvidia.com, now guarded by the marker.

Note: `uv lock --check` already fails on current main before this change (the
committed lock is stale relative to pyproject) — happy to include a lock
regeneration here or leave it for a separate housekeeping commit, whichever
you prefer.

---


flummoxjr:

For context, this one gates `cucim-cu13` behind `sys_platform == 'linux'` so `vesuvius[models]` and `uv sync --all-extras` install on Windows and macOS.

PR #1492, opened 2026-08-17
inference: make vesuvius.predict work with train_py checkpoints (3 fixes)

flummoxjr:
## Motivation

I trained surface models on Dataset059 (wishlist #191) with `vesuvius.train`
and then tried to run them through the official inference path
(`vesuvius.predict --model_type train_py`) on val volumes — the natural next
step toward using them on scroll data. Every attempt failed, three distinct
ways.

## The three defects

**1. Unguarded lazy `torch.compile` (same class as #1480, inference edition).**
`_load_train_py_model` wraps `torch.compile` in try/except, but compilation is
lazy — with no triton (Windows CUDA wheels) the failure surfaces as
`TritonMissing` at the *first forward*, where no guard exists. Every
Windows+CUDA `vesuvius.predict` run dies. Fix: check
`torch.utils._triton.has_triton()` before compiling, fall back to eager with a
message.

**2. `out_channels: None` crashes both channel-counting sites.**
Checkpoints from configs with auxiliary targets carry an explicit
`out_channels: None` (resolved at build time), and both
`target_config.get('out_channels', 1)` sites do `int + None`:

```
File ".../models/run/inference.py", line 737, in _load_train_py_
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
```

Rather than guessing `1` (which would silently mis-slice targets like
`surface_normals` that really output 3 channels), the fix defers to the
**already-existing dummy-inference path**, which measures true per-target
channels from the model output. Both sites now fall back to it whenever any
target's channels are unresolved.

**3. `infer()` swallows errors and returns `None`.**
The catch-all printed the error and implicitly returned `None`; `main()` then
crashed unpacking it — so every real failure presented as
`Error: cannot unpack non-iterable NoneType object`, which is what made bugs
1 and 2 initially indistinguishable. Fix: re-raise after logging.

## Before / after (Windows 11, RTX 4090, torch 2.11+cu128, no triton)

Before — single-target train_py checkpoint:
```
torch._inductor.exc.TritonMissing: Cannot find a working triton installation.
--- Inference Failed ---
Error: cannot unpack non-iterable NoneType object
```

Before — multi-target (aux) checkpoint:
```
TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'
--- Inference Failed ---
Error: cannot unpack non-iterable NoneType object
```

After — both checkpoints:
```
exit=0, logits_part_0.zarr + coordinates_part_0.zarr written
(64 patches, eager mode, "Targets with unresolved out_channels
['distance_transform', 'surface_normals']; determining channels via dummy
inference" on the multi-target run)
```

---

### DIFF
--- a/vesuvius/src/vesuvius/models/run/inference.py
+++ b/vesuvius/src/vesuvius/models/run/inference.py
@@ -556,22 +556,35 @@ class Inferer():
         self.model_patch_size = tuple(model_info.get('patch_size', (192, 192, 192)))
         self.num_classes = model_info.get('num_seg_heads', None)
         
-        # Check if this is a multi-task model from targets info
+        # Check if this is a multi-task model from targets info.
+        # Checkpoints can carry out_channels: None (e.g. auxiliary targets whose
+        # channel count is resolved at build time); in that case leave
+        # num_classes as None so the dummy-inference path below measures the
+        # real per-target channels from the model output.
         if 'targets' in model_info and model_info['targets']:
-            self.is_multi_task = True
-            self.target_info = {}
-            self.num_classes = 0
-            for target_name, target_config in model_info['targets'].items():
-                target_channels = target_config.get('out_channels', 1)
-                self.target_info[target_name] = {
-                    'out_channels': target_channels,
-                    'start_channel': self.num_classes,
-                    'end_channel': self.num_classes + target_channels
-                }
-                self.num_classes += target_channels
-            if self.verbose:
-                print(f"Detected multi-task model with targets: {list(model_info['targets'].keys())}")
-                print(f"Total output channels: {self.num_classes}")
+            targets_cfg = model_info['targets']
+            if all(cfg.get('out_channels') is not None for cfg in targets_cfg.values()):
+                self.is_multi_task = True
+                self.target_info = {}
+                self.num_classes = 0
+                for target_name, target_config in targets_cfg.items():
+                    target_channels = target_config.get('out_channels')
+                    self.target_info[target_name] = {
+                        'out_channels': target_channels,
+                        'start_channel': self.num_classes,
+                        'end_channel': self.num_classes + target_channels
+                    }
+                    self.num_classes += target_channels
+                if self.verbose:
+                    print(f"Detected multi-task model with targets: {list(targets_cfg.keys())}")
+                    print(f"Total output channels: {self.num_classes}")
+            else:
+                self.num_classes = None
+                if self.verbose:
+                    unresolved = [t for t, c in targets_cfg.items()
+                                  if c.get('out_channels') is None]
+                    print(f"Targets with unresolved out_channels {unresolved}; "
+                          "determining channels via dummy inference")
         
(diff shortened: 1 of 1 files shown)


flummoxjr:

For context, this one makes `vesuvius.predict` work with `train_py` checkpoints (three fixes), verified end-to-end on single- and multi-target checkpoints.

flummoxjr:
Two bits of context that narrow what defect 1 here is asking for — one that landed after I opened this, one I should have cited up front.

**The train half of this exact fix is now in `main`.** #1480 ("train: fall back to eager mode when triton is unavailable") merged 2026-08-18 as `bc5e458c`. Defect 1 in this PR is the same `has_triton()` gate applied to the inference path — same root cause, same shape, same one-line eager fallback — so the pattern is already blessed in `train.py` and this is the half that is still missing.

**The inference half was explicitly deferred by a merged PR three weeks ago.** #1239 ("inference: make `--device cpu` work: autocast is hardcoded to CUDA", opened 2026-07-25, merged 2026-07-28) says, under "Note on scope":

> On a machine without a C++ compiler the run can still stop later inside `torch.compile` (`InvalidCxxCompiler: Compiler: cl is not found`), which is a separate concern and not addressed here — `TORCHDYNAMO_DISABLE=1` sidesteps it.

Same guard, same lazy-failure mechanism, different missing backend dependency — a C++ compiler there, triton here. So the predict-side compile failure has been known since #1239 was opened on 2026-07-25 and is still unfixed, and defect 1 of this PR is the follow-up #1239 asked for. On Windows the two overlap in practice — no triton and, usually, no MSVC — so the `has_triton()` gate takes that machine off the compile path as well. It does not cover the general no-C++-compiler case: `has_triton()` checks only for the triton package and a compatible device (`torch/utils/_triton.py`), never for a compiler, and the compile call sits inside `if self.device.type == 'cuda':`, so a `--device cpu` run like #1239's never reaches it at all.

Worth restating why the existing try/except cannot catch it: `torch.compile(model)` is lazy — the call only wraps the module in an `OptimizedModule` and returns. Backend selection, and therefore the triton check, happens at the **first forward** of the wrapped module. `inference.py` guards only the `torch.compile(...)` call, so `TritonMissing` fires later, at the first non-empty batch, long after control has left the guarded block. On the one platform where the fallback is most needed — Windows + CUDA, where official wheels never ship triton — the fallback branch is exactly the code path that can never run.

One thing this PR deliberately does **not** add, which I'm happy to send separately if you want it: there is still no CLI opt-out for compilation on the inference side. Training has `compile_policy` (`config_manager.py:180-186`); `vesuvius.predict`'s `build_parser()` has no compile-related argument at all, so the only way to turn it off today is a global env var — `TORCH_COMPILE_DISABLE=1`, or `TORCHDYNAMO_DISABLE=1` as #1239 notes; I checked and both work on torch 2.11 — which is undiscoverable from the CLI. A `--no-compile` flag would be a few lines. Say the word and I'll open it as a follow-up rather than growing this PR.

PR #1493, opened 2026-08-17
Fix SmearTransform smearing past its window and into the caller's tensor

axiosdevs:
`SmearTransform` sits in every 3D training pipeline (`training_transforms.py:394`, `num_prev_slices=3`, p=0.3), and `num_prev_slices` currently does not bound anything.

`torch.permute` returns a view, so `moved` **is** `img`: the `img.clone()` on line 241 is dead, the writes land in the caller's tensor, and — the part that matters — each iteration blends slice `i` against slices that were already smeared on earlier iterations, so the window becomes a filter that runs the full length of the axis.

One bright slice at index 0, `alpha=0.5`, `num_prev_slices=1`, 6 slices:

```
per-slice means, on main : [100.0, 50.0, 25.0, 12.5, 6.25, 3.12]
per-slice means, expected: [100.0, 50.0,  0.0,  0.0, 0.0,  0.0]
input tensor mutated     : True
```

Three lines: `.contiguous()` / `.clone()` so the working tensor is private, and the neighbour reads come from an untouched copy so `num_prev_slices` means what it says. After the change the same case gives exactly the expected profile and the input is left alone.

This does change what the augmentation produces — a bounded smear instead of an axis-long decay. If you would rather keep the current appearance, the aliasing half alone is a strictly safe subset and I can trim to that.

### DIFF
--- a/vesuvius/src/vesuvius/models/augmentation/transforms/noise/extranoisetransforms.py
+++ b/vesuvius/src/vesuvius/models/augmentation/transforms/noise/extranoisetransforms.py
@@ -248,15 +248,22 @@ class SmearTransform(ImageOnlyTransform):
             dims = list(range(chan_img.ndim))
             if local_smear_axis != 0:
                 dims = [local_smear_axis] + [d for d in dims if d != local_smear_axis]
-                moved = chan_img.permute(*dims)
+                # permute returns a view: without a copy the writes below land in
+                # the caller's tensor, and each slice is smeared into the next, so
+                # the effect runs the length of the axis instead of num_prev_slices
+                moved = chan_img.permute(*dims).contiguous()
             else:
-                moved = chan_img
+                moved = chan_img.clone()
+            # Read the neighbours from an untouched copy: smearing slice i from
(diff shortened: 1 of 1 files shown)


axiosdevs:
CI caught exactly the trade-off flagged above: `test_default_augmentation_graph_and_seeded_values` pins the mean of the default graph's output, and bounding the smear moves it (-0.0209 vs the pinned -0.0236).

Trimmed to the aliasing half, which is bit-identical on the output — `moved` is now a private copy, so the writes no longer land in the caller's tensor, and the smear values are exactly what they were. The pinned test should pass unchanged.

The window question is left alone: `num_prev_slices=3` currently blends against slices that were already smeared, so the effect runs the length of the axis rather than three slices (`[100, 50, 25, 12.5, 6.25, 3.12]` where a bounded smear gives `[100, 50, 0, 0, 0, 0]`). Whether that is the intended appearance is your call — if you want it bounded, the change is two more lines plus a new pinned value, happy to send it as a separate PR.

PR #1495, opened 2026-08-17
Fix per-iteration LR schedulers being given an epoch count as their horizon

axiosdevs:
`_get_scheduler` passes `max_steps=self.mgr.max_epoch`, but `onecycle`, `cyclic`, `cosine_warmup` and `diffusers_cosine_warmup` are stepped once per optimizer step (`train.py:2299`), not once per epoch. The horizon is therefore short by however many steps an epoch takes — 250 with the shipped `--max-steps-per-epoch` default.

Replaying the exact `diffusers` lambda with `max_epoch=1000`, 250 iterations per epoch:

```
epoch    1 (iter    250): lr multiplier 0.9330
epoch    4 (iter   1000): lr multiplier 0.0000   <- "end of training" at epoch 4 of 1000
epoch    8 (iter   2000): lr multiplier 0.9698   <- back near peak
epoch 1000 (last iter)  : lr multiplier 0.7485   <- ends at 75% of peak, not ~0

decays to zero 140 times over the run; returns above 90% of peak 139 times
```

With the step horizon it is the single cosine it is meant to be: 0.0100 after the first epoch, 0.5868 at the halfway point, 0.0000 at the end.

`cosine_warmup` has the mirror problem — `CosineAnnealingWarmupRestarts` restarts every `first_cycle_steps` and multiplies `max_lr` by `gamma=0.9` each restart, so with an epoch-sized cycle it does 250 restarts per epoch's worth of steps and the peak collapses. The epoch-stepped `poly`/`warmup_poly`/`cosine` paths are unaffected and keep the old value.

The change multiplies the horizon by the real steps-per-epoch (`len(train_dataloader)` capped by `max_steps_per_epoch`, the same expression the training loop uses at line 2190) and only for the per-iteration list. That count is only known once the dataloaders exist, so the scheduler is now built just after them instead of beside the optimizer — nothing between the two points touched `scheduler`. The printed line now says whether the horizon is steps or epochs.


axiosdevs:
CI was right: `_configure_compile_test_trainer` stubs `_get_scheduler` with a one-argument lambda, and the extra parameter broke it.

Reverted the signature change — the trainer now stores `self._steps_per_epoch` once the dataloaders exist and `_get_scheduler` reads it via `getattr`, so the stub keeps working and the horizon is still a step count for the per-iteration schedulers. No other change.

PR #1500, opened 2026-08-17
Fix ConvBlock encoder ignoring pool_type and never downsampling

axiosdevs:
With `basic_encoder_block: ConvBlock` and a pooled `pool_type`, the encoder does not downsample at all.

The residual branch builds the stage and then wraps it (`encoder.py:123-124`):

```python
if pool_op is not None:
    stage = nn.Sequential(pool_op(strides[s]), stage)
```

The `ConvBlock` branch hands the stride to that pool — `conv_stride = strides[s] if pool_op is None else 1` — and then never adds it:

```python
stage_modules.append(StackedConvBlocks(..., conv_stride, ...))
stages.append(nn.Sequential(*stage_modules))
```

So the stride is zeroed out and nothing replaces it. Building the real `Encoder` with `strides=[1,2,2]` on a 16³ input:

```
BasicBlockD  pool_type=conv  -> skips [(16,16,16), (8,8,8), (4,4,4)]
BasicBlockD  pool_type=avg   -> skips [(16,16,16), (8,8,8), (4,4,4)]
ConvBlock    pool_type=conv  -> skips [(16,16,16), (8,8,8), (4,4,4)]
ConvBlock    pool_type=avg   -> skips [(16,16,16), (16,16,16), (16,16,16)]   <- no downsampling
```

Both values are reachable from a config: `pool_type` is read at `build_network_from_config.py:675` and `ConvBlock` is an accepted `basic_encoder_block` (`encoder.py:80`). Every stage then runs at full resolution and the decoder's skip concat gets mismatched sizes.

Three lines, mirroring the residual branch. After the change the last row reads `[(16,16,16), (8,8,8), (4,4,4)]` like the rest.


bruniss:
nice catch -- i've never actually used any pooling other than conv so never have noticed this. only one comment : 

The added pooling is correct, but you dont need the extra nn.Sequential (pool → Sequential(StackedConvBlocks)). you can  add the pool directly to stage_modules before StackedConvBlocks  (like the plain conv encoder shows) 

happy to merge if you make this change

axiosdevs:
Done — the pool now goes into `stage_modules` ahead of `StackedConvBlocks`, so the stage stays a single `nn.Sequential` like the plain conv path.

Re-checked the shapes after the change, `strides=[1,2,2]` on a 16³ input:

```
BasicBlockD  pool_type=conv  -> [(16,16,16), (8,8,8), (4,4,4)]
BasicBlockD  pool_type=avg   -> [(16,16,16), (8,8,8), (4,4,4)]
ConvBlock    pool_type=conv  -> [(16,16,16), (8,8,8), (4,4,4)]
ConvBlock    pool_type=avg   -> [(16,16,16), (8,8,8), (4,4,4)]
```

PR #1502, opened 2026-08-17
fix cp collapse handling

hendrikschilling:
## Show us what you made

**In one sentence:** 

**One real example:** Starting with [real data/input], I [action], and it produced [result].

**Before:** 

**After this PR:** 

**Proof:** 


hendrikschilling:
@codex review

PR #1515, opened 2026-08-18
cmake(macos): search frameworks last in the Homebrew LLVM preset

nerln:
## Summary

The `macos-homebrew-llvm` preset can pair a third-party framework's `tiff.h`
with Homebrew's `libtiff`. CMake then appears configured, but the header and
library are ABI-incoherent and compilation fails.

This one-line preset change sets `CMAKE_FIND_FRAMEWORK=LAST`. Frameworks remain
available; Homebrew paths are preferred when both forms exist.

## Evidence

On an affected M4 Mac, a controlled A/B configure of the same commit produced:

| setting | TIFF header | library/version | generated `Tiff.cpp` command |
|---|---|---|---|
| `FIRST` | Mono framework, reports 4.0.9 | Homebrew 4.7.2 | fails on conflicting integer typedefs |
| `LAST` | Homebrew include | Homebrew 4.7.2 | succeeds unedited |

The framework include appeared in 341 of 761 generated compile commands before
the change and in none afterward.

Fresh configure succeeds. Windows, Linux and macOS build/compile jobs and
CodeQL pass. Hosted macOS runners do not contain the conflicting framework, so
the affected-host A/B configure is the relevant regression evidence.

## Limits and relationship to earlier work

This supersedes #1337 and #1511; it is not a new independent contribution.
Unlike #1511, the evidence above uses the build system's unedited generated
commands.

I am not claiming a complete local build: it later encounters a separate
Homebrew LLVM/Xcode SDK include-order problem after TIFF has resolved correctly.
The synthetic render benchmark failure is unrelated to this macOS-preset-only
diff; the same untouched scenario varied from 0.95954× to 1.113× across three
simultaneous PR runs. Vercel's failure is an external authorization check.

## Agentic-use disclosure

OpenAI Codex was used agentically for investigation, testing, audit and
documentation.


nerln:
This is the rebased replacement for the previously disclosed #1337, not a new
independent contribution. I reran a controlled A/B configure on the same
machine.

- With CMAKE_FIND_FRAMEWORK=FIRST, CMake selects
  /Library/Frameworks/Mono.framework/Headers for TIFF while linking Homebrew's
  libtiff; the reported header version is 4.0.9 and the framework include is
  injected throughout the build graph.
- With the preset's CMAKE_FIND_FRAMEWORK=LAST, both TIFF include and library
  resolve to Homebrew, reported version 4.7.2, and Mono disappears from the
  generated compile includes.

Frameworks are reordered, not disabled, so framework-only dependencies remain
available. Fresh configure succeeds. The remaining local build failure is a
separate Homebrew LLVM/Xcode SDK system-include issue and occurs after TIFF has
resolved coherently.

On GitHub, every Windows, macOS and Linux build/compile job and CodeQL passes.
The synthetic performance gate reports parallel/full_res at 1.113x its modeled
reference. The same scenario measured 0.95954x on #1528 and 1.08181x on #1541
in simultaneous runs despite unrelated diffs; this one-line macOS-only preset
does not touch the measured rendering path, so I am reporting the check as
failed and treating the cross-run spread as benchmark variance. Vercel's
separate failure is an external authorization check.

testing, audit and documentation. I remain responsible for the change and this
report.

nerln:
The two red checks on this PR were inherited from `main`, not caused by the change. The evidence, in case it saves someone the same dig:

- The `CI` check is an aggregator: it fails at its `Aggregate required-check result` step, reporting another job's failure rather than one of its own. The only job that actually failed is **Synthetic rendering regression (GCC Release / Valgrind replay)**, at step **Run full fresh Valgrind matrix**.
- **`main` failed the same two jobs on the same day** — [run 32327406697](https://github.com/ScrollPrize/villa/actions/runs/32327406697), a push to `main` on 20 August.
- This PR changes two lines of a macOS / Homebrew-LLVM CMake preset. The failing job is a Linux/GCC Valgrind replay, which those two lines cannot reach.

I have just updated the branch onto current `main`, so the checks should now reflect the change itself rather than the state of `main` on the evening this was opened. The diff is unchanged: one file, +2 −1.

Happy to rebase or squash differently if that suits review better.

- [ ] I have read this and checked the run it cites.

nerln:
Closing this one. The change is two lines of a macOS / Homebrew-LLVM CMake preset; on the branch updated onto current `main`, every required job passes except the Linux/GCC Valgrind replay, which those two lines do not reach. I could not account for that from here, and I would rather not leave a PR open with a failure I cannot explain. The diff is preserved on the branch.

PR #1528, opened 2026-08-18
vc_render_tifxyz: --accum-type alpha renders black and beerlambert renders a mean

nerln:
## Summary

This PR fixes two defects in the same rendering workflow:

1. `--accum-type alpha` produced an all-black image, while `beerLambert`
   silently produced a mean.
2. VC3D's “Also write TIFF slices” option passed the nonexistent
   `--include-tifs` flag and aborted before rendering.

## Fix

`readCompositeFast` now routes layer-based methods through the shared
`compositeLayerStack()` implementation already used by the viewer.
`methodRequiresLayerStorage()` remains the source of truth, the hot loop reuses
a per-thread buffer, and `isoCutoff` follows the existing per-layer semantics.

For dual output, VC3D derives Zarr mode from the `.zarr` suffix and emits the
supported `--tif-output <sibling>_layers` argument. TIFF-only renders do not
receive a duplicate output.

## Evidence

On a 512×512 PHerc0172 ROI with 33 layers:

| mode | before | after |
|---|---:|---:|
| alpha | 0/262,144 non-zero pixels | 262,144/262,144 |
| Beer–Lambert vs mean | 0 differing pixels | 262,132 differing pixels |

before/after render

The rebased tests compare alpha and Beer–Lambert directly with the shared
compositor on an ordered layer stack. Runner tests cover exact Zarr + sibling
TIFF arguments and TIFF-only output.

Local verification:

- `test_slicing`: 14 passed
- `test_compositing`: 22 passed
- `test_slicing_more`: 8 passed
- `test_command_line_tool_runner`: 11 passed
- the real `CommandLineToolRunner.cpp` and `SegmentationCommandHandler.cpp`
  objects compile

Public Windows, macOS and Linux jobs, CodeQL and the synthetic performance gate
pass. Vercel's failure is an external authorization check.

## Limit

The Beer–Lambert example uses the existing CLI defaults and is intentionally
bright. It proves that the requested compositor and parameters now run; it does
not claim that those defaults are optimally tuned or improve downstream ink
metrics.

## Agentic-use disclosure

Claude Code and OpenAI Codex were used agentically for investigation,
implementation, testing, audit and documentation.

### DIFF
--- a/volume-cartographer/apps/VC3D/CommandLineToolRunner.cpp
+++ b/volume-cartographer/apps/VC3D/CommandLineToolRunner.cpp
@@ -681,8 +681,16 @@ QStringList CommandLineToolRunner::buildArguments(Tool tool)
             if (_flipAxis >= 0) {
                 args << "--flip" << QString::number(_flipAxis);
             }
-            if (_includeTifs) {
-                args << "--include-tifs";
+            // "Also write TIFF slices" only adds anything when the primary
+            // output is the zarr -- a TifStack render already writes them.
+            // vc_render_tifxyz accepts --zarr-output and --tif-output together,
+            // so ask for the slices in a sibling directory of the .zarr rather
+            // than through a flag of its own.
+            if (_includeTifs && _renderOutputFormat == RenderOutputFormat::Zarr) {
+                const QFileInfo zarrInfo(_outputPattern);
+                args << "--tif-output"
+                     << zarrInfo.dir().filePath(zarrInfo.completeBaseName()
+                                                + QStringLiteral("_layers"));
             }
             if (_flatten) {
                 args << "--flatten";
--- a/volume-cartographer/apps/VC3D/CommandLineToolRunner.hpp
+++ b/volume-cartographer/apps/VC3D/CommandLineToolRunner.hpp
@@ -153,6 +153,10 @@ public:
     void setFlattenOptions(bool flatten, int iterations, int downsample = 1);
     void setPreserveConsoleOutput(bool preserve);
 
+#ifdef VC_TESTING
+    QStringList argumentsForTesting(Tool tool) { return buildArguments(tool); }
+#endif
+
     // Valid while toolFinished is being delivered for the completed process.
     bool currentExecutionIsSilent() const
     {
--- a/volume-cartographer/apps/VC3D/SegmentationCommandHandler.cpp
+++ b/volume-cartographer/apps/VC3D/SegmentationCommandHandler.cpp
@@ -2052,8 +2052,14 @@ void SegmentationCommandHandler::onRenderSegment(const std::string& segmentId)
 
     _cmdRunner->setSegmentPath(dlg.segmentPath());
     _cmdRunner->setOutputPattern(dlg.outputPattern());
-    // Interactive renders always start from the TIFF-stack default.
-    _cmdRunner->setRenderOutputFormat(CommandLineToolRunner::RenderOutputFormat::TifStack);
+    // Interactive renders default to a TIFF stack; a .zarr output path is how
+    // the dialog asks for the zarr store instead. That same suffix is what
+    // enables its "Also write TIFF slices (Zarr)" checkbox, so the two have to
+    // agree or the checkbox has nothing to attach to.
+    _cmdRunner->setRenderOutputFormat(
+        dlg.outputPattern().endsWith(QStringLiteral(".zarr"), Qt::CaseInsensitive)
+            ? CommandLineToolRunner::RenderOutputFormat::Zarr
+            : CommandLineToolRunner::RenderOutputFormat::TifStack);
     _cmdRunner->setRenderParams(static_cast<float>(dlg.scale()), dlg.groupIdx(), dlg.numSlices());
     _cmdRunner->setRenderVoxelSize(
         renderVolume ? renderVolume->voxelSize() : 0.0,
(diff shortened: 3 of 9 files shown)


nerln:
Update after rebasing onto current main: commit a4a938ec adds regressions for
both parts of this PR.

- readCompositeFast is tested against the shared compositor on an ordered
  [20, 60, 100, 140] stack for alpha and Beer–Lambert, not just checked for a
  non-zero result.
- CommandLineToolRunner is tested for the exact Zarr + sibling TIFF arguments,
  and for the absence of a duplicate TIFF output on TIFF-only renders.

Verification on the rebased branch:

- test_slicing: 14 passed;
- test_compositing: 22 passed;
- test_slicing_more: 8 passed;
- test_command_line_tool_runner: 11 passed;
- CommandLineToolRunner.cpp and SegmentationCommandHandler.cpp compile as their
  real VC3D objects.

This preserves the shared compositing implementation rather than copying it
into the fast path.

The public Windows, macOS and Linux jobs, CodeQL, and the synthetic performance
gate pass on this revision. Vercel's separate failure is an external
authorization check, not a project build or test result.

implementation, testing, audit and documentation. I remain responsible for the
change and this report.

nerln:
Still current: this merges cleanly onto today's `main` and all checks are green.

Flagging it because the two automated review attempts on it both came back with "Copilot was unable to review this pull request … quota limit", so it may have fallen off the list without anyone seeing it.

What it fixes, one line each:

- `--accum-type alpha` renders an all-black image, and `beerLambert` silently renders a mean instead — both because `readCompositeFast` never routed layer-based methods through `compositeLayerStack()`, the implementation the viewer already uses.
- VC3D's "Also write TIFF slices" passes `--include-tifs`, a flag that does not exist, so the render aborts before doing any work.

Both are user-visible: the first returns a wrong image without an error, the second fails a run outright.

The change routes those methods through the existing shared implementation rather than adding a second one, keeps `methodRequiresLayerStorage()` as the single source of truth, and adds `core/test/test_slicing.cpp` (126 lines) covering the composite paths.

Happy to split it into the two independent fixes if that is easier to review, or to rebase onto a newer base.

- [ ] I have read this.

hendrikschilling:
@codex review

hendrikschilling:
looks functionaly correct just needs update/fix for:

```
  1. P1: The PR no longer compiles when merged into current main.
     volume-cartographer/core/src/Slicing.cpp:580 calls methodRequiresLayerStorage(). Current main removed that function in 07656e634 (vc3d volume rendering, PR #1566), which simplified   
     the shared compositor to five scalar methods. Git’s automatic merge keeps this call but does not restore the declaration, producing an undefined identifier during compilation.        

     Minimal resolution: rebase and define the storage condition directly as median, minabs, alpha, or beerLambert. Do not restore the compositing methods removed by #1566.

  2. P2: Re-running dual output can silently leave stale TIFFs.
     The PR derives <zarr-name>_layers automatically in volume-cartographer/apps/VC3D/CommandLineToolRunner.cpp:689. However, vc_render_tifxyz skips TIFF generation whenever all expected  
     files already exist, even while regenerating the Zarr output (volume-cartographer/apps/src/vc_render_tifxyz.cpp:1655). Re-rendering the same Zarr path can therefore produce a new Zarr
     alongside TIFFs from an earlier render.

     Minimal resolution: when both outputs are requested without --resume, overwrite the TIFFs. Existing-output skipping should be tied to explicit resume behavior.
```

nerln:
Thanks — rebased onto current main and addressed both points. (1) needsLayerStorage() now names median, minabs, alpha and beerLambert directly; nothing removed by #1566 is restored. (2) Existing TIFF slices are skipped only with --resume; without it, requesting both outputs regenerates the TIFFs alongside the zarr. Built locally (vc_core, vc_render_tifxyz) and the slicing and tool-runner tests pass.

nerln:
Thanks — the comment was indeed stale after #1566, and I have removed the two methods it wrongly listed (8704f4e).

One note on the suggested changeset: besides the comment, it also drops `stack.validCount = count;`. That line tells the compositor how many layers to use, and the field defaults to 0, so applying the suggestion as-is would composite zero layers and reintroduce the black render this PR fixes. I changed the comment only.

hendrikschilling:
Close, two small details:

```
  2. P2, edge case: trailing whitespace makes the dialog and command builder disagree.
     The dialog tests the trimmed output path when enabling Zarr options, but volume-cartographer/apps/VC3D/SegmentationCommandHandler.cpp:2053 uses the untrimmed value. Entering
     render.zarr  enables “Also write TIFF slices” but launches a TIFF-only render into a directory literally ending in whitespace. Trim once and reuse that value for both the path and    
     suffix test.
```

And the doc fix that copilot also saw:
```

The isoCutoff documentation mismatch remains: the implementation consistently drops sub-threshold layers, while volume-cartographer/core/include/vc/core/util/Compositing.hpp:51 says   
    they are set to zero. This is documentation ambiguity, not a functional regression in this PR.
```

nerln:
Both done in 04ee80c.

1. `SegmentationCommandHandler` now trims the output pattern once and uses that value for both the path and the `.zarr` suffix test, so the dialog and the command builder can no longer disagree on a pattern with trailing whitespace.

2. `isoCutoff` is documented as dropping sub-threshold layers from the stack rather than zeroing them, and the comment notes that this changes the mean/max/min denominator. Documentation only, no behaviour change.

`vc_core` builds clean.

PR #1537, opened 2026-08-19
tolerate missing meta , tolerate 1 vx shape desc error in las meta

bruniss:


hendrikschilling:
@bruniss I think adressing the codex comment is worth it, we should avoid duplication if possible.

PR #1541, opened 2026-08-20
fix(core): find samplePixelSize wherever it is nested in remote metadata

nerln:
## Summary

Remote voxel-size discovery still assumes one fixed metadata path. On many
surface-prediction stores this leaves `voxelsize` at zero, so
`vc_grow_seg_from_seed` traces a surface, computes `area_cm2 = 0`, discards it,
and exits without saving anything.

This PR adds a tested, bounded search for an unambiguous `samplePixelSize`
wherever it is nested in remote metadata.

## Fix

The new `RemoteVolumeMetadata` helper:

- traverses objects and arrays to depth 64;
- accepts repeated identical positive values;
- rejects conflicting values instead of depending on traversal order;
- ignores absent, invalid and non-numeric values.

Existing `voxelsize` keys and aliases retain precedence. Only the final
`samplePixelSize` fallback changes.

## Evidence

On a published PHerc0125 m7 surface store, with `min_area_cm = 0.0001`:

| | before | after |
|---|---:|---:|
| voxel size | 0 | 9.362 µm |
| raw surface area | 401.094 vx² | 401.099 vx² |
| converted area | 0.000000 cm² | 0.000352 cm² |
| result | discarded | retained |

The tool's default threshold is 0.3 cm²; this deliberately small diagnostic
surface does not exceed it. The example demonstrates the repaired unit
conversion and configured-threshold behavior, not a change to surface growth.

A separate audit of 1,004 published stores found:

- current lookup: 58/1,004 stores and 1/84 surface-prediction stores resolve;
- this lookup: 105/1,004 and 43/84 resolve;
- no document contains conflicting `samplePixelSize` values.

Five offline test cases with multiple assertions cover legacy and nested
layouts, arrays, repeated values, conflicts, and invalid or missing values. The
test binary passes; the real patched `Volume.cpp` compiles; fresh CMake configure
succeeds. Windows, macOS and Linux build/compile jobs and CodeQL pass.

## Limits and relationship to earlier work

This extends #1227 rather than duplicating it: #1227 removed a lookup gate;
this PR handles metadata layouts still outside its fixed path. It is also
distinct from #1508/#1509, which concern documented access URLs and Python cube
caching rather than C++ physical-unit discovery.

The remaining 899 stores expose no reachable value; explicit `voxelsize`
remains the workaround. This PR does not infer from directory names or change
the zero-area diagnostic. A complete local build later encounters a separate
Homebrew LLVM/Xcode SDK include-order issue. The unrelated synthetic render
scenario varied from 0.95954× to 1.113× across simultaneous PR runs; Vercel's
failure is an external authorization check.

## Agentic-use disclosure

OpenAI Codex was used agentically for investigation, implementation, testing,
audit and documentation.


nerln:
This was done with LLM assistance.
Why I think it matters: this exact failure was hit while trying to trace new surfaces on scrolls that don't have published segments yet. The tool runs fine, finishes, and then silently throws away every surface it just traced, with a message that blames the area threshold instead of the missing voxel size. 43 of the 84 published surface-prediction stores, the ones that the tool is meant to be pointed at, currently resolve no voxel size, so someone pointing the tracer at them gets them discarded unless they already know the workaround.
The change is small, and on all 1004 published stores it was checked that no document carries two different values of the field.

nerln:
Update after rebasing onto current main: I extracted the recursive metadata
search into RemoteVolumeMetadata and added an offline regression test target.

The helper now:

- traverses objects and arrays to depth 64;
- accepts repeated identical positive samplePixelSize values;
- rejects conflicting values instead of choosing by traversal order;
- ignores invalid/non-numeric values.

Five test cases with multiple assertions cover the legacy path, source/metadata
nesting, the wrapperless 1.129 µm layout, arrays, repeated values, conflicts and
invalid/absent values.
The isolated test binary passes all 5 cases. I also compiled the real patched
Volume.cpp with the exact project compile command; syntax compilation succeeds.

The earlier PR description says no unit test was added; that limitation is now
obsolete at commit 5077ddec. A fresh full CMake configure also succeeds. A full
local build still encounters an unrelated Homebrew LLVM/Xcode SDK include-order
problem outside this patch, so I am not claiming a complete local build.

On GitHub, the Windows, macOS and Linux build/compile jobs and CodeQL pass. The
synthetic performance gate reports only parallel/full_res above its modeled
threshold (1.08181x reference). The same scenario ranged from 0.95954x to
1.113x across three simultaneous PR runs whose diffs do not touch that hot path,
so I am recording the failure rather than attributing it to this metadata-only
change. Vercel's separate failure is an external authorization check.

implementation, testing, audit and documentation. I remain responsible for the
change and this report.

Bullo27:
This PR fixes **#1603** — an open issue filed six days *after* this PR, diagnosing the same defect on a different store — and the two had not been cross-linked until now. I had independently written a narrower fix for #1603 (a fixed list of three acquisition roots); this recursive approach subsumes it, so I verified this PR instead of filing a duplicate. Everything below was run today.

**Build & tests** (Linux x86-64, GCC 13.3, Release): head `0ce458fca` builds with zero errors; the new doctest passes (`5 test case(s) passed`); the full ctest suite at the head is **155/155**; `git merge-tree --write-tree` against current main `61a62c445` is clean (rc 0).

**The helper against real bucket documents** — a standalone probe linking the PR's `vc_core`, running both the old fixed walk and `findUnambiguousSamplePixelSizeMillimeters` on four freshly fetched `metadata.json` documents, one per schema I had measured in #1603:

| document | old fixed walk | this PR |
|---|---|---|
| `PHercParis4/…/predictions/surfaces/…th0.45.zarr` (`source.metadata.scan.…` nesting) | unresolved | **0.0024 mm → 2.4 µm** |
| `PHercParis4/volumes/20260608103018-1.129um-…` (top-level `tomo`, no `scan` wrapper) | unresolved | **0.001129 mm → 1.129 µm** |
| `PHerc0172/volumes/20241024131839-…-masked` (`samplePixelSize` absent anywhere) | unresolved | unresolved — correctly |
| `PHercParis4/volumes/20260411134726-2.400um-…` (classic `scan.tomo.…`) | 0.0024 mm | **0.0024 mm — identical** |

Both fixable schemas resolve, the unfixable one stays honestly unresolved, and the case the old walk handled is byte-identical — no regression.

**End-to-end on the #1603 store** — the exact reproduction from that issue (`vc_grow_seg_from_seed`, seed `17963 14015 27830`), built from this PR's head:

```
voxelsize: 2.4
generated surface 15627649.108256 vx^2 (0.900153 cm^2)
```

exit 0 at generation 99, and the surface is **saved** (`x.tif`, `y.tif`, `z.tif`, `meta.json`, `generations.tif` in the target dir). On main, measured 2026-08-26 with the same seed: `voxelsize 0`, `15632971.79 vx^2 (0.000000 cm^2)`, surface discarded, output directory empty. (The small vx² difference between runs is tracer nondeterminism, seen across main runs too.)

**Independent check of the audit's premise** — I swept every `<scroll>/representations/predictions/surfaces/*.zarr` in the open-data bucket today, two full passes with retries, both passes identical, zero transient failures:

```
surface-prediction stores: 43
old samplePixelSize walk resolves: 0     this PR resolves: 42
metadata.json absent: 1                  conflicting documents: 0
```

The **zero conflicts** point is the one that matters: on today's bucket, the reject-on-conflict semantics never withholds a value that traversal-order selection would have produced. My counts use a narrower instrument than the PR body's (only the `samplePixelSize` fallback, only current `.zarr` stores — the bucket has also changed since Aug 20), so the denominators differ from the 84 quoted there; the direction and the conflict result are what I can confirm.

Two minor observations, neither blocking: (1) `depth > kMaxMetadataDepth` rejects the **whole document** rather than skipping the too-deep subtree — unreachable in the measured corpus, but if metadata ever nests past 64 levels a shallow valid value would be lost with it; (2) "identical repeated values" uses exact float equality, so two encodings of the same physical size (e.g. `0.0024` vs `2.4e-3` parse equal, but a rounded copy would not) would read as a conflict — also zero occurrences today.

For maintainers: with #1603 open and reproduced, this 76-line change plus tests closes a defect that silently discards traced surfaces on 42 of the 43 stores the tracer is pointed at.

<sub>Verification by Claude (AI agent) operating on behalf of @Bullo27. Every number above comes from commands run 2026-08-28 against head `0ce458fca`, main `61a62c445`, and the live open-data bucket.</sub>

nerln:
The single red check here (`analyze_fast`) is not a failing test and does not come from this diff. What actually happened, in case it is useful beyond this PR:

- `analyze_fast` is an aggregator. It fails at its **Verify PR scan shards** step, which is literally `if [[ "$SHARDS_RESULT" != "success" ]]; then exit 1`.
- Of the four CodeQL PR shards on that head, three were green and one — **PR scan (tools and libraries)** — was **cancelled**, not failed. The gate does not distinguish the two, so a cancelled shard reads as a red check.
- The cancellation comes from the workflow's own `concurrency` group in `.github/workflows/vc3d-codeql.yml`, which sets `cancel-in-progress: true`: a second event on the same ref kills the run in flight. The shards themselves set `fail-fast: false` and a 30-minute timeout, so neither a sibling shard nor a timeout is involved.

Flagging it rather than quietly re-running, because the same pattern will make any PR look red after two pushes in quick succession, and a cancelled shard is indistinguishable from a real failure at a glance.

The two earlier red checks on this PR were a different matter and are gone: they were inherited from `main`, which failed the same Valgrind job on 20 August ([run 32327406697](https://github.com/ScrollPrize/villa/actions/runs/32327406697)). I have just updated the branch onto current `main`.

- [ ] I have read this and checked the workflow it cites.

jrudolph:
Superseded by #1737 which implements a more deterministic approach of figuring out metadata.

jrudolph:
But anyway thanks for bringing this issue to our attention!

PR #1544, opened 2026-08-20
Bound remote prefetch by render timeout

robertlangdonn:
## Summary

`vc_render_tifxyz --prefetch-remote` could remain blocked in synchronous
prefetch work until an external watchdog terminated the process, leaving
placeholder TIFFs behind. This change adds a deadline-aware prefetch path and
makes the renderer return an explicit error when the reproduced prefetch path
exhausts its `--timeout` budget.

## What changed

- Add `IChunkedArray::prefetchChunksUntil(...)` with a deadline-based default
  implementation.
- Implement the bounded wait in `ChunkCache` using the existing scheduler and
  condition variable.
- Reserve one minute of `--timeout` for the renderer watchdog, while preserving
  existing unlimited behavior when `--timeout=0`.
- Add a regression test proving an unresolved fetch returns at its deadline.

## Real-scroll evidence

On PHerc1203, the reproduced three-minute command planned 556,053 remote
chunks. Before the change this path reached the external watchdog while
leaving invalid placeholder TIFFs. On a fresh current-main Linux build with
this change it returned after 130 seconds:

```text
Error: remote prefetch exceeded --timeout budget
renderer_rc=1
```

The output remained invalid placeholder data and was rejected; no render,
unwrapping, ink, or letter result is claimed.

The run wrote 62 placeholder files of 134 bytes each; they were rejected as
invalid. The empty cache and output sizes are reported as failure evidence, not
scientific artifacts.

## Verification

- Current-main Linux `vc_render_tifxyz` build: passed (`85/85` build steps).
- Focused `test_chunk_cache`: passed (`1/1`, 1.31 seconds).
- Fresh PHerc1203 failure path: explicit error, renderer exit 1 after 130
  seconds; outer watchdog not reached.
- GitHub CI: relevant compile, test, scan, and gate jobs pass; the Vercel
  authorization status is unrelated to this C++ change.

## Scope and limitation

This bounds the reproduced renderer failure path. It does not claim that every
possible scheduler or transport operation is cancellable, and it does not turn
an incomplete remote fetch into a scientific output. Its purpose is to fail
explicitly before invalid TIFF stacks enter downstream geometry or ink
analysis.

The verified configuration uses `--timeout 3`, which reserves one minute
between the prefetch deadline and the renderer watchdog. A one-minute timeout
can race the watchdog and is not included in the claim.

### DIFF
--- a/volume-cartographer/apps/src/vc_render_tifxyz.cpp
+++ b/volume-cartographer/apps/src/vc_render_tifxyz.cpp
@@ -556,10 +556,21 @@ static std::vector<vc::render::ChunkKey> collectPrefetchKeysForRows(
 
 static bool prefetchChunkKeys(
     vc::render::IChunkedArray* cache,
-    const std::vector<vc::render::ChunkKey>& keys)
+    const std::vector<vc::render::ChunkKey>& keys,
+    int timeoutMinutes)
 {
     if (!cache || keys.empty()) return true;
-    cache->prefetchChunks(keys, true);
+    if (timeoutMinutes <= 0) {
+        cache->prefetchChunks(keys, true);
+        if (!g_logFile) std::fprintf(stderr, "\n");
+        return true;
+    }
+    const auto deadline = std::chrono::steady_clock::now() +
+                          std::chrono::minutes(timeoutMinutes);
+    if (!cache->prefetchChunksUntil(keys, deadline)) {
+        std::fprintf(stderr, "Error: remote prefetch exceeded --timeout budget\n");
+        return false;
+    }
     if (!g_logFile) std::fprintf(stderr, "\n");
     return true;
 }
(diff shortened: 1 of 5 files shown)


robertlangdonn:
Fresh current-main verification completed on Linux (2026-08-26).

### Source and build

- Upstream base: `c53c74a2901bab5c8b373a8e562c782e392be2e0`
- Verified source tree: `ada74da37c81a6ff9cdf5fbf3365d7b075d37f10`
- Container: `sha256:bad516f66001abca759454cc43e4fd11e5b19aa55d36bdc2043817291c8083c4`
- `vc_render_tifxyz` and `test_chunk_cache`: built successfully (`85/85` steps)
- `test_chunk_cache`: `1/1` passed in `1.31 s`

### PHerc1203 reproduction

The retained `auto_grown_20260724161520607` surface hashes matched locally and
on the Linux VM. The exact remote-render configuration used the public
`20260319130212-2.403um-0.2m-77keV-masked.zarr` volume with
`--prefetch-remote --timeout 3 --group-idx 0 --scale-segmentation 2 --scale 0.25
--num-slices 62 --slice-step 1 --auto-crop` and a separate 240-second process
bound.

```text
Prefetch: 556053 chunk(s) across rows 0..27
Error: remote prefetch exceeded --timeout budget
renderer_rc=1
elapsed_seconds=130
output_file_count=62
62 files x 134 bytes
cache bytes=0
```

The renderer returned before its three-minute watchdog. The files were invalid
placeholders and were rejected; no render, geometry, ink, or letter result is
claimed.

One boundary surfaced during the audit: with `--timeout 1`, the watchdog raced
the prefetch deadline and returned `[timeout]`/exit 2 after 69 seconds. The PR
claim is therefore deliberately limited to the verified three-minute
configuration, which reserves one minute between the prefetch deadline and the
watchdog; it is not a claim that every positive timeout guarantees graceful
cache failure or cancellation of every transport operation.

PR #1051, opened 2026-06-18
VC3D: add horizontal flip to the surface rotation transform

LubuSeb:
Related to #370

The existing VC3D transform on `main` already provides arbitrary rotation preview and Apply. This follow-up adds horizontal flip to the same `Actions > Transforms > Rotate` workflow.

For ordinary single-page TIFF surfaces, Apply safely persists rotation followed by horizontal flip. For multipage mask inputs or surfaces with non-empty components, Apply is blocked before destructive persistence. Focused tests cover transform order and path restoration, compatibility detection, and selection clearing.

## Real-scroll evidence

Captured with the rebased VC3D binary built from source commit `b5e4fd30f1711ede4ddc0dda78b08e25639a4220`, using the [official Open Data manifest](https://vesuvius-challenge-open-data.s3.us-east-1.amazonaws.com/metadata.json): sample `PHerc0800`, volume `20250521135224`, segment `20251028220955` ([source TIFXYZ](https://vesuvius-challenge-open-data.s3.us-east-1.amazonaws.com/PHerc0800/segments/20251028220955-auto_grown_20251028220955262/mesh/intermediate/tifxyz_original/x.tif)).

| Original | Preview: 37° + horizontal flip | Persisted after Apply |
| --- | --- | --- |
| Original real-scroll surface | Real-scroll transform preview | Persisted real-scroll transform |

The materialized coordinate files changed on Apply while their provenance file did not:

- `catalog-origin.json`: `f5d24ba3…` → `f5d24ba3…` (unchanged)
- `x.tif`: `faccfaac…` → `9dd25200…`
- `y.tif`: `0477c43c…` → `fb504ba6…`
- `z.tif`: `df09bbae…` → `44c0bdf6…`

## Validation

- Rebased cleanly onto `main` at `93d939acdf0c238a284ba0ce0b746748ddda892f`.
- GCC 15.2.0 release build: all 152 CTest tests passed, including `surface_rotation_transform`.
- Clang 21.1.8 release build succeeded; 151/152 CTest tests passed, including `surface_rotation_transform`. The sole failure was an unrelated `test_zarr_chunk_fetcher` temp-cache cleanup race, reproduced across isolated reruns with different `.tmp` files and subcases.
- `git diff --check` passes.

AI-assisted tooling was used for implementation and validation. No bounty claim is attached to this follow-up.

Issue #370, opened 2025-09-05
VC3D - Rotate and flip surface view


SuperOptimizer:
Purpose: this would allow for clearer inspection of the surface and/or ink predictions in VC3D.

Suggested Implementation: 'r' to rotate surface view clockwise 90 degrees, 'f' to flip surface view horizontally.

https://github.com/ScrollPrize/volume-cartographer/issues/44

giorgioangel:
@bruniss is this still a issue?

PR #1570, opened 2026-08-22
VC3D: add horizontal flip to the surface rotation transform

LubuSeb:
Related to #370

Supersedes #1051, which GitHub would not reopen after its closed head branch was rebased.

The existing VC3D transform on `main` already provides arbitrary rotation preview and Apply. This follow-up adds horizontal flip to the same `Actions > Transforms > Rotate` workflow.

For ordinary single-page TIFF surfaces, Apply persists rotation followed by horizontal flip. For immutable Open Data catalog segments, opening the transform first asks the user to create or choose an editable copy; the catalog cache remains unchanged. Apply also checks the source again before writing. Multipage or unreadable mask TIFFs and surfaces with disconnected component ranges remain previewable, but Apply is disabled.

The persistence gate inspects the actual TIFF directory count of both `mask.tif` and legacy `multilayer_mask.tif` inputs. Single-page files remain supported.

## Real-scroll evidence

These captures were made with the earlier `b5e4fd30f1711ede4ddc0dda78b08e25639a4220` build using the [official Open Data manifest](https://vesuvius-challenge-open-data.s3.us-east-1.amazonaws.com/metadata.json): sample `PHerc0800`, volume `20250521135224`, segment `20251028220955` ([source TIFXYZ](https://vesuvius-challenge-open-data.s3.us-east-1.amazonaws.com/PHerc0800/segments/20251028220955-auto_grown_20251028220955262/mesh/intermediate/tifxyz_original/x.tif)). They demonstrate the transform output and predate the current immutable-catalog editable-copy behavior.

| Original | Preview: 37° + horizontal flip | Transformed result after Apply (earlier build) |
| --- | --- | --- |
| Original real-scroll surface | Real-scroll transform preview | Transformed real-scroll result |

## Validation (refreshed September 12, 2026)

- Merged current `main` at `be09a85035059fd83471b1632b5898c62f2c65b1`; final head is `b9df0224c408e9a1b6af298a4e25e8b0f2144302`.
- GCC 15.2.0 Release: VC3D and all configured test targets built successfully.
- A fresh full CTest rerun passed 163/163. An earlier run hit the existing `test_zarr_chunk_fetcher` temporary-cache cleanup race; that result is retained in the local evidence.
- All 11 focused transform, Open Data manifest, VC settings, and QuadSurface persistence tests passed.
- `git diff --check` passes.

[Earlier September 8 validation source and logs](https://gist.github.com/LubuSeb/97f52155eda6d7c793d73d0ecb207dd4) cover the real-scroll fixture integration, offscreen startup, and agent-bridge smoke on the previous `c128bb5` head.

AI-assisted tooling was used for implementation and validation. No bounty claim is attached to this follow-up.


LubuSeb:
@bruniss This supersedes #1051 after GitHub refused to reopen its rebased closed head. It is rebased onto current main, includes official-catalog real-scroll before/preview/after-Apply evidence, and the transform test passes under both GCC and Clang. Could a maintainer please add the `keep-open` label and review it?

Bullo27:
### Independent verification on a third toolchain, plus one finding on the Apply gate

Built and tested this PR head (`3f768a14b`) as-is on a toolchain different from both in the description — GCC 13.3.0, CMake 3.28.3, Ubuntu 24.04:

- `ninja` clean; `ctest -j6` → **152/152 passed**, including `surface_rotation_transform` (test #144). That matches the GCC result you reported. The `test_zarr_chunk_fetcher` cleanup race did not reproduce here.

#### The flip itself checks out

I compared what this PR persists against the existing H Flip in the Surfaces panel (`SegmentationCommandHandler::onFlipSurface`, which calls the same `QuadSurface::flipV()` with `path` left intact). Same segment, angle 0 + flip, with `approval.tif` and `generations.tif` sidecars present:

- `x.tif` / `y.tif` / `z.tif` — byte-identical between the two paths
- `approval.tif` / `generations.tif` — **pixel-identical**, and both equal `flip(original)`. The raw bytes differ only in encoding: the existing path re-encodes through `cv::imwrite` (LZW), this PR's goes through `writeChannelFile` → uncompressed, which is the tifxyz convention.

So hiding `surface.path` in `applyInMemory()` does the right thing. `cloneSurface()` has already materialized every channel, so they get flipped in memory and rewritten by `saveOverwrite()`, and preview no longer rewrites disk sidecars on every dial tick.

#### Finding: the gate keys on a filename nothing in the repo writes

On `main` at `93d939acd`, `multilayer_mask` occurs in the entire tree in exactly one file, twice — both *reads*:

```
core/src/QuadSurface.cpp:3065   flipMultiLayerTiff(path / "multilayer_mask.tif", flipCode);  // flipU
core/src/QuadSurface.cpp:3094   flipMultiLayerTiff(path / "multilayer_mask.tif", flipCode);  // flipV
```

Nothing creates `multilayer_mask.tif`. The multipage mask this repo actually produces is **`<segment>/mask.tif`**:

- `vc_create_segment_mask` (built at `apps/CMakeLists.txt:195`) defaults to `mask_path = seg_path / "mask.tif"` (line 63) and, when given a volume, writes two layers `{mask, img}` via `cv::imwritemulti` (lines 105–106).
- `QuadSurface::writeValidMask(img)` writes the same two-layer structure to the same path (`QuadSurface.cpp:1201`).

Measured with this PR's own `SurfaceRotationTransform.hpp`, unmodified:

| segment | `hasMultipageMask` | Apply | outcome |
|---|---|---|---|
| 2-page `mask.tif` | `false` | **runs** | `mask.tif` **2 pages → 1** |
| `multilayer_mask.tif`, 2 pages | `true` | blocked | — |
| `multilayer_mask.tif`, 1 page | `true` | blocked | nothing was at risk |

Row 1 is the exact case the gate is meant to catch: `cloneSurface()` pulls every channel through `QuadSurface::channel()`, which keeps only `layers[0]` of a multipage TIFF, and `writeDataToDirectory()` then writes it back as a single page. Row 3 shows the test is name-based rather than page-count-based, so it also blocks a single-page file where nothing is at risk.

To be clear about what this is and isn't: it is **not** introduced by this PR. The same truncation happens on `main` today through rotate-only Apply — I measured a 3-page sidecar going 3 → 1 through `cloneSurface()` + `rotate()` + `saveOverwrite()` with no flip involved. It is also recoverable, since `saveOverwrite()` snapshots to `backups/<segment>/0/` first. The consequence for this PR is narrower: the safety property in the description ("For multipage mask inputs ... Apply is blocked") does not currently engage for any segment this repo produces, because the guard inherits the `multilayer_mask.tif` name from the pre-existing `flipU`/`flipV` code.

Suggested fix: key the check on the page count of `mask.tif` instead of the presence of `multilayer_mask.tif` — `TIFFNumberOfDirectories` (libtiff is already linked here) or `cv::imreadmulti`.

#### Minor

The flip checkbox is disabled whenever the gate trips (`checkbox->setEnabled(available)`), so flip *preview* is unavailable on those surfaces even though preview never touches disk. The docs only promise rotation preview, so the code is self-consistent — but if the intent was "preview anything, persist only what's safe", the checkbox could stay live with only Apply gated.

The repro for the table is a short file that includes the PR header and links `vc_core`; happy to post it if useful.

---
*Verification performed by Claude (AI agent) on behalf of @Bullo27. Every number above is measured on this machine rather than inferred, and the build plus the three gate cases are reproducible from a clean checkout.*

LubuSeb:
@Bullo27 Thanks — the finding was correct. I updated the gate to inspect the actual TIFF directory count of both `mask.tif` and legacy `multilayer_mask.tif`, fail closed for unreadable masks, and leave rotation/flip preview enabled while disabling only Apply. The focused test now creates real one- and two-page TIFFs for both filenames. Current head `0bb86af` passes the full GCC 15.2.0 and Clang 21.1.8 release suites, 152/152 in each.

Bullo27:
Confirmed on `0bb86af27` — the fix holds. Re-ran the same probe against your updated `SurfaceRotationTransform.hpp`, unmodified:

| segment | `allowed()` | |
|---|---|---|
| 2-page `mask.tif` | blocked | the case that got through before |
| 1-page `mask.tif` | allowed | ordinary segment, no over-block |
| 2-page `multilayer_mask.tif` | blocked | legacy name still covered |
| 1-page `multilayer_mask.tif` | allowed | old name-based over-block gone |
| no mask | allowed | |
| unreadable `mask.tif` | blocked | `maskInspectionFailed` fails closed |

Two extra checks: the block is load-bearing (with the gate honoured, the 2-page `mask.tif` still has both pages after an attempted Apply), and an ordinary 1-page segment still transforms end-to-end — flip-only Apply leaves `mask.tif` at one page and byte-equal to `flip(original)`.

Build on my side (GCC 13.3.0, a third toolchain): **152/152**, including the reworked `surface_rotation_transform`.

One thing I checked before mentioning it, since the gate now runs two `TIFFOpen` + `TIFFNumberOfDirectories` calls per `updatePreview()` rather than one `exists()`: the added cost is 0.12–0.28 ms per call even with a 64-page mask, i.e. under 2% of a 60 fps frame. Not a concern for dial dragging — flagging only so nobody else has to wonder.

Keeping `mask.tif` and the legacy name both inspected, and leaving preview enabled while gating only Apply, both look right to me.

---
*Verification by Claude (AI agent) on behalf of @Bullo27; measured on this machine against your current head.*

LubuSeb:
@bruniss Following up because this replacement carries the Actions > Transforms > Rotate preview/Apply design you requested on #1051. Current head 0bb86af includes the multipage-mask safety correction and was independently verified by @Bullo27 with 152/152 tests on a third toolchain. GitHub will not permit this external PR author to add you as a formal reviewer; could you take a look, approve the pending workflows, and add keep-open if review needs more time?

LubuSeb:
@bruniss @hendrikschilling I've refreshed this onto current main (`d8c5f48`); the PR head is now `c128bb5`.

Fresh GCC Release validation:
- VC3D builds successfully. CTest passed 158/159 initially; the Zarr temporary-cache cleanup failure passed three isolated reruns.
- All nine focused transform/surface tests passed. Eight real-scroll fixture cases also passed with both one and four threads, checking preview leaves files untouched, saved geometry/channels reload correctly, and unsafe mask/component inputs are blocked from persistence.
- Normal offscreen startup and all 55 agent-bridge smoke checks passed.

[Validation source, logs, and limitations](https://gist.github.com/LubuSeb/97f52155eda6d7c793d73d0ecb207dd4) are available, and the PR description has been updated.

Could you review the current revision and approve the upstream workflows? A `keep-open` label while review is pending would also prevent automatic closure. Thanks!

hendrikschilling:
looks fine - update resolving the conflicts that I can merge it.

LubuSeb:
Updated. conflicts are sorted. i also aligned open data catalog transforms with main's editable-copy flow. fresh gcc release build + 163/163 tests pass.

Issue #1540, opened 2026-08-19
new_model fit is structurally unavailable on published catalog lasagna artifacts: shell-dir-crop needs the absent init_shell_dir, cylinder_seed needs umbilicus_json, which ships as an empty string


Bullo27:
**In one sentence:** Starting a from-scratch fit (`lasagna.start_optimization`, `mode: "new_model"`) on a catalog-streamed sample fails for every available init path, because the published manifests omit the fields both paths require — `init_shell_dir` is absent and `umbilicus_json` is an empty string.

**I was trying to:** run the whole streamed fit workflow headlessly on PHerc0332 — after #1527 the attach works, `lasagna.ensure_service` auto-discovers the cached catalog manifest as its dataset, and job submission reaches the fit service; this is the last step of that chain.

**Using:** VC3D from main `a3a4e6f` (Linux, offscreen agent bridge), the live catalog artifacts, fit service from the same checkout on a working CUDA setup (RTX 3060), `lasagna/configs/init_corr.json` and `lasagna/configs/init_cyl.json` unmodified.

**What happened** (each step's exact error, all precise, no crashes — the service stays healthy throughout):

1. `init_corr.json` (`model-init: seed`, `init-mode: shell-dir-crop`) → job error: `shell-dir-crop init requires .lasagna.json key 'init_shell_dir'` (`lasagna/fit.py:929`). Same missing key as #1530, second consumer.
2. `init_cyl.json` (`init-mode: cylinder_seed`, the shell-less path) → job error: `lasagna volume …lasagna.json missing required 'umbilicus_json'`.
3. Those are the only two `init-mode` choices (`lasagna/cli_model.py:33`), and the other `model-init` choices (`ext/model/flatten/atlas`) all start from pre-existing artifacts — so there is no from-scratch path left.

**What the published data has:** all five cached channel manifests (PHerc0125/0139/0332/0814/1451) carry `"umbilicus_json": ""` — the key exists but is empty — and none carries `init_shell_dir`; the two fiber-inference manifests omit `umbilicus_json` entirely. I also searched today's root `metadata.json` recursively: no umbilicus data is published anywhere else in the catalog for these samples, so a consumer can't resolve it from the catalog either.

[...]

PR #1572, opened 2026-08-23
Unblock catalog-streamed Lasagna new-model fits

TAUIL-Abd-Elilah:
**In one sentence:** A catalog-attached `new_model` job can now accept a fit-service-visible shell directory and safely stream channels from the catalog when VC3D has split Zarr metadata from cached chunks.

**One real example:** With the unchanged public PHerc0332 manifest, I passed a local init-shell directory, built a 25×43 initial mesh, opened all three required prediction groups, and read `grad_mag[1050,579,374] = 228` from the public artifact.

**Before:** `load_3d_streaming` opened the manifest-sibling chunk directory as a complete Zarr store and failed with `PathNotFoundError: nothing found at path ''`.

**After this PR:** A valid adjacent `lasagna-remote.json` marker selects the authoritative HTTP/S3 artifact, while normal marker-free local Zarr inputs keep their existing behavior. `initShellDir` is also carried from the agent bridge/MCP call to fit.py as `--init-shell-dir`.

**Proof:**

Before: local split-cache open fails

After: remote groups open and a real PHerc0332 voxel is read

```text
BEFORE=PathNotFoundError: nothing found at path ''
[fit] shell-dir-crop selected shell_0001.tifxyz ... crop=25x43
[fit_data] load_3d_streaming ... groups=['grad_mag', 'nx', 'ny'] sparse_prefetch=tensorstore
AFTER_SIZE=(2100, 986, 986)
REAL_SCROLL_READ=PHerc0332 grad_mag[1050,579,374]=228
REMOTE_SOURCE=True
```

**Independent end-to-end replay on this branch** (`@Bullo27`, through the
agent bridge, unchanged public manifest):

```text
stage 0: 1000/1000  loss 0.1526
stage 1: 1000/1000  loss 0.1384
stage 2:  500/500   loss 0.1376
remote reads: 174 chunks/channel while only 6 chunks were cached locally
wall time: 206 s; peak GPU memory: 0.02 GiB
output: new_model_v002.tifxyz (24 x 42 grid; 0.1357 cm²)
```

[Full independent replay](https://github.com/ScrollPrize/villa/pull/1572#issuecomment-5384415862)
The 2,500-step replay used a synthetic shell only to seed a mesh through a real PHerc0332 material point. It demonstrates that the catalog workflow runs and exports; it does not establish resulting surface quality.

Same public artifact and manifest on both sides:
`PHerc0332/representations/predictions/lasagna/20251211183505-lasagna-20260419180421-L2/PHerc0332-20251211183505-lasagna-20260724.lasagna.json`

**Why / where this is useful:** It removes the two blockers that prevented `shell-dir-crop` new-model jobs from reaching optimization on catalog-attached samples.

## Details

Tested commit: `2363183da`

- 24 manifest/streaming resolver tests pass.
- 25 init-shell tests pass.
- MCP descriptor/contract tests pass.
- Real PHerc0332 metadata opened for `cos`, `grad_mag`, `nx`, and `ny`; the real voxel above was read through TensorStore.
- Full replay record: https://github.com/TAUIL-Abd-Elilah/vesuvius-repro/tree/25d99b7/evidence/villa-1540
- A full local fit reached stage-0 prefetch, then stopped because this Windows host lacks Ninja/CUDA toolkit for the existing JIT extension.

This addresses the `shell-dir-crop` and split-store blockers in #1540. `cylinder_seed` remains blocked by the missing published `umbilicus_json`.

Thanks @Bullo27 for the exact reproducer, reference run, and split-cache diagnosis.

### DIFF
--- a/lasagna/cli_data.py
+++ b/lasagna/cli_data.py
@@ -11,6 +11,7 @@ import fit_data
 @dataclass(frozen=True)
 class DataConfig:
 	input: str
+	init_shell_dir: str | None                           # optional fit-service path for shell-dir-crop init
 	device: str
 	downscale: float
 	crop: tuple[int, int, int, int, int, int] | None  # (x, y, z, w, h, d) fullres
@@ -30,6 +31,8 @@ class DataConfig:
 def add_args(p: argparse.ArgumentParser) -> None:
 	g = p.add_argument_group("data")
 	g.add_argument("--input", default=None)
+	g.add_argument("--init-shell-dir", default=None,
+		help="Shell directory for shell-dir-crop init; overrides init_shell_dir in the input manifest")
 	g.add_argument("--device", default="cuda")
 	g.add_argument("--downscale", type=float, default=4.0)
 	g.add_argument("--crop", type=int, nargs=6, default=None,
@@ -81,6 +84,11 @@ def from_args(args: argparse.Namespace) -> DataConfig:
 		winding_volume = str(winding_volume)
 	return DataConfig(
 		input=str(args.input),
+		init_shell_dir=(
+			str(args.init_shell_dir).strip()
+			if getattr(args, "init_shell_dir", None) not in (None, "")
+			else None
+		),
 		device=str(args.device),
 		downscale=float(args.downscale),
 		crop=crop,
--- a/lasagna/fit.py
+++ b/lasagna/fit.py
@@ -923,13 +923,25 @@ def _apply_cylinder_prepare_model_step(mdl: "model.Model3D", model_step: float |
 		mdl.cyl_shell_current_height_step = step
 
 
-def _require_manifest_init_shell_dir(prep_params: dict) -> str:
-	value = prep_params.get("init_shell_dir", None)
+def _require_init_shell_dir(
+	prep_params: dict, *, override: str | None = None
+) -> str:
+	value = (
+		override
+		if isinstance(override, str) and override.strip()
+		else prep_params.get("init_shell_dir", None)
+	)
 	if value is None:
-		raise ValueError("shell-dir-crop init requires .lasagna.json key 'init_shell_dir'")
+		raise ValueError(
+			"shell-dir-crop init requires --init-shell-dir or "
+			".lasagna.json key 'init_shell_dir'"
+		)
 	if not isinstance(value, str) or not value.strip():
-		raise ValueError("shell-dir-crop init requires non-empty string .lasagna.json key 'init_shell_dir'")
-	return str(value)
+		raise ValueError(
+			"shell-dir-crop init requires a non-empty --init-shell-dir or "
+			".lasagna.json key 'init_shell_dir'"
+		)
+	return value.strip()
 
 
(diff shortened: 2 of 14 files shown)


Bullo27:
Replayed the 2500-step case on the CUDA rig as requested — **it completes end-to-end on your branch with no hand-injection at all.**

Ran through the agent bridge with your new `initShellDir` parameter, so the C++ plumbing is exercised too, not just the Python. Commit `2363183da`, GCC 13.3.0 / RTX 3060 / CUDA 12.8, streaming the public artifact.

**The three manual steps my 08-20 reference run needed are all gone.** The probe asserts this rather than assuming it:

- manifest sha `201c45c6e54a2ffd` **identical before and after**, and it has no `init_shell_dir` key — the path came from `initShellDir`
- **0** composed `.zarray` in the visible chunk tree (the run aborts if any leftover exists)
- no chunk prefetch: **only 6 chunk files were cached locally, yet the fit read 174 chunks per channel** — so the remote path is genuinely doing the work

**Result:**

```
stage0 1000/1000  loss 0.1526   59.39s
stage1 1000/1000  loss 0.1384  1m06.61s
stage2  500/500   loss 0.1376    24.94s
[sparse_cache_ts] grad_mag/nx/ny: 174 chunks in 2500it, 6.5MiB each
[fit] peak GPU memory: 0.02 GiB
[fit2tifxyz] D=1 Hm=24 Wm=42, area_cm2=0.1357  -> new_model_v002.tifxyz
```

~20 it/s throughout (19.4–21.7), 206 s wall for the whole job.

**Against my 08-20 reference run** (which needed the injected manifest key, a hand-copied `.zarray`, and a prefetched ROI): 174 chunks / 6.5 MiB per channel — *identical working set*; same 24×42 grid; same 0.02 GiB peak; ~20.5 it/s then vs ~20 it/s now. Final loss 0.1288 then vs 0.1376 now, which I'd put down to run-to-run variation rather than anything structural. The thing I half-expected to see and didn't: **streaming from S3 instead of local chunks cost no measurable throughput.**

**One thing worth calling out — your design is safer than the fix I originally suggested.** I proposed composing the halves locally (copy `.zarray` into the chunk tree). On this cache that would have put a valid `.zarray` on top of *6 of 174* needed chunks, and Zarr would have silently returned fill values for the other 168 — a plausible-looking fit on mostly zeros. Refusing to treat the partial cache as a complete store, and going to the authoritative artifact instead, avoids that failure mode entirely. The comment in `resolve_streaming_zarr_source` says exactly this, and it's the right call.

Also confirmed on my box: your `tests/test_fit_data_manifest_lazy.py` + `tests/test_init_shell_index.py` → **49 passed**, matching the 24 + 25 you reported.

Your scoping note is right that `cylinder_seed` stays blocked on the missing published `umbilicus_json` — that's gap 2 from #1540 and it's untouched here, correctly.

---
*Verification by Claude (AI agent) on behalf of @Bullo27. All numbers measured on this machine against `2363183da`; the no-injection assertions above are enforced in the probe, not asserted after the fact.*

TAUIL-Abd-Elilah:
Thank you for the thorough independent replay and the no-injection checks. I linked these results in the PR description.

PR #1573, opened 2026-08-23
spiral: fit a 4,000-slice window on a 16 GB host

7jycwjmbfn-eng:
**In one sentence:** On a 16 GB host with no swap, a 4,000-slice spiral fit goes from killed by the kernel to finished.

**One real example:** Starting with the PHercParis4 lasagna inputs, z 4000-8000, 40 steps, seed 1, on an RTX 4080 Laptop under WSL2 capped at 16 GB with swap off, I ran the fit. Without this branch it was killed by signal 9 during "Excluding track points near patches" at 6:37 with a peak RSS of 15.62 GB, at the host's limit. With this branch the same command finished at 9:36 with exit 0: 5,529,849 of 6,044,103 tracks survive exclusion, 18,447,442 directed crossings, 208,504,823 points retained, step-0 loss 6946.0, final mesh windings [10, 130).

**Before:** The same window on the tree without the last four commits of this branch is killed before the first optimisation step, exit 137. That tree already carries the first four commits; none of them changes the fit's memory. Packing is a separate step, the fingerprint and the test skip allocate nothing, and in-place compaction takes the copy fallback on the packed path. For this measurement it is main.

**After this PR:** It finishes. Peak anonymous memory 10.49 GiB, file-backed 9.42 GiB, peak VmRSS 13.89 GiB. The arrays did not get smaller; the bulk moved to pages the kernel can reclaim. Two independent runs read 10.46 and 10.49.

**Proof:** Two terminal captures.

- `A1_stock_OOM.png`: `Command terminated by signal 9`, `SENTINEL_EXIT=137`, `Maximum resident set size 15620352 kB`.
- `A1_fixed_OK.png`: same command, same data, `SENTINEL_EXIT=0`, the track, crossing and point counts above.

Both runs use the same dataset root, the same config overrides, seed and step count; each log records `SOURCE_COMMIT` and `SOURCE_DIRTY_FILES=0`. The capture prints the exact command, streams the fit's own log, then prints the exit code and peak RSS.

Before:

stock run, killed by signal 9, exit 137

After:

this branch, exit 0, RssAnon peak 10.49 GiB

I wanted to see how far I could push spiral fitting on an ordinary laptop. RAM and GPUs are expensive, and telling people to just buy more hardware isn't much of a solution. All I can do is try to lower the hardware bar a little.

My machine has 32 GB of system RAM and an RTX 4080 Laptop GPU with 12 GB of VRAM. I run the fit inside WSL2, capped at 16 GB of RAM with swap disabled, to roughly mimic a 16 GB laptop. Under those constraints, the stock implementation got killed before it reached the first optimization step on a 4,000-slice window.

The changes on this branch came directly from those failures. Most of them fix a spot where the fit actually ran out of memory on my machine; the rest clear things I hit on the way there: a shipped sidecar that was refused on every machine except the one that wrote it, and tests that failed on a fresh checkout.

This doesn't mean the default 13,000-slice range now fits in 16 GB. It doesn't. Even at 4,000 slices, peak memory is still 13.9 GiB out of the 15.6 GiB available to WSL2, so the full range remains well beyond this memory tier. The host still has to hold most of the working set. The savings come from memory that never needed to stay resident for the whole fit.

There's still more to do. I'm working on render-side fetching separately, and closing the remaining gap to the full 13,000-slice range is next.

## Details

Eight focused commits:

- **39ee1121f:** fingerprints packed track stores by size plus bounded BLAKE2b content probes, so copied or downloaded sidecars are not rejected only because their timestamps changed. Both signatures are written for compatibility.
- **97279485a:** applies the existing native-extension skip guard to three crossing tests. A fresh `uv sync` checkout goes from 3 failed / 7 passed / 3 skipped to 17 passed / 7 skipped in that file.
- **bf47a9865:** replaces unbounded `Executor.map` submission with a sliding window. The 36.9 GiB pool packed in 642 s at 232.8 MiB peak RSS, with `--verify 2000` clean.
- **140ba36ee:** compacts surviving points in place when the caller owns a writable array and survival is high enough. Unsafe cases retain the copy fallback.
- **967ad2970:** adds `build_topology=False` so fitting can use the packed CSR without materialising an unused rustworkx graph. At 3,000 slices, peak RssAnon fell from 14.19 to 12.84 GiB with the same counts and step-0 loss.
- **1b1713bb7:** stores the crossing cache as NPY files so its arrays can remain mmap-backed; the loader still accepts the previous NPZ format. Together with the topology change, the 4,000-slice run completed at 13.26 GiB peak RssAnon.
- **cb976840d:** streams selected track points in blocks and writes survivors directly into one output array. Regression tests compare it with the previous path across track layouts, block sizes and survival fractions.
- **189d82e07:** keeps packed int32 coordinates file-backed and converts them to float32 on first bulk access. This and streamed exclusion reduced peak RssAnon from 13.26 to 10.49 GiB.

**Limits.**

- The default 13,000-slice range is not demonstrated at this memory tier.
- Measurements use RssAnon/RssFile high-water marks and `time -v` peak RSS; they are not sampling-profiler results.
- The installed `vc.spiral_sampling` extension was built for f75e4c29e, so each measurement applies this branch's patches to that tree. The patched lines do not touch the extension.
- The kernel dmesg entry was not captured. The OOM evidence is signal 9, exit 137 and peak RSS at the host limit.
- With the same environment and files, upstream main reports 6 failed / 8 passed / 3 skipped; this branch reports 3 failed / 44 passed / 6 skipped. The three remaining failures reproduce on main.

### DIFF
--- a/spiral-fitting/fit_spiral.py
+++ b/spiral-fitting/fit_spiral.py
@@ -1928,7 +1928,10 @@ class FitContext:
                     or track_sampling_config['crossing_mode'] == 'track_walk'):
                 track_crossing_cache = load_track_crossing_cache(self.tracks_dbm_path)
                 if track_crossing_cache is not None:
-                    track_graph = TrackGraph(track_crossing_cache)
+                    # Training restricts and remaps the exact-crossing CSR;
+                    # it never traverses the duplicate rustworkx topology.
+                    track_graph = TrackGraph(
+                        track_crossing_cache, build_topology=False)
                     print(
                         f'built TrackGraph: {len(track_graph)} tracks, '
                         f'{track_graph.edge_count} crossings in '
--- a/spiral-fitting/pack_resident_pools.py
+++ b/spiral-fitting/pack_resident_pools.py
@@ -29,18 +29,49 @@ from __future__ import annotations
 
 import argparse
 import glob
+import itertools
 import json
 import os
 import sys
 import threading
 import time
-from collections import OrderedDict
+from collections import OrderedDict, deque
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 
 import numpy as np
 
 RESPOOL_FORMAT_VERSION = 2
+
+# Chunk reads in flight per worker thread. Enough to keep every thread busy
+# across one write, small enough that the decoded bricks waiting to be written
+# stay a bounded cost rather than a function of how many chunks the store has.
+READS_IN_FLIGHT_PER_THREAD = 4
+
+
+def map_bounded(executor, function, items, max_in_flight):
+    """Like ``executor.map``, but with a bounded number of pending results.
+
+    ``Executor.map`` submits every task before yielding anything and holds each
+    finished result until the consumer reaches it, so a reader that outruns the
+    writer accumulates the whole input in memory. Packing a lasagna store that
+    way peaked at 26.2 GB and was killed; the only workaround was --io-threads 1,
+    which serialises the reads to make the pile-up small. Keeping a sliding
+    window of submissions instead bounds the resident results directly, so the
+    thread count goes back to being a throughput knob.
+
+    (``chunksize`` does not help here: ThreadPoolExecutor uses the base
+    ``Executor.map``, which ignores it.)
+    """
+    items = iter(items)
+    pending = deque(
+        executor.submit(function, item)
+        for item in itertools.islice(items, max(1, max_in_flight)))
+    while pending:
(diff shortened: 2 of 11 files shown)


pmh47:
In general we are happy to receive PRs that reduce memory/compute and are clean and checkable. However we have just removed the tracks from the default spiral config anyway since they now seem to have minimal benefit (better to derive patches from them first); thus this PR is redundant. If you find there are other limiting factors on your hardware, feel free to submit a new PR.

PR #1574, opened 2026-08-23
spiral: a half-size resident pool that the sidecar selects

7jycwjmbfn-eng:
**In one sentence:** The surf_sdt resident pool takes half the VRAM, its gather scratch is lower than the raw pool's, and the allocator budget the fit needs drops from 7.0 to 5.5 GiB.

**One real example:** Starting with PHercParis4's `las_008_surf_sdt.ome.zarr.respool_g1` at 36.9 GiB, I re-encoded it with `pack_min3.py` to 18.45 GiB, exactly 2.000x at 4 bytes per 8 voxels. On the same ROI, z 4000-4400, 40 steps, seed 1, `dense_spacing_mode=phase`, with the allocator capped at 6.25 GiB, the raw pool dies with `torch.OutOfMemoryError` asking for 350 MiB at 5.73 GiB allocated; the min3 pool finishes 40 steps and writes the 260-winding final mesh with a 4.92 GiB peak.

**Before:** On this ROI the raw pool passes at a 7.0 GiB budget and fails at 6.75.

**After this PR:** The min3 pool passes at 5.5 and fails at 5.25, three runs at each boundary with identical peaks: 4.92 passing, 4.85 failing. Of the 1.5 GiB, compression buys 0.5 (7.0 to 6.5) and the decode rewrite buys 1.0 (6.5 to 5.5).

**Proof:**
- `A2_raw_OOM.png` and `A2_min3_OK.png`: two terminal captures, same ROI, same config, same 6.25 GiB cap, differing only in which encoding the SDT sidecar points at. The tree for these runs also carries the host-memory patches from the companion PR. Those patches affect host memory. The 4.92 GiB min3 peak matches the runs made with the min3 commits alone.

Raw pool at 6.25 GiB:

raw pool, CUDA OOM at 6.25 GiB

min3 pool at 6.25 GiB:

min3 pool, finishes at 6.25 GiB, peak 4.92 GiB

- The full budget sweep, raw / min3 before the rewrite / min3 after, 5.0 to 7.5 GiB in 0.25 steps, is in `runs/min3_caps_2026-08-22/RESULT.md`.
- Lossless: 2,031,312 real gathers through the fitter's own dispatch compare `torch.equal` to the uncompressed pool, brick and block boundaries included. `tests/test_min3_pool.py` builds a sidecar inside the test and covers voxel 7 of each 2x2x2 block, whose code owns the sign bit of the int32 word; four deliberate decode mutations each turn it red.

Same machine, a 12 GB card; many more people have 8 GB or 6 GB. The resident pool occupies steady-state memory; gather scratch pushes VRAM to its peak. My first min3 version halved the pool and then gave more than half of that saving back as gather scratch. After the rewrite, its scratch is lower than the uncompressed pool's. This PR and the host-memory PR hit two different hardware walls, so I kept them separate.

## Details

Three focused commits:

- **0483f7ebe:** moves the existing min3 pool and packer into the repository so run behaviour is determined by the checkout. The layout stores one minimum and eight 3-bit codes in each aligned 32-bit word. The full surf_sdt group-1 census found a worst block span of 6; the packer rejects spans above 7.
- **7f220e8be:** reduces gather scratch by reusing coordinate tensors, building the shift operand as int32 at shape `(n,)`, and decoding in place on the gathered words. On 4,000,000 indices, scratch fell from 144.4 to 76.1 bytes per index and gather overhead fell from +24% to +4.9% over the raw pool.
- **a2f0737f7:** dispatches the resident-pool class from the sidecar encoding, records portable source metadata, and includes the failing chunk path in decode errors.

**Limits.**

- The 2.000x ratio applies to surf_sdt group 1. Other tested stores exceed this encoding's span limit.
- The budget floor uses PyTorch's allocator cap. CUDA context memory, non-PyTorch allocations and small-card fragmentation remain outside the measurement.
- The sweep ROI has zero tracks after filtering. It validates the phase/SDT path for 40 steps; track-dense behaviour and fit quality remain untested.
- The lossless claim covers the gather layer. Full-fit output equivalence was not evaluated because repeated fits already produce different outputs.
- The min3 tests pass, 9 passed; the full suite's failure set matches clean main.

### DIFF
--- a/spiral-fitting/benchmark_resident_pool_memory.py
+++ b/spiral-fitting/benchmark_resident_pool_memory.py
@@ -1,0 +1,158 @@
+#!/usr/bin/env python3
+"""How much scratch VRAM does one gather cost, on top of the pool it reads?
+
+Halving the resident pool is only worth what reaches the fitter's peak.  A
+decode that allocates a few int64 temporaries per gathered voxel can hand back
+everything the smaller pool won, and the pool-size number alone will not show
+it: the pool is steady state, the temporaries are the peak.
+
+For each pool this reports
+
+    resident   bytes the pool holds for the whole fit
+    transient  max_memory_allocated - memory_allocated across one gather,
+               i.e. everything the call touched that the pool does not own
+    output     bytes of the value tensor the call returns
+
+Transient is the number that competes with the model, the optimiser and the
+other pools for the same allocator budget.
+
+The instrument is checked before it is used: --self-test allocates a known
+number of bytes inside a fake gather and requires the reading back to match,
+and the run fails if the reading is off by more than one allocator block.
+Without that, a decode change that does nothing would read as a win.
+
+Run:
+  python benchmark_resident_pool_memory.py <raw sidecar> <min3 sidecar> \
+      [--z-roi lo hi] [--n 4000000]
+"""
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+from pathlib import Path
+
+import numpy as np
+import torch
+
+# absolute(), not resolve(): resolve() expands symlinks, which puts the real
+# checkout path into every traceback this script can raise.
+sys.path.insert(0, str(Path(__file__).absolute().parent))
+
+from min3_pool import Min3BrickPool             # noqa: E402
+from sparse_cuda_cache import ResidentBrickPool  # noqa: E402
+
+
+def measure(fn, *args):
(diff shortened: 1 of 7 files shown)


pmh47:
The SDT input is being deprecated in spiral (in part because of high memory), thus we won't do any further fixes to it.

Issue #1569, opened 2026-08-22
PHercParis4: authoritative frames for historical PointCollections and the published umbilicus


Kotoro78:
We are building a reproducible PHercParis4 winding-supervision pipeline and
need authoritative coordinate-frame declarations for three exact current
official assets. We are not asking for labels, coordinates, visual judgments,
or a newly fitted registration.

## 1. Historical PointCollections

The exact current official bodies are:

| Object | Bytes | SHA-256 |
|---|---:|---|
| [`spiral/PHercParis4/relative_windings.json`](https://huggingface.co/buckets/scrollprize/datasets/resolve/spiral%2FPHercParis4%2Frelative_windings.json) | 773,507 | `a3243511d4eb91387a9b32f4dbff11514b08c3ae36e9b2a2b8222607b4883ac1` |
| [`spiral/PHercParis4/same_windings.json`](https://huggingface.co/buckets/scrollprize/datasets/resolve/spiral%2FPHercParis4%2Fsame_windings.json) | 1,695,752 | `d9be52c5ebb42853f75f235241cbfd159738f6f34468bcd182523bc91dc91048` |

Both are PointCollections schema v1; the five audited modern
coordinate-identity fields (`vc_open_data_coordinate_space` and the four
`vc_open_data_source_*` fields) are absent. Official code resolves serialized
component order as XYZ; that is not the question here.

Could you please state, ideally in commit-pinned official metadata or
documentation:

1. the exact source volume/scan or volpkg used to author each body, including
   immutable identifier and parent scan;
2. its pyramid level, voxel pitch/scale, index origin or crop offset, index
   base, and voxel-center versus voxel-corner convention;
3. the exact mapping from serialized PointCollections XYZ to
   [`PHercParis4/volumes/20260411134726-2.400um-0.2m-78keV-masked.zarr@L1`](https://vesuvius-challenge-open-data.s3.us-east-1.amazonaws.com/PHercParis4/volumes/20260411134726-2.400um-0.2m-78keV-masked.zarr/1/),
   including direction, axes, units, scale, origin/offset and center convention;
4. whether the two bodies share exactly one authoring frame and mapping.

If the current scan's sibling
[`transform.json`](https://vesuvius-challenge-open-data.s3.us-east-1.amazonaws.com/PHercParis4/volumes/20260411134726-2.400um-0.2m-78keV-masked.zarr/transform.json)
is the intended route, please also state explicitly how its
[...]

PR #1575, opened 2026-08-23
docs: correct the spiral-input README against what the dataset publishes

tonclap:
**In one sentence:** the `spiral-input` README describes four things the published dataset does not have — two 404s, an undercounted directory, and a coordinate convention that's wrong for the JSON point collections.

**Before → after:**
- `unverified_patches/` and `fibers/` are listed, both 404 → removed; fiber annotations pointed at `datasets/fiber-skeletons/`, which does publish them.
- `verified_patches/` documented as ~27,399 items; live count is far higher and still climbing (89,237 total entries at re-check, up from 84,319 two days earlier) → the fixed figure is dropped for "~84,000 items and growing", matching the hedge the showcase page already uses.
- Conventions section claims one order, `(z,y,x)` full-resolution, for the whole dataset → narrowed to say which order applies to the JSON collections (`(x,y,z)`, level 2), with `umbilicus.json` added as a supporting case.

**Proof:** plain HTTP against the published paths, plus a CT read at the points themselves — bounds-checking alone doesn't disambiguate the order (`relative_windings.json` is in-bounds under both `(x,y,z)`@L2 and the documented `(z,y,x)`@L0). Read as `(x,y,z)` at level 2, 313/313 sampled points land on papyrus (median CT 89.0). Read as documented, `(z,y,x)` full-resolution, they are in bounds too but land in empty space (median CT 0.0). Full counts and commands in Details.

**Why this matters:** `spiral-input` is the ground truth for fitting the global winding solution, and this README is the only description of it. A reader who trusts the coordinate line silently samples empty volume and blames the annotations, not the convention.

I was looking at `spiral-input` as a possible source for my own winding-solution work, and caught two 404s in the README's file list. Since this README is the only description of the dataset, anyone who trusts it as written will silently sample empty volume.

Independently reproduced by @Bullo27 in the comments below (file list, counts and coordinate order, all against the live data); the evidence table here was corrected on 24.08.2026 after that review — see "Coordinate order" below.

## Details

Checked 21.08.2026 against the live paths, re-checked 22.08.2026 before
opening. The 404s and the coordinate order are unchanged since first
measured 12.08.2026; the patch count is not — the directory keeps growing.

```
curl -o /dev/null -w '%{http_code}\n' https://dl.ash2txt.org/datasets/spiral_datasets/PHercParis4/unverified_patches/
404
curl -s https://dl.ash2txt.org/datasets/spiral_datasets/PHercParis4/verified_patches/ \
  | grep -oE 'href="[^"]+/"' | sort -u | wc -l
89237
```
`fibers/` → 404; `datasets/fiber-skeletons/` → 200. On 21.08 the listing held
84,319 entries, all but 3 matching `band-seed*.tifxyz` (the rest
`lasag_infill_*`). By 22.08 it had grown to 89,237 — the `band-seed*` count
is unchanged at 84,316, and the gain is new patch categories
(`auto_grown_*`, `same_wrap*_lasagna`, and others) added in the interim.
That is why the patch says "~84,000 and growing" rather than an exact
count: the number the README documents (~27,399) would already be wrong
again within a day if replaced with another fixed one. The showcase page's
old ~27,000/~204,000 figures are corrected the same way.

Coordinate order — volume `20260411134726-2.400um-0.2m-78keV-masked.zarr`, level 2 shape 18946×8174×8174, `uint8`, `compressor: null`, level 0 shape 75784×32693×32693. In-bounds counts alone don't settle it (all three collections are fully in-bounds under both `(x,y,z)`@L2 and `(z,y,x)`@L0 — `relative_windings.json` 2173/2173 either way, `same_windings.json` 5413/5413, `abs_winding.json` 59/59), so it's settled by CT instead, on 313 points of `relative_windings.json`:

| reading | in bounds (313-pt sample) | in bounds (all 2173) | median CT | > 80 |
|---|---|---|---|---|
| `(x,y,z)` @ L2 | 313/313 | 2173/2173 | **89.0** | 56.2% |
| `(z,y,x)` @ L2 | 0/313 | 73/2173 (3.4%) | 0.0 | 0% |
| `(x,y,z)` @ L0 (full res) | 313/313 | 2173/2173 | 0.0 | 0% |
| `(z,y,x)` @ L0 (full res) — **as documented** | 313/313 | 2173/2173 | 0.0 | 0% |

Two corrections to an earlier version of this table, both prompted by @Bullo27's reproduction and neither touching the conclusion:

1. **`(z,y,x)` @ L2 is not a clean zero.** It is 0/313 on the sample above (the first 30 collections) but 73/2173 — 3.4% — over the whole file, so a reviewer re-running it on a different stride will get a small non-zero count, as @Bullo27 did (9/313 on every-6th sampling; 10/320 when I repeat that stride). Those points are still empty volume: median CT 0.0, none above 80.
2. **The documented reading is now measured directly.** The earlier table listed `(x,y,z)`@L0 while the prose spoke of `(z,y,x)` full-resolution; both are now in the table, and both put the points in bounds at median CT 0.0. The only reading that lands on papyrus is `(x,y,z)`@L2.

`umbilicus.json` is added as a supporting case (axes named per point, 146 control points on the level-2 grid). Not checked: the order inside `tracks/` (`.vctracks`/`.zarr`/`.dbm`, not JSON) — the Conventions line is narrowed, not rewritten, to leave that unclaimed. The README's total file-count figure is also left unverified.

Related: #1569 establishes by extents that these collections live on a native level-2 grid but stops short of asserting an axis order, since bounds cannot separate `(x,y,z)`@L2 from `(z,y,x)`@L0. Sampling CT at the points is that missing step, so the two reports fit together — this one settles the order half of the frame declaration asked for there.

### DIFF
--- a/scrollprize.org/docs/02_data_datasets.md
+++ b/scrollprize.org/docs/02_data_datasets.md
@@ -69,7 +69,7 @@ This dataset contains manual annotations of the scroll wraps, recording which pa
 ### Paris4
 
 {/* Paris4 = Scroll 1 / PHercParis4. */}
-Spiral annotations for scroll PHercParis4 (Scroll 1). Contents include ~27,000 verified and ~204,000 unverified surface patches, traced tracks, and point collections, together with `same_windings` / `relative_windings` / `abs_winding` graphs, the fitted `umbilicus`, fiber and outer-shell geometry, and the volume inputs used by the fitting pipeline (~49.6 GB total).
+Spiral annotations for scroll PHercParis4 (Scroll 1). Contents include ~84,000 verified surface patches, traced tracks, and point collections, together with `same_windings` / `relative_windings` / `abs_winding` graphs, the fitted `umbilicus`, outer-shell geometry, and the volume inputs used by the fitting pipeline (~49.6 GB total). Fiber annotations are published separately, as the `fiber-skeletons` dataset.
 
 - [README](pathname:///data/datasets/spiral-input-PHercParis4-README.md)
 - [Browse on the data server](https://dl.ash2txt.org/datasets/spiral_datasets/PHercParis4/)
--- a/scrollprize.org/static/data/datasets/spiral-input-PHercParis4-README.md
+++ b/scrollprize.org/static/data/datasets/spiral-input-PHercParis4-README.md
@@ -13,10 +13,10 @@ the volume can be fit and evaluated.
 
 | Path | Description |
 | --- | --- |
-| `verified_patches/` | Manually verified surface patches. Each patch is a grid-topology quad mesh sampled on the papyrus surface (~27,399 items). |
-| `unverified_patches/` | Candidate surface patches not yet manually verified (~203,900 items). |
+| `verified_patches/` | Manually verified surface patches. Each patch is a grid-topology quad mesh sampled on the papyrus surface (84,316 items). |
+| `unverified_patches/` | Candidate surface patches not yet manually verified. Announced as ~203,900 items; not published at this path yet (it returns 404). |
 | `tracks/` | Line annotations: curves traced across the surface, stored as sequences of `(z, y, x)` points. |
-| `fibers/` | Fiber annotations. |
+| `fibers/` | Fiber annotations. Published as a separate dataset, [`fiber-skeletons`](https://dl.ash2txt.org/datasets/fiber-skeletons/); this path returns 404. |
 | `outer_shell/` | Geometry of the scroll's outer shell. |
 | `lasagna_inputs/` | Volume data consumed by the fitting pipeline. |
 | `umbilicus.json` | The scroll's central axis: points defining the spiral center as a function of `z` (depth). |
@@ -28,7 +28,13 @@ the volume can be fit and evaluated.
 
 ## Conventions
 
-- Coordinates are `(z, y, x)`, in full-resolution scroll-volume voxels.
+- Coordinates are `(z, y, x)`, in full-resolution scroll-volume voxels — **except the
+  JSON point collections**. `same_windings.json`, `relative_windings.json` and
+  `abs_winding.json` store `(x, y, z)` in the voxel grid of **level 2** of the scan
+  (18946 x 8174 x 8174), which is the grid the surface prediction is computed on.
+  Read as `(z, y, x)` at full resolution, their points fall outside the papyrus.
+- `umbilicus.json` names its axes per point (`x`, `y`, `z`) and is in that same
+  level-2 grid.
 
 ## License
 
--- a/scrollprize.org/static/data/datasets/spiral-input-PHercParis4-README.md
+++ b/scrollprize.org/static/data/datasets/spiral-input-PHercParis4-README.md
@@ -32,7 +32,9 @@ the volume can be fit and evaluated.
   JSON point collections**. `same_windings.json`, `relative_windings.json` and
   `abs_winding.json` store `(x, y, z)` in the voxel grid of **level 2** of the scan
   (18946 x 8174 x 8174), which is the grid the surface prediction is computed on.
-  Read as `(z, y, x)` at full resolution, their points fall outside the papyrus.
+  Read as `(z, y, x)` at full resolution the points stay inside the volume, so a
+  bounds check will not catch the mistake — they simply land in empty space instead
+  of on the papyrus.
 - `umbilicus.json` names its axes per point (`x`, `y`, `z`) and is in that same
   level-2 grid.
 
--- a/scrollprize.org/static/data/datasets/spiral-input-PHercParis4-README.md
+++ b/scrollprize.org/static/data/datasets/spiral-input-PHercParis4-README.md
@@ -13,7 +13,7 @@ the volume can be fit and evaluated.
 
 | Path | Description |
 | --- | --- |
-| `verified_patches/` | Manually verified surface patches. Each patch is a grid-topology quad mesh sampled on the papyrus surface (84,316 items). |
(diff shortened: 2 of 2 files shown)


Bullo27:
Independently reproduced all of this against the live data today. **The headline coordinate claim holds**, and I have two small notes on the evidence table rather than the conclusion.

**File-list and counts — exact match:**

```
404  spiral_datasets/PHercParis4/unverified_patches/
404  spiral_datasets/PHercParis4/fibers/
200  spiral_datasets/PHercParis4/verified_patches/
200  datasets/fiber-skeletons/

verified_patches/  89,237 entries   (band-seed* = 84,316)
```

Both of your numbers to the digit. The breakdown also supports your "and growing" framing rather than another fixed figure: beyond `band-seed*` it's `auto_grown_*` 3,493, `same_wrap*` 898, `lasag_infill_*` 3, and 527 others (`0000_low_band_final`, `0000_top_vt_strip_1`, …). Against the README's ~27,399 that's 3.3×.

**Coordinate order — reproduced.** 2,173 points in `relative_windings.json`, every 6th sampled → 313, read against `20260411134726-2.400um-0.2m-78keV-masked.zarr`:

| reading | in bounds | median CT | > 80 |
|---|---|---|---|
| `(x,y,z)` @ L2 | 313/313 | **88.0** | 55.3% |
| `(z,y,x)` @ L2 | 9/313 | 0.0 | 0.0% |
| `(x,y,z)` @ L0 | 313/313 | 0.0 | 0.0% |
| `(z,y,x)` @ L0 | 313/313 | 0.0 | 0.0% |

Same conclusion as yours, and the 88.0 / 55.3% vs your 89.0 / 56.2% is just a different sample stride.

Two notes on the table itself:

1. **`(z,y,x)` @ L2 is 9/313 in bounds on my sample, not 0/313.** Nothing turns on it — median 0.0, none above 80 — but the row reads as a clean zero and a reviewer re-running it may not get one.
2. **Your prose and your table disagree about which row is the documented reading.** The text says "read as documented, `(z,y,x)` full-resolution", while the table's third row is `(x,y,z)` @ L0. I measured both, and both give 313/313 in bounds at median 0.0 — so the conclusion is unaffected either way, but the two descriptions should probably be made to agree before review.

Also worth recording: the file is served identically from `dl.ash2txt.org/datasets/spiral_datasets/PHercParis4/` and from the HF bucket at `spiral/PHercParis4/` — both 773,507 bytes — so the correction covers the same object under both routes.

**Cross-link:** this closes a gap I explicitly left open in #1569. There I showed the same three collections' extents are consistent with a native level-2 grid (umbilicus occupies 96.3% of L2 height vs 48.1% of L1), but I said plainly that bounds evidence "cannot fix origin/crop/index-base/centre-vs-corner" and stopped short of asserting an axis order. Sampling CT at the points is exactly the disambiguation bounds can't do — `relative_windings.json` is fully in bounds under both `(x,y,z)`@L2 and `(z,y,x)`@L0, so only the intensities separate them. Kotoro78 asked for authoritative frame declarations in #1569 and this answers the order half of that for the JSON collections; worth linking the two.

Your scoping is right that `tracks/` is left unclaimed — I didn't check those either.

---
*Verification by Claude (AI agent) on behalf of @Bullo27. Measured today against the live paths; the CT figures are from my own sample, not re-stated from the PR.*

tonclap:
Thank you — both notes were right, and both are now fixed in the description. Neither moves the conclusion, but both would have cost a reviewer time.

**1. `(z,y,x)` @ L2 is not a clean zero.** Your reading and mine differ only in the sample, and I can reproduce yours exactly: on every-6th sampling I get 10/320 in bounds; over the whole file it is **73/2173 (3.4%)**. My 0/313 came from the first 30 collections, which happen to contain none of them — a stride artifact presented as a property of the data. The table now carries an all-points column next to the sample column, so the row no longer reads as an absolute. Those points are still empty volume: median CT 0.0, none above 80.

**2. Prose vs. table — you found a real gap, not just an inconsistency.** The table was missing the reading the prose was talking about, so I measured it rather than relabelling the row. Same 313-point sample, same volume:

| reading | in bounds (sample) | in bounds (all 2173) | median CT | > 80 |
|---|---|---|---|---|
| `(x,y,z)` @ L2 | 313/313 | 2173/2173 | **89.0** | 56.2% |
| `(z,y,x)` @ L2 | 0/313 | 73/2173 (3.4%) | 0.0 | 0% |
| `(x,y,z)` @ L0 | 313/313 | 2173/2173 | 0.0 | 0% |
| `(z,y,x)` @ L0 — as documented | 313/313 | 2173/2173 | 0.0 | 0% |

Same as yours: both full-resolution readings are fully in bounds at median 0.0, and only `(x,y,z)`@L2 lands on papyrus.

That also exposed a wording problem in the patch itself, so I pushed one more commit (efbc47f): the README line said the misread points "fall outside the papyrus", which reads as *out of bounds* — the one thing they are not. It now says plainly that at full resolution the points stay inside the volume under either order, so a bounds check passes and only the intensities separate the readings. That is the reader this line has to reach.

**On #1569** — agreed, and I've linked it from the description. Your extents result is what makes the level-2 half of this credible from a second direction, and CT-at-the-points is precisely the step you said bounds could not take. Between the two, the order half of the frame declaration Kotoro78 asked for is answered for the JSON collections; `tracks/` remains unclaimed by both of us, and I'd rather leave it that way than narrow the Conventions line on a guess.

Noted too that the file is byte-identical across `dl.ash2txt.org` and the HF bucket — worth having on the record, since the correction then covers both routes.

*Measurements above are my own, re-run today against the live paths; work done with AI assistance (Claude) under my direction, as disclosed in the linked package.*

pmh47:
The readme does need an update & corrections at some point but the proposed changes are somewhat wrong (wrt coordinates etc) and for the paths/contents the data is churning fast anyway so not worthwhile right now

PR #1577, opened 2026-08-23
VC3D: allow catalog fiber atlases with local shells

### LINKED ISSUE #1530: fiber.create_atlas is structurally unavailable on catalog-streamed samples: no published lasagna manifest carries init_shell_dir, and there is no local-shell override
**In one sentence:** Building a fiber atlas on a catalog-streamed sample always fails with `Lasagna manifest is missing init_shell_dir`, because none of the published lasagna manifests carries that key and `fiber.create_atlas` offers no way to supply a shell directory yourself.

**I was trying to:** drive the fiber → atlas workflow headlessly over the agent bridge on PHerc0332 streamed from the Open Data catalog — the path #1527 just unblocked (catalog lasagna attach now works, verified cold on PHerc0125 in #1355).

**Using:** VC3D built from main `bf57b86` (Linux, `QT_QPA_PLATFORM=offscreen`, agent bridge); the live `metadata.json` and lasagna artifacts fetched today for PHerc0125 / PHerc0139 / PHerc0332 / PHerc0814 / PHerc1451.

**What happened:** the earlier steps all work — `catalog.open_sample {"resources":{"kinds":["lasagna"]}}` attaches all four channel volumes, `fiber.import` of a v1 fiber returns `{"imported": 1}` — and then:

```
fiber.create_atlas {"fiberId": "1"}
→ {"code": -32005, "data": {"detail": "Lasagna manifest is missing init_shell_dir"},
   "message": "Atlas creation failed"}
```

**What I expected or needed:** a way to reach atlas creation on catalog data — either the manifest carrying the field, or a parameter to point atlas creation at a shell directory I have locally.

**Evidence / reproduction:**

- All six lasagna manifests VC3D itself cached from the live catalog (5 samples incl. both PHerc0332 variants) have exactly the keys `base_shape_zyx, grad_mag_encode_scale, grad_mag_factor, groups, source_to_base, umbilicus_json, version` (the `las-sd1` variant swaps `umbilicus_json` for `provenance`) — none has `init_shell_dir`.
- The requirement is `core/src/Atlas.cpp:3795`, which throws exactly this error, then requires the directory to exist and contain `shell_*.tifxyz`.
- The producer *can* emit the key — `lasagna/fit_data.py:1057` writes it via `init_shell_dir_abs_path()` — but as an **absolute local path**, which is meaningless once published, so it is understandable that the published artifacts drop it. The C++ side already supports manifest-relative resolution (`core/test/test_atlas.cpp`: "Atlas manifest init_shell_dir resolves relative to lasagna manifest"), but no published artifact ships a shell alongside.

**Why it matters, and the cheap unlock:** most scrolls have no shell at all yet (cf. the context in #1220), so "publish shells everywhere" is not generally possible. An optional `initShellDir` parameter on `fiber.create_atlas` (bridge + GUI path), overriding the manifest lookup, would make atlas creation reachable on catalog-streamed projects for anyone who has a shell locally — e.g. the published `spiral/PHercParis4` `outer_shell`, or shells they fitted themselves. Happy to send that PR if you want it.

Adjacent, not duplicate: #1436 tracks the *other* publish-side gap on this surface (fibers artifacts with `creation_info: null`, which is now their sole attach blocker since #1527 fixed the marker ordering); #1379 is about the shell pack's own `scale` field, not its reachability from atlas creation.

## Details

Found while probing the newly unblocked catalog-lasagna surface headlessly. Analysis and runs done with AI assistance (Claude) under my direction; the transcript lines above are from actual bridge sessions against the live catalog.

TAUIL-Abd-Elilah:
**In one sentence:** Catalog-streamed Lasagna datasets can now create, display, and remap fiber atlases using a local shell without editing the published manifest.

**One real example:** On the unchanged public PHerc0332 catalog manifest, the issue reporter passed a local wrapped shell to `fiber.create_atlas`; it returned a complete displayed atlas, then `atlas.open` and `atlas.remap` both succeeded without `pred_dt`.

**Before:** Atlas creation stopped with `Lasagna manifest is missing init_shell_dir`. Hand-injecting that field reached creation, but display then stopped because the published dataset has no `pred_dt`.

**After this PR:** An absolute `initShellDir` can be supplied without changing the manifest. Display and remap work without `pred_dt`; snap ranking remains unavailable and rejects immediately with a clear error.

**Proof:** Independent PHerc0332 replay, including the unchanged-manifest check, negative control, complete artifact, display/remap, snap guard, and 151/151 tests: https://github.com/ScrollPrize/villa/pull/1577#issuecomment-5385808699

All code-relevant GitHub checks pass on Linux, Windows, and macOS: https://github.com/ScrollPrize/villa/actions/runs/32636827471

**Why / where this is useful:** This unblocks the catalog workflow after fitting a fiber: users can create and inspect its atlas directly against streamed scroll data.

## Details

Fixes #1530. Tested commit: `ec5cf5f3b63a77ba2ffe6e2d3ff49bcaafef2539`.

The existing manifest fallback is unchanged. An explicit shell override must be absolute and exist. Pred-snap attachments are optional only for display/remap; ranking still requires `pred_dt` and its GUI controls stay disabled when that channel is absent.

The real-scroll replay above was run independently by @Bullo27 against this exact commit.

### DIFF
--- a/volume-cartographer/apps/VC3D/CWindow.cpp
+++ b/volume-cartographer/apps/VC3D/CWindow.cpp
@@ -222,6 +222,22 @@ resolvedLasagnaForState(const CState* state)
         *state->vpkg(), state->currentVolumeId());
 }
 
+bool stateLasagnaHasPredDt(const CState* state)
+{
+    try {
+        const auto resolved = resolvedLasagnaForState(state);
+        if (!resolved || resolved->manifestPath.empty() ||
+            !std::filesystem::exists(resolved->manifestPath)) {
+            return false;
+        }
+        const auto manifest = vc::lasagna::LasagnaDatasetManifest::parseFile(
+            resolved->manifestPath);
+        return manifest.groupForChannel("pred_dt") != nullptr;
+    } catch (...) {
+        return false;
+    }
+}
+
 enum class SurfaceOverlayItemKind {
     Folder,
     Surface,
@@ -2065,7 +2081,8 @@ QDockWidget* createAtlasOverviewDock(QWidget* parent)
     auto* rankSnap = new QPushButton(QObject::tr("Rank via Fit-Service"), content);
     rankSnap->setObjectName(QStringLiteral("atlasRankSnapButton"));
     rankSnap->setToolTip(QObject::tr(
-        "Run atlas pred-snap ranking through the Lasagna fit-service job queue."));
+        "Run atlas pred-snap ranking through the Lasagna fit-service job queue. "
+        "Requires a selected Lasagna dataset with a pred_dt channel."));
     rankSnap->setEnabled(false);
     actions->addWidget(rankSnap);
 
@@ -2100,6 +2117,8 @@ QDockWidget* createAtlasFiberDock(QWidget* parent)
 
     auto* optimize = new QPushButton(QObject::tr("Optimize snap cands"), content);
     optimize->setObjectName(QStringLiteral("atlasOptimizeSnapCandidatesButton"));
+    optimize->setToolTip(QObject::tr(
+        "Requires a selected Lasagna dataset with a pred_dt channel."));
     optimize->setEnabled(false);
     layout->addWidget(optimize);
 
@@ -4925,19 +4944,9 @@ void CWindow::updateAtlasFiberDocks()
     auto* optimize = _atlasWorkspaceFiberDock->widget()->findChild<QPushButton*>(
         QStringLiteral("atlasOptimizeSnapCandidatesButton"));
     auto updateOptimizeEnabled = [&]() {
-        bool hasManifest = false;
-        if (_state && _state->vpkg()) {
-            try {
-                const auto resolved = resolvedLasagnaForState(_state);
-                hasManifest = resolved && !resolved->manifestPath.empty() &&
-                              std::filesystem::exists(resolved->manifestPath);
-            } catch (...) {
-                hasManifest = false;
-            }
-        }
         if (optimize) {
             optimize->setEnabled(_currentAtlasDir.has_value() &&
-                                 hasManifest &&
+                                 stateLasagnaHasPredDt(_state) &&
                                  LasagnaServiceManager::instance().isRunning());
         }
     };
@@ -5099,6 +5108,7 @@ void CWindow::updateAtlasSearchDocks()
     }
     const bool hasSnapshots = !snapshots.empty();
     const bool atlasUsable = _currentAtlasDir.has_value() && atlasLoadError.isEmpty();
+    const bool predSnapAvailable = stateLasagnaHasPredDt(_state);
     const QString atlasName = _currentAtlasDir
         ? QString::fromStdString(_currentAtlasName.empty()
               ? _currentAtlasDir->filename().string()
@@ -5144,7 +5154,7 @@ void CWindow::updateAtlasSearchDocks()
             continue;
         }
         if (auto* rankSnap = dock->widget()->findChild<QPushButton*>(QStringLiteral("atlasRankSnapButton"))) {
-            rankSnap->setEnabled(atlasUsable);
+            rankSnap->setEnabled(atlasUsable && predSnapAvailable);
         }
         if (auto* remap = dock->widget()->findChild<QPushButton*>(QStringLiteral("atlasRemapButton"))) {
             remap->setEnabled(atlasUsable);
@@ -5312,6 +5322,14 @@ bool CWindow::optimizeAtlasSnapCandidatesHeadless(QString* errorMessage,
     }
     const std::filesystem::path manifestPath = resolvedLasagna->manifestPath;
     const double workingToBaseScale = resolvedLasagna->workingToBaseScale;
+    try {
+        const auto manifest = vc::lasagna::LasagnaDatasetManifest::parseFile(manifestPath);
+        if (manifest.groupForChannel("pred_dt") == nullptr) {
+            return fail(tr("The selected Lasagna dataset has no pred_dt channel."));
+        }
+    } catch (const std::exception& ex) {
+        return fail(QString::fromStdString(ex.what()));
+    }
 
     auto& manager = LasagnaServiceManager::instance();
     if (manager.isExternal()) {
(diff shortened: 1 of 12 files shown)


Bullo27:
Ran the PHerc0332 catalog replay — **every claim in the description checks out on the unchanged published manifest.** Built `ec5cf5f3b` locally (GCC 13.3.0) rather than waiting on CI.

The manifest is the same one that blocked this in #1530: keys `base_shape_zyx, grad_mag_encode_scale, grad_mag_factor, groups, source_to_base, umbilicus_json, version`, sha `201c45c6e54a2ffd` **identical before and after the run** — no injection.

**1. Negative control first — the manifest fallback is intact.** Before testing the override I called `fiber.create_atlas` *without* it, and it still fails exactly as before:

```
{"code": -32005, "data": {"detail": "Lasagna manifest is missing init_shell_dir"},
 "message": "Atlas creation failed"}
```

Worth stating explicitly, since a passing positive test alone wouldn't catch an override that had accidentally become mandatory.

**2. With `initShellDir`, on that same manifest:**

```
{"atlasDir": ".../projects/atlases/fiber_1_2", "displayed": true}
```

Artifact is complete — `base_mesh/shell_probe.tifxyz`, `mappings/fibers`, `links.json`, `metadata.json`. On 08-19 I could only reach this by hand-injecting the key.

**3. The pred_dt relaxation is genuinely exercised.** I checked the published channel set rather than assuming: `groups = ['cos', 'grad_mag', 'nx', 'ny']` — **no `pred_dt`**. With that dataset selected:

```
atlas.open   -> {"opened": true}
atlas.remap  -> {"remapped": true}
```

Both of which is what #1530 hit the wall on.

**4. Snap ranking is still gated, and the synchronous part holds:**

```
atlas.optimize_snap_candidates
-> {"code": -32005,
    "data": {"detail": "The selected Lasagna dataset has no pred_dt channel."},
    "message": "Snap-candidate ranking failed to start"}
returned in 0.0s
```

Immediate, names the missing channel, and never reaches a worker — matching "rejects synchronously instead of failing later in a worker".

**5. `ctest` on the branch: 151/151 passed** (my build's usual count; nothing regressed).

So from my side the catalog replay you listed as pending is done and green. Happy to re-run against a later head if you push more — the harness is scripted now, so it's cheap.

---
*Verification by Claude (AI agent) on behalf of @Bullo27. Measured on this machine against `ec5cf5f3b`; the no-injection and negative-control assertions are enforced in the probe rather than checked after the fact.*

TAUIL-Abd-Elilah:
Thanks, this closes the functional checks. Could you add a screenshot of the before error and the successful displayed atlas? CONTRIBUTING asks for real-scroll before/after visual proof.

Bullo27:
Happy to help, but I'd push back gently on the framing: that checkbox is a *personal* attestation by the PR author, so captures produced on my machine can't be what you tick it against. Anything I supply is independent third-party verification sitting alongside your evidence, not a substitute for it.

The good news is you can very likely produce it yourself. **The atlas path needs no CUDA and no JIT** — that was only the fit in #1572. My replay ran with an empty environment:

```python
b = Bridge("atlas1577", env_extra={})   # no CUDA_HOME, no nvcc on PATH
```

So the Ninja/CUDA gap that blocked you on #1572 doesn't apply here. If you can build VC3D on your host you can run this end to end.

**Recipe**, all offscreen, no display needed:

```
QT_QPA_PLATFORM=offscreen VC3D --agent-bridge --agent-bridge-name <name>

catalog.open_sample  {"sample": "PHerc0332", "resources": {"kinds": ["lasagna"]}}
volume.select        {"volumeId": "<20251211183505-… base volume>"}
fiber.list           -> pick a fiberId
fiber.create_atlas   {"fiberId": "<id>"}                      # BEFORE
fiber.create_atlas   {"fiberId": "<id>", "initShellDir": DIR} # AFTER
screenshot.capture   {"target": "seg yz", "filePath": "…png"}
```

Two things that will save you time:

- `screenshot.capture` resolves `target` through `surfName()`, so the only valid values are `segmentation`, `seg xz`, `seg yz`. In the default offscreen layout `seg xz` returns `-32009 target widget is not currently visible`; `seg yz` works.
- `DIR` must contain a **wrapped** shell as `shell_*.tifxyz` — a flat patch is rejected with `atlas init shell is not explicitly wrapped: first and last columns differ`, and the fiber's seed has to sit on real material. I recovered both from the artifact my 08-19 run left at `projects/atlases/fiber_1` (`base_mesh/shell_probe.tifxyz`, and the fiber under `mappings/fibers/`), which is worth knowing if you hit those gates.

On the "before" specifically: it isn't really a visual state. Without the override the call fails at the RPC boundary with `-32005 Lasagna manifest is missing init_shell_dir` and nothing is drawn, so a screenshot of it is an empty viewport — the JSON error is the stronger artifact, and it's already in my previous comment. The meaningful capture is the *after*: the atlas displayed in the pane.

I do have that capture locally. I can't attach images through the API path I post with, and I'd rather not push evidence for your PR into my own repos unasked — so if a maintainer wants it hosted, say the word and I'll sort it. Otherwise your own run is the better artifact for the checkbox.

---
*Claude (AI agent) on behalf of @Bullo27.*

hendrikschilling:
Currently the team is not actively using the atlas tab - and the shell based lasagna tracing might be removed at some point in favor of the newer spiral and fiber based efforts.
That does not mean I won't merge this work, but you need to give me some outlook what you want to achieve with this that will help the unwrapping, or comments from others that use this.
E.g. from the progress price guidelines:
`Resolve outstanding bugs in tools that people are using, and that you are using yourself, evidenced by before/after screenshots, logs, etc.`
Currently this looks like a random bug not some directed effort towards some goal. Prove me wrong!

TAUIL-Abd-Elilah:
Thanks — this is useful context. #1577 came from the real PHerc0332 catalog replay in #1530 and was meant to complete the downstream create/display/remap stage after #1572's catalog-streamed fit. The current evidence shows that workflow succeeds, but I do not yet have evidence that the Atlas tab is actively used or that this change improves unwrapping, and I do not want to overstate that.

If this path is being retired, which concrete handoff into the current Spiral/fiber workflow would be worth targeting? Otherwise I am happy not to add maintenance burden.

TAUIL-Abd-Elilah:
Closing this to free a contributor slot after the maintainer feedback above. The PHerc0332 replay and tests are real and remain in the history, but I do not currently have evidence that the Atlas tab is actively used or that this shell-based path advances current unwrapping; keeping it open would add maintenance burden without meeting that bar.

I am redirecting the slot to #1588, an actively encountered stale-runtime-image defect with direct real-scroll before/after evidence. #1572 remains open for the active catalog-streamed fitting path.

Issue #1547, opened 2026-08-20
[data QA] PHerc0139 w045/w046 duplicate substantially the same traced surface


nerln:
I found one cross-segment duplicate in the published tifxyz catalog:

- PHerc0139/20260126000000-w045_2026012619
- PHerc0139/20260325000000-w046_20260325

After a four-cell grid alignment, their valid masks are identical and 89,503 of
109,823 vertices (81.4975%) are bit-identical in x, y and z float32 values. The
remaining 20,320 vertices form one connected region. Their published ink maps,
aligned independently, correlate at r=0.810 over 10.3 Mpx; adjacent controls are
+0.008 and -0.051.

A geometry scan of the 19 August public catalog covered 188 surfaces and 5,142
ordered pairs. This is the only pair classified as duplicate by the declared
rule in that snapshot: coincident fraction 0.818, versus 0.127 for the largest
non-duplicate anywhere else (0.006 for the next consecutive pair on PHerc0139).

The nominal w046→w047 gap measures like a two-sheet step. As supporting evidence,
the published m7 surface-prediction model has a distinct interior response on
85/86 rays in the released manifest's 40–55 voxel subset, versus 2/28 for a
one-sheet control and 94/95 and 102/102 for two-sheet controls. I treat that as
model evidence, not physical ground truth; this was not preregistered.

Reproduction, manifests, tests and machine-readable results:
https://github.com/nerln/vesuvius-ladder

The compact duplicate check downloads about 8 MB and verifies every asset by
size and SHA-256. The end-to-end ray command validates 361 Zarr chunks
(112,892,827 bytes); a clean-cache run regenerated the committed JSON
byte-for-byte (SHA-256 13733677d4901e9c33277e7e8ab2c84aba1e4af4de6a8efd0eefc799fee470ea).

Suggested actions:

1. verify and, if appropriate, quarantine or annotate w046 and its derived
   renders/ink outputs;
2. verify whether a distinct winding 46 needs tracing;
3. consider recording coincidence magnitude in vc_seg_add_overlap, which
   currently reports this pair among ordinary overlaps without distinguishing it.

Known limits: the 29.9 cm² repeated-area figure in the report is an estimate
(repeated-cell fraction × reported w046 area), not direct area integration. The
whole-corpus snapshot does not vendor all 188 inputs; the two PHerc0139 headline
experiments are asset-pinned.

Agentic-use disclosure: I used OpenAI Codex agentically for investigation,
[...]

PR #1578, opened 2026-08-23
Add quantitative pair reports to vc_seg_add_overlap

TAUIL-Abd-Elilah:
**In one sentence:** Adds an optional JSON report showing how much of each source surface lies within the existing overlap tolerance of each target.

**One real example:** On the pinned original PHerc0139 w044-w047 TIFXYZ files, w045->w046 measured 83.51% and w046->w045 measured 83.47%; the largest preregistered adjacent control was 2.31%.

The earlier #1547 figures use nearest-vertex distance; this report uses `vc_seg_add_overlap`'s existing point-to-triangle test, which also increases control coverage, so the percentages are not directly comparable.

**Before:** `vc_seg_add_overlap` only wrote a boolean-style overlap list. All six pairs in this four-window set were listed, so the unusual w045/w046 pair looked like an ordinary small contact.

**After this PR:** `--report-json PATH` writes matched and queried point counts plus directed coverage fractions while leaving the default output and `overlapping.json` behavior unchanged.

**Proof:** The thresholds were [posted before the run](https://github.com/ScrollPrize/villa/issues/1547#issuecomment-5386343015). The [immutable replay](https://github.com/TAUIL-Abd-Elilah/vesuvius-repro/tree/a2565599772605901ec515d75ba0cb7dd918f335/evidence/villa-1547) passed every gate: 36.07x positive/control separation, byte-identical reports with 1 and 4 workers, and byte-identical legacy output with and without the report.

Directed PHerc0139 source-point coverage: the w045/w046 pair is about 83.5% in both directions; adjacent controls are at most 2.31%

All code-relevant GitHub checks pass: https://github.com/ScrollPrize/villa/actions/runs/32646992978

**Why / where this is useful:** It lets surface reviewers distinguish broad near-duplication from incidental contact before accepting data for downstream training or analysis.

## Details

- Uses the existing 2-unit point-to-target-triangle test and stride-1 target index.
- Counts every queried source point only when `--report-json` is requested; the default fast path is unchanged.
- Emits stable relative paths and deterministic output across worker counts.
- Omits zero-hit rows and states that zero and self pairs are excluded.
- Rejects output paths that alias source data and deduplicates canonical directory aliases.
- Adds focused metric and CLI tests for real output behavior, synthetic geometry, worker determinism, legacy compatibility, and invalid inputs.

Addresses #1547.

### DIFF
--- a/volume-cartographer/apps/src/vc_seg_add_overlap.cpp
+++ b/volume-cartographer/apps/src/vc_seg_add_overlap.cpp
@@ -1,6 +1,8 @@
 #include "vc/core/util/QuadSurface.hpp"
 #include "vc/core/util/SurfacePatchIndex.hpp"
 
+#include "vc_seg_add_overlap_metrics.hpp"
+
 #include <algorithm>
 #include <atomic>
 #include <chrono>
@@ -8,6 +10,7 @@
 #include <cstdlib>
 #include <exception>
 #include <filesystem>
+#include <fstream>
 #include <iomanip>
 #include <iostream>
 #include <limits>
@@ -32,6 +35,7 @@ constexpr float kOverlapTolerance = 2.0f;
 struct TargetInfo {
     std::string key;
     std::string id;
+    std::string relativePath;
     fs::path path;
 };
 
@@ -39,7 +43,10 @@ struct SourceScanResult {
     fs::path sourcePath;
     std::string sourceKey;
     std::string sourceId;
+    std::string sourceRelativePath;
     std::vector<std::pair<std::string, std::string>> hits;
+    std::map<std::string, size_t> matchedPointsByTarget;
+    size_t validPoints = 0;
     size_t queriedPoints = 0;
 };
 
@@ -48,6 +55,7 @@ struct Config {
     fs::path sourceRoot;
     size_t requestedWorkers = 0;
     size_t pointStride = 1;
+    fs::path reportPath;
     bool help = false;
 };
 
@@ -62,6 +70,60 @@ std::string canonical_key(const fs::path& path)
         p = path;
     }
     return p.lexically_normal().generic_string();
+}
+
+std::string relative_report_path(const fs::path& path, const fs::path& root)
+{
+    std::error_code ec;
+    fs::path relative = fs::relative(path, root, ec);
+    if (ec) {
+        return path.filename().generic_string();
+    }
+    const std::string value = relative.lexically_normal().generic_string();
+    return value.empty() ? "." : value;
+}
+
+void validate_report_path(const fs::path& reportPath,
+                          const std::vector<fs::path>& targetDirs,
+                          const std::vector<fs::path>& sourceDirs)
+{
+    if (reportPath.empty()) {
+        return;
+    }
+
+    fs::path parent = reportPath.parent_path();
+    if (parent.empty()) {
+        parent = fs::current_path();
+    }
+    std::error_code ec;
+    if (!fs::is_directory(parent, ec) || ec) {
+        throw std::runtime_error(
+            "Report parent is not a directory: " + parent.string());
+    }
+
+    const std::string reportKey = canonical_key(reportPath);
+    auto rejectAlias = [&](const fs::path& dir) {
+        for (const char* name : {
+                 "meta.json", "x.tif", "y.tif", "z.tif", "overlapping.json"}) {
+            const fs::path dataPath = dir / name;
+            std::error_code equivalentEc;
+            const bool equivalent = fs::exists(reportPath, equivalentEc) &&
+                !equivalentEc &&
+                fs::equivalent(reportPath, dataPath, equivalentEc) &&
+                !equivalentEc;
+            if (canonical_key(dataPath) == reportKey || equivalent) {
+                throw std::runtime_error(
+                    "Report path would overwrite surface data: " +
+                    reportPath.string());
+            }
+        }
+    };
+    for (const fs::path& dir : targetDirs) {
+        rejectAlias(dir);
+    }
+    for (const fs::path& dir : sourceDirs) {
+        rejectAlias(dir);
+    }
 }
 
 bool is_tifxyz_dir(const fs::path& dir)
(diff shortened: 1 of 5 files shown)


Bullo27:
Independently reproduced the pre-registered replay on `ee5aa5bc8`. **Every figure matches to the digit, and all six gates pass.** I also chased down why these numbers differ from the ones in #1547, since that's the first thing a reviewer will trip on.

**Inputs are the same bytes you pinned.** Re-fetched all 16 `tifxyz_original` files for w044–w047 from the public bucket: **16/16 SHA-256 match** nerln's `data_manifest.json`, so the live catalog is unchanged since the 19 Aug snapshot.

**Pre-registered gates** (from your comment posted *before* the run):

| | mine | yours | gate | |
|---|---|---|---|---|
| w045→w046 | 83.5126% | 83.5126% | >75% | PASS |
| w046→w045 | 83.4661% | 83.4661% | >75% | PASS |
| largest adjacent control | 2.3137% | 2.3137% | <5% | PASS |
| separation | 36.07× | 36.07× | >15× | PASS |

Determinism, on my own runs rather than re-stated:

```
report, 1 worker vs 4 workers      -> byte-identical (sha 6a5b6ef621762945)
overlapping.json, with vs without --report-json -> identical for all four windows
```

`ctest`: **153/153 passed**, including the two new tests (`test_seg_add_overlap_metrics`, `test_seg_add_overlap_cli`).

Full directed set for reference — 12 pairs, 438,181 queried points:

```
w045 -> w046   83.5126%   91716/109823      w044 -> w045    2.1704%
w046 -> w045   83.4661%   91665/109823      w045 -> w044    2.3137%
w046 -> w044    1.6517%                     w046 -> w047    0.4316%
w047 -> w045    0.4942%                     w047 -> w046    0.5108%
w045 -> w047    0.4152%                     w044 -> w046    1.5753%
w047 -> w044    0.0018%                     w044 -> w047    0.0009%
```

**Reconciling with #1547 — worth stating in the PR, because the numbers look contradictory otherwise.** nerln's pinned `within_2` expectations are w045→w046 **81.9819%**, w044→w045 **0.0263%**, w046→w047 **0.0100%**. Your tool reports 83.5126%, 2.1704%, 0.4316% for the same pairs and the same tolerance.

I measured the cause rather than guessing: recomputing with **nearest-vertex** distance instead of point-to-triangle reproduces nerln's three numbers exactly —

```
pair          vertex(mine)   triangle(#1578)   nerln within_2
w045->w046      81.9819%         83.5126%         81.9819%
w044->w045       0.0263%          2.1704%          0.0263%
w046->w047       0.0100%          0.4316%          0.0100%
```

So both are right and they measure different things, exactly as your metric block says (`euclidean point-to-target-triangle`). Since point-to-triangle ≤ point-to-nearest-vertex, triangle coverage must be the larger of the two, and it is.

The part I'd flag: **the triangle metric lifts the controls far more than the positives** — the duplicate pair moves +1.5 points (81.98→83.51), while the adjacent control moves 82× (0.026→2.17). Separation therefore falls from ~3,100× under nearest-vertex to 36× here. Still decisive, but anyone carrying the #1547 intuition over will expect a thousand-fold gap and find a thirty-fold one. A line in the report or the PR text naming the metric difference would save that confusion.

---
*Verification by Claude (AI agent) on behalf of @Bullo27. All figures measured on this machine against `ee5aa5bc8`; the vertex-vs-triangle comparison is my own computation, not restated from either thread.*

nerln:
Thanks. The vertex-vs-triangle reconciliation is right, and confirmed in source: ladder's within_2 is nearest-vertex distance (cKDTree.query in check_duplicate.py), so point-to-triangle reading higher everywhere is expected. Naming the metric in the #1578 report text sounds right.

@TAUIL-Abd-Elilah: no overlap on our side. The ladder pins at 7ae4dff are the right ones. I'll take a look at #1578. If a pinned-snapshot manifest helps your CI, ladder's data_manifest.json is stable.

nerln:
Reviewed the diff. This addresses #1547 as reported, and the parts I checked are sound.

What I verified by reading the source:

- **The metric names its own definition in the output.** `directed_source_point_surface_coverage`, defined in the report as `matched_source_points / queried_source_points`, and the computation matches that string exactly — `queried_source_points` is the sampled count the fraction divides by. A number that carries its own definition is worth a lot here, because the pair reports will be read months later by someone who did not run them.
- **Sampling is a fixed stride and the stride is recorded** in `parameters.point_stride`, so a report is exactly reproducible from its own contents rather than from the command that produced it.
- **The invariants throw rather than clamp**: `matched > queried` and `queried > valid` both raise, and a zero denominator yields 0.0 instead of a NaN that would propagate silently into a summary.
- **Directory symlinks within a collection are de-duplicated by canonical path**, which is what stops one physical segment being counted twice and coverage exceeding 1.
- **`--report-json` refuses to write over surface data** (`meta.json`, `x/y/z.tif`, `overlapping.json`), checked both by canonical key and by `fs::equivalent`.

One suggestion, not a blocker: with a stride above 1 the coverage fraction is an estimate of the whole-surface figure. The report already carries everything a reader needs to see that (`valid_source_points`, `queried_source_points`, the stride), so the remaining gap is only that nothing states how stable the estimate is across strides. A single line in the PR description — the same pair at stride 1, 2 and 4 — would let a reviewer accept the default without having to reason about it.

I did not build or run this; the above is a source read of the diff at its current head.

TAUIL-Abd-Elilah:
@nerln Thanks 
I ran the requested sensitivity check on the same pinned `w045 -> w046` pair at `ee5aa5bc8`:

| point stride | matched / queried | coverage |
|---:|---:|---:|
| 1 | 91,716 / 109,823 | 83.5126% |
| 2 | 45,861 / 54,912 | 83.5173% |
| 4 | 22,940 / 27,456 | 83.5519% |

The full spread is **0.0393 percentage points**. This pair is therefore stable across the tested sampling strides; stride 1 remains the default.

nerln:
Recomputed all three: 91,716/109,823 = 83.5126%, 45,861/54,912 = 83.5173%, 22,940/27,456 = 83.5519%, spread 0.0393 pp. That settles the stride question — the coverage figure is a property of the pair, not of the sampling, and stride 1 as default costs nothing.

From my side as the author of #1547 this is complete: the pair I reported is measured, the metric carries its own definition in the output, the stride is recorded in the report, and the invariants raise rather than truncate. Nothing further is outstanding for me.

- [x] I personally recomputed the three coverage figures above.

hendrikschilling:
Looks good, update and resolve the conflicts then we can merge it.

TAUIL-Abd-Elilah:
Rebuilt on current main — this had gone conflicting.

The only conflict was `core/test/CMakeLists.txt`: #1729 added an `if(TARGET vc_render_tifxyz)` block between this branch's two insertion points. Both sets of additions are kept, upstream's logging test and this branch's two test registrations, and I checked no upstream line was dropped in the resolution. The other four files were untouched by main since May, so the PR diff is still the same five files.

Still happy to split this if the size is what's holding it up — the metrics header plus its unit test stands alone without the CLI contract test.

PR #1579, opened 2026-08-23
Show a coordinate frame gizmo at the top right of each viewport.

spelufo:
image

New gizmo to show world coordinate frame. Matches how the coords printed on the statsbar increase. Derived directly from the plane's _M transform.

The scale bar also wasn't showing for me and getting rid of that dpr code fixed it.

I changed the negative handle of the plane slicing overlay controller to draw in black, which is ugly, but I wanted to make it asymetrical, to see what's going on. Probably shouldn't belong in this PR.

### DIFF
--- a/volume-cartographer/apps/VC3D/CWindow.cpp
+++ b/volume-cartographer/apps/VC3D/CWindow.cpp
@@ -8231,6 +8231,17 @@ void CWindow::CreateWidgets(void)
         chkAxisOverlays->setChecked(showOverlays);
         connect(chkAxisOverlays, &QCheckBox::toggled, this, &CWindow::onAxisOverlayVisibilityToggled);
     }
+    // Deliberately not persisted: the coordinate frame gizmo is on by default
+    // every session, matching the checkbox's designer state.
+    if (auto* chkCoordinateFrames = ui.chkCoordinateFrames) {
+        connect(chkCoordinateFrames, &QCheckBox::toggled, this, [](bool show) {
+            for (auto* manager : ViewerManager::allManagers()) {
+                if (manager) {
+                    manager->setShowCoordinateFrames(show);
+                }
+            }
+        });
+    }
     if (auto* btnResetRot = ui.btnResetAxisRotations) {
         connect(btnResetRot, &QPushButton::clicked, this, &CWindow::onResetAxisAlignedRotations);
     }
--- a/volume-cartographer/apps/VC3D/VCMain.ui
+++ b/volume-cartographer/apps/VC3D/VCMain.ui
@@ -1052,6 +1052,19 @@
        </property>
       </widget>
      </item>
+     <item row="11" column="0" colspan="4">
+      <widget class="QCheckBox" name="chkCoordinateFrames">
+       <property name="text">
+        <string>Show coordinate frames</string>
+       </property>
+       <property name="checked">
+        <bool>true</bool>
+       </property>
+       <property name="toolTip">
+        <string>Draw the X/Y/Z reference frame in the corner of the slice views</string>
+       </property>
+      </widget>
+     </item>
     </layout>
    </widget>
    </widget>
--- a/volume-cartographer/apps/VC3D/ViewerManager.cpp
+++ b/volume-cartographer/apps/VC3D/ViewerManager.cpp
@@ -370,6 +370,7 @@ VolumeViewerBase* ViewerManager::initializeChunkedViewer(CChunkedVolumeViewer* c
         bool showNormals = settings.value(viewer::SHOW_SURFACE_NORMALS, viewer::SHOW_SURFACE_NORMALS_DEFAULT).toBool();
         baseViewer->setShowSurfaceNormals(showNormals);
     }
+    chunkedViewer->setShowCoordinateFrame(_showCoordinateFrames);
 
     {
         using namespace vc3d::settings;
@@ -780,6 +781,16 @@ void ViewerManager::setShowSurfaceNormals(bool show)
     forEachBaseViewer([show](VolumeViewerBase* viewer) {
         if (viewer) {
             viewer->setShowSurfaceNormals(show);
+        }
+    });
+}
+
+void ViewerManager::setShowCoordinateFrames(bool show)
+{
+    _showCoordinateFrames = show;
+    forEachBaseViewer([show](VolumeViewerBase* viewer) {
+        if (auto* chunkedViewer = dynamic_cast<CChunkedVolumeViewer*>(viewer)) {
+            chunkedViewer->setShowCoordinateFrame(show);
         }
     });
 }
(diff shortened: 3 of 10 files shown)


Bullo27:
The scale-bar line in your description is doing more work than it looks like. **That `dpr` code was a real HiDPI bug, and removing it is the correct fix — not a simplification that might regress high-DPI users**, which is what a reviewer seeing `devicePixelRatioF()` deleted would reasonably suspect. I reproduced it headlessly both ways.

**Mechanism.** On `main`, `CVolumeViewerView::drawForeground` calls `p->resetTransform()` (line 87), which puts the painter in *logical* viewport pixels — Qt applies the device pixel ratio itself at rasterisation. The next lines then apply it a second time by hand:

```cpp
int bottom = static_cast<int>(vpH * dpr) - M;      // line 93
p->drawLine(M, bottom, M + _cachedBarPx, bottom);  // QPen(Qt::red, 2)
```

At dpr = 2 that puts `bottom` at roughly twice the viewport height, so the bar is drawn well below the visible area. `_cachedBarPx` (`barUm * m11 * dpr`) and `setPointSizeF(12 * dpr)` are double-applied the same way, so even when visible it would be twice too long in twice-sized type. Note the `dpr` factors cancel in `barUm` itself, so only the drawn geometry is wrong, not the chosen length — which is why this reads as "sometimes just missing" rather than "wrong number".

**Measured, offscreen, same build, counting red pixels in the pane:**

| | `QT_SCALE_FACTOR=1` | `QT_SCALE_FACTOR=2` |
|---|---|---|
| `main` (`8b59dde58`) | bar present (354 red px) | **absent — 0 red px anywhere** |
| this PR (`71f0dfeb6`) | bar present | **bar present** |

At scale 2 on `main` the pane renders fine and the status text visibly doubles, so the scale factor is in effect and the render is healthy — the bar is simply drawn off-screen. That matches "wasn't showing for me" exactly.

**The gizmo renders.** In a plane pane (`seg yz`) it draws in the top-right at both scale factors, alongside the scale bar:

- scale 1 — small axis cross top-right, `1 cm` bar bottom-left
- scale 2 — same, both scaled up

It does *not* appear in the `segmentation` surface pane, which I assume is intended given it's derived from the plane's `_M`. Flagging only in case a frameless surface view was meant to show one too.

Two caveats on my side, so nothing here is over-read: my offscreen window is a fixed *device* size, so raising the scale factor shrinks the logical viewport and changes each element's relative size and the chosen bar length (`200 µm` → `100 µm`). I'm therefore only claiming **presence vs absence**, not layout proportions — those artefacts are my harness, not your change. And I haven't checked the gizmo's axis *directions* against the statusbar coordinates; I only confirmed it draws.

**On the part you flagged yourself** — the plane-slicing change is one line, still in the diff:

```
volume-cartographer/apps/VC3D/overlays/PlaneSlicingOverlayController.cpp
+        handleStyle.brushColor = Qt::NoBrush;
```

Since you said it probably shouldn't belong, dropping it would make this a clean gizmo + scale-bar-fix PR. Might also be worth promoting the scale-bar fix in the title or description — it's an independently useful fix that's currently buried in an aside, and it'd otherwise land unmentioned in the changelog.

---
*Verification by Claude (AI agent) on behalf of @Bullo27. Measured on this machine against `main` `8b59dde58` and PR head `71f0dfeb6`.*

Bullo27:
Heads-up: **#1566 just merged as `07656e634`, and it lands on top of two of the files this PR touches.** Worth rebasing before going further.

```
git diff --stat 8b59dde58 07656e634 -- <files this PR modifies>
  CWindow.cpp                                |  32 +-
  volume_viewers/CChunkedVolumeViewer.cpp    | 743 ++++++++++++++++++---
```

`CChunkedVolumeViewer.cpp` in particular was substantially rewritten there, and this PR adds to it, so that's where a conflict is most likely.

Two things that should reduce the worry, both checked rather than assumed:

**The scale-bar fix is still needed.** #1566 did *not* touch `CVolumeViewerView.cpp`, so the double-applied `dpr` at line 54 is untouched on `main` at `07656e634` — the HiDPI bug I measured earlier survives the merge, and this PR is still the thing that fixes it.

**The two "gizmos" aren't duplicates**, which was my first worry when I saw the filename. #1566's `CameraGizmoWidget` is an interactive camera pad for the volumetric composite mode — azimuth dial, elevation and perspective gauges, dragged with the mouse, *centered* in the viewport. Yours is a passive world-axis indicator derived from the plane's `_M`, drawn *top-right*. Different jobs, different positions, no functional overlap that I can see.

The only thing I'd raise is naming: once both land, the codebase has `CameraGizmoWidget` and a "coordinate frame gizmo" meaning quite different things in the same viewer. Might be worth a more specific name here — an axis/orientation indicator — purely to keep them distinguishable for whoever reads this next.

I also checked whether the newly merged rendering code repeats the dpr mistake, and it doesn't — neither `CameraGizmoWidget` nor `VolumetricCompositor` calls `devicePixelRatio`, so line 54 remains the only live instance.

---
*Claude (AI agent) on behalf of @Bullo27. Checked against `main` at `07656e634`.*

Bullo27:
Correction to my previous comment: **I said this needed a rebase and that a conflict was likely in `CChunkedVolumeViewer.cpp`. That was wrong — it merges cleanly.** I inferred conflict risk from file overlap instead of running the merge, which I should have done before saying it.

Measured now against `main` at `de3c24943`, which includes both #1566 and #1459 (merged since):

- `git merge-tree --write-tree de3c24943 <this branch>` → clean, rc 0. Same against `07656e634`.
- GitHub agrees — the API reports `mergeable: true`. (`mergeable_state: unstable` is just the `Vercel` context, which fails on every outside contributor's PR in this repo.)

Two smaller corrections while I'm here: this branch's actual merge base is **`97621bcf4`** ("Fiber multiwrap unwrap fix"), not `8b59dde58` — GitHub's `base.sha` is the base *branch tip*, not the merge base — and the true diff is 11 files, +253/−56.

**What still stands:**

- **The scale-bar fix is still needed.** `CVolumeViewerView.cpp` has now survived *two* merges untouched — `git log 8b59dde58..de3c24943 -- <that file>` is empty, so neither #1566 nor #1459 went near it. The double-applied `dpr` is still live on current main at lines 54, 72, 74, 76 and 93, and this PR is still what fixes it.
- **#1566 still rewrote the file you add to.** `CChunkedVolumeViewer.cpp` took +654/−89 there; this PR adds +31. A clean textual merge isn't the same as still being correct on top of a rewrite that size, so it's worth a read even though git is happy — that part of my earlier comment I'd still act on, just not for the reason I gave.
- #1459 landed a few hours ago and also touches `CWindow.cpp`, which you modify, but the hunks sit ~700 lines apart and it doesn't change the clean result.

For reference, I rebuilt current main: suite green at **154/154** on `de3c24943` (up from 152 — #1459 adds `fiber_network_layout` and `fiber_map_staleness`, both passing).

---
*Claude (AI agent) on behalf of @Bullo27. Measured on `de3c24943`.*

pmh47:
@spelufo could you rebase this, and make the big black circle a bit less big and black? just a blur shadow on the arrows would be enough without covering too many voxels

spelufo:
How's this? I removed the circle.
image

PR #1580, opened 2026-08-24
ink_detection: report a prepared input whose scale does not match its recipe

nerln:
**In one sentence:** ink inference can now tell whether the input it was handed was prepared at a scale the recipe trains at, using metadata `prepare_9um_isotropic_input` already writes.

**One real example:** Starting from a 2.399 µm rendered surface volume for PHerc0139 `w035`, I prepared the model input with `--level 0` instead of the documented `--level 2`, ran the shipped `aligned21_hybrid_3d2d` recipe, and scored the output against the published ink labels. The input was four times finer in-plane than the scale the recipe trains at, **and no stage of the pipeline said so**. Holding the depth orientation correct and changing only `--level`, that mistake costs **AUC 0.637 / 0.641 against 0.949 / 0.955** (two seeds).

**Before:** nothing downstream reads the provenance. `prepare_9um_isotropic_input` records `format`, `source`, `source_level`, `source_shape_zyx`, `source_z_slice` and `z_pool` on its output group, but across the repository `source_level` appears only in that writer, in the test asserting it was written, and in two unrelated files. Inference never looks at any of it, so an input at the wrong in-plane scale is indistinguishable from a correct one at run time.

**After this PR:** inference resolves the in-plane micrometres of the recorded source at the recorded level, compares them with the scales the recipe trains at, and reports the numbers on a mismatch:

```
w035_9um.zarr: input is 2.399 um in-plane; this recipe trains at 9.596 or 9.362 um
(nearest 9.362, factor 0.256). Prepared from ink_ctrl2/w035_sv.zarr at level 0.
```

**Proof.** Both inputs were produced by `prepare_9um_isotropic_input` from the same source volume and window, differing only in `--level`:

| input | in-plane | check |
|---|---|---|
| `--level 2` | 9.596 µm | silent |
| `--level 0` | 2.399 µm | reports the mismatch |
| published surface volume (not produced by this preparer) | — | silent, no recipe tag |

The silence on the first is a passing check, not a skipped one.

**What the mistake costs.** The scale fault in this incident was accompanied by a second one — the render traversed the surface normal in the opposite order to the labels — so the two were varied independently. Two seeds, same window and labels, AUC over the whole crop:

| | depth order as rendered | depth order reversed |
|---|---|---|
| `--level 0` (2.399 µm) | 0.555 / 0.538 | **0.637 / 0.641** |
| `--level 2` (9.596 µm) | 0.511 / 0.530 | **0.949 / 0.955** |

Published surface volume, for reference: 0.979. AUC is computed exactly, as the rank statistic over every pixel of the window against the binary published labels; a threshold-sweep approximation gives figures around 0.01 lower on the same rasters, so the method matters when reproducing these. Restricting the same computation to the annotated region moves every figure in the same direction and does not change the comparison — 0.645 / 0.671 against 0.968 / 0.976 for the two reversed-order cells.

With the depth order correct, changing only `--level` costs **0.31 of AUC**, and the two seeds agree to the third digit in both cells.

Neither fault explains the failure alone. Correcting the depth order alone moves 0.555 to 0.637; correcting the scale alone moves 0.555 to 0.511, which is noise between two values at chance. Only both together reach 0.949. That interaction is why the original investigation could not diagnose either fault from results alone: fixing one and re-running still looks like failure.

**What to look at:** the check reads only metadata that was already on disk during the incident. It adds no new keys, defines no new format, requires no change to how inputs are produced, and never reads an array — from the source store it reads the metadata document only.

Regression, whole `ink_detection` suite, same environment and command, measured on this branch and on the commit it is rebased onto:

| | failed | passed | skipped | collection errors |
|---|---|---|---|---|
| without this PR | 2 | 20 | 1 | 16 |
| with this PR | 2 | 34 | 1 | 16 |

The fourteen additional passes are exactly this PR's tests. The two failures are the same two before and after — this PR neither fixes nor touches them — and the 16 collection errors are optional dependencies missing in the test environment, identical in both runs.

**Why / where this is useful:** anyone running the shipped ink recipe on their own renders. The failure it catches is silent by construction — the model returns a confident, plausible, wrong map, and a negative result is indistinguishable from a misconfigured one.

### Motivation

A pre-registered ink protocol on PHerc. MAN5 — criterion fixed before any data, thirteen jobs, displaced negative controls, and a blind read by a human — returned a clean negative and was about to be reported as a null. The pre-registration had also prescribed a positive control on a segment with published labels, and that step had not been run. Running it returned a clean negative there too, on documented letters.

Working out why took longer than it should have, because there was more than one fault and they hid each other, as the table above shows. Correcting it then introduced a third of the same family: the corrected input was made by resampling the existing one in place while keeping its attributes, so the file went on describing data it no longer contained — and this check reports it, correctly, for that reason.

Every defence in that protocol sits downstream of all three, and each behaved exactly as it would have in front of a real null. This PR addresses the one of the three for which the repository does not already have an answer. The depth convention is documented — `--flip-normals`, and `--direction both` when a segment's orientation is unknown — and we had not read it.

### How the scale is resolved

`source_level` alone is not a scale, so the check consults the store named in `source`:

1. **OME metadata first.** A rendered surface volume records micrometre axes and a scale per dataset, so the level the input was built from names its own in-plane sampling directly — `9.596` at level 2, `2.399` at level 0, for a 2.399 µm render.
2. **Otherwise a recorded voxel size**, read in the order `Volume::normalize()` uses: a numeric `voxelsize`, then the alternative spellings (`voxel_size_um`, `voxelSizeUm`, `pixel_size_um`, `pixelSizeUm`, `resolution_um`), then the beamline `samplePixelSize`, which is recorded in millimetres. That figure is then scaled by the pyramid level.

Only metadata is read, never an array.

**The recipe trains at two scales, not one, so the check accepts either.** `aligned21_fixed_scroll_prior.json` — described in the docs as the readable contract behind the recipe's `datasets` block — declares 29 representations in two source families: 24 `public_2p4_level2_zmean4` (2.399 µm renders pooled at level 2, so 9.596 µm) and 5 `native_9p362_level0` (native 9.362 µm renders used at level 0). The tutorial states the same split. `0139:w035` appears in both, the same physical segment at two scales. The check therefore accepts an input within 2% of either figure, and reports which one it was nearest to. The two acceptance windows overlap, and the nearest wrong pyramid level is 50% away, so nothing legitimate falls between them and nothing wrong falls inside.

If the source cannot be opened, or records no scale, that outcome is kept **distinct from a scale error**: the check falls back to comparing the recipe tag with the recorded level — which needs nothing but the input itself — and says in the message that the absolute scale was not checked, naming the reason, and noting that a native 9.362 µm render prepared at level 0 is legitimate and would be reported by that fallback.

### Default behaviour

**The check reports by default**; `--strict-input-scale` turns the same finding into a refusal.

Reporting is the right default for a check that lives in someone else's inference path. A refusal's blast radius falls on users whose inputs were prepared by another route, or whose recorded `source` has since moved — and a message carrying the actual numbers is enough to act on. The incident above would have been stopped just as effectively by the warning as by a refusal, because the numbers are in it. If maintainers prefer refusal as the default, inverting the flag is a one-line change.

In a folder run, a refusal skips that segment and records it — `LOGGER.warning` plus `skipped_count`, the same convention the loop twenty lines above already uses for a segment whose input zarr is missing — rather than aborting the remaining segments. The refusal uses a dedicated `InputScaleRefused(ValueError)` so that the loop catches this case and not unrelated errors.

### Limits and relationships

- **The recipe tag is a constant, so the tag-vs-level fallback can mis-report the native family.** `prepare_9um_isotropic_input` writes `format: FORMAT_TAG` unconditionally, with no branch on the level, so every input it produces claims the level-2 recipe — including one correctly prepared at level 0 from a native 9.362 µm render. When the source can be consulted this is irrelevant, because the absolute scale settles it; when it cannot, the fallback reports such an input, and says so in its own message.
- **`source` is recorded as it was typed, so a relative path stops resolving elsewhere.** `prepare_9um_isotropic_input` stores the input path verbatim; when that is relative, a consumer run from a different working directory cannot open it. The check then reports `absolute scale not checked: source unreachable`, which is correct but weaker than it needs to be. We hit this on the first run of the check against our own incident artefacts.
- **The check trusts the recorded provenance.** If an array is resampled or rewritten after preparation while its attributes are preserved, the metadata describes the old data and the check reports on the stale record. This is not hypothetical: it happened to one of our own corrected inputs, which we resampled in place and which the check therefore reports.
- **Only one recipe tag is known** (`level2-zmean4-21slice-v1`). A new recipe needs one line in the mapping; an unknown tag is not an error and is skipped untouched.
- **A source that cannot be consulted degrades the check, it does not fail it.** A public-catalogue sweep found stores returning `NoSuchKey`, so "unreachable" had to be a distinct outcome rather than a crash or a false alarm.
- The check does not validate the surface against its CT volume. That is #1463, and it should run first.
- It does not check depth orientation. The tutorial already documents `--flip-normals` for reproducing the published renders' orientation, and `--direction both` for the case where a segment's orientation is unknown; we did not read that before losing the experiment above, and there is nothing to add to it.
- **Precedent rather than neighbourhood:** #1483 fixed a training/inference divergence in this same path — inference doing something different from training — and was merged. This is that shape again, with the pyramid level as the diverging variable. #1229 fixed the same class one stage upstream, scale metadata describing the wrong level. #1541 (open) makes voxel-size discovery robust to nested remote metadata; this change is independent of it and does not depend on it landing.

### Agentic disclosure

Investigation, implementation, testing and drafting used Claude Code and OpenAI Codex under my direction. Every number above comes from runs on public data.

### DIFF
--- a/vesuvius/src/vesuvius/ink_detection/inference/infer.py
+++ b/vesuvius/src/vesuvius/ink_detection/inference/infer.py
@@ -38,6 +38,8 @@ from vesuvius.ink_detection.inference.inference_runtime import (
 from vesuvius.ink_detection.models.model import make_model
 from vesuvius.ink_detection.volume_io import (
     ZARR_V3,
+    InputScaleRefused,
+    check_flat_input_scale,
     open_volume,
     open_volume_root,
     select_volume_level,
@@ -964,6 +966,13 @@ def infer_single_zarr(
     """Run one direction over one surface-volume input and replace a TIFF."""
 
     root = open_volume_root(input_zarr)
+    mismatch = check_flat_input_scale(
+        root,
+        source=str(input_zarr),
+        strict=bool(getattr(args, "strict_input_scale", False)),
+    )
+    if mismatch is not None:
+        LOGGER.warning("%s", mismatch)
     resolution = "0" if hasattr(root, "shape") else str(args.resolution)
     volume = select_volume_level(root, resolution, source=str(input_zarr))
     depth_first, depth, height, width, chunk_h, chunk_w = _volume_axes(volume)
@@ -1239,14 +1248,21 @@ def infer_folder(
                 )
                 skipped_count += 1
                 continue
-            infer_single_zarr(
-                args=args,
-                input_zarr=input_zarr,
-                configured_model=configured_model,
-                device=device,
-                output_tiff=prediction_dir / f"{name_prefix}{date}.tif",
-                layer_direction=direction,
-            )
+            try:
+                infer_single_zarr(
+                    args=args,
+                    input_zarr=input_zarr,
+                    configured_model=configured_model,
+                    device=device,
+                    output_tiff=prediction_dir / f"{name_prefix}{date}.tif",
+                    layer_direction=direction,
+                )
+            except InputScaleRefused as exc:
+                # Same convention as the missing-input case above: one bad segment is
+                # skipped and recorded, it does not abort the rest of the folder.
+                LOGGER.warning("Skipping %s: %s", segment_dir, exc)
+                skipped_count += 1
+                continue
             ran_count += 1
     LOGGER.info(
         "Folder run complete. segments_ran=%d segments_skipped=%d",
@@ -1336,6 +1352,14 @@ def parse_args(argv: Sequence[str] | None = None):
         "--amp-dtype",
         choices=("auto", "default", "fp16", "bf16"),
         default="auto",
+    )
+    parser.add_argument(
+        "--strict-input-scale",
+        action="store_true",
+        help=(
+            "refuse to run when a prepared input's scale does not match what its "
+            "recipe trains at, instead of reporting it and continuing"
+        ),
     )
     parser.add_argument("--tta-mirror", action="store_true")
     parser.add_argument("--tta-batch-size", type=int)
--- a/vesuvius/src/vesuvius/ink_detection/inference/infer.py
+++ b/vesuvius/src/vesuvius/ink_detection/inference/infer.py
@@ -1239,14 +1239,7 @@ def infer_folder(
             LOGGER.warning("Skipping %s: %s", segment_dir, exc)
             skipped_count += 1
             continue
-        try:
-            report_input_scale(args, input_zarr)
-        except InputScaleRefused as exc:
-            # Same level and same convention as the missing-input case above: one bad
-            # segment is skipped and recorded, and the rest of the folder still runs.
-            LOGGER.warning("Skipping %s: %s", segment_dir, exc)
-            skipped_count += 1
-            continue
+        scale_reported = False
         for direction in resolve_run_directions(args.direction):
             name_prefix = (
                 f"{prefix}{segment_dir.name}_{checkpoint_stem}_{direction}_"
@@ -1266,6 +1259,19 @@ def infer_folder(
                 )
                 skipped_count += 1
                 continue
+            if not scale_reported:
+                # Once per input rather than once per direction, and only once some
+                # direction is actually going to run: a segment whose predictions all
+                # exist is skipped without opening its store at all. On a refusal the
+                # whole segment is skipped and recorded, which is the level the
+                # missing-input case above uses.
+                try:
+                    report_input_scale(args, input_zarr)
+                except InputScaleRefused as exc:
+                    LOGGER.warning("Skipping %s: %s", segment_dir, exc)
+                    skipped_count += 1
+                    break
+                scale_reported = True
             infer_single_zarr(
                 args=args,
                 input_zarr=input_zarr,
--- a/vesuvius/src/vesuvius/ink_detection/inference/infer.py
+++ b/vesuvius/src/vesuvius/ink_detection/inference/infer.py
@@ -966,13 +966,6 @@ def infer_single_zarr(
     """Run one direction over one surface-volume input and replace a TIFF."""
 
     root = open_volume_root(input_zarr)
-    mismatch = check_flat_input_scale(
-        root,
-        source=str(input_zarr),
-        strict=bool(getattr(args, "strict_input_scale", False)),
-    )
-    if mismatch is not None:
-        LOGGER.warning("%s", mismatch)
     resolution = "0" if hasattr(root, "shape") else str(args.resolution)
     volume = select_volume_level(root, resolution, source=str(input_zarr))
     depth_first, depth, height, width, chunk_h, chunk_w = _volume_axes(volume)
@@ -1156,6 +1149,23 @@ def resolve_run_directions(direction: str) -> tuple[str, ...]:
     return ("forward", "reverse") if direction == "both" else (direction,)
 
 
+def report_input_scale(args, input_zarr) -> None:
+    """Compare a prepared input's recorded scale with its recipe's, once per input.
+
+    Called before the direction loop rather than inside `infer_single_zarr`, which runs
+    once per direction: the finding describes the input, not the traversal, and
+    `--direction both` would otherwise report it twice.
+    """
+
+    mismatch = check_flat_input_scale(
+        open_volume_root(input_zarr),
+        source=str(input_zarr),
+        strict=bool(getattr(args, "strict_input_scale", False)),
+    )
+    if mismatch is not None:
+        LOGGER.warning("%s", mismatch)
+
+
 def resolve_single_output_path(
     output_tiff: Path,
     *,
@@ -1229,6 +1239,14 @@ def infer_folder(
             LOGGER.warning("Skipping %s: %s", segment_dir, exc)
             skipped_count += 1
             continue
+        try:
+            report_input_scale(args, input_zarr)
+        except InputScaleRefused as exc:
+            # Same level and same convention as the missing-input case above: one bad
+            # segment is skipped and recorded, and the rest of the folder still runs.
+            LOGGER.warning("Skipping %s: %s", segment_dir, exc)
+            skipped_count += 1
+            continue
         for direction in resolve_run_directions(args.direction):
             name_prefix = (
                 f"{prefix}{segment_dir.name}_{checkpoint_stem}_{direction}_"
@@ -1248,21 +1266,14 @@ def infer_folder(
                 )
                 skipped_count += 1
                 continue
-            try:
-                infer_single_zarr(
-                    args=args,
-                    input_zarr=input_zarr,
-                    configured_model=configured_model,
-                    device=device,
-                    output_tiff=prediction_dir / f"{name_prefix}{date}.tif",
-                    layer_direction=direction,
-                )
-            except InputScaleRefused as exc:
-                # Same convention as the missing-input case above: one bad segment is
-                # skipped and recorded, it does not abort the rest of the folder.
-                LOGGER.warning("Skipping %s: %s", segment_dir, exc)
-                skipped_count += 1
-                continue
+            infer_single_zarr(
+                args=args,
+                input_zarr=input_zarr,
+                configured_model=configured_model,
+                device=device,
+                output_tiff=prediction_dir / f"{name_prefix}{date}.tif",
+                layer_direction=direction,
+            )
             ran_count += 1
     LOGGER.info(
         "Folder run complete. segments_ran=%d segments_skipped=%d",
@@ -1404,6 +1415,11 @@ def main(argv: Sequence[str] | None = None) -> int:
     if args.folder is not None:
         infer_folder(args, configured, device=device)
     else:
+        try:
+            report_input_scale(args, args.input_zarr)
+        except InputScaleRefused as exc:
+            LOGGER.error("%s", exc)
+            return 1
         for direction in resolve_run_directions(args.direction):
             infer_single_zarr(
                 args=args,
--- a/vesuvius/src/vesuvius/ink_detection/volume_io.py
+++ b/vesuvius/src/vesuvius/ink_detection/volume_io.py
@@ -6,7 +6,7 @@ import hashlib
 import json
 import os
 from pathlib import Path
-from typing import Any
+from typing import Any, Mapping
 
 import aiohttp
 import numpy as np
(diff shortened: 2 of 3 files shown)


Bullo27:
Independent reproduction, different machine and a different Python environment. Everything below I ran rather than read off the diff. My `main` is `07656e634`, which is exactly this PR's base, so nothing here is rebased.

## The regression claim reproduces — in a stronger form than the table states

My environment has different optional dependencies from yours, so the absolute counts differ. The **deltas are identical**:

| | failed | passed | collection errors |
|---|---|---|---|
| without the PR (`07656e634`) | 1 | 15 | 17 |
| with the PR (`66da84d7d`) | 1 | 29 | 17 |

**+14 passes, and the one failure is the same test on both sides** (`test_volume_io.py::test_parent_populated_dataset_cache_misses_after_fork`). `test_volume_io.py` alone goes 13 → 27 collected. That your absolute numbers and mine disagree while the delta agrees exactly is the useful part: the 14 additions, and "this PR neither fixes nor touches the pre-existing failures", are environment-independent.

For anyone else reproducing: the suite needs `pynrrd` to collect at all (`vesuvius/__init__` → `data.volume` → `import nrrd`), and `vesuvius` resolves as a *namespace* package to the repo's top-level `vesuvius/` directory unless you put `vesuvius/src` on `PYTHONPATH` first.

## The advertised message is byte-identical to what the code emits

Built a prepared input carrying your attrs, against an OME source at `[2.399, 4.798, 9.596]`:

```
w035_9um.zarr: input is 2.399 um in-plane; this recipe trains at 9.596 or 9.362 um (nearest 9.362, factor 0.256). Prepared from ink_ctrl2/w035_sv.zarr at level 0.
```

Byte-for-byte the string in the PR body. Negative controls stay silent as claimed: level-2 → `None`, and a native 9.362 µm render at level 0 → `None`.

## Repo-side claims, all checked

- **`source_level`** — 9 hits across 4 files: the writer, its test, and `lasagna/omezarr_pyramid.py` + `lasagna/optimizer.py`. Exactly "two unrelated files".
- **`aligned21_fixed_scroll_prior.json`** — 29 representations, 24 `public_2p4_level2_zmean4` + 5 `native_9p362_level0`, `w035` present in both families.
- **Arithmetic** — 2.399 × 4 = 9.596; nearest *in relative distance* is 9.362 (0.744 vs 0.750); factor 0.2562; the nearest wrong pyramid level is 50% away; the two 2% windows do overlap, [9.175, 9.549] and [9.404, 9.788], so "nothing legitimate falls between them" holds.
- **The `Volume::normalize()` ordering** — checked against `volume-cartographer/core/src/Volume.cpp:109-148`. The Python mirrors it faithfully: scan-merge first, then the same six spellings in the same order, then the four nested containers, then `scan/tomo/acquisition/detector/samplePixelSize` × 1000.

## One thing I'd raise — minor, not a correctness bug

The new `except InputScaleRefused` sits at the **direction** level, inside `for direction in resolve_run_directions(...)`. Its comment cites "the missing-input case above", but that one (`infer.py:1219` on main) is at the **segment** level. Since `resolve_run_directions("both") == ("forward", "reverse")`, measured on your branch:

- default reporting, `--direction both`, both directions actually running: **2 identical warnings for one input**
- `--strict-input-scale` in a folder run: **`skipped_count` += 2 for one segment**, where the case it cites would add 1

Nothing aborts and the rest of the folder still runs, so the stated goal holds — it just matches the *other* existing convention (the already-have-prediction skip at `1233`, which is also per-direction) rather than the one named. I mention it mainly because this PR itself recommends `--direction both` when a segment's orientation is unknown, so the headline output double-prints under exactly the flag it recommends. In folder mode the second print is skipped if that direction already has a prediction; the single-input path in `main()` (`infer.py:1384`) always runs both, so it double-prints unconditionally.

Separately, `main()` doesn't catch `InputScaleRefused`, so `--strict-input-scale` on a single input surfaces as a traceback rather than as the clean message. Arguably right — refusing *should* stop a single-input run — but the message is the valuable half and it currently arrives wrapped in a stack trace.

Neither point affects the check itself, which does what the PR says it does.

---
*Claude (AI agent) on behalf of @Bullo27. Measured on `main` at `07656e634` and this PR at `66da84d7d`; the two review points are read off the call sites and then confirmed by running them.*

khj1222:
Nice measurement — varying `--level` and the depth order independently is what makes it
convincing, since the two faults arrived together in the incident.

A supporting data point from the other end of the same failure, offered as evidence for the
PR rather than as a request: I have been running `prepare_9um_isotropic_input` across the
whole `ink_9um` corpus this month.

Where your check would have caught me: it would not have, and I think that is the
interesting part rather than an objection. Both input families I trained and scored on are
scales this recipe trains at:

- **aligned** — built with `prepare_9um_isotropic_input --level 2` from the public 2.399 µm
  volumes with 4× z mean pooling, i.e. 9.596 µm. Your table's first row, correct usage.
- **native** — the published 9.362 µm surface volumes, not produced by the preparer.
  Your third row: silent, no recipe tag.

Both pass your check, and both should. They are still not interchangeable at transfer time.

PHerc0139 is the only scroll in the corpus where the same physical segments exist in both
families, so it is a controlled comparison. Holding out all of PHerc0139 and retraining the
recipe on the remaining 15 representations (2 seeds, 7 checkpoints each, best of that grid,
threshold swept per cell), scored against each segment's own trivial-classifier lower bound
2p/(1+p) — the bound differs slightly between the two renders, which is why I compare
margins rather than raw F1:

| segment | aligned F1 (margin) | native F1 (margin) | Δ margin |
|---|---|---|---|
| w035 | 0.740 (+0.260) | 0.679 (+0.203) | +0.057 |
| w039 | 0.654 (+0.169) | 0.585 (+0.103) | +0.066 |
| w040 | 0.703 (+0.088) | 0.652 (+0.040) | +0.048 |
| w041 | 0.734 (+0.145) | 0.703 (+0.117) | +0.028 |

Aligned wins 4/4, +0.028 to +0.066. That arm's training corpus is entirely aligned-family,
so the reading I take is domain match: transfer into the family you trained on is
consistently better, and some of what looks like cross-scroll difficulty is scan and
preprocessing domain difference rather than the scroll.

**Caveats, because this is not a replication of your measurement.** Different metric (F1
margin, not AUC), cross-scroll rather than same-scroll, best-of-grid selection applied
identically to both families, and the two families ship their own label sets — so
representation is not perfectly isolated from label transfer. Against your 0.95 → 0.64 the
effect I see is small. The point I would make for the PR is the direction, not the size:
the catastrophic case you catch is silent today, and the residual within-spec case is
measurable too, which makes the provenance you are already recording worth surfacing rather
than only comparing.

Raw numbers, if useful — the grid this table comes from, one row per arm × segment × step:
https://github.com/khj1222/vesuvius-challenge/blob/main/runs/ink9um_scorecard/no0139_matrix.csv
(`pherc0139-wNNN` rows are aligned, bare `wNNN` are native; `loso42`/`loso43` are the
held-out arm, `ref42`/`ref43` the reference models that trained on these segments). The
arm generator is `tools/make_ink9um_config.py` in the same repo. The write-up is in Korean
at the moment; happy to put the method in English here if anyone wants it.

Bullo27:
@khj1222 — I pulled your `no0139_matrix.csv` and recomputed the table from it. **Every published figure reproduces exactly**: all eight F1s, all eight margins, all four Δ margins, and the 4/4.

**To be clear about what that does and doesn't establish:** I recomputed your table *from your own published grid*, so this checks that the table faithfully summarises the CSV — it is not an independent reproduction of your training runs, which I have no way to repeat. What follows is new analysis of your numbers, not new measurement.

Two things I checked because the CSV supports them and they seemed worth knowing.

## Your selection rule isn't doing the work

All eight winning cells come from `loso43`, none from `loso42`, and the rule is a max over 14 cells — so the obvious worry is best-of-grid inflation. It survives everything I threw at it:

| selection rule | aligned wins | mean Δ margin |
|---|---|---|
| max over both seeds × 7 ckpt (yours) | 4/4 | +0.0498 |
| max, `loso42` only | 4/4 | +0.0528 |
| max, `loso43` only | 4/4 | +0.0498 |
| mean over both seeds × 7 ckpt | 4/4 | +0.0405 |
| median over both seeds × 7 ckpt | 4/4 | +0.0407 |

And with no selection at all — each of the 7 checkpoints taken on its own, both seeds pooled — aligned wins **4/4 at every single checkpoint**, mean Δ from +0.0296 to +0.0497. The seed concentration is harmless: `loso42` alone gives a *larger* mean gap than your rule does.

## The reference arms are a control for your interpretation, and they support it

You offer domain match as "the reading I take". The competing reading is duller — that aligned renders are simply easier to score, in which case the gap should appear whether or not the model has seen these segments. Your `ref` arms test exactly that:

| arm | aligned wins | mean Δ margin |
|---|---|---|
| `loso42`/`loso43` (held out) | **4/4** | **+0.0498** |
| `ref42`/`ref43` (trained on these segments) | **0/4** | **−0.0056** |

Per segment, ref: w035 −0.0063, w039 −0.0069, w040 −0.0055, w041 −0.0036. The gap doesn't shrink — it collapses to roughly zero and tips marginally the other way, consistently across all four.

That's the shape domain match predicts and "aligned renders are intrinsically better" does not: an advantage that exists only when the family is unseen. Worth putting in the comment, because it moves this from an interpretation you prefer to one with a control behind it.

One caveat I can't resolve from the CSV: **what the `ref` arms trained on.** You say the held-out arm's corpus is entirely aligned-family, but not the reference arms'. If ref is also aligned-only and merely includes these segments, the control isolates segment familiarity; if ref saw both families, it isolates family exposure directly. Both readings point the same way — the intrinsic-quality explanation is ruled out either way — but the mechanism differs, so it's worth stating.

Absolute margins for context, since the ref arms gain a lot in both families (w035: loso +0.260/+0.203 → ref +0.497/+0.503; w040: +0.088/+0.040 → +0.349/+0.355) — the collapse of the *gap* is not a ceiling effect flattening two saturated numbers.

---
*Claude (AI agent) on behalf of @Bullo27. Recomputed from `no0139_matrix.csv` as published (392 rows, 4 arms × 7 checkpoints); no models were trained or re-scored on my side.*

nerln:
Thanks both — the two actionable findings are fixed, and I re-verified the current head (`2baa98e`) by running it rather than re-reading the diff.

**Review fix 1 — one report per input, not per direction** (`ce94337`). The check moved out of `infer_single_zarr`, which runs once per direction, into `report_input_scale`, called once above the direction loop. `--direction both` emits one diagnostic, and a refusal under `--strict-input-scale` skips the segment once rather than twice.

**Review fix 2 — and that first fix was itself wrong in a way CI caught** (`2baa98e`). Hoisting the call above the loop made it run for segments whose predictions all exist, which are skipped without their store ever being opened; on a fixture whose `.zarray` is empty that raises `JSONDecodeError` from `open_volume_root`, well before any scale logic. `test_folder_mode_logs_existing_prediction_and_summary` failed on both Python versions. The call now sits inside the loop behind a `scale_reported` flag, placed after the existing-prediction skip, which holds all three properties at once: once per input, only when a direction is actually going to run, and a fully skipped segment is never opened.

**Strict single-input** now logs the diagnostic and returns a non-zero exit code instead of surfacing `InputScaleRefused` as a traceback.

**Regression.** Measured on three machines with different optional dependencies. The absolute counts differ; what is environment-independent is that **the failing set is identical with and without the PR** in all three, and none of those tests is touched here.

| environment | without | with | delta |
|---|---|---|---|
| mine, on the PR base `8b59dde5` | 2 failed / 20 passed / 16 errors | 2 / 34 / 16 | +14 |
| @Bullo27, on `07656e634` | 1 / 15 / 17 | 1 / 29 / 17 | +14 |
| a third, with `torch`/`timm`/`pynrrd` present | 4 / 122 | 4 / 135 | +13 |

The third environment collects far more of the suite and lands one short of the other two; I have not chased which of the added tests does not collect there, and I would rather say so than round it to fourteen. One of the four failures in that environment is intermittent, which is why a single run can read three.

@khj1222's representation-family result looks real, and @Bullo27's recomputation of the grid holds — including that it survives per-seed, per-checkpoint, mean and median selection. It is also, by construction, outside this PR: both families are valid-scale inputs and both pass this check, which is the intended behaviour, and the native 9.362 µm case has a test here asserting exactly that. I've opened #1582 so that the question of what provenance identifies a representation family can be decided without this scale fix depending on it.

Unless there's a blocker I've missed, I believe #1580 is ready for review.

PR #1581, opened 2026-08-24
Scroll 3 is catalogued under 54 keV but its only volume is the 53 keV…

EazyHood:
**In one sentence:** You can now open Scroll 3 with the energy its data actually has.

**One real example:** Starting with Scroll 3 (PHerc332) on `dl.ash2txt.org`, I called
`Volume(type="scroll", scroll_id=3, energy=53, resolution=7.91)` and it produced a
`ValueError` — even though the only volume the server publishes for that scroll is
`53keV_7.91um_Scroll3.zarr`.

**Before:**

```
>>> from vesuvius import Volume
>>> v = Volume(type="scroll", scroll_id=3, energy=53, resolution=7.91)
Error reading or parsing config file .../configs/scrolls.yaml: URL not found in config
for scroll=3, energy=53, resolution=7.91
ValueError: URL not found in config for scroll=3, energy=53, resolution=7.91
```

And the inverse, which is the part that worries me more — asking for an energy that does not
exist succeeds and silently hands back the 53 keV volume:

```
>>> v = Volume(type="scroll", scroll_id=3, energy=54, resolution=7.91)
>>> v.shape()
(9778, 3550, 3400)          # this is 53keV_7.91um_Scroll3.zarr
```

**After this PR:**

```
>>> v = Volume(type="scroll", scroll_id=3, energy=53, resolution=7.91)
>>> v.shape()
(9778, 3550, 3400)
>>> patch = v[5000, 1700:1764, 1700:1764]
>>> patch.shape, patch.dtype, patch.min(), patch.max(), round(float(patch.mean()), 1)
((64, 64), dtype('uint8'), 0, 216, 49.9)
```

**Proof:** the two blocks above are the same call before and after the one-line change, on the
same machine and the same remote volume. The directory listing at
`https://dl.ash2txt.org/full-scrolls/Scroll3/PHerc332.volpkg/volumes_zarr_standardized/`
contains exactly one entry, `53keV_7.91um_Scroll3.zarr` — there is no 54 keV volume to point at.
Scroll 5, whose volume is also a 53 keV scan, is already keyed `"53"` in the same file.

I also re-checked the four scrolls I did not touch, and they open unchanged:

```
scroll 1 @ 54keV -> (14376, 7888, 8096)
scroll 2 @ 54keV -> (14428, 10112, 11984)
scroll 4 @ 54keV -> (11174, 3340, 3440)
scroll 5 @ 53keV -> (21000, 6700, 9100)
```

**Why / where this is useful:** anyone selecting scroll volumes by energy — comparing scans
across scrolls, or scripting over `list_files()` — currently has to ask for the wrong number to
get Scroll 3, and gets no warning that the volume they received was scanned at a different
energy from the one they requested.

      stated data.

## Details

Tested commit: `de3c249`. One-line change in
`vesuvius/src/vesuvius/install/configs/scrolls.yaml`: the Scroll 3 block's energy key goes from
`"54"` to `"53"`. No URL is edited and no other scroll is touched — the diff is +1/-1.

Environment: Windows 11, Python 3.14.5, minimal install (`uv pip install -e .` plus
`nest-asyncio`, which `list_files()` needs and is not in the base dependencies — that part is
already tracked in #1329, so it is not included here).

### DIFF
--- a/vesuvius/src/vesuvius/install/configs/scrolls.yaml
+++ b/vesuvius/src/vesuvius/install/configs/scrolls.yaml
@@ -27,7 +27,7 @@
 
 # Scroll 4 - tested 03/24/25
 "4":
-  "54":
+  "53":
     "7.91":
       volume: "https://dl.ash2txt.org/full-scrolls/Scroll4/PHerc1667.volpkg/volumes_zarr/20231117161658.zarr/"
 
--- a/vesuvius/src/vesuvius/install/configs/scrolls.yaml
+++ b/vesuvius/src/vesuvius/install/configs/scrolls.yaml
@@ -21,7 +21,7 @@
 
 # Scroll 3 -  tested 03/24/25
 "3":
-  "54":
+  "53":
     "7.91":
       volume: "https://dl.ash2txt.org/full-scrolls/Scroll3/PHerc332.volpkg/volumes_zarr_standardized/53keV_7.91um_Scroll3.zarr/"
 


Bullo27:
Verified — and **the same defect is present on Scroll 4**, which this PR doesn't touch.

## Your Scroll 3 claim reproduces

On `main` at `de3c24943`, the Scroll 3 block is keyed `"54"` while the URL under it is `.../53keV_7.91um_Scroll3.zarr/` — a direct key-vs-URL disagreement. The server publishes exactly one standardized volume there, `53keV_7.91um_Scroll3.zarr/`, and the volpkg's raw volumes are 53 keV @ 3.24 µm, 53 keV @ 7.91 µm, and 70 keV @ 3.24 µm. **No 54 keV Scroll 3 scan exists**, so relabelling is the right fix rather than repointing.

## The method matters, and it turns up a second instance

A filename only carries keV for some entries — Scrolls 4 and 5 point at bare-timestamp zarrs (`20231117161658.zarr`, `20241024131838.zarr`), so a name-based check can't see them. I checked every entry against the energy the **data itself declares**, in `volumes/<id>/meta.json`:

| scroll | config key | data declares | |
|---|---|---|---|
| 1 | 54 | 54 | OK |
| 2 | 54 | 54 | OK |
| 3 | 54 | **53** | fixed by this PR |
| 4 | 54 | **53** | **not fixed here** |
| 5 | 53 | 53 | OK |

Scroll 4's configured volume `20231117161658` declares `'PHerc1667, 53 keV, 7.91 um, merged'` — I read it three times, same answer each time. And there's no 54 keV Scroll 4 volume to point at either: that volpkg holds exactly two raw volumes, 88 keV @ 3.24 µm and 53 keV @ 7.91 µm. So it's your fix again, one block lower — `"54"` → `"53"` for Scroll 4.

Scrolls 1 and 2 declare 54 keV and match their keys, so the check discriminates; it isn't flagging everything.

## Why the four-scroll re-check couldn't catch it

Your verification line was `scroll 4 @ 54keV -> (11174, 3340, 3440)` — and that shape *is* the 53 keV 7.91 µm volume. So rather than showing Scroll 4 is fine, that line is itself an instance of the silent substitution you describe for Scroll 3: you asked for 54 and got 53 keV data. Opening successfully only proves the URL resolves; it can't prove the label is right. Reading the volume's own `meta.json` is what separates the two.

Nothing here changes the one-line diff you've already made — it just means the same one-liner is needed twice, and your "before/after" framing covers both.

---
*Claude (AI agent) on behalf of @Bullo27. Measured against `main` at `de3c24943` and live `dl.ash2txt.org` metadata; every energy above is read from the volume's own `meta.json`, not from a filename.*

EazyHood:
Thank you — you are right on both counts, and the second one is the more useful of the two.

I have pushed the Scroll 4 fix, so the PR is now two lines instead of one.

On the method: my check compared the config key against the keV in the filename, which is
exactly the wrong source. It cannot see Scrolls 4 and 5 at all, and — as you point out — my own
`scroll 4 @ 54keV -> (11174, 3340, 3440)` line was not a regression check passing, it was the
bug happening in front of me. Opening successfully only proves the URL resolves. I read it as
proof the label was right, which is precisely the substitution I was complaining about for
Scroll 3.

I re-checked against the volumes' own metadata rather than take it on trust, and it agrees with
you:

| scroll | config key | `meta.json` name | |
|---|---|---|---|
| 1 | 54 | `PHercParis4 54keV stitched_part_1` | ok |
| 2 | 54 | `PHercParis3 54keV stitched_part_1` | ok |
| 3 | 54 | `PHerc332, 53 keV, ..., merged` | fixed |
| 4 | 54 | `PHerc1667, 53 keV, 7.91 um, merged` | fixed now |
| 5 | 53 | `P.Herc. 172, 53keV, 7.91um` | ok |

With both keys corrected:

```
scroll 1 @ 54keV -> (14376, 7888, 8096)
scroll 2 @ 54keV -> (14428, 10112, 11984)
scroll 3 @ 53keV -> (9778, 3550, 3400)
scroll 4 @ 53keV -> (11174, 3340, 3440)
scroll 5 @ 53keV -> (21000, 6700, 9100)

Volume(scroll_id=4, energy=53)[6000, 1600:1664, 1600:1664]
-> (64, 64) uint16, min 12598, max 57122, mean 23546.6
```

and asking for 54 keV on Scrolls 3 or 4 now raises instead of quietly handing back 53 keV data,
which is the behaviour I wanted in the first place.

Bullo27:
Merged as `c53c74a29`, and it took both halves — Scroll 3 and the Scroll 4 twin. Re-audited the whole file against what the data declares, not just the two changed lines:

| scroll | config | volume's own `meta.json` | |
|---|---|---|---|
| 1 | 54 | 54 | ok |
| 2 | 54 | 54 | ok |
| 3 | **53** | 53 | ok — fixed here |
| 4 | **53** | 53 | ok — fixed here |
| 5 | 53 | 53 | ok |

**0 mismatches across all 5 entries.** Scroll 4 resolves through the raw sibling `PHerc1667.volpkg/volumes/20231117161658/meta.json`, which reads `"PHerc1667, 53 keV, 7.91 um, merged"` — so the new `53` is what the data says, and `scrolls.yaml` is now internally consistent end to end rather than just locally patched.

Nice one — thanks for carrying the second scroll along with the first.

---
*Claude (AI agent) on behalf of @Bullo27. Measured against `main` at `c53c74a29` and `dl.ash2txt.org` on 2026-08-25, with retries on each fetch (transient S3 errors have produced false negatives for me before). Scrolls 3 and 5 point at standardized zarrs, so their energies are read from the raw sibling volume matched on voxel size.*

PR #1584, opened 2026-08-24
spiral: the CUDA torch index is gated to Linux, so Windows installs a CPU-only build

EazyHood:
**Scope, so the title is not read as more than it is:** on Windows this brings CUDA torch, not
the full GPU path. `triton` publishes no Windows wheel — 3.7.1 ships 12 files, all manylinux —
so it correctly stays `sys_platform == 'linux'` in the lock, and the fused kernels in
`flow_triton.py` and `gap_triton.py` take the eager fallback there. Both guard their import
already, so this is a performance ceiling rather than a failure.

**In one sentence:** `uv sync` in `spiral-fitting` now gives Windows the same CUDA torch it
already gives Linux, instead of a CPU-only build.

**One real example:** Starting with `spiral-fitting/pyproject.toml` as it is on `main`, I
installed torch on a Windows machine with a working CUDA device and got `2.12.1+cpu` with
`torch.cuda.is_available() == False`.

**Before:**

`tool.uv.sources` sends torch to the `pytorch-cu128` index only under
`sys_platform == 'linux'`, so on Windows the resolver falls through to PyPI. The committed
lockfile already records this:

```
{ name = "torch", version = "2.11.0",       source = pypi.org,              marker = "sys_platform != 'linux'" }
{ name = "torch", version = "2.11.0+cu128", source = download.pytorch.org,  marker = "sys_platform == 'linux'" }
```

PyPI's Windows torch wheels are CPU-only. For the version the lockfile actually resolves,
`torch-2.11.0-cp314-cp314-win_amd64.whl` on PyPI is 115 MB (114,552,300 bytes), against
2771 MB (2,770,971,557 bytes) for `torch-2.11.0+cu128-cp314-cp314-win_amd64.whl` in the
cu128 index. Not a like-for-like download increase, though: on Linux the CUDA runtime
arrives as separate `nvidia-*` packages rather than bundled into the wheel. On a machine
with a working GPU:

```
torch              : 2.12.1+cpu
torch.version.cuda : None
cuda.is_available  : False

>>> torch.zeros(4, device='cuda')
AssertionError: Torch not compiled with CUDA enabled
```

**After this PR:**

Same machine, same `>=2.11,<2.13` pin, only the index differs:

```
torch              : 2.11.0+cu128
torch.version.cuda : 12.8
cuda.is_available  : True
GPU                : NVIDIA GeForce RTX 5060 Laptop GPU
compute capability : sm_120
VRAM               : 8.5 GB
>>> (torch.randn(1024, 1024, device='cuda') @ torch.randn(1024, 1024, device='cuda')).sum()
tensor(12439.3, device='cuda:0')
```

**Proof:** the two blocks above are the same machine, minutes apart, differing only in which
index torch came from. The wheel the fix needs is one the lockfile already pins for Linux, and
it exists for Windows: `torch-2.11.0+cu128-cp314-cp314-win_amd64.whl` is published in the same
`cu128` index. Note that only `2.11.0` has `cp314` `cu128` wheels — `2.12.x` does not — which is
exactly the version the lockfile already resolves to.

Re-locking changes markers only. I diffed package versions across the whole lockfile before and
after: 121 packages before, 121 after, and **no package changes version**. macOS keeps resolving
from PyPI as it does today.

**Why / where this is useful:** anyone running `uv sync` on Windows today gets a GPU fitting
pipeline that cannot see the GPU, with no error at install time — the failure only surfaces
later as `Torch not compiled with CUDA enabled`, or as a fit that silently runs on CPU.

      stated data.

## Details

Tested on Windows 11, Python 3.14.5, driver 610.62, RTX 5060 Laptop (sm_120, 8 GB).

The change is one marker in `spiral-fitting/pyproject.toml`:

```diff
-    { index = "pytorch-cu128", marker = "sys_platform == 'linux'" },
+    { index = "pytorch-cu128", marker = "sys_platform == 'linux' or sys_platform == 'win32'" },
```

plus the regenerated `uv.lock`, which afterwards reads:

```
{ name = "torch", version = "2.11.0",       source = pypi.org,             marker = "sys_platform != 'linux' and sys_platform != 'win32'" }
{ name = "torch", version = "2.11.0+cu128", source = download.pytorch.org, marker = "sys_platform == 'linux' or sys_platform == 'win32'" }
```

I did not touch the `vesuvius` subproject, which declares its torch differently.


EazyHood:
Some context, since this is a packaging change.

I am on Windows and I was setting up to try a spiral fit locally. I installed the deps, and
before starting anything long I checked that torch could actually see my card — and it could
not. It took me a while to work out that it was not my driver or my card: the install had simply
given me a CPU build, because the cu128 index is gated to Linux.

What made me open a PR rather than just work around it locally is that nothing tells you. There
is no warning at install time, and my GPU is fine — you only find out later, either from
`Torch not compiled with CUDA enabled` or from a fit that is mysteriously slow because it is on
CPU. On a pipeline whose whole point is GPU fitting that seems worth failing loudly about, or
better, just not doing.

I checked the wheel exists before proposing this: it is the same version your lockfile already
pins for Linux.

Bullo27:
Checked the four claims here that don't need a Windows box, and they all hold. One packaging note that does need acting on, plus a scope note.

## Verified

| claim | result |
|---|---|
| main gates cu128 to Linux | `pyproject.toml:41` on `de3c24943` = `{ index = "pytorch-cu128", marker = "sys_platform == 'linux'" }` |
| the Windows cu128 wheel exists | `torch-2.11.0+cu128-cp314-cp314-win_amd64.whl` → **HTTP 200, 2771 MB** (and the `cp314t` one, same size) |
| PyPI's Windows wheel is the small CPU build | `torch-2.11.0-cp314-cp314-win_amd64.whl` → **115 MB**. Your `torch.version.cuda is None` on real hardware is the direct proof; the 115 MB vs 2771 MB contrast corroborates it |
| "121 packages before, 121 after, no package changes version" | **confirmed, and a bit stronger**: 122 `[[package]]` blocks / 121 distinct names on both sides (torch appears twice by design), no name added or removed, and **no name's version set changes** |

One framing caveat on the sizes: 2771 MB vs Linux's 820 MB isn't apples-to-apples, because on Linux the CUDA runtime arrives as separate `nvidia-*` / `cuda-toolkit` packages while the Windows wheel bundles it. Not a like-for-like download increase.

## This PR carries #1581's commits

Its diff against main is **three** files, not two — `vesuvius/src/vesuvius/install/configs/scrolls.yaml` rides along. Measured:

```
$ git log --oneline de3c24943..pr1584
516ea5608 spiral: the CUDA torch index is gated to Linux, so Windows installs a CPU build
7d3176710 Scroll 4 has the same key/data disagreement as Scroll 3
3661de5b6 Scroll 3 is catalogued under 54 keV but its only volume is the 53 keV scan

$ git merge-base --is-ancestor pr1581 pr1584   # → yes
```

The blob is byte-identical to #1581's (`d404344ee26f`). The branch was cut from the #1581 branch rather than from main. Your own "I did not touch the `vesuvius` subproject" is how I noticed — that sentence is true of the change you *made* and false of the diff you *submitted*.

It matters because reviewers are already requested here: as it stands, merging this merges the scroll relabel too, without it being reviewed on its own merits, and leaves #1581 as an empty PR. This rebase drops the two scroll commits and keeps yours; I ran it in a throwaway worktree:

```
$ git rebase --onto de3c24943 7d3176710 fix/windows-cuda-torch-index
# → 1 commit, 2 files (pyproject.toml + uv.lock), the marker change intact
```

The merge base is already `de3c24943`, so nothing else is stale.

## Scope note: triton stays Linux-only, and that's correct

After this PR Windows gets `torch 2.11.0+cu128`, but `triton` keeps `marker = "sys_platform == 'linux'"` in the lock. That is right, not an oversight — I checked before suggesting otherwise: upstream `triton` publishes **no Windows wheel at all** (latest 3.7.1 on PyPI is manylinux-only; the cu128 index carries 164 triton wheels, zero for Windows). Same for the `nvidia-*` and `cuda-toolkit` entries, which the Windows wheel bundles instead.

The consequence is worth stating in the PR rather than leaving for someone to discover: on Windows the fused kernels in `flow_triton.py` and `gap_triton.py` stay off and those paths take the eager fallback. Nothing breaks — both modules already guard with `try: import triton / except ImportError: _HAS_TRITON = False`, and the call sites go through `rk4_triton_available()` / `gap_triton_available()`. So "Windows now gets CUDA torch" is accurate, and "Windows now gets the full GPU path" would not be.

Small consequence of that: the comment on those two imports, `# pragma: no cover - triton is present on CUDA installs`, stops being true for Windows CUDA installs once this lands.

---
*Claude (AI agent) on behalf of @Bullo27. Measured against `main` at `de3c24943` and your `516ea5608`; wheel sizes are `Content-Length` from HEAD requests, the lockfile comparison parses both `uv.lock` revisions, and the rebase was executed in a throwaway worktree, not inferred.*

EazyHood:
> Checked the four claims here that don't need a Windows box, and they all hold. One packaging note that does need acting on, plus a scope note.
> 
> ## Verified
> claim	result
> main gates cu128 to Linux	`pyproject.toml:41` on `de3c24943` = `{ index = "pytorch-cu128", marker = "sys_platform == 'linux'" }`
> the Windows cu128 wheel exists	`torch-2.11.0+cu128-cp314-cp314-win_amd64.whl` → **HTTP 200, 2771 MB** (and the `cp314t` one, same size)
> PyPI's Windows wheel is the small CPU build	`torch-2.11.0-cp314-cp314-win_amd64.whl` → **115 MB**. Your `torch.version.cuda is None` on real hardware is the direct proof; the 115 MB vs 2771 MB contrast corroborates it
> "121 packages before, 121 after, no package changes version"	**confirmed, and a bit stronger**: 122 `[[package]]` blocks / 121 distinct names on both sides (torch appears twice by design), no name added or removed, and **no name's version set changes**
> One framing caveat on the sizes: 2771 MB vs Linux's 820 MB isn't apples-to-apples, because on Linux the CUDA runtime arrives as separate `nvidia-*` / `cuda-toolkit` packages while the Windows wheel bundles it. Not a like-for-like download increase.
> 
> ## This PR carries #1581's commits
> Its diff against main is **three** files, not two — `vesuvius/src/vesuvius/install/configs/scrolls.yaml` rides along. Measured:
> 
> ```
> $ git log --oneline de3c24943..pr1584
> 516ea5608 spiral: the CUDA torch index is gated to Linux, so Windows installs a CPU build
> 7d3176710 Scroll 4 has the same key/data disagreement as Scroll 3
> 3661de5b6 Scroll 3 is catalogued under 54 keV but its only volume is the 53 keV scan
> 
> $ git merge-base --is-ancestor pr1581 pr1584   # → yes
> ```
> 
> The blob is byte-identical to #1581's (`d404344ee26f`). The branch was cut from the #1581 branch rather than from main. Your own "I did not touch the `vesuvius` subproject" is how I noticed — that sentence is true of the change you _made_ and false of the diff you _submitted_.
> 
> It matters because reviewers are already requested here: as it stands, merging this merges the scroll relabel too, without it being reviewed on its own merits, and leaves #1581 as an empty PR. This rebase drops the two scroll commits and keeps yours; I ran it in a throwaway worktree:
> 
> ```
> $ git rebase --onto de3c24943 7d3176710 fix/windows-cuda-torch-index
> # → 1 commit, 2 files (pyproject.toml + uv.lock), the marker change intact
> ```
> 
> The merge base is already `de3c24943`, so nothing else is stale.
> 
> ## Scope note: triton stays Linux-only, and that's correct
> After this PR Windows gets `torch 2.11.0+cu128`, but `triton` keeps `marker = "sys_platform == 'linux'"` in the lock. That is right, not an oversight — I checked before suggesting otherwise: upstream `triton` publishes **no Windows wheel at all** (latest 3.7.1 on PyPI is manylinux-only; the cu128 index carries 164 triton wheels, zero for Windows). Same for the `nvidia-*` and `cuda-toolkit` entries, which the Windows wheel bundles instead.
> 
> The consequence is worth stating in the PR rather than leaving for someone to discover: on Windows the fused kernels in `flow_triton.py` and `gap_triton.py` stay off and those paths take the eager fallback. Nothing breaks — both modules already guard with `try: import triton / except ImportError: _HAS_TRITON = False`, and the call sites go through `rk4_triton_available()` / `gap_triton_available()`. So "Windows now gets CUDA torch" is accurate, and "Windows now gets the full GPU path" would not be.
> 
> Small consequence of that: the comment on those two imports, `# pragma: no cover - triton is present on CUDA installs`, stops being true for Windows CUDA installs once this lands.
> 
> _Claude (AI agent) on behalf of @Bullo27. Measured against `main` at `de3c24943` and your `516ea5608`; wheel sizes are `Content-Length` from HEAD requests, the lockfile comparison parses both `uv.lock` revisions, and the rebase was executed in a throwaway worktree, not inferred._

All three land. Taking them in order.

**The scroll commits.** Caught the same thing myself shortly before your comment and
force-pushed the rebase — the PR now shows one commit and two files, `pyproject.toml` and
`uv.lock`. Your comment is measured against `516ea5608`; the head is `949c745a5`. Your reading
of how it happened is exactly right, including that my "I did not touch the vesuvius subproject"
was true of the change and false of the diff. That is a good way to put it and I will check the
diff rather than my intent next time.

**The wheel sizes.** Yours are better than mine — I quoted 123 MB from the PyPI JSON for 2.12.1
rather than the 2.11.0 the lock actually resolves. 115 MB vs 2771 MB is the right pair. And your
apples-to-apples caveat is fair: the Linux CUDA runtime arrives as separate `nvidia-*` packages,
so this is not a like-for-like download increase.

**Triton.** I had not looked at this and it belongs in the description. Confirmed here: triton
3.7.1 publishes 12 files, none for Windows — manylinux aarch64 and x86_64 only. Both call sites
guard with `try: import triton / except ImportError: _HAS_TRITON = False`, so nothing breaks,
but the fused paths in `flow_triton.py` and `gap_triton.py` take the eager fallback on Windows.
I have added this to the PR description:

> On Windows this brings CUDA torch, not the full GPU path. `triton` publishes no Windows wheel,
> so it correctly stays `sys_platform == 'linux'` in the lock, and the fused kernels in
> `flow_triton.py` and `gap_triton.py` fall back to eager. Both already guard their import, so
> this is a performance ceiling rather than a failure. It also means the
> `# pragma: no cover - triton is present on CUDA installs` comment on those two imports stops
> being accurate for Windows CUDA installs once this lands.

I have left the pragma comments alone rather than widen this PR; happy to fix them here or in a
separate one, whichever you prefer.

Issue #1583, opened 2026-08-24
get_ink_metrics.py: the fork guard is dead code, and on Windows every fold dies after the expensive prep


EazyHood:
**What I was trying to do:** work out which `spiral-fitting` entry points can start on native
Windows before committing to a full run on this machine.

**What happened:** `get_ink_metrics.py` forces the fork start method at lines 137-140:

```python
try:
    multiprocessing.set_start_method('fork', force=True)
except RuntimeError:
    pass
```

On Windows `fork` does not exist, and `set_start_method` raises `ValueError`, not
`RuntimeError`, so the guard does not catch it:

```
>>> import multiprocessing, sys
>>> sys.platform, multiprocessing.get_all_start_methods()
('win32', ['spawn'])
>>> multiprocessing.set_start_method('fork', force=True)
ValueError: cannot find context for 'fork'
```

CPython raises it from `context.get_context()`, which is a `ValueError` subclass path — the
`RuntimeError` arm of `set_start_method` only fires when a context is already set.

**Shortest fix:** widen the guard to `except (RuntimeError, ValueError):`. The comment above it
already explains the intent is "restore fork where available", so falling through to the
platform default is the behaviour it wants.

**Honest scope note:** I verified the exception type on Windows (Python 3.14.5) but did not
reach this line through a real run — `get_ink_metrics.py` validates its `ink_dir` first and
needs `render_ink.py` output I do not have. So I am reporting it rather than opening a PR. If
you would rather see a PR with a full run behind it, say so and I will get there.

Bullo27:
Confirmed. Adding three things your report doesn't have: a run of the guard rather than a reading of it, the fact that the `except RuntimeError` arm is inert on *every* platform (not just Windows), and where the failure actually surfaces — which isn't at startup.

## Reproduced by running it

I have no Windows box, so I emulated the only thing that matters here: on win32 CPython builds `multiprocessing.context._concrete_contexts` with `spawn` as its sole key. I reduced it to exactly that (and swapped the default context to match), then executed **lines 137-140 verbatim as extracted from `main` at `de3c24943`** — not a paraphrase.

| guard variant | fork-less (emulated win32) | this Linux box (control) |
|---|---|---|
| as shipped | **`ValueError: cannot find context for 'fork'`, exit 1** | falls through, start method `fork`, exit 0 |
| `except (RuntimeError, ValueError):` | caught, start method `spawn`, exit 0 | `fork`, exit 0 |
[...]

PR #1585, opened 2026-08-24
get_ink_metrics: ask for fork only where fork exists

EazyHood:
**In one sentence:** the ink-metrics folds now start on platforms without `fork`, instead of
dying after the expensive prep with their traceback hidden in a log file.

**One real example:** Starting with `get_ink_metrics.py` as it is on `main`, running the guard
at lines 137-140 on Windows raises `ValueError: cannot find context for 'fork'` — which the
`except RuntimeError` does not catch.

**Before:**

```python
try:
    multiprocessing.set_start_method('fork', force=True)
except RuntimeError:
    pass
```

Two things are wrong with that guard, and the second is the one that bites:

1. `force=True` makes the `RuntimeError` arm unreachable. CPython raises `RuntimeError` only on
   the "context has already been set" path, which is guarded by `and not force`. Verified:
   calling `set_start_method('spawn', force=True)` twice raises nothing, while calling it twice
   without `force` raises `RuntimeError: context has already been set`. So the `try` is inert on
   every platform.
2. What a fork-less platform raises is `ValueError`, from `get_context`. On this Windows box
   `multiprocessing.get_all_start_methods()` is `['spawn']`, and the block above gives
   `ValueError: cannot find context for 'fork'`.

Where that surfaces matters: line 138 is inside `run_worker`, reached only through the
`__worker__` branch at line 668, which line 538 spawns. So the script does start — it lists the
tiles, concatenates the strip and writes `<name>_0000.png` — and only then does every fold die.
Line 544 redirects each fold's output to `logs/fold_N.log`, so the operator sees
`fold 0 (GPU 0) FAILED (rc=1)` and a log tail rather than the exception.

**After this PR:**

```python
if 'fork' in multiprocessing.get_all_start_methods():
    multiprocessing.set_start_method('fork', force=True)
```

**Proof:** the same block extracted from the file and executed on Windows, before and after:

```
old guard -> ValueError: cannot find context for 'fork'
new guard -> no exception, start method stays 'spawn'
```

Where `fork` exists the check passes and the call runs exactly as today, so the behaviour the
comment at 134-136 is protecting is unchanged.

**Why / where this is useful:** anyone running the ink metrics on a platform without `fork`
currently pays for the tile concatenation and the strip write, then loses every fold, and the
reason is not in the terminal.

**One trade this does not hide:** on a fork-less platform the folds now run under `spawn`, which
is exactly what the comment at 134-136 was avoiding — nnU-Net re-imports the module in each pool
worker. The module has a proper `if __name__ == '__main__'` guard, so that is safe, but it does
re-import torch per worker. I have not measured that cost: I have the GPU but not the
Dataset001_Ink model or a render to feed it. So this turns "every fold dies with a hidden
traceback" into "the folds run, possibly slower than intended on that platform". Better, but a
trade rather than parity.

      stated data.

## Details

Tested on Windows 11, Python 3.14.5. Diff is +4/-3 in
`spiral-fitting/get_ink_metrics.py`. Context for the discussion is in #1583, where @Bullo27
established the dead-`except` framing and the call-site location, both of which I re-checked
here.

The other three `set_start_method` sites outside `deprecated/` request `'spawn'`, which exists
everywhere, so their identical `except RuntimeError` shape is harmless — this is the only live
site asking for a start method that may not exist.


EazyHood:
This came out of #1583, so most of the reasoning is over there rather than repeated here.

Short version of how I got to it: I am on Windows and I was working out which parts of
spiral-fitting I could actually run before committing to anything long. This one looked fine —
it starts, it does the prep — and then every fold failed with nothing useful in the terminal.
The traceback is in the fold logs, which is where I finally found it.

@Bullo27 is the reason this PR uses the capability check rather than a wider `except`: they
pointed out the `except RuntimeError` cannot fire at all with `force=True`, which I had missed
and then confirmed. The trade at the end of the description is theirs too — I would rather ship
it with that stated than pretend the two start methods are equivalent.

Bullo27:
The fix itself is right, and I can add the arm you can't run: `test_guard_still_selects_fork_where_fork_exists` is `skipif`'d on Windows, so it had never executed until now. On Linux it passes.

But **the other test fails on Linux, including on Python 3.14** — the version `spiral-fitting/pyproject.toml` requires. Real pytest run, your branch `eee9072a2`, in a clean worktree:

```
$ python3.14 -m pytest tests/test_fork_start_method_guard.py -v
test_guard_does_not_raise_without_fork ............ FAILED
test_guard_still_selects_fork_where_fork_exists ... PASSED

>       assert result.stdout.strip() == "spawn"
E       AssertionError: assert 'forkserver' == 'spawn'
1 failed, 1 passed
```

## Why

The emulation *replaces* the default context:

```python
ctx._default_context = ctx.DefaultContext(ctx._concrete_contexts['spawn'])
```

but `multiprocessing.get_all_start_methods` and `multiprocessing.get_start_method` are module-level names bound **at import** to the bound methods of the *original* `DefaultContext` instance. Rebinding `ctx._default_context` doesn't touch them, so they keep answering from the object you replaced. Measured on both 3.14.6 and 3.13.11:

```
get_all_start_methods is bound to new ctx? False
get_start_method       is bound to new ctx? False
```

So on 3.14 the guard is skipped correctly, but the readout still reports the real default, `forkserver` — hence `'forkserver' != 'spawn'`.

## Fix: mutate the existing context instead of replacing it

```python
ctx._concrete_contexts = {'spawn': ctx._concrete_contexts['spawn']}
ctx._default_context._default_context = ctx._concrete_contexts['spawn']
ctx._default_context._actual_context = None
```

Same file, same run:

```
test_guard_does_not_raise_without_fork ............ PASSED
test_guard_still_selects_fork_where_fork_exists ... PASSED
2 passed
```

## And the control still fires

A green test only counts if it can go red. With the emulation fixed, I reverted `get_ink_metrics.py` to main's `try/except RuntimeError` guard and re-ran — `test_guard_does_not_raise_without_fork` fails with `ValueError: cannot find context for 'fork'`. So the corrected test is genuinely protecting your change rather than passing vacuously.

## One version caveat

On **≤3.13** the corrected emulation still isn't enough: `get_all_start_methods()` there doesn't consult the patched `_concrete_contexts` at all, so it keeps advertising `['fork', 'spawn', 'forkserver']` and the guard tries fork anyway. Measured on 3.13.11 and 3.12.3. Since `requires-python = ">=3.14"` this is moot for CI, but if you want the test to survive being run on an older interpreter, add `multiprocessing.get_all_start_methods = lambda: ['spawn']` to the emulation block too — with that line it passes on 3.12, 3.13 and 3.14.

Worth noting the failure mode is invisible from your side: on Windows the emulation is a no-op (there genuinely is no fork), so the first test passes trivially and the second is skipped. It only shows up on a fork-capable machine.

Nice touch extracting the guard from the source by locating the call and widening by indentation — that survives the `try` → `if` rewording, which a copied literal wouldn't.

---
*Claude (AI agent) on behalf of @Bullo27. Runs on Linux against your `eee9072a2` in a throwaway worktree, CPython 3.14.6 / 3.13.11 / 3.12.3; the pytest output above is quoted from the actual run, and the fixed-emulation result was produced by editing only the four emulation lines.*

EazyHood:
> The fix itself is right, and I can add the arm you can't run: `test_guard_still_selects_fork_where_fork_exists` is `skipif`'d on Windows, so it had never executed until now. On Linux it passes.
> 
> But **the other test fails on Linux, including on Python 3.14** — the version `spiral-fitting/pyproject.toml` requires. Real pytest run, your branch `eee9072a2`, in a clean worktree:
> 
> ```
> $ python3.14 -m pytest tests/test_fork_start_method_guard.py -v
> test_guard_does_not_raise_without_fork ............ FAILED
> test_guard_still_selects_fork_where_fork_exists ... PASSED
> 
> >       assert result.stdout.strip() == "spawn"
> E       AssertionError: assert 'forkserver' == 'spawn'
> 1 failed, 1 passed
> ```
> 
> ## Why
> The emulation _replaces_ the default context:
> 
> ```python
> ctx._default_context = ctx.DefaultContext(ctx._concrete_contexts['spawn'])
> ```
> 
> but `multiprocessing.get_all_start_methods` and `multiprocessing.get_start_method` are module-level names bound **at import** to the bound methods of the _original_ `DefaultContext` instance. Rebinding `ctx._default_context` doesn't touch them, so they keep answering from the object you replaced. Measured on both 3.14.6 and 3.13.11:
> 
> ```
> get_all_start_methods is bound to new ctx? False
> get_start_method       is bound to new ctx? False
> ```
> 
> So on 3.14 the guard is skipped correctly, but the readout still reports the real default, `forkserver` — hence `'forkserver' != 'spawn'`.
> 
> ## Fix: mutate the existing context instead of replacing it
> ```python
> ctx._concrete_contexts = {'spawn': ctx._concrete_contexts['spawn']}
> ctx._default_context._default_context = ctx._concrete_contexts['spawn']
> ctx._default_context._actual_context = None
> ```
> 
> Same file, same run:
> 
> ```
> test_guard_does_not_raise_without_fork ............ PASSED
> test_guard_still_selects_fork_where_fork_exists ... PASSED
> 2 passed
> ```
> 
> ## And the control still fires
> A green test only counts if it can go red. With the emulation fixed, I reverted `get_ink_metrics.py` to main's `try/except RuntimeError` guard and re-ran — `test_guard_does_not_raise_without_fork` fails with `ValueError: cannot find context for 'fork'`. So the corrected test is genuinely protecting your change rather than passing vacuously.
> 
> ## One version caveat
> On **≤3.13** the corrected emulation still isn't enough: `get_all_start_methods()` there doesn't consult the patched `_concrete_contexts` at all, so it keeps advertising `['fork', 'spawn', 'forkserver']` and the guard tries fork anyway. Measured on 3.13.11 and 3.12.3. Since `requires-python = ">=3.14"` this is moot for CI, but if you want the test to survive being run on an older interpreter, add `multiprocessing.get_all_start_methods = lambda: ['spawn']` to the emulation block too — with that line it passes on 3.12, 3.13 and 3.14.
> 
> Worth noting the failure mode is invisible from your side: on Windows the emulation is a no-op (there genuinely is no fork), so the first test passes trivially and the second is skipped. It only shows up on a fork-capable machine.
> 
> Nice touch extracting the guard from the source by locating the call and widening by indentation — that survives the `try` → `if` rewording, which a copied literal wouldn't.
> 
> _Claude (AI agent) on behalf of @Bullo27. Runs on Linux against your `eee9072a2` in a throwaway worktree, CPython 3.14.6 / 3.13.11 / 3.12.3; the pytest output above is quoted from the actual run, and the fixed-emulation result was produced by editing only the four emulation lines._



You are right, and thank you for running it — I could not have caught this here.

Fixed and pushed. The emulation now mutates the existing default context instead of replacing
it, with your `get_all_start_methods` line included so it also holds on <= 3.13.

I confirmed the mechanism on this box before changing anything, since it is platform-independent:

```
>>> multiprocessing.get_all_start_methods.__self__ is ctx._default_context
True
>>> ctx._default_context = ctx.DefaultContext(ctx._concrete_contexts['spawn'])
>>> multiprocessing.get_all_start_methods.__self__ is ctx._default_context
False
```

So the module-level names kept answering from the object I had replaced. The guard was skipped
correctly and the readout was not.

The part worth recording is your last paragraph, because it is the actual failure here: on
Windows the emulation is a no-op — there genuinely is no fork — so the first test passes
trivially and the second is skipped. I shipped a test I had no way of exercising, and reported
"validated both ways" on the strength of a run that could only ever go one way. The control I
ran did fire, but only against the guard, never against the emulation itself.

Re-verified after the fix, on Windows: with main's `try/except RuntimeError` restored,
`test_guard_does_not_raise_without_fork` fails with `ValueError: cannot find context for 'fork'`
again. That still leaves the Linux arm unverified from here, so if you have it in a worktree
already I would rather you confirm it than take my word for it.

Bullo27:
Confirmed on Linux — the corrected emulation is right. Ran your head `5a61899d5`, both arms.

## Green on the versions that matter

```
$ python3.14 -m pytest tests/test_fork_start_method_guard.py -v     # CPython 3.14.6
test_guard_does_not_raise_without_fork ............ PASSED
test_guard_still_selects_fork_where_fork_exists ... PASSED
2 passed

$ python -m pytest tests/test_fork_start_method_guard.py -v         # CPython 3.13.11
2 passed
```

3.13 is the arm that actually exercises your `get_all_start_methods` override, since that's the version where `_concrete_contexts` isn't consulted — so the belt-and-braces is doing real work, not sitting inert. (I couldn't add 3.12: the system interpreter here has no `pip` and I didn't want to start installing into it. 3.13 covers the ≤3.13 path.)

## The control fires on both

With `main`'s `try/except RuntimeError` restored over your `get_ink_metrics.py`, same test file, same command:

```
3.14.6   test_guard_does_not_raise_without_fork  FAILED
3.13.11  test_guard_does_not_raise_without_fork  FAILED
     E   ValueError: cannot find context for 'fork'
     E   assert 1 == 0
```

and `test_guard_still_selects_fork_where_fork_exists` stays PASSED in the control — correct, since main's guard handles the fork-present arm fine and only the fork-less arm is the defect. Restoring your version returns it to `2 passed`. So on Linux the test discriminates in exactly the direction it's supposed to, and it can go red.

## Your extractor really is spelling-agnostic

Worth recording because it's the property that stops the test drifting from the code. I pointed `_extract_guard()` at both versions and printed what it pulled:

```
PR head (if-form)              main (try/except form)
if 'fork' in multiprocessing…  try:
    multiprocessing.set_st…        multiprocessing.set_st…
                               except RuntimeError:
                                   pass
```

Both extracted cleanly, which is why the control above works at all — the harness picked up main's differently-shaped guard without modification.

## Still merges clean

`git merge-tree --write-tree origin/main 5a61899d5` → rc 0 against current `main` (`54814ac6a`); merge-base `de3c24943`. Only the two files.

---

On the wider point in your note — I'd rather not let you carry that alone. I closed an issue of my own tonight for a version of the same error: I measured carefully and then asserted a conclusion my measurements didn't actually license. Shipping a test you couldn't exercise and then noticing it yourself is the better half of that trade.

---
*Claude (AI agent) on behalf of @Bullo27. Measured against PR head `5a61899d5` with `main` at `54814ac6a`, on CPython 3.14.6 and 3.13.11, Linux. Both files extracted from git at that SHA and run outside the repo tree, so nothing in the checkout was modified.*

EazyHood:
Confirmed here too, and I owe you a correction on the one sentence you were kind about.

I told you I had no way of exercising the Linux arm on this box. That was wrong. Docker
Desktop was already installed here and running Linux containers; I had simply never
checked. So the honest version is not "I could not run it", it is "I decided I could not
run it without spending two minutes finding out". I ran it tonight in throwaway
containers off `5a61899d5`, both files extracted from git and executed outside the
checkout, so nothing in the tree was touched.

## Your result reproduces, on three interpreters

| CPython | PR head | control: `main`'s `try/except RuntimeError` |
|---|---|---|
| 3.14.7 | 2 passed | `test_guard_does_not_raise_without_fork` FAILED, `ValueError: cannot find context for 'fork'` |
| 3.13.15 | 2 passed | same failure |
| 3.12.14 | 2 passed | same failure |

`test_guard_still_selects_fork_where_fork_exists` stays PASSED in every control, which is
correct: main's guard only breaks the fork-less arm. The 3.14 and 3.13 patch releases are
different from the ones you ran, and 3.12.14 covers the arm you could not add.

## The `get_all_start_methods` line, checked by deleting it

You said the override is doing real work on <= 3.13 rather than sitting inert. I did not
want to take that on reasoning, so I removed just that line and re-ran the same three:

| CPython | with the line removed |
|---|---|
| 3.14.7 | 2 passed, so it is genuinely inert there |
| 3.13.15 | fork-less arm FAILED |
| 3.12.14 | fork-less arm FAILED |

And the reason, printed inside 3.13 right after the emulation block:

```
get_all_start_methods() after emulating: ['fork', 'spawn', 'forkserver']
```

The guard therefore tries fork, `_concrete_contexts` no longer has it, and it raises. That
is your diagnosis, confirmed by mutation rather than by argument: belt and braces on 3.14,
load bearing on 3.12 and 3.13.

## Still merges clean

`git merge-tree --write-tree origin/main HEAD` returns rc 0 against `54814ac6a`, merge base
`de3c24943`, +129/-3 across the two files.

On your closing paragraph: taken, and it lands better for not being softened. The part I
want to keep from this is narrower than "I shipped a test I could not exercise". It is that
I asserted a limitation of my own setup without testing it, in the same comment where I was
apologising for asserting a result without testing it. That one is cheap to fix.

bruniss:
happy to merge if you remove the giant test

EazyHood:
> happy to merge if you remove the giant test

Removed it — the PR is now just the four-line change in `get_ink_metrics.py`.

@Bullo27, sorry to drop the file after you ran both arms on Linux; your run is what confirmed the
corrected emulation, and if the guard ever regresses the test is in this thread's history rather
than gone.

PR #1587, opened 2026-08-24
ink_detection: make cache_max_gb bound the cache directory, not each process

### LINKED ISSUE #1586: ink_detection: the on-disk volume cache settles at ~2x --cache-max-gb because the budget is enforced twice
**In one sentence:** `--cache-max-gb` / `volume_cache_max_gb` is enforced twice by two accountants that cannot see each other's books, so the ink-detection on-disk volume cache settles at roughly **twice** the budget you ask for.

**I was trying to:** Screen PHerc1203 for First Letters with the released `scrollprize/ink_9um` checkpoints — pull `tifxyz` segments, render surface volumes off the 9.362 µm masked scroll volume straight from the public S3 bucket, then run ink inference over the rendered stack. Rendering a single `auto_grown` segment reads tens of thousands of chunks out of a 6844×6844×18977 volume, so I turned on the documented on-disk chunk cache (`--cache-dir` / `--cache-max-gb`, `vesuvius/docs/ink_detection.md:482-489`) rather than re-fetch the same chunks for every mesh tile.

An earlier pass of mine filled the pod's 20 GB `/workspace` and died with `OSError: [Errno 28] No space left on device` — my own render wrapper had left its cache budget at its 100 GB default, which is on me, not on villa. That is what sent me to instrument the budget. On the rerun I set a deliberately conservative `--cache-max-gb 4` and instrumented the cache directory between passes — and the directory does not stop at 4 GB.

**Using:**

| | |
|---|---|
| villa | `volume_io.py` as of `37e300d3`; added by #1456 (`960d76ac`, 2026-08-14) and unchanged since, so the pod — a shallow clone of `main` on 2026-08-17 — and the workstation ran identical code. |
| Where it first bit | Ubuntu 22.04 RunPod pod, `/workspace` 20 GB, `--cache-max-gb 4` |
| Where the small numbers below were measured | Windows 11 (10.0.26200), same bucket, same volume, 30,000,000 B budget |
| Python / zarr (workstation, where the small numbers below were measured) | 3.14.7 / 3.3.0 |
| s3fs / fsspec / aiobotocore (workstation) | 2026.7.0 / 2026.7.0 / 3.9.0 |
| zarr / aiobotocore (pod, where it first bit) | 3.2.1 / 3.7.0 — `cache_store.py`'s LRU-init path is identical in 3.2.1 and 3.3.0 |
| Data | `s3://vesuvius-challenge-open-data/PHerc1203/volumes/20250820131727-9.362um-1.2m-113keV-masked.zarr`, level 0, uint8, chunks `128×128×128` (2,097,152 B/chunk), read anonymously |

The documented invocation this budget pattern comes from (`vesuvius/docs/ink_detection.md:482-489`), with my 4 GB budget substituted — this is the shape the docs tell users to run, not a transcript of my own run. My pod measurements came from a small wrapper of mine calling the same `open_volume(..., cache_dir=..., cache_max_gb=4)` boundary:

```bash
uv run --extra models python -m vesuvius.ink_detection.inference.infer_full3d_tifxyz \
  /workspace/segments/<segment> \
  /workspace/ink_9um/checkpoints/native-best.pth \
  /workspace/predictions/<segment>.ome.zarr \
  --tta \
  --cache-dir /workspace/volcache \
  --cache-max-gb 4 \
  --gpus 0
```

**What happened:** The cache directory overshoots the budget permanently and never comes back under it. Re-opening a directory that is at or under budget costs a further full budget on top of what is already there — 1.96× measured below. From the next run on it holds a steady state and stays there: 1.82× (54,525,952 B / 26 files) on the harness below, byte-identical over five further consecutive runs. It is a steady state, not a transient, and every process that reuses the directory re-establishes it.

On the pod, one `auto_grown` segment render at a 4,000,000,000 B budget:

```
pass 1: 3,999,538,943 B / 1,907 files      (1.00x — correct)
pass 2: 5,979,366,890 B / 2,854 files      (1.49x)
```

Reproduced in isolation on a second machine at a 30,000,000 B budget, run twice against the same cache directory (verbatim run 2):

```
before open: (29360128, 14) budget: 30000000
after open : (29363409, 17)
  read 0: on_disk=(58720256, 28) CacheStore.current_size=29360128 tracked_keys=14
  read 1: on_disk=(58720256, 28) CacheStore.current_size=29360128 tracked_keys=14
  read 2: on_disk=(58720256, 28) CacheStore.current_size=29360128 tracked_keys=14
```

**58,720,256 bytes on disk against a 30,000,000-byte budget — 1.96×** — while the cache reports itself as holding 29,360,128 bytes / 14 keys and the directory holds 28 files. This is a single process with a single reader; `README.md:36-41` already notes that *concurrent* workers "can cause a brief shared cache-directory overshoot", but this is neither concurrent nor brief.

**What I expected or needed:** The directory bounded by the budget — which is what the docs describe. `vesuvius/src/vesuvius/ink_detection/README.md:36-41` says the directory is "swept oldest-first to 90% when opened, then Zarr `CacheStore` applies a process-local LRU", a model that predicts a ceiling of ~1× the budget.

All three places that suggest a number are affected. `vesuvius/docs/ink_detection.md:488` shows `--cache-max-gb 100` in the copy-paste example, which on this mechanism occupies ~190 GB; the tutorial ships `"volume_cache_max_gb": 120` (`scrollprize.org/docs/07_tutorial5.md:327`) and repeats the same 120 on the CLI path at `07_tutorial5.md:385`, which occupies ~228 GB. A user who provisions exactly what the docs told them to ask for runs out of disk part-way through a long render.

**Evidence / reproduction:** Self-contained, needs only the public bucket. Run it **twice** — run 1 is a cold cache and behaves correctly, run 2 is the warm-cache case that overshoots.

```python
# budget_repro.py
import os, sys
from pathlib import Path
from vesuvius.ink_detection.volume_io import open_volume, read_bbox_with_padding

VOL = ("s3://vesuvius-challenge-open-data/PHerc1203/volumes/"
       "20250820131727-9.362um-1.2m-113keV-masked.zarr")
CACHE, BUDGET_GB = sys.argv[1], 0.03            # 30,000,000 byte budget

def disk(p):
    total = files = 0
    for d, _, fs in os.walk(p):
        for f in fs:
            try:
                total += (Path(d) / f).stat().st_size; files += 1
            except OSError:
                pass
    return total, files

print("before open:", disk(CACHE), "budget:", int(BUDGET_GB * 1e9))
vol = open_volume(VOL, 0, cache_dir=CACHE, cache_max_gb=BUDGET_GB)
print("after open :", disk(CACHE))
cz, cy, cx = (int(c) for c in vol.chunks[:3])
for i in range(3):                               # papyrus-bearing region
    z0, y0, x0 = 5000 + i * 260, 3400 + i * 260, 3400 + i * 260
    read_bbox_with_padding(vol, (z0, y0, x0, z0 + 2*cz, y0 + 2*cy, x0 + 2*cx), fill_value=0)
    print(f"  read {i}: on_disk={disk(CACHE)} "
          f"CacheStore.current_size={vol.store.cache_info()['current_size']} "
          f"tracked_keys={vol.store.cache_info()['cached_keys']}")
```

```
python budget_repro.py /tmp/volcache      # run 1 (cold)
python budget_repro.py /tmp/volcache      # run 2 (warm)  <-- overshoots
```

### Mechanism

`vesuvius/src/vesuvius/ink_detection/volume_io.py` (`open_volume_root`) hands the *same* budget to two independent enforcement paths:

| file:line (`37e300d3`) | code | effect |
|---|---|---|
| `volume_io.py:117-119` | `maximum_bytes = int(float(cache_max_gb) * 1e9)` | the budget, in bytes |
| `volume_io.py:124-127` | `snapshot = _cache_snapshot(cache_path)` … `_evict_to_watermark(snapshot, maximum_bytes)` | villa sweeps the directory to 0.9 × budget **at open, and only if it is already over budget** (the `sum(...) > maximum_bytes` guard on line 126); an at- or under-budget directory is left untouched |
| `volume_io.py:138-142` | `CacheStore(store=…, cache_store=LocalStore(cache_path), max_size=maximum_bytes)` | the **same** budget is handed to zarr's `CacheStore` |
| `zarr/experimental/cache_store.py:125` | `self._state = _CacheState()` | `CacheStore` builds its LRU in memory and starts it **empty** in every new process; it never scans `cache_store` on construction |
| `zarr/experimental/cache_store.py:196-226` (`_track_entry`) | — | only keys written *by this process* ever enter `_state.key_sizes` / `_state.cache_order` |
| `zarr/experimental/cache_store.py:228-232` (`_update_access_order`) | `if entry_key in self._state.cache_order:` | a cache **hit** on a file left by an earlier run is counted in `hits` but is a no-op for LRU tracking |
| `zarr/experimental/cache_store.py:155-167` (`_accommodate_value`) | `while … and self._state.cache_order:` | eviction only ever considers this process's own keys |

So the `0.9 × budget` of files the sweep deliberately retained is invisible to the store that is supposed to enforce the budget. The new process spends a *further* full `max_size` on top of them before its first eviction, and because `_update_access_order` no-ops on inherited files, reading an inherited chunk never registers it either. There are two regimes. While the directory is still at or under budget the open-time sweep does not fire at all — it is guarded by `sum(...) > maximum_bytes` on `volume_io.py:126` — so the process inherits everything already there and adds a full budget on top; that is the 1.96× above. Once the directory is over budget the sweep cuts it to the 0.9 watermark and the process again adds a full budget on top: steady state is `0.9 × budget` retained plus `1.0 × budget` freshly written ≈ **1.9× the requested budget**. The plateau is `⌊0.9×budget/chunk⌋ + ⌊budget/chunk⌋` chunks, so it measures 1.82× on the 30 MB harness below, where one chunk is 7% of the budget, and approaches 1.9× at the budgets the docs recommend.

Two consequences worth naming separately:

1. **Disk.** The number in the config is not the number on disk.
2. **Cache quality.** Inherited files are ranked only by mtime and reuse never promotes them, so the next open-time sweep evicts oldest-first regardless of how hot a chunk was during the run.

Pass 2 on the pod reads 1.49× rather than the 1.9× plateau because a real render's chunk set does not sweep the volume cleanly: the second pass re-requests many chunks the first pass still holds (cheap hits that write nothing) while adding new ones, so the directory climbs toward the plateau over several passes instead of jumping to it. The 30 MB harness, which does read the same set twice, jumps straight past the plateau to 1.96× on its second run and settles onto its 1.82× plateau from the third run on.

### Fix

A PR is open against this issue. It makes the open-time sweep hand the files it retained to the `CacheStore` this process just built, registering them oldest-first in that store's LRU before the first read goes through it. `max_size` then stays at the full `cache_max_gb` and means what the docs say — a bound on the directory, not an allowance per process — and reusing a chunk inherited from an earlier run promotes it properly instead of leaving it ranked by download order.

Same cache directory, same reads, same 30,000,000 B budget:

| | directory before open | after 3 reads | vs budget |
|---|---|---|---|
| current `main` | 29,360,128 B / 14 files | **58,720,256 B / 28 files** | **1.96×** |
| with the fix | 29,360,128 B / 14 files | 29,360,128 B / 14 files | 0.98× |

Crop means are identical across both (`114.8 / 109.0 / 104.5` from the same papyrus region), so only the eviction schedule differs. The patched opener also *recovers* an already-overshot directory: pointed at the 58,720,256 B directory `main` had just produced, it swept and seeded to 25,165,824 B and finished the same three reads at 29,360,128 B in one open.

Honest about the cost: there is no public seam for this. `cache_info()` is read-only and `CacheStore.__init__` takes no starting state, so the seed writes to `CacheStore._state`. In the PR that access is guarded — if the expected attributes are absent the seed is skipped and `open_volume_root` raises a `RuntimeWarning` naming the zarr version and the budget it can no longer enforce, so it degrades to today's behaviour loudly rather than to a silently unbounded cache. The proper home for this is `CacheStore.__init__` scanning its own `cache_store`; I am reporting that upstream separately, and once it lands the villa-side seed becomes a no-op that can be deleted.

I also tried the more obvious shape — keep the sweep and give `CacheStore` only `max(budget - retained, 0)` — and rejected it: it shrinks the working set exactly when the cache is warm, and at the bottom it stops *evicting* rather than stops *caching*. The PR carries that comparison in full.

### Two smaller things in the same area, happy to fold in or leave out

1. **`--cache-max-gb` is not range-checked.** In `vesuvius/src/vesuvius/ink_detection/inference/infer_full3d_tifxyz.py` it is declared at line 150 but is not covered by the `checks` block at 153-165 that validates the other numeric flags, so `--cache-max-gb -5` gets past argparse and only fails once the volume is opened, inside `open_volume_root` (`volume_io.py:120-121`), after the checkpoint and config have already been loaded. (No argument in that parser carries `help=` text at all, so that is a whole-file gap rather than something specific to these two flags.)

2. **An upstream `CacheStore` bug this budget path can trip.** In `zarr/experimental/cache_store.py:279-293`, `_cache_miss` writes the value to the cache store and *then* calls `_track_entry`; the byte-range branch rolls the insertion back when the value is too large (286-293), but the full-key branch (280-282) does not. If `max_size` is ever smaller than a single chunk, every chunk is written to disk and never tracked or evicted. Measured on this volume with a 1,000-byte budget: 165,678,289 bytes on disk after three reads, `current_size` = 263, `evictions` = 0, still climbing linearly. That one is zarr's, not villa's — I'm reporting it there — but this code path has no floor under it if a user sets a very small `--cache-max-gb`.

This came out of screening PHerc1203 for the First Letters prize — rendering every published segment against the 9.362 µm volume straight from S3, where the chunk cache is the difference between a $5 screen and a $50 one. On 20 GB pod disks I would rather the budget be honest than large: a 4 that quietly means 8 is how a screening run dies at 90%.

---

*Disclosure: I used Claude (Anthropic) to help investigate this and draft the text. The runs, the measurements, the before/after numbers and the fix were produced and checked in my own sessions against real scroll data, and I have reviewed and edited everything above.*

flummoxjr:
Fixes #1586.
**In one sentence:** Makes `cache_max_gb` bound the ink-detection cache *directory* instead of granting each process its own full budget, so the number in your config is the number on disk.

**One real example:** Starting with the 5,979,366,890-byte cache directory a real PHerc1203 `auto_grown` render had just produced under a `--cache-max-gb 4` budget, I re-ran the same render with this PR applied, and it produced a **3,999,538,943-byte** directory in a single open — the render exited 0 and the surface volume was intact (shape `(21, 4639, 5359)`, mid-slice mean 77.04 DN, 65.6% non-zero).

**Before:** The budget is enforced twice, by two accountants who cannot see each other's books. `open_volume_root` sweeps the cache directory to 0.9 × budget from its own filesystem snapshot, then hands the *same* budget to zarr's `CacheStore`, whose LRU ledger is in-memory and starts empty in every process. The retained 90% is invisible to the store that is supposed to enforce the budget, so the process spends a further full `max_size` on top of it before its first eviction. Steady state ≈ **1.9× the requested budget**, re-established by every process that reuses the directory. Reading an inherited chunk does not even register it, because `_update_access_order` no-ops on keys it has never seen.

**After this PR:** The sweep hands the files it retained to the `CacheStore` this process just built, registering them oldest-first in that store's LRU before the first read goes through it. `max_size` stays at the full `cache_max_gb` and means what the docs already say it means: a bound on the directory, not an allowance per process. Reuse of an inherited chunk promotes it properly, so eviction follows actual access rather than original download order. The patched opener also **recovers** a directory that is already over budget, rather than merely preventing the next overshoot.

**Proof:** Three measurements on this fix's before/after boundary. **Sources are labelled — they are not the same machine or the same data.**

| # | source | data | budget | before (on disk) | vs budget | after | vs budget |
|---|---|---|---|---|---|---|---|
| 1 | **RunPod pod**, Ubuntu 22.04, `/workspace` 20 GB — *real data* | one `auto_grown` PHerc1203 segment render, `s3://vesuvius-challenge-open-data/PHerc1203/volumes/20250820131727-9.362um-1.2m-113keV-masked.zarr` | 4,000,000,000 B | pass 1: 3,999,538,943 B / 1,907 files<br>pass 2: **5,979,366,890 B / 2,854 files** | 1.00×<br>**1.49×** | same dir, same render, same budget: **5,979,366,890 B → 3,999,538,943 B / 1,907 files** in a single open | **1.00×** |
| 2 | **workstation**, same volume over anonymous S3 — *also real data* | same PHerc1203 volume, level 0, papyrus-bearing region, 3 bbox reads | 30,000,000 B | 29,360,128 B / 14 files at open → **58,720,256 B / 28 files** after the reads | **1.96× (first warm re-open)** | 29,360,128 B / 14 files at open → 29,360,128 B / 14 files after the reads | **0.98×** |
| 3 | **local synthetic**, no network | 40-chunk uncompressed `LocalStore` volume in a temp dir, 4 sequential re-opens | 81,920 B | pass 1: 81,920 B<br>pass 2: 163,840 B<br>pass 3-4: 155,648 B | 1.00×<br>**2.00×**<br>**1.90× (steady)** | 81,920 B on every pass | 1.00× |

Rows 1 and 2 are both real scroll data at budget scales three orders of magnitude apart — 4 GB on the pod that actually filled, 30 MB on a workstation where the arithmetic is small enough to read at a glance. They agree on the mechanism and bracket the effect. Row 3 removes the network and makes the chunk sizes exact; it is what the regression test in this PR asserts.

Row 2 is the cleanest apples-to-apples — **same starting directory, same reads, same budget**: `main` takes 29,360,128 B to 58,720,256 B, the patched opener leaves it at 29,360,128 B. Crop means are identical across both (`114.8 / 109.0 / 104.5`), so only the eviction schedule differs. Voxel data is unchanged in all three; row 3's harness asserts byte-equality against the source on every pass.

**Why / where this is useful:** Anyone rendering or inferring off an S3-backed scroll volume with `--cache-dir` turned on — which the docs recommend for exactly this workload — currently has to provision twice the disk they asked for, and a long render that dies at 90% costs far more than a smaller cache does.

It also makes both documented numbers honest. `vesuvius/docs/ink_detection.md:488` puts `--cache-max-gb 100` in the copy-paste example (~190 GB on this mechanism), and `scrollprize.org/docs/07_tutorial5.md:327` ships `"volume_cache_max_gb": 120`, with line 349 describing it as "an on-disk LRU cache (capped at `volume_cache_max_gb`)". On `main` that cap is per process against an empty ledger, so following the tutorial on a 128 GB disk occupies ~228 GB and fails. With this PR the documented sentence becomes true and neither example needs its number changed, which is why I left them alone.

### Mechanism

| file:line (`37e300d3`) | code | effect |
|---|---|---|
| `volume_io.py:117-119` | `maximum_bytes = int(float(cache_max_gb) * 1e9)` | the budget, in bytes |
| `volume_io.py:124-127` | `snapshot = _cache_snapshot(cache_path)` … `_evict_to_watermark(snapshot, maximum_bytes)` | sweeps the directory to 0.9 × budget at open |
| `volume_io.py:138-142` | `CacheStore(store=…, cache_store=LocalStore(cache_path), max_size=maximum_bytes)` | the same budget handed to zarr's `CacheStore` |
| `zarr/experimental/cache_store.py:125` | `self._state = _CacheState()` | LRU ledger starts empty in every process; never scans `cache_store` |
| `zarr/experimental/cache_store.py:196-226` | `_track_entry` | only keys written by this process enter the ledger |
| `zarr/experimental/cache_store.py:228-232` | `if entry_key in self._state.cache_order:` | a hit on an inherited file is unrankable |
| `zarr/experimental/cache_store.py:155-167` | `_accommodate_value` | eviction only considers this process's own keys |

### The change

- `_evict_to_watermark` returns the surviving snapshot entries instead of a byte total (no caller used the total).
- New `_seed_cache_store_lru(store, cache_path, snapshot)` fills `cache_order`, `key_sizes`, `key_insert_times` and `current_size` from those entries, mapping each path back to its store key with `path.relative_to(cache_path).as_posix()` (`LocalStore` stores key `k` at `root / k`, `zarr/storage/_local.py:213`).
- `open_volume_root` seeds **before** `zarr.open()`, because `zarr.open()` itself fetches metadata through the store and those writes must be charged against the bytes already on disk.

| file | change |
|---|---|
| `vesuvius/src/vesuvius/ink_detection/volume_io.py` | `+84 / -3`: `_evict_to_watermark` returns survivors; new `_seed_cache_store_lru`; `open_volume_root` seeds and warns |
| `vesuvius/tests/ink_detection/test_volume_io.py` | `+112`: three regression tests |
| `vesuvius/docs/ink_detection.md` | the cache paragraph now describes the seed and is honest about the concurrent case |
| `vesuvius/src/vesuvius/ink_detection/README.md` | same, in short form |

No new dependencies, no public-flag changes, no change to `cache_max_gb` semantics other than making them true.

### Why this design, and not the headroom variant

The obvious alternative — keep the sweep and give `CacheStore` only `max(budget - retained, 0)` — makes the on-disk invariant true too, but buys it with two real regressions:

- **It shrinks the working set exactly when the cache is warm.** The sweep leaves 90% of the budget in place, so headroom collapses to 10% of the budget; to get usable headroom back you have to delete cache you were about to reuse (retaining half the budget means re-downloading up to 60 GB per process start at the tutorial's `120`). A warm start would get a *smaller* working set than a cold one, which is backwards.
- **At the bottom it silently disables eviction, not caching.** With `max_size` below one value, `_cache_miss` writes to the cache store at `zarr/experimental/cache_store.py:281` and *then* calls `_track_entry`, and unlike the byte-range branch at 286-293 the full-key branch does not roll back. Every chunk lands on disk, none is tracked, `evictions` stays 0, and the directory grows without limit. (Measured separately at a 1,000-byte budget: 165,678,289 B on disk after three reads, `evictions=0`, still climbing.) The headroom design walks toward that cliff by construction, since headroom shrinks as the cache fills.

Evicting far enough to free a whole budget avoids both, but throws away the cross-run reuse the disk cache exists to provide — the entire reason `--cache-dir` is worth turning on for an S3-backed render. Seeding keeps the full budget usable, keeps the warm cache, and fixes the LRU-promotion bug as a side effect.

### The part that is not villa's to fix

`_seed_cache_store_lru` writes to `CacheStore._state` because zarr exposes no public seam for pre-populating the LRU — `cache_info()` is read-only and `__init__` takes no starting state. The correct home for this is `CacheStore.__init__` scanning its `cache_store` (or accepting a seed callback); I have filed that upstream as zarr-developers/zarr-python#4282.

Until that lands the access is guarded rather than assumed: `_seed_cache_store_lru` returns `False` if the expected attributes are absent, and `open_volume_root` raises a `RuntimeWarning` naming the installed zarr version and the budget it can no longer enforce. It degrades to today's behaviour, loudly — never to a silently disabled or silently unbounded cache. If the upstream fix lands, this seed becomes a harmless no-op re-assert of values zarr already computed, and can be deleted.

**Known limitation, now stated honestly in the docs rather than as "brief":** several workers sharing one cache directory *at the same time* each enforce the budget against their own view, so a concurrent fleet can still overshoot until the next open-time sweep. This PR fixes the sequential case — the train-then-infer and render-then-rerun patterns that actually filled the disk.

### Test evidence

Run from the repo root with a Zarr-3 environment (these paths skip on Zarr 2, which cannot use the disk cache at all):

```
python -m pytest vesuvius/tests/ink_detection/test_volume_io.py -q
```

| test | asserts |
|---|---|
| `test_reopening_reuses_one_budget_instead_of_granting_a_new_one` | three sequential opens of a 40-chunk volume with a 20-chunk budget; after each, `0 < on_disk <= cache_info()["current_size"] <= budget`, i.e. the directory is bounded by the store's own ledger and the ledger by the budget. Reads are asserted byte-identical to the source on every pass |
| `test_seeded_entries_evict_oldest_first_and_survive_reuse` | seeded entries carry mtime order (the older file is the lexicographically *later* one, so a path-ordered seed would evict the wrong one); reading a seeded entry promotes it, and the next admission evicts the other |
| `test_unseedable_cache_store_is_reported_not_silently_overfilled` | the guard returns `False` on a store with no `_state`, and `open_volume_root` raises `RuntimeWarning` instead of proceeding quietly |

The three new tests against **unpatched** `volume_io.py`:

```
FAILED test_reopening_reuses_one_budget_instead_of_granting_a_new_one
E       assert 163840 <= 81920
FAILED test_seeded_entries_evict_oldest_first_and_survive_reuse
FAILED test_unseedable_cache_store_is_reported_not_silently_overfilled
3 failed, 13 deselected
```

With this PR:

```
3 passed, 13 deselected in 3.08s
```

Whole-file and neighbours, on this branch rebased onto `37e300d3`:

| suite | result |
|---|---|
| `vesuvius/tests/ink_detection/test_volume_io.py` | 13 passed, 2 failed, 1 skipped |
| `vesuvius/tests/data/test_chunk_cache.py` + `vesuvius/tests/neural_tracing/test_disk_cache_store.py` | 7 passed, 1 skipped |
| `test_data_foundation.py`, `test_dependency_wiring.py`, `test_flat_inference.py`, `test_native_full3d_inference.py` | 53 passed |

The 2 failures are **pre-existing and Windows-only artifacts of my measurement machine**, present identically before and after this change; villa targets Ubuntu/macOS (`AGENTS.md §1.5`):

- `test_open_sweep_prunes_oldest_recursively_and_leaves_under_budget` sets mtimes with `os.utime(path, ns=(1,1))`/`(2,2)`/`(3,3)`; NTFS keeps 100 ns granularity anchored at 1601, so all three collapse to `st_mtime_ns == 0` and the eviction order falls through to the path tiebreak. Fails deterministically.
- `test_shared_cache_multiprocess_reads_are_not_torn` fails with `PermissionError(13)`: Windows will not replace a file another process has open, which is what `LocalStore`'s `.partial` install does. This one is **flaky rather than deterministic** — it passed in 1 of 3 runs here.

The new tests avoid the first trap deliberately: they set mtimes at whole-second resolution, with a comment saying why.

> I hit this while screening PHerc1203 for First Letters on 20 GB pods, where a render that dies at 90% costs far more than a smaller cache does. I would rather the budget be honest than large.

---

*Disclosure: I used Claude (Anthropic) as a coding assistant to investigate this and draft the text and the patch. The pod runs, the measurements, the before/after numbers and the design decision were produced and checked in my own sessions against real scroll data, and I have reviewed and edited everything above.*

### DIFF
--- a/vesuvius/docs/ink_detection.md
+++ b/vesuvius/docs/ink_detection.md
@@ -90,13 +90,19 @@ not totals across the cache root.
 
 When a volume opens with a budget, the code scans that volume's cache
 subdirectory once. An over-budget directory is pruned oldest-mtime-first to at
-most 90% of the budget; an at- or under-budget directory is left alone. Zarr's
-CacheStore then applies a process-local steady-state LRU. Multiple workers may
-share the directory safely because LocalStore installs complete values
-atomically, but independent process accounting means the directory can briefly
-overshoot its budget before a later open-time sweep. Every process constructs
-its own source store after it starts; a remote fsspec store is never inherited
-as the active transport across a fork.
+most 90% of the budget; an at- or under-budget directory is left alone. The
+files that survive the scan are then registered, oldest first, in the LRU that
+Zarr's CacheStore builds for this process. Zarr starts that LRU empty in every
+new process and never scans the cache directory itself, so without the seed a
+re-open would spend a second full budget on top of whatever the sweep retained;
+with it, one budget bounds the directory across sequential runs, and reusing a
+file inherited from an earlier run refreshes its LRU position instead of
+leaving it ranked by its original mtime. Multiple workers may share the
+directory safely because LocalStore installs complete values atomically, but
+each worker still enforces the budget against its own view, so a directory
+written by several simultaneous processes can overshoot until the next
+open-time sweep. Every process constructs its own source store after it starts;
+a remote fsspec store is never inherited as the active transport across a fork.
 
 ## Shipped aligned-corpus configs
 
--- a/vesuvius/src/vesuvius/ink_detection/README.md
+++ b/vesuvius/src/vesuvius/ink_detection/README.md
@@ -36,9 +36,10 @@ Paste volume paths in their correct local, `s3://`, or `https://` form; the
 package does not rewrite URLs. The optional compressed-chunk disk cache requires
 Zarr 3; uncached reads support Zarr 2.18.7 and Zarr 3. Its budget is per volume:
 an over-budget volume subdirectory is swept oldest-first to 90% when opened,
-then Zarr CacheStore applies a process-local LRU. Concurrent workers use atomic
-LocalStore installs, but their independent accounting can cause a brief shared
-cache-directory overshoot.
+and the files that survive the sweep are registered in this process's Zarr
+CacheStore LRU so the budget bounds the directory rather than each process.
+Concurrent workers use atomic LocalStore installs, but their independent
+accounting can still cause a shared cache-directory overshoot while they run.
 
 ## Shipped configs
 
--- a/vesuvius/src/vesuvius/ink_detection/volume_io.py
+++ b/vesuvius/src/vesuvius/ink_detection/volume_io.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 import hashlib
 import json
 import os
+import time
+import warnings
 from pathlib import Path
 from typing import Any
 
@@ -39,9 +41,15 @@ def _cache_snapshot(cache_dir: Path) -> list[tuple[int, int, Path]]:
 
 def _evict_to_watermark(
     snapshot: list[tuple[int, int, Path]], max_bytes: int
-) -> int:
+) -> list[tuple[int, int, Path]]:
+    """Delete oldest entries until the snapshot fits under the watermark.
+
+    Returns the entries that survived, still sorted oldest-mtime-first, so the
+    caller can hand exactly the retained set to ``_seed_cache_store_lru``.
+    """
     total = sum(size for _, size, _ in snapshot)
     target_bytes = 0.9 * max_bytes
+    evicted = 0
     for _, size, path in snapshot:
         if total <= target_bytes:
             break
@@ -50,7 +58,63 @@ def _evict_to_watermark(
         except FileNotFoundError:
             pass
         total -= size
-    return total
+        evicted += 1
+    return snapshot[evicted:]
+
+
+def _seed_cache_store_lru(
+    store: Any, cache_path: Path, snapshot: list[tuple[int, int, Path]]
+) -> bool:
+    """Register already-cached files in a fresh CacheStore's LRU accounting.
+
+    ``CacheStore`` builds its LRU bookkeeping in memory and starts it empty in
+    every new process: it never scans ``cache_store`` on construction, so files
+    left by an earlier run are invisible to it. It therefore writes a full
+    ``max_size`` of new data on top of whatever the open-time sweep retained,
+    and a hit on a pre-existing file neither counts toward the budget nor
+    refreshes that file's LRU position. Seeding the accounting from the files
+    that are actually on disk makes ``max_size`` a true bound on the cache
+    directory rather than a per-process allowance, and lets reuse of an
+    inherited entry keep it alive.
+
+    The alternative of shrinking ``max_size`` to the remaining headroom is
+    worse: the sweep leaves 90% of the budget in place, so headroom collapses
+    to 10% (or to 0 when the directory is already at budget), and a
+    ``max_size`` below one chunk makes ``CacheStore`` write every value to disk
+    and then decline to track it -- an unbounded, never-evicted cache. Evicting
+    far enough to free a whole budget would instead throw away the cross-run
+    reuse the disk cache exists to provide.
+
+    This reaches into ``CacheStore._state`` because zarr exposes no public seam
+    for pre-populating the LRU (``cache_info()`` is read-only). The proper
+    home for this is ``CacheStore.__init__``; until it lands upstream, the
+    access is guarded and reported rather than assumed.
+
+    Returns True when the accounting was seeded.
+    """
+    state = getattr(store, "_state", None)
+    required = ("cache_order", "key_sizes", "key_insert_times", "current_size")
+    if state is None or not all(hasattr(state, name) for name in required):
+        return False
+    monotonic_now = time.monotonic()
+    wall_now = time.time()
+    total = 0
+    for mtime_ns, size, path in snapshot:
+        try:
+            key = path.relative_to(cache_path).as_posix()
+        except ValueError:
+            continue
+        # snapshot is oldest-first, so insertion order is already LRU order.
+        state.cache_order[key] = None
+        state.key_sizes[key] = size
+        # Age entries by mtime so a finite max_age_seconds would still expire
+        # them; with the "infinity" default this value is never read.
+        state.key_insert_times[key] = monotonic_now - max(
+            0.0, wall_now - mtime_ns / 1e9
+        )
+        total += size
+    state.current_size = total
+    return True
 
 
 def load_volume_auth(auth_json_path: str | Path | None) -> tuple[str, str] | None:
@@ -121,10 +185,13 @@ def open_volume_root(
             raise ValueError("cache_max_gb must be nonnegative or None")
         cache_path = disk_cache_subdir(path_text, Path(cache_dir))
         cache_path.mkdir(parents=True, exist_ok=True)
+        retained: list[tuple[int, int, Path]] = []
         if maximum_bytes is not None:
             snapshot = _cache_snapshot(cache_path)
             if sum(size for _, size, _ in snapshot) > maximum_bytes:
-                _evict_to_watermark(snapshot, maximum_bytes)
+                retained = _evict_to_watermark(snapshot, maximum_bytes)
+            else:
+                retained = snapshot
         if is_remote:
             remote_options = dict(storage_options)
             remote_options["skip_instance_cache"] = True
@@ -140,6 +207,20 @@ def open_volume_root(
             cache_store=LocalStore(cache_path),
             max_size=maximum_bytes,
         )
+        # Seed before the first read: zarr.open() itself fetches metadata
+        # through this store, and those writes must be charged against the
+        # bytes already on disk rather than against an empty budget.
+        if maximum_bytes is not None and not _seed_cache_store_lru(
+            store, cache_path, retained
+        ):
+            warnings.warn(
+                "installed zarr CacheStore exposes no LRU state to seed "
+                f"(zarr {zarr.__version__}); the volume cache budget of "
+                f"{maximum_bytes} bytes is enforced only by the open-time "
+                "sweep, so this process may grow the cache directory past it",
+                RuntimeWarning,
+                stacklevel=2,
+            )
         return zarr.open(store=store, mode="r")
 
     if is_remote and ZARR_V3:
--- a/vesuvius/tests/ink_detection/test_volume_io.py
+++ b/vesuvius/tests/ink_detection/test_volume_io.py
@@ -171,6 +171,118 @@ def test_open_sweep_prunes_oldest_recursively_and_leaves_under_budget(
 
 
 @pytest.mark.skipif(not _ZARR_V3, reason="volume disk cache requires Zarr 3")
+def test_reopening_reuses_one_budget_instead_of_granting_a_new_one(tmp_path):
+    chunk_bytes = 16 * 16 * 16
+    chunk_count = 40
+    shape = (16, 16, 16 * chunk_count)
+    expected = (
+        np.arange(chunk_bytes * chunk_count, dtype=np.int64)
+        .astype(np.uint8)
+        .reshape(shape)
+    )
+    source_path = tmp_path / "volume.zarr"
+    source = zarr.open_group(source_path, mode="w")
+    # Uncompressed chunks make every cached file exactly chunk_bytes long, so
+    # the budget below holds exactly half of this volume.
+    source.create_array(
+        "0",
+        shape=shape,
+        chunks=(16, 16, 16),
+        dtype="uint8",
+        compressors=None,
+    )
+    source["0"][:] = expected
+
+    budget_bytes = chunk_bytes * (chunk_count // 2)
+    cache_root = tmp_path / "cache"
+    cache_path = disk_cache_subdir(str(source_path), cache_root)
+
+    for _ in range(3):
+        # Each open builds a new CacheStore whose in-memory LRU starts empty,
+        # which is exactly what a second training or inference process sees.
+        root = open_volume_root(
+            source_path,
+            cache_dir=cache_root,
+            cache_max_gb=budget_bytes / 1e9,
+        )
+        assert np.asarray(root["0"][:]).tobytes() == expected.tobytes()
+        on_disk = sum(
+            size for _, size, _ in volume_io._cache_snapshot(cache_path)
+        )
+        assert 0 < on_disk <= budget_bytes
+        # Every cached file is a tracked full-key entry, so the store's own
+        # accounting can only exceed the directory (current_size also covers
+        # CacheStore's in-memory byte-range entries, which never reach disk).
+        # Sandwiching the directory under it is what makes max_size a real
+        # bound; before the seed, on_disk ran to twice current_size.
+        assert on_disk <= root.store.cache_info()["current_size"] <= budget_bytes
+
+
+@pytest.mark.skipif(not _ZARR_V3, reason="volume disk cache requires Zarr 3")
+def test_seeded_entries_evict_oldest_first_and_survive_reuse(tmp_path):
+    from zarr.experimental.cache_store import CacheStore
+    from zarr.storage import LocalStore, MemoryStore
+
+    cache_path = tmp_path / "cache"
+    files = [cache_path / "c" / "1", cache_path / "c" / "0"]
+    # Whole seconds, not raw nanoseconds: NTFS keeps 100 ns granularity and
+    # anchors at 1601, so sub-microsecond mtimes near the epoch collapse to 0
+    # and the intended age ordering disappears.
+    for timestamp, path in enumerate(files, start=1):
+        path.parent.mkdir(parents=True, exist_ok=True)
+        path.write_bytes(b"x" * 4)
+        os.utime(path, ns=(timestamp * 10**9, timestamp * 10**9))
+
+    source = MemoryStore()
+    store = CacheStore(
+        store=source,
+        cache_store=LocalStore(cache_path),
+        max_size=10,
+    )
+    assert volume_io._seed_cache_store_lru(
+        store, cache_path, volume_io._cache_snapshot(cache_path)
+    )
+    assert store.cache_info()["current_size"] == 8
+
+    async def scenario():
+        from zarr.core.buffer.core import default_buffer_prototype
+
+        prototype = default_buffer_prototype()
+        # Reading the older seeded entry must promote it, so the newer one is
+        # what the next admission evicts. "c/1" is the older file here, so an
+        # order taken from paths instead of mtimes would evict the wrong one.
+        assert (await store.get("c/1", prototype)).to_bytes() == b"x" * 4
+        await source.set("c/2", prototype.buffer.from_bytes(b"y" * 4))
+        assert (await store.get("c/2", prototype)).to_bytes() == b"y" * 4
+
+    asyncio.run(scenario())
+
+    assert files[0].exists()
+    assert not files[1].exists()
+    assert store.cache_info()["current_size"] == 8
+    assert sum(size for _, size, _ in volume_io._cache_snapshot(cache_path)) == 8
+
+
+@pytest.mark.skipif(not _ZARR_V3, reason="volume disk cache requires Zarr 3")
+def test_unseedable_cache_store_is_reported_not_silently_overfilled(
+    tmp_path, monkeypatch
+):
+    class Stateless:
+        pass
+
+    assert not volume_io._seed_cache_store_lru(Stateless(), tmp_path, [])
+
+    monkeypatch.setattr(volume_io.zarr, "open", lambda *, store, mode: store)
+    monkeypatch.setattr(
+        volume_io, "_seed_cache_store_lru", lambda *args, **kwargs: False
+    )
+    with pytest.warns(RuntimeWarning, match="enforced only by the open-time"):
+        open_volume_root(
+            "unseedable.zarr", cache_dir=tmp_path / "cache", cache_max_gb=1e-6
+        )
+
+
+@pytest.mark.skipif(not _ZARR_V3, reason="volume disk cache requires Zarr 3")
 def test_shared_cache_multiprocess_reads_are_not_torn(tmp_path):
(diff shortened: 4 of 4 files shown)


Bullo27:
Independently reproduced the bug and verified this fix on a different platform — Linux, CPython 3.13.11, **zarr 3.2.1** against your Windows 11 / 3.14.7 / 3.3.0. Same bucket, same volume, your harness from #1586 unmodified.

## The bug reproduces byte-identically

| run | on disk | vs 30,000,000 B budget |
|---|---|---|
| 1 (cold) | 29,360,128 B / 14 files | 0.98× |
| 2 (warm) | **58,720,256 B / 28 files** | **1.957×** |
| 3-5 | 54,525,952 B / 26 files, byte-identical | 1.817× |

Those are your numbers to the byte, including the 1.96× spike and the 1.82× plateau. Throughout, `CacheStore.current_size` stays at 29,360,128 / 14 keys while the directory holds nearly double — the store believing it is within budget is the whole bug in one line.

That also settles something you asserted but, per your own table, only measured on 3.3.0: that `cache_store.py`'s LRU-init path is identical in 3.2.1. It is — 3.2.1 produces your 3.3.0 numbers exactly.

## The fix holds, including the recovery claim

Same directory, `main` first and then this branch:

```
MAIN     run 1 (cold)  ->  29,360,128 / 14
MAIN     run 2 (warm)  ->  58,720,256 / 28      1.96x
PATCHED  on that same overshot dir:
         after open    ->  25,169,105 / 15      (swept + seeded)
         after 3 reads ->  29,360,128 / 14      0.98x
PATCHED  run 2         ->  29,360,128 / 14      0.98x
PATCHED  fresh, cold + 2 warm -> 29,360,128 / 14 throughout
```

So the recovery path you described works on real data on a second platform — my 25,169,105 vs your 25,165,824 differs only by the metadata written during open.

## Your new tests pass on Linux

```
1 failed, 15 passed        (this branch)
1 failed, 12 passed        (main, same file)
```

All three new tests pass, and the delta is clean: **+3 passes, and the single failure is the same one before and after** — `test_parent_populated_dataset_cache_misses_after_fork`, which this PR neither fixes nor touches. I've seen that same test fail on this box in unrelated work and haven't chased whether it's environmental here, so I'd treat it as pre-existing rather than as a signal about this PR.

Worth noting `test_seeded_entries_evict_oldest_first_and_survive_reuse` passes here specifically: its whole-second `os.utime` workaround is motivated by NTFS's 100 ns granularity and 1601 epoch, so it was reasonable to wonder whether it was load-bearing on a filesystem with different granularity. It isn't — the ordering assertion holds on ext4 too.

## On the private `_state` access

Since that is the obvious reviewer objection, I checked the range rather than assuming: `zarr>=2.18.7,<4` currently resolves to at most **3.3.0** — the newest zarr 3.x on PyPI is 3.3.0, which you already tested. So the private shape is covered across every version the pin can actually select today, and the exposure is future 3.x releases only, which is what the `RuntimeWarning` guard is for. (On the zarr 2 end of the pin the cache path is unreachable anyway — `volume_io.py:109` raises `NotImplementedError`.)

## Doc numbers check out

`vesuvius/docs/ink_detection.md:488` is `--cache-max-gb 100`, and `scrollprize.org/docs/07_tutorial5.md:327` / `:385` are `120` in both the config and CLI forms — so the three places you flag do recommend exactly those numbers, and at the 1.9× plateau they describe ~190 GB and ~228 GB of actual disk.

---
*Claude (AI agent) on behalf of @Bullo27. Every figure above is from runs on this machine against the live public bucket — `main` at `51c9f8dac` and this PR at `refs/pull/1587/head` in a throwaway worktree; the pytest counts are quoted from those two runs.*

flummoxjr:
Thank you — a byte-identical reproduction on a different OS, Python and zarr version is exactly the verification this needed, and "your numbers to the byte" is more than I could have asked for.

One consequence of your zarr 3.2.1 datapoint worth recording here: the double-accounting spans at least 3.2.1 → 3.3.0, so this isn't a recent zarr regression villa happened to catch — it's the standing contract between the sweep and `CacheStore`. I've added your version datapoint to the upstream report (zarr-developers/zarr-python#4282).

PR #1593, opened 2026-08-24
scrollprize.org eligible volumes

pmh47:
Show prize-eligible volumes in data-browser too (and source both that and prizes page list from shared json). No change to actual eligibility!


Bullo27:
Checked the "No change to actual eligibility!" claim rather than taking it on trust, since this is the one list on the site people plan work around. **It holds — 13/13, byte-identical.**

I parsed the `<details>` block being deleted from `34_prizes.md` (decoding each hardcoded Neuroglancer state to recover the zarr it actually opens) and diffed it against `prizeEligibility.json`:

| check | result |
|---|---|
| `(scroll, volume)` pairs | 13/13 identical, **order preserved** |
| rendered label (`s.label` vs old text) | 13/13 identical |
| scan text (`scan.name` vs old link text) | 13/13 identical |
| zarr the link opens | 13/13 identical |

Worth noting the old entries were internally split — link text was the *scan* id (`20250720091415`) while the URL pointed at the *volume* id (`20250821151825`). The JSON records the volume id and re-derives the scan name from the index, which reproduces both correctly.

## The part I'd actually flag for review

`EligibleVolumes` resolves each entry by walking `scans[].volumes[]` — a structure that only landed yesterday in #1591. If any eligible volume weren't reachable that way, the entry silently degrades to a bare `<code>` with **no Neuroglancer link**, and the page would still look plausible. So I checked that path explicitly, on both index sources:

- **bundled snapshot** (`static/data_browser/index.json`): 13/13 resolve.
- **live `metadata.min.json`** — the source `useAtlasData` *prefers* at runtime, so it's what visitors get: 13/13 resolve. Every eligible volume is present in its sample's `volumes`, carries a `scan_id`, that scan exists in the sample's `scans` map, and has a zarr origin.

Also confirmed `grand-prize-2027` matches a real `id` in the `prizes:` frontmatter (a typo there would render "Eligible scroll volumes (0)" with no other symptom), and that `display` is non-empty for all 45 samples, so the generated layer names don't come out as `undefined …`.

Repro for the live half:

```bash
curl -s https://vesuvius-challenge-open-data.s3.us-east-1.amazonaws.com/metadata.min.json \
| jq -r '.samples | to_entries[] | . as $s | $s.value.volumes // {} | to_entries[]
         | select(.key | IN("20250821151825","20250821151635","20250821151803","20250821151750",
                            "20251110183117","20250821151737","20250521135224","20250821151723",
                            "20250821151701","20250820131727","20250521120456","20250521151220",
                            "20250821151648"))
         | "\($s.key) \(.key) scan_id=\(.value.scan_id) linked=\((.value.scan_id) as $x | $s.value.scans | has($x))"'
```

---
*Claude (AI agent) on behalf of @Bullo27. Measured against `3ea5214ce` with base `2c477013e`, and against live `metadata.min.json` on 2026-08-25. I did not build the site — the label/scan-text/zarr comparisons are computed by applying the component's own resolution logic and `dataAccess`'s rewrite rules to the index, not read off a rendered page.*

PR #1595, opened 2026-08-24
Fix Windows crash: request int64 explicitly in LegacyNumpyRandom.randint

Pizzakira:
**In one sentence:**
This lets anyone on Windows actually run the spiral fitter, which today stops with an error before its first optimizer iteration.

**One real example:** 
Starting with a scroll spiral dataset on Windows 11, I ran fit_spiral.py, and it loaded the whole dataset, printed the first PROGRESS line, then died with ValueError: high is out of bounds for int32. With this change the same command keeps optimizing.

**Before:** 
The run loads tracks, lasagna fields and the shell table, prints "PROGRESS Optimizing 0/30,000 iterations", then raises ValueError: high is out of bounds for int32 from losses.py in _sample_patch_points. It is not an edge case: prefetch_enabled() returns False unless FIT_SPIRAL_PREFETCH=1 is set, so this is the path a Windows user gets by default.
 The fitter cannot start at all.

**After this PR:**
The same command starts optimizing and keeps running.

**Proof:** 
Two lines in a fresh python session on Windows reproduce the failure with no data and no checkout at all:

    import numpy as np
    np.random.randint(0, np.iinfo(np.int64).max)
    ValueError: high is out of bounds for int32

What to look at is the bound. RandomState.randint defaults to the C long, which is 32 bits on Windows and 64 bits on Linux, and losses.py asks for a 64-bit seed with seed = int(rng.randint(0, np.iinfo(np.int64).max)). The identical call succeeds on Linux, which is why this has been invisible.

Same settings, before and after, for the part that matters to reviewers, that existing sampling is untouched :

    np.random.seed(12345); [np.random.randint(0, 1000) for _ in range(5)]
    before: [482, 485, 285, 129, 420]
    after:  [482, 485, 285, 129, 420]

A bound of exactly np.iinfo(np.int32).max also gives the same value before and after. int64 is only requested when the bound does not fit in int32, so every existing small-range call keeps the historical bit stream, which is the stated purpose of LegacyNumpyRandom.

Anyone who tries the spiral fitter on Windows. Right now they hit this in the first second of optimization, after a long dataset load, with an error that points at numpy rather than at the cause.
With the fix they can run fits and report real results instead of a portability bug.

## Details

Tested on Windows 11, Python 3.14.2, NumPy 2.5.2, villa at commit de139a6, before spiral-fitting moved to the repo root.
The dataset was a scoll spiral root with tracks, lasagna normals and grad_mag, a measured outer shell and four patches.

Call chain of the failure:

    fit_spiral.py in step
    losses.py in get_patch_and_umbilicus_losses
    losses.py in _sample_patch_batch      return build(prefetch.LegacyNumpyRandom)
    losses.py in build
    losses.py in _sample_patch_points     seed = int(rng.randint(0, np.iinfo(np.int64).max))
    numpy/random/mtrand.pyx in RandomState.randint
    ValueError: high is out of bounds for int32

The traceback above is shortened to the call chain, with only the two source lines that matter kept.

Why the fix goes in LegacyNumpyRandom and not in losses.py: the prefetch path passes a _GeneratorShim, which delegates to Generator.integers and already defaults to int64, and that shim takes no dtype argument. Patching the call site would therefore break the other path. losses.py also uses LegacyNumpyRandom unconditionally in _sample_requested_patch_rows, so a call-site patch would need to be repeated there.

The same latent pitfall exists in StepPrefetcher.np_rng, which calls np.random.randint(0, 2 ** 63 - 1) without a dtype. That would raise on Windows for anyone who sets FIT_SPIRAL_PREFETCH=1. This PR fixes it too.

Limitations : I have only tested this on Windows and I have not run the full fit to completion on Linux, so I cannot claim a before and after comparison of final results. The argument that results are unchanged rests on the bit stream check above rather than on a completed run.

Tooling note : I used an LLM assistant (Claude Opus 5) while tracking this down. The crash, the diagnosis and the reproduction are from my own machine, running the spiral fitter on a real dataset.

cc @nvining

### DIFF
--- a/spiral-fitting/prefetch.py
+++ b/spiral-fitting/prefetch.py
@@ -61,23 +61,9 @@ class LegacyNumpyRandom:
     # identical to the historical behaviour).
 
     random = staticmethod(np.random.random)
+    randint = staticmethod(lambda *a, **k: np.random.randint(*a, dtype=np.int64, **k))
     uniform = staticmethod(np.random.uniform)
     choice = staticmethod(np.random.choice)
-
-    @staticmethod
-    def randint(low, high=None, size=None):
-        # RandomState.randint defaults to the C long, which is 32-bit on Windows. Callers
-        # asking for a 64-bit seed -- losses.py:_sample_patch_points does
-        #     seed = int(rng.randint(0, np.iinfo(np.int64).max))
-        # -- therefore raise "ValueError: high is out of bounds for int32" at the first
-        # optimizer iteration. The C long is 64-bit on Linux, which is why this is invisible
-        # there. int64 is requested only when the bound does not fit in int32, so every
-        # existing small-range call keeps the exact historical bit stream, which is the
-        # stated purpose of this class.
-        bound = low if high is None else high
-        if bound is not None and int(bound) > np.iinfo(np.int32).max:
-            return np.random.randint(low, high, size=size, dtype=np.int64)
-        return np.random.randint(low, high, size=size)
 
 
 def _iter_tensors(value):
@@ -115,8 +101,6 @@ class StepPrefetcher:
         if rng is None:
             # Seeded once from the global stream: deterministic given the
             # deterministic first-use order, then independent of it.
-            # Same 32-bit C long pitfall as in LegacyNumpyRandom.randint above:
-            # without an explicit dtype this raises on Windows.
             seed = int(np.random.randint(0, 2 ** 63 - 1, dtype=np.int64))
             rng = _GeneratorShim(np.random.Generator(np.random.PCG64(seed)))
             self._np_rngs[key] = rng
--- a/spiral-fitting/prefetch.py
+++ b/spiral-fitting/prefetch.py
@@ -61,9 +61,23 @@ class LegacyNumpyRandom:
     # identical to the historical behaviour).
 
     random = staticmethod(np.random.random)
-    randint = staticmethod(np.random.randint)
     uniform = staticmethod(np.random.uniform)
     choice = staticmethod(np.random.choice)
+
+    @staticmethod
+    def randint(low, high=None, size=None):
+        # RandomState.randint defaults to the C long, which is 32-bit on Windows. Callers
+        # asking for a 64-bit seed -- losses.py:_sample_patch_points does
+        #     seed = int(rng.randint(0, np.iinfo(np.int64).max))
+        # -- therefore raise "ValueError: high is out of bounds for int32" at the first
+        # optimizer iteration. The C long is 64-bit on Linux, which is why this is invisible
+        # there. int64 is requested only when the bound does not fit in int32, so every
+        # existing small-range call keeps the exact historical bit stream, which is the
+        # stated purpose of this class.
+        bound = low if high is None else high
+        if bound is not None and int(bound) > np.iinfo(np.int32).max:
+            return np.random.randint(low, high, size=size, dtype=np.int64)
+        return np.random.randint(low, high, size=size)
 
 
 def _iter_tensors(value):
@@ -101,7 +115,9 @@ class StepPrefetcher:
         if rng is None:
             # Seeded once from the global stream: deterministic given the
             # deterministic first-use order, then independent of it.
-            seed = int(np.random.randint(0, 2 ** 63 - 1))
+            # Same 32-bit C long pitfall as in LegacyNumpyRandom.randint above:
+            # without an explicit dtype this raises on Windows.
+            seed = int(np.random.randint(0, 2 ** 63 - 1, dtype=np.int64))
             rng = _GeneratorShim(np.random.Generator(np.random.PCG64(seed)))
             self._np_rngs[key] = rng
         return rng


pmh47:
please shorten to a 1-line change. Doesn't have to preserve exact semantics, there are other sources of non-determinism anyway

Bullo27:
Reviewed by running your actual adapter (extracted from `685e5ec6b` and exercised standalone). **The fix is correct and complete** — it covers both sites, and the stream-preservation claim holds end-to-end.

## Verified

| check | result |
|---|---|
| the failing call now succeeds | `LegacyNumpyRandom.randint(0, iinfo(int64).max)` → `7692698082559361259` |
| small-bound stream through the adapter | `seed(12345)` → `[482, 485, 285, 129, 420]` — **identical** raw vs adapter, i.e. I reproduce your published numbers independently |
| `high=None` form (`randint(5)`) | adapter passes `high=None` explicitly; draws identical to `np.random.randint(5)` |
| bound exactly `iinfo(int32).max` | stays on the default branch, identical |
| bound `int32max + 1` | takes the int64 branch, matches the Linux default stream |
| second site | `prefetch.py` `np_rng` seed now `dtype=np.int64` — the path that would have bitten anyone with `FIT_SPIRAL_PREFETCH=1` |
| scope | exactly one call reaches the adapter (`losses.py:180`); `_sample_requested_patch_rows` routes into the same `_sample_patch_points`, so both consumers are covered by the one change |
| merge | `git merge-tree --write-tree origin/main 685e5ec6b` → rc 0 on `54814ac6a`, one file |

The `high=None` case is the one I'd have most expected a wrapper to break, since you now always pass `high` positionally. It doesn't.

## One measurement you may want, on the conditional

I checked whether the dtype actually perturbs the stream, across 11 bounds × 7 seeds (2 … `int32max`), comparing default vs `int32` vs `int64`:

```
combinations tested: 77
mismatches (default vs int32 vs int64): 0
```

So an **unconditional** `dtype=np.int64` would preserve the historical stream too — the conditional isn't load-bearing for that purpose, and the simpler two-line form would be equally safe if a reviewer prefers less code. Keeping it is perfectly defensible as defensive coding; I'm only flagging that the comment's reasoning ("int64 is requested only when the bound does not fit in int32, **so** every existing small-range call keeps the exact historical bit stream") reads as though the conditional is what preserves the stream, when the stream appears to be dtype-independent anyway.

**Caveat:** that's measured on Linux only, where the default is already int64. The case the conditional really protects is a *Windows* small-bound call (default int32) — my int32-vs-int64 comparison is strong evidence it's an algorithm-level property rather than a platform one, but you're the one with the box, and it's the same two lines to confirm there.

## One latent edge, if you keep the conditional

`int(bound)` narrows the accepted input relative to the function it replaces:

```python
>>> LegacyNumpyRandom.randint(0, np.array([5, 10, 20]))
TypeError: only 0-dimensional arrays can be converted to Python scalars
>>> np.random.randint(0, np.array([5, 10, 20]))     # what it did before
array([ 4,  9, 14])
```

Not a live problem — the single adapter call site passes a scalar — so this is a note for whoever next reuses the class, not a change request. `bound = np.max(bound)` or a `np.ndim(bound) == 0` check would close it if you'd rather it stay a drop-in.

## Regression coverage

Nothing here needs it to merge, but if a reviewer asks: this can't be caught on a Linux runner by a direct test, since the call simply works there. #1585 hit the same wall for the Windows fork guard and solved it by emulating the platform difference in a subprocess — that's the pattern if you ever want this pinned in CI.

---
*Claude (AI agent) on behalf of @Bullo27. Measured on `main` at `54814ac6a` with your head `685e5ec6b`, CPython 3.13.11 / NumPy 2.4.2, Linux. **Verified at the adapter level, not by running a fit** — I have no dataset or built `vc_spiral` binding here, so your Windows run remains the only end-to-end evidence. Your NumPy is 2.5.2 and mine 2.4.2.*

Pizzakira:
It is done, but with two lines, one per call site, no conditional.

@Bullo27 was right that it wasnt load-bearing : on Windows, where the default is
int32, the stream is identical with and without dtype=np.int64 across 77 bound/seed
combinatons : zero mismatches.

Second line is StepPrefetcher.np_rng, same missing dtype on the FIT_SPIRAL_PREFETCH=1
path. I could cut it to one, but that would leave the same crash sitting in the
prefetch path for whoever turns that flag on, and it is one word in the same file.

Bullo27:
Verified the form that actually merged, since it isn't the one I reviewed — `4e9601f19` ships the unconditional pass-through:

```python
randint = staticmethod(lambda *a, **k: np.random.randint(*a, dtype=np.int64, **k))
```

That's the simpler variant, and it resolves both of my earlier notes at once: no `int(bound)`, so array bounds broadcast again (`L.randint(0, np.array([5,10,20]))` → `[0, 4, 12]`, which the conditional version rejected). Re-checked against the merged source:

| check | result |
|---|---|
| the originally failing call | `randint(0, iinfo(int64).max)` → returns a value |
| stream vs raw default | 7 bounds × 5 seeds = 35 combos, **0 mismatches** |
| `high=None` (`randint(5)`) | identical draws to `np.random.randint(5)` |
| `size` kwarg **and** positional | both fine |
| `np_rng` seed at `:104` | carries `dtype=np.int64` |

## One latent edge this form introduces

Since `dtype` is now baked in, a caller passing it explicitly collides:

```python
>>> LegacyNumpyRandom.randint(0, 10, dtype=np.int32)
TypeError: numpy.random.RandomState.randint() got multiple values for keyword argument 'dtype'
```

Nothing does this today — the only call through the adapter is `losses.py:180`, and I grepped `spiral-fitting/` for `randint(...dtype` and found only the definition itself plus the direct `np.random.randint` at `prefetch.py:104`. So it's a note for whoever next reuses the class, not a problem with the merge. `k.setdefault("dtype", np.int64)` would make it a true drop-in if that ever matters.

Also worth recording for the "bitwise identical" contract: on Windows the *values* are unchanged (that's the 35-combo check above, and the 77-combo one in my earlier comment), but an array result's **dtype** now comes back `int64` where the platform default would have been `int32`. The single call site wraps the result in `int(...)`, so nothing observes it.

---
*Claude (AI agent) on behalf of @Bullo27. Measured against merged `4e9601f19`, CPython 3.13.11 / NumPy 2.4.2, Linux — the adapter was extracted from the merged source and exercised directly. Still not a Windows run: the platform-default dtype claim is inference from the same mechanism as before, not something I observed there.*

PR #1596, opened 2026-08-25
tracer: fix OmpThreadPointCol shared-state race

aviad12g:
### Motivation

While reproducing a real-scroll PHerc0813 tracing failure, I found that
`OmpThreadPointCol::next()` resets one entry of `_thread_points` outside its
OpenMP critical region. Inside that region, `extract_point_min_dist()` reads the
complete shared vector. A worker can therefore write an element while another
worker reads that same element without synchronization.

The pre-patch production class produces the expected write/read report under
GCC ThreadSanitizer. A matched mutex-based isolation test reports the race in
the buggy arm and completes one million calls cleanly when the reset and the
full-vector read share one lock.

### Change

- Move `_thread_points[t_id] = {-1, -1}` into the existing OpenMP critical
  region.
- Add an actual-class four-worker stress test that checks the published-point
  minimum-distance contract.
- Make `OMPThreadPointCollection.hpp` self-contained for `cv::Point2i` so the
  test also builds in the default Release/PCH-off configuration.

This is intentionally a race-only change. It does not replace the separate
`rand()` candidate-start path and does not claim deterministic multithreaded
tracing.

### Validation

- Actual production-class test: 100/100 consecutive executions passed in an
  Ubuntu 24 GCC/OpenMP environment with warnings treated as errors.
- The same test compiled and passed with Clang 18 plus the repository's serial
  OpenMP compatibility shim, also with warnings treated as errors.
- Default Release with `VC_USE_PCH=OFF` and `VC_TESTING=ON`: the standalone
  test target built and its focused CTest passed after adding the direct
  OpenCV types include.
- Exact one-thread class sequence: 4,096/4,096 points byte-identical before and
  after the patch (SHA-256
  `496e14ab69483abc3479079fbdf95e64ece56042702976e8529f7d6685c5c42c`).
- Paired synchronization-shape TSan test: buggy arm reports the race; fixed arm
  completes 1,000,000 calls cleanly.
- Bounded actual-class microbenchmark: 92.093 ms pre-patch versus 89.373 ms
  patched per run (-2.953%); no measured regression.
- Official `builder-ubuntu-26.04`: affected targets built and the registered
  production-class CTest passed.
- Full-tracer one-thread gate: two runs per arm produced byte-identical X, Y,
  Z, and generations before and after the patch.
- Full-tracer four-thread replay with both known RNG streams controlled: five
  unique XYZ hash sets in five patched runs; the generation TIFF was common.
  The patch fixes the race but does not restore multithread determinism.
- Full-tracer timing: serial internal mean +0.153%; four-thread median internal
  time +11.275% and median wall time +10.0%, with higher patched variance.

The registered actual-class test is a concurrency contract/stress test, not a
deterministic red/green regression for this schedule-dependent race. Independent
verification by @Bullo27 ran the unfixed implementation 260 times without a
failure; the paired sanitizer/mutex reproducer above is the discriminating
race evidence. The same independent check also found identical point counts
between main and patched code over 10 distance/thread configurations and 1,200
trials per configuration, supporting the bounded race-only claim.

Raw TSan over OpenMP is not used as patched-clean evidence because the local
libgomp runtime does not expose the critical-region synchronization to TSan.

### Explicit limitation

This patch fixes a confirmed shared-state race. It does **not** claim
deterministic multithreaded tracing. In separate unchanged-binary experiments,
four-thread output remained scheduling-sensitive even after the two observed
RNG inputs were controlled. That behavior remains follow-up work.

### Performance disclosure

The class-level microbenchmark did not predict the supported full-tracer
four-thread result. In one local Docker-on-macOS five-run-per-arm set, patched
four-thread tracing had a 10% higher median wall time and substantially higher
variance. This small local sample is not sufficient to characterize fleet
performance, so the result is disclosed as a potential regression rather than
a precise performance estimate. The serial path remains unchanged within
measurement noise.

A bounded two-phase-lock alternative preserved byte-identical serial output
and reduced the local four-thread wall median from 44 to 42 seconds, but its
internal median improved only about 0.9% and remained about 10.3% above the
unpatched baseline. That was not a clear enough improvement to justify the
extra synchronization complexity, so this draft retains the simpler patch.

### Draft checklist

- [x] Race-only source change
- [x] Actual-class concurrent contract/stress test (not a red/green race test)
- [x] Default Release/PCH-off standalone test build
- [x] Focused Linux/GCC/OpenMP stress test
- [x] One-thread actual-class byte-identity check
- [x] Bounded class-level performance comparison
- [x] Supported VC builder: affected targets
- [x] Patched full-tracer four-thread A/B
- [x] Patched one-thread byte-identity replay
- [x] End-to-end tracer performance comparison


Bullo27:
Built and ran this against a local VC3D tree. The race analysis looks right to me, and the disclosure in the description is unusually careful. Two things to flag before review, one of them blocking, plus one result that supports your "race-only" claim.

## 1. Blocking: the new test target doesn't compile in the default configuration

Real build, `main` at `54814ac6a` with your head `49d62fa29`, stock configure (Release, `VC_USE_PCH=OFF`):

```
$ ninja -C build/local-verify test_omp_thread_point_collection
core/include/vc/core/util/OMPThreadPointCollection.hpp:20:12: error: 'Point2i' in namespace 'cv' does not name a type
core/include/vc/core/util/OMPThreadPointCollection.hpp:58:9: error: 'Point2i' in namespace 'cv' does not name a type
core/test/test_omp_thread_point_collection.cpp:60:50: error: 'class OmpThreadPointCol' has no member named 'next'
```

The header isn't self-contained: it returns `cv::Point2i` (`:20` and `:58`) but includes only `<opencv2/core/matx.hpp>`, which supplies `Vec2i` and not `Point2i`. That has never mattered, because its only two includers — `GrowPatch.cpp` and `GrowSurface.cpp` — pull in fuller OpenCV headers first. **Your test is the first TU to include it alone, which is what exposes it.**

I'd guess your builder had the PCH on: `vc_add_test` does `target_precompile_headers(... REUSE_FROM vc_test_pch)` when `VC_USE_PCH` is set, which would supply `Point2i` without the TU asking. `VC_USE_PCH` defaults **OFF** (`CMakeLists.txt:74`) and is forced ON only for `CMAKE_BUILD_TYPE=QuickBuild` (`:113`) — so a QuickBuild passes and a default Release doesn't. I can't see your configure line, so that's a guess at the cause; the failure itself is reproducible here.

One line fixes it, and makes the header self-contained as a bonus:

```cpp
#include <opencv2/core/matx.hpp>
#include <opencv2/core/types.hpp>   // cv::Point2i
```

With that added: builds clean, and the **full suite is 155/155** (`ctest -j6`, 58s — the 154 that were green on `main`, plus your new one).

## 2. The new ctest doesn't discriminate the fix

Worth knowing since it's registered as the regression test for this change. With the include added so both arms compile, I reverted **only** `next()` to main's racy form and left your test untouched:

```
patched next()   : 20/20 runs passed
racy next()      : 200/200 runs passed, 0 failed
racy, OMP_NUM_THREADS=16 : 60/60 passed
```

260 runs on the buggy arm, no failures. So it's a useful concurrency smoke test of the published-point contract, but it won't catch a regression of *this* fix. Part of the reason looks structural: the per-wave `#pragma omp barrier` and `#pragma omp single` serialise almost exactly the window the race lives in, so the unsynchronised write and the full-vector read rarely overlap.

That's not an argument against merging — your discriminating evidence is the paired TSan/mutex harness described in the description. It just isn't the thing that landed in the repo, and the checklist item "Actual-class concurrent test" reads as though it is.

## 3. Supporting your "race-only" claim

I tried to show the change alters behaviour and **failed to**, which is a point in your favour. Moving the reset inside the critical region means a thread's previous point stays visible to other threads until it acquires the lock, where before it was cleared first — so I expected the patched arm to hand out points more conservatively. Measured across 10 configurations (`dist` ∈ {8, 20, 30, 40, 50} × {4, 8} threads, 120 trials each, identical seeds):

```
main mean points handed out : 4096.00   (every configuration)
patched                     : 4096.00   (every configuration)
```

Identical throughout. `extract_point_min_dist` scans the whole candidate list against at most `nthreads-1` blocked points, so the constraint essentially never binds and nothing is dropped either way. Consistent with "fixes the race, changes nothing else."

---
*Claude (AI agent) on behalf of @Bullo27. Measured on Ubuntu 24.04, GCC 13.3, OpenCV 4 (`/usr/lib/x86_64-linux-gnu/cmake/opencv4`), stock `cmake -S volume-cartographer -B build` (Release, PCH off). The build failure, the 155/155, and the 260 control runs were all executed here; my tree was restored to clean `main` afterwards. I did **not** reproduce your TSan report or the PHerc0813 tracing failure — no ThreadSanitizer run and no scroll data on this box, so the race evidence remains yours.*

aviad12g:
@Bullo27 Thank you — confirmed and fixed in `81881c3d`. I added the direct `<opencv2/core/types.hpp>` include, then rebuilt the standalone target in the official builder with `CMAKE_BUILD_TYPE=Release`, `VC_USE_PCH=OFF`, and `VC_TESTING=ON`; the focused CTest passes.

I also relabeled the landed test in both source and the PR description as a concurrency contract/stress test, not a deterministic red/green regression. The paired sanitizer/mutex reproducer remains the discriminating race evidence. I incorporated your 260 unfixed control runs and the 10-configuration point-count result as independent limits/support for the race-only claim.

I also tried the bounded two-phase-lock alternative suggested by the timing result. It preserved byte-identical serial output and narrowed the local timing spread, but improved the internal median only about 0.9% versus this simpler patch and remained about 10.3% above baseline, so I did not add the extra synchronization complexity.

hendrikschilling:
Looks fine to me. Update resolving the merge conflicts than we can merge this.

aviad12g:
@hendrikschilling Updated in `04cdf1897b7114422b39f94c7c4b15abaff148f1`: merged current main (`be09a850`) into this branch without a force-push. GitHub now reports no merge conflicts.

The only conflict was in `core/test/CMakeLists.txt`. The resolution keeps our test registration alongside the upstream surface-area and voxel-size tests. The race-fix header and stress-test source are **byte-identical to the reviewed `81881c3d` revision**; no new algorithm or test logic was added.

Fresh local checks on the merged source, using the existing Ubuntu 26.04 builder and OpenCV 4.10:

- GCC 15.2 with real OpenMP: direct test compilation and execution passed.
- Clang 21.1.8 with the repository's serial OpenMP shim: direct test compilation and execution passed.
- Actual CMake target `test_omp_thread_point_collection`: configured, built and linked successfully with `CMAKE_BUILD_TYPE=Release`, `VC_USE_PCH=OFF`, and `VC_TESTING=ON` (apps/UI tracer/flatboi/Python bindings disabled).
- Merge audit and `git diff --check` passed; the diff against current main remains the original three files.

**Completed follow-up, 12 September:** After restarting the local Docker daemon, the registered CTest passed **1/1**, followed by **100/100 consecutive repetitions with zero failures**, on the same already-built Release/PCH-off binary and unchanged `04cdf189` source. The executable's real `libgomp` linkage was verified.

```sh
ctest --test-dir /out/build -R '^test_omp_thread_point_collection

AI-assisted maintenance and validation. --output-on-failure
ctest --test-dir /out/build -R '^test_omp_thread_point_collection

AI-assisted maintenance and validation. --repeat until-fail:100 --output-on-failure
```

The 100 repetitions took 14.63 seconds in the bounded local container; this is a test-run duration, not a tracer-performance comparison. Binary SHA-256: `f6a2a55f0f49cd4f004c369beae803e938229c64afa6afe1e35201e9c6789372`.

The earlier Docker-blocked attempts remain recorded as canceled, not passes. These successful runs are still concurrency contract/stress checks, **not a discriminating red/green race regression**. The earlier full-tracer, timing, nondeterminism and sanitizer disclosures remain unchanged; those experiments were not rerun.

AI-assisted maintenance and validation.

**CI follow-up:** The required CI aggregate is red because the [Linux VC3D GUI/agent-bridge smoke job](https://github.com/ScrollPrize/villa/actions/runs/34697470832/job/103563335627) failed at `c3_first_conn_still_answers` (10-second RPC timeout) and `process_alive` (`running=False`), after compilation succeeded. The base-test job passed. I have not established whether this GUI failure is pre-existing or flaky, and have not changed unrelated code or weakened the checks. This still needs investigation or a maintainer rerun before treating CI as clear.

PR #1602, opened 2026-08-26
VC3D: Fiber Map shows every fiber, placed by inferred winding

skyward7187:
# VC3D: Fiber Map shows every fiber, placed by inferred winding

## What this does

The Fiber Map tab previously showed only the largest manually-linked
networks, each unrolled in its own panel. This PR replaces the panels with
**one global map containing every fiber**, where the x-axis is the fiber's
true winding about the umbilicus: `x = (s·θ + 2π·k) · r_ref`, with the
integer turn offset `k` per fiber **inferred from how H and V fibers cross**.

## The winding solver (`FiberWindingSolver`, new, Qt-free)

Papyrus structure drives the inference: horizontal fibers lie on the front
of the sheet, verticals behind, so an H fiber on the same winding as a V
fiber passes between it and the umbilicus. Each transversal H×V crossing
(detected in unwrapped (θ, z) coordinates over every 2π translate, so the
integer turn gap is exact by construction) classifies by the sign of
ΔR = r_H − r_V:

- **inside** (ΔR ≤ 0): same winding or further in — a weak constraint,
  `W_h ≤ W_v`, which densest packing collapses to equality;
- **outside** (ΔR > 0): strictly outward, `W_h ≥ W_v + 1`.

There is deliberately **no same-winding tie band and no pitch model**: the
scroll is too crumpled for either (wrap spacing ~0.85 mm in the dense band,
only a few sheet thicknesses). Confidence is asymmetric — the nearly
unfalsifiable weak inside stays high at any margin, the strict outside earns
confidence with radial margin. Manual links are integer equalities with
residual-based confidence, and they are *droppable evidence* like crossings:
a link wrong by one turn loses to correct crossings via a seen-count
discount in the repair loop.

Contradictions (positive cycles in the difference-constraint graph) are
repaired deterministically by dropping the weakest edge per detected cycle;
the feasible system packs densest-from-below (missing windings are never
guessed), equality-locked blocks then coordinate-ascend on a local
radial-ordering cost (the one radius fact that survives crumpling: windings
stay radially ordered along a ray), and constraint-islands anchor by the
same ordinal cost — or are honestly reported **unresolved** when nothing
anchored is nearby. Per-fiber canonical gauges make the whole solve
invariant to each fiber's arbitrary unwrap branch.

## Honest error reporting

A dropped crossing becomes a red map ring only when (a) **both fibers are
trusted** — at least one fiber-model-traced span, with absent provenance
given the benefit of the doubt exactly as the renderer draws it; pure
control-point interpolation can be off by whole turns between controls, so
its contradictions are expected noise — and (b) **the final map actually
violates the dropped constraint** (computed exactly, in integer turns, from
the crossing's translate and the solved offsets): greedy repair routinely
sacrifices innocent edges whose constraints the final placement satisfies
anyway. On PHercParis4 these two gates, together with the sign-only
classification, took the declared errors from 1339 candidate drops down to
~10 genuine inward regressions. Sheet-drift tags need distinct violated
evidence; suspect links require trusted geometry on both ends. The dock
shows per-fiber winding ranges and anchor provenance (crossings / radius /
radius? / unresolved · drift?), and unplaceable fibers are listed rather
than silently dropped.

## Workspace

- One scene: winding-numbered gridlines (the number IS the winding
  coordinate), the familiar link dots/suspect treatment, red rings at
  violated drops.
- Dock grouped by linked network, largest → smallest, then unlinked fibers,
  inner → outer; selection glows the whole linked network (20 px soft halo)
  behind an unchanged selected-fiber highlight.
- Fiber label chips hide when a winding maps to under 180 px on screen.
- Staleness machinery preserved and extended: a grid change (commonly just a
  transiently displayed volume — the annotation frame follows the current
  volume) now marks the map stale and heals itself when the volume switches
  back, instead of destroying the layout; only a package change clears.
  Staleness a rebuild genuinely fixes (fiber or umbilicus changes)
  triggers an **automatic Update** — debounced, visible-only, backed by a
  light 1 s poll so an idle-but-visible map notices too — with the banner
  reading "updating…". The obsolete minFibers/maxNetworks settings and
  their staleness arm are removed.
- A translucent progress bar sweeps across the triggering button during a
  rebuild, advancing at real phase boundaries (the rebuild is synchronous,
  so the sweep is stepped, not animated).
- Opening and closing a line-annotation session without edits no longer
  rewrites the fiber file or invalidates derived views (`saveSessionAsFiber`
  skips no-op saves via serializer-level comparison, decided before any
  branch-metadata side effects; non-skips log their first differing field).

## Performance: memoized rebuilds

Two toolbar buttons. **Update** memoizes per-fiber preparation and
per-(H,V)-pair crossing detection in `GlobalLayoutCache`, keyed on chained
content digests of exactly what each artifact consumes (never generation
counters; fileName identity; links deliberately outside the fiber digest).
The per-pair detection refactor makes cached and fresh builds the *same
computation*, so results are identical by construction — regression-tested
by digest deep-compare, minimal-recomputation counters, invalidation
granularity, chirality-flip, empty shards, and repair-tie-break replay.
**Full rebuild** drops the caches and, when inputs match the last memoized
Update, verifies the output digest — a mismatch disables memoization for
the session. An exact hoisting of the ascent's k-independent invariants
halved the solve.

PHercParis4 (452 fibers, ~1M points, 50 500 pairs): cold 2.3 s, warm Update
1.1 s, single-fiber edit 1.1 s (50 250 pairs reused) — vs ~10 s before this
PR. Every optimization step was verified bit-identical on this dataset.

Also fixed en route: opening/closing an *unedited* line-annotation session
no longer rewrites the fiber file and bumps the fiber generation
(`saveSessionAsFiber` skips no-op saves via serializer-level comparison) —
this was invalidating the map on every fiber open.

## Verification

- 4 test suites, 71 slots: `test_fiber_winding_solver` (28; deliberately
  crumpled spiral fixtures, both chiralities, non-finite-input safety),
  `test_fiber_global_layout` (17, incl. the cache exactness matrix),
  `test_fiber_network_layout` (9; panel path preserved bit-identically
  through the refactor), `test_fiber_map_staleness` (17).
- The plan and the completed work each went through three independent
  Codex reviews (full-context, zero-context, and specialist adversarial
  briefs) plus a verification round; all confirmed findings are fixed and
  regression-tested.
- Real-data validation throughout on PHercParis4 (452 fibers): all fibers
  placed, winding→radius monotone through the dense band, deterministic
  across runs.

## Notes for reviewers

- The panel-mode `buildLayout` is untouched in behavior; it now shares
  preparation seams with the global path (bit-identical, tested).
- The solve stays deliberately uncached: warm-starting would change repair
  trajectories and break exactness. It is the 1 s floor; background
  rebuild is the natural follow-up if that matters.
- GUI-level flows (package-switch cache reset, Full-rebuild verify,
  auto-update scheduling, the progress overlay) are compile- and
  read-verified; the solver/layout logic carries the tests.

https://claude.ai/code/session_01K3HGTzfhCF2MMVVJQts9yt


Bullo27:
Built and ran this on Linux in the default configure, then went after three of the invariants the header asserts — using a synthetic spiral written against the public API rather than the PR's own fixtures, so the checks are independent of the tests under review. Short version: **everything I could check holds**, and I failed to break the two structural claims the memoized cache rests on.

Environment: Ubuntu 24.04, g++, Ninja, `CMAKE_BUILD_TYPE=Release`, `VC_USE_PCH=OFF`, `VC_BUILD_APPS=ON`, `VC_TESTING=ON`. Branch head `92badb603`, base `c53c74a29` — which is `main` exactly, so no merge staleness to work around.

## Build and suites

| check | result |
|---|---|
| `ninja` (default configure) | exit **0**, no failures |
| warnings in PR-touched files | **0** (the only two in the whole build are third-party: libigl `slim.h`, pastix `graph_compute_projection.c`) |
| `ctest` total | 154 → **156** (your two new registrations) |
| `ctest -j6` | **156/156 passed, 0 failed** |

Slot counts match your body exactly:

| suite | measured | claimed |
|---|---|---|
| `fiber_winding_solver` | 28 passed / 0 failed | 28 |
| `fiber_global_layout` | 17 passed / 0 failed | 17 |
| `fiber_network_layout` | 9 passed / 0 failed | 9 |
| `fiber_map_staleness` | 17 passed / 0 failed | 17 |
| | **71** | 71 |

## Independent probe of the documented invariants

I built a synthetic scroll — `r(ψ) = 2000 + 354·ψ/2π` voxels (≈0.85 mm wrap spacing, your dense-band figure), 6 H fibers each sweeping ~1.6 turns at staggered z, 7 V fibers sweeping z at fixed angle — in two variants: one where H and V radii coincide on a shared winding, and one where H rides a sheet-thickness inward as the papyrus model says it should. Both yield 68 crossings. Instrument check first: a probe that detects nothing proves nothing, so the crossing count and the surviving-constraint count are both asserted non-zero before anything else runs.

Checks 1 and 2 below were run on the **coincident-radius** variant (66 of 68 crossings surviving) — the harder of the two, for the reason the last section explains.

**1. Shard-order invariance** — *"Shards may arrive in any order; they are assembled in canonical (hFiber, vFiber) order, which reproduces the plain overload's constraint order exactly."*

This is the claim the whole memoization argument rests on, so I tested it adversarially: built all 42 pair shards myself via `canonicalizeTrace` + `detectPairCrossings`, then called the sharded `solveWindings` overload with the shards **shuffled**, comparing a full field-by-field digest of `SolveResult` (placements, every crossing's `n`/`deltaR`/`confidence`/`kind`/`status`/`violationTurns`, dropped links, link errors) against the plain overload.

- natural order == plain overload ✅
- **8 independently shuffled orders, all byte-identical to the plain overload** ✅

**2. Determinism** — 6 repeated solves in-process are byte-identical. Across 5 *separate processes* the only variation in the suite binaries is QTest's own wall-clock totals line (`201ms` vs `191ms`); mask durations and the outputs are identical. Also identical across four `QT_HASH_SEED` values. That is structurally supported by the code: `FiberWindingSolver.cpp` contains no `QHash`/`QSet`/`unordered_*` at all — only ordered `std::map`/`std::set` — so there is no per-process hash-order to leak in.

**3. Gauge invariance** — *"Per-fiber canonical gauges make the whole solve invariant to each fiber's arbitrary unwrap branch."*

Added an independent random integer number of turns to each fiber's `theta` (`θ_i += 2π·m_i`, `m_i ∈ [-4, 4]`) and required every fiber's `windingLo` to be **unchanged**, not merely shifted. On the separated-radius variant, across 8 random gauge assignments: **8/8 fully invariant, zero fibers moved, `droppedCrossingCount` 0 throughout.** (On the coincident-radius variant it does not fully hold — that is the next section, and it is the reason I ran the other two checks there.)

**4. Non-finite input safety** — injected NaN and ±Inf into `theta`, `radius` and `z` of both an H and a V fiber (6 cases). No crash, no hang, every returned `windingLo`/`windingHi` finite.

## One characterisation, offered as information rather than a defect

Invariance (3) holds on geometry where H and V are physically separated. It degrades when they are not.

My first fixture built H and V radii from the *same* spiral formula, so every same-winding crossing sat at **ΔR = 0 exactly** — precisely the boundary your sign-only classifier splits on. There, the 2π add/subtract leaves a ~1e-15 residue in ψ (the gauge cancels algebraically but not in floating point), enough to flip `inside`/`outside`:

| fixture | min \|ΔR\| | gauge shifts fully invariant | `dropX` |
|---|---|---|---|
| H and V radii identical | 0 vx | 5/8 — one fiber moved a **full winding** | 1–3 |
| H one sheet inward | 60 vx | **8/8** | 0 |
| H further inward | 120 vx | **8/8** | 0 |

I don't think this is a bug, and I want to be clear about why I'm reporting it anyway: your header argues explicitly for sign-only classification with no tie band, because a band "manufactures constraints out of sub-band radial noise". The table is just the quantified edge of that trade-off — it says the method is conditioned on H and V actually being resolvably separated in radius, and that at exact coincidence a single fiber can land one winding out. On real data that separation is the physical premise of the whole approach, so the degenerate column should be unreachable; it took a deliberately unphysical fixture to produce it.

## Two things I suspected and could not substantiate

Worth recording so nobody re-runs them:

- **Hash-order nondeterminism.** `FiberNetworkLayout.cpp` uses `std::unordered_map` in 8 places, which is where I expected iteration order to leak into results. It doesn't reach the solver (Qt-free, ordered containers only), and the collection sites normalise with `std::sort` before use. No effect found.
- **Cross-process nondeterminism.** My first pass hashed whole test-binary stdout across runs and reported 2–4 distinct hashes. That was QTest's timing line, not the computation — after masking durations the outputs are identical. False alarm on my side.

## Limits

- **No GUI run.** You flag the GUI-level flows (package-switch cache reset, Full-rebuild verify, auto-update scheduling, progress overlay) as compile- and read-verified only; I have not exercised them either, so that part of the PR is still unreviewed by anyone at runtime.
- **No real-data replication.** I have no PHercParis4 project here, so the 452-fiber numbers, the 2.3 s cold / 1.1 s warm timings, and the 1339-candidate-drops → ~10-genuine-regressions reduction are yours alone — I neither confirm nor dispute them.
- The synthetic scroll is smooth and uncrumpled; it exercises the solver's logic and invariants, not its behaviour under real sheet drift.

Happy to share the probe harness (~200 lines, links only `FiberWindingSolver.cpp`, no Qt) if it would be useful as a regression test — the shard-shuffle and gauge-shift checks in particular seem worth having in-tree, since both are properties that would fail silently.

---
*Claude (AI agent) on behalf of @Bullo27. All results measured on Ubuntu 24.04 / g++ / Ninja with the default configure (Release, `VC_USE_PCH=OFF`), against branch head `92badb603` on base `c53c74a29`, on 2026-08-26.*

hendrikschilling:
I'd recommend not attacking the performance risk (4.) until after this PR as a separate follow up.

```
  1. P1: VC3D does not compile with its normal PCH. The lambda named emit conflicts with Qt’s emit macro, producing const auto = .... Both Linux VC3D and Windows fail here:
     FiberWindingSolver.cpp:115 (https://github.com/ScrollPrize/villa/blob/92badb6030e756ba776b0515c56adeffc37cea18/volume-cartographer/apps/VC3D/FiberWindingSolver.cpp#L115). Rename it to
     emitBranch. The independent reviewer used VC_USE_PCH=OFF, which is why they missed this.

  2. P1: automatic updates synchronously freeze the GUI. The timer calls rebuildLayout(false) directly on the GUI thread FiberMapWorkspace.cpp:891
     (https://github.com/ScrollPrize/villa/blob/92badb6030e756ba776b0515c56adeffc37cea18/volume-cartographer/apps/VC3D/FiberMapWorkspace.cpp#L891), and the full snapshot, solver and layout
     execute synchronously FiberMapWorkspace.cpp:991
     (https://github.com/ScrollPrize/villa/blob/92badb6030e756ba776b0515c56adeffc37cea18/volume-cartographer/apps/VC3D/FiberMapWorkspace.cpp#L991). The PR’s own measurements are 1.1       
     seconds warm and 2.3 seconds cold. The progress repaint() shows intermediate states but does not process interaction. A stale visible map therefore automatically stalls all VC3D      
     interaction. Computation should run in a worker from an immutable snapshot, with generation validation before applying the result.

  3. P2: the new no-umbilicus test has an out-of-bounds fixture. arcPoints(..., 0.0, 2.0) produces 400 points, indexed 0..399, but the test requests control index 400
     test_fiber_global_layout.cpp:719
     (https://github.com/ScrollPrize/villa/blob/92badb6030e756ba776b0515c56adeffc37cea18/volume-cartographer/apps/VC3D/test/test_fiber_global_layout.cpp#L719), which is dereferenced       
     unchecked by makeFiber() test_fiber_global_layout.cpp:84
     (https://github.com/ScrollPrize/villa/blob/92badb6030e756ba776b0515c56adeffc37cea18/volume-cartographer/apps/VC3D/test/test_fiber_global_layout.cpp#L84). This explains the Linux      
     base-test abort. Release builds silently running through this UB do not validate the intended no-umbilicus behavior.

  4. P2 performance risk: selection still scales linearly with all displayed geometry. Every scene click scans every segment of every fiber FiberMapWorkspace.cpp:1653
     (https://github.com/ScrollPrize/villa/blob/92badb6030e756ba776b0515c56adeffc37cea18/volume-cartographer/apps/VC3D/FiberMapWorkspace.cpp#L1653). Selection then deletes and recreates   
     glow paths for every other fiber in the linked network FiberMapWorkspace.cpp:1776
     (https://github.com/ScrollPrize/villa/blob/92badb6030e756ba776b0515c56adeffc37cea18/volume-cartographer/apps/VC3D/FiberMapWorkspace.cpp#L1776). That was less exposed when only a few  
     top networks were shown; it now scales to the entire map. A scene-space index and one reusable combined glow path per network would avoid this.

```

skyward7187:
@hendrikschilling Thanks for the review — all three items are addressed in the four commits just pushed (`0e7c554` … `ee3bc8b`), and #4 is deferred per your recommendation.

**1. `emit` lambda (P1)** — renamed to `emitBranch` (0e7c554). Reproduced your failure mode first: compiling the TU with `-include QtCore/QObject` produces exactly the reported error, and is clean after the rename. Also verified with a scratch `-DVC_USE_PCH=ON` build of the VC3D target, which now links.

**2. Synchronous rebuild freeze (P1)** — rebuilds now run on a dedicated single-thread pool from an immutable snapshot (d7881d1, hardened in ee3bc8b). The shape follows your suggestion:

- The GUI thread keeps only the snapshot capture and the scene/tree publication; conversion, digesting, and the solve run in the worker. The job owns everything it touches — including the memoization cache, which is moved out of the workspace for the flight — so the two threads share no mutable state.
- Generation validation before applying, as requested: a monotonic epoch (bumped by clears/package switches/shutdown) plus apply-time comparison of package generation, the full annotation frame (grid *and* voxel size), and fiber/umbilicus generations + fingerprint. A result whose inputs changed mid-flight is discarded and a follow-up update is coalesced; its content-keyed cache is retained so the re-run stays warm.
- The request/coalesce/pending logic lives in `FiberMapRebuildQueue`, a small pure state machine with its own QtTest suite (9 tests). Buttons disable during a build and a marquee sweeps the active button; the status line now reports the true GUI stalls (snapshot + publish) separately from worker time — on PHercParis4 (452 fibers) the GUI-thread cost is now tens of ms instead of 1.1–2.3 s.

**3. Out-of-bounds fixture (P2)** — the crossing fixture uses `{100, 399}`, and `makeFiber()` in both fiber layout test files now throws on any bad control index so a fixture mistake fails loudly in every build type; a test exercises the guard (965ff34).

**4. Selection scaling** — left alone for now, as you suggested.

Verification: all 11 fiber suites pass in Release (the 5 touched ones also in Debug), the PCH-ON VC3D build links, and the real-data harness produces bit-identical per-fiber turns before/after (the async change touches no solver/layout code beyond the rename).

Bullo27:
Followed this through the merge. Two things: a scope correction on my earlier comment, and confirmation that `e1d7bf0a1` is green.

## Correction: my verification did not cover the config that mattered

My earlier comment reported a clean build against branch head `92badb603` in the **default** configure (`VC_USE_PCH=OFF`). That was accurate as scoped, but the merged commit contains a fix I could not have caught:

```cpp
-    const auto emit = [&](std::size_t begin, std::size_t end) {
+    // Named to survive Qt's `emit` macro: with the project PCH, Qt headers
+    // reach even this Qt-free TU, and a lambda named `emit` fails to parse.
+    const auto emitBranch = [&](std::size_t begin, std::size_t end) {
```

I went back and built both configs at both commits to establish exactly what my review missed:

| | `VC_USE_PCH=OFF` (default) | `VC_USE_PCH=ON` |
|---|---|---|
| head I reviewed, `92badb603` | clean, 156/156 | **FAILS** |
| merged, `e1d7bf0a1` | clean, **157/157** | clean, **157/157** |

The failure at `92badb603` under PCH is exactly one object:

```
FAILED: apps/VC3D/CMakeFiles/VC3D.dir/FiberWindingSolver.cpp.o
  ... -include .../CMakeFiles/VC3D.dir/cmake_pch.hxx ...
FiberWindingSolver.cpp:115:21: error: expected unqualified-id before '=' token
  115 |     const auto emit = [&](std::size_t begin, std::size_t end) {
```

So the fix was necessary and the comment on it is exactly right — the shared PCH puts Qt headers in front of a translation unit that includes no Qt itself, and `emit` is a macro.

## The generalisable bit

I had been treating `VC_USE_PCH=OFF` as the strict configuration, on the reasoning that PCH-off catches headers that are not self-contained. Measuring it, that is only half true: **neither config dominates.**

- **PCH off** catches a TU that forgets an include it depends on.
- **PCH on** catches a TU whose identifiers collide with something the carrier header drags in — `emit`, `signals`, `slots`, `foreach` — which PCH-off cannot see, because without the PCH those macros never reach the file.

A Qt-free source file is precisely where this bites, since nothing in the file itself suggests Qt macros are in scope. Worth knowing for anything else landing under `apps/VC3D/` that is deliberately Qt-free, as `FiberWindingSolver.cpp` is.

For what it is worth, the PCH-on build of `e1d7bf0a1` is clean across the whole tree, so no other translation unit currently carries this collision.

## Merged state

`e1d7bf0a1` on Ubuntu 24.04 / g++ / Ninja:

- default configure: `ninja` exit 0, **`ctest -j6` 157/157 passed, 0 failed**
- `VC_USE_PCH=ON`: `ninja` exit 0, zero failed objects, **157/157**
- ctest went 154 → 157: `fiber_winding_solver`, `fiber_global_layout`, and `fiber_map_rebuild_queue`, the last from the `FiberMapRebuildQueue.hpp` work added after the head I reviewed

The three invariants I probed earlier — shard-order invariance, determinism, and gauge invariance — were measured against `92badb603`. `FiberWindingSolver.hpp` is byte-identical between that head and the merge, and the only `.cpp` change is the `emit` → `emitBranch` rename, so those results carry over unchanged; I am flagging the provenance rather than re-asserting them as fresh measurements.

## Limits

- I did not re-run the invariant probe against the merged tree — see the note directly above for why I think it carries, and treat it as inherited rather than re-measured if that distinction matters to you.
- `FiberMapRebuildQueue` and the ~500 lines of `FiberMapWorkspace.cpp` added after my review are **unreviewed by me**; my earlier comment does not cover them. `test_fiber_map_rebuild_queue` passes, but I have not read that code.
- Still no GUI run, and still no real-data replication of the PHercParis4 figures.

---
*Claude (AI agent) on behalf of @Bullo27. Measured on Ubuntu 24.04 / g++ / Ninja against `92badb603` and merged `e1d7bf0a1`, both `VC_USE_PCH` settings, on 2026-08-26.*

PR #1606, opened 2026-08-26
vesuvius: tifxyz write_metadata recomputes the bbox from valid vertices

Nieuwlaar:
**In one sentence:** `TifxyzWriter.write_metadata()` writes the bbox of the surface you hand it, so the Python tifxyz API can finally correct a wrong stored bbox instead of copying it back to disk.

**One real example:** Starting with `same_wrap002204_lasagna`, one of the PHercParis4 verified patches with a stale meta.json bbox reported in #1272, I called `read_tifxyz()` and then `TifxyzWriter(...).write_metadata(surface)`, and the patch's meta.json came out with the bbox of its actual vertices.

**Before:** the rewrite was a no-op. The poisoned box went straight back to disk, and a z-range query still selected the patch nowhere near where it lives.

```
patch same_wrap002204_lasagna  (164x173 grid, 24,272 invalid vertices)
  stored meta.json bbox              [[-0.2, -0.2, -0.2], [3412.2, 2957.0, 13247.7]]
  bbox of its valid vertices         [[2800.9, 2240.5, 12565.3], [3412.2, 2957.0, 13247.7]]

after read_tifxyz() -> TifxyzWriter.write_metadata():
  meta.json bbox                     [[-0.2, -0.2, -0.2], [3412.2, 2957.0, 13247.7]]
  repaired: False

list_tifxyz(z_range=(1000, 2000)): 1 of 1 patch selected (the patch really spans z [12565.3, 13247.7], so the right answer is 0)

after translating x by +100 in memory and calling write_metadata():
  meta.json bbox                     [[-0.2, -0.2, -0.2], [3412.2, 2957.0, 13247.7]]
  x_min reflects the edit: False
```

**After this PR:** same patch, same calls.

```
patch same_wrap002204_lasagna  (164x173 grid, 24,272 invalid vertices)
  stored meta.json bbox              [[-0.2, -0.2, -0.2], [3412.2, 2957.0, 13247.7]]
  bbox of its valid vertices         [[2800.9, 2240.5, 12565.3], [3412.2, 2957.0, 13247.7]]

after read_tifxyz() -> TifxyzWriter.write_metadata():
  meta.json bbox                     [[2800.9, 2240.5, 12565.3], [3412.2, 2957.0, 13247.7]]
  repaired: True

list_tifxyz(z_range=(1000, 2000)): 0 of 1 patch selected (the patch really spans z [12565.3, 13247.7], so the right answer is 0)

after translating x by +100 in memory and calling write_metadata():
  meta.json bbox                     [[2900.9, 2240.5, 12565.3], [3512.2, 2957.0, 13247.7]]
  x_min reflects the edit: True
```

**Proof:** the two blocks are the same script on a copy of the same published patch, run against `main` at `c53c74a29` and against this branch. The recomputed box also matches, to the decimal, what my out-of-tree audit tool derives for this patch by replaying VC3D's loader validity rules, which is an independent check that the recomputation is the right box and not just a different one.

**Why / where this is useful:** anyone using the `vesuvius.tifxyz` Python API to edit surfaces (translate, crop, repair coordinates, or repair a stale box like the 106 patches in #1272) and then persist the metadata. Today that silently writes the pre-edit box, and downstream z filtering acts on it.

**Details of the failure mode:** `write_metadata()` carried `surface.bbox` verbatim, which is whatever meta.json said at load time:

```python
use_bbox = bbox if bbox is not None else surface.bbox
```

The full `write()` path already recomputes via `compute_grid_bounds`, so only the direct-metadata path leaked stale values.

## Details

Branch `fix-tifxyz-write-metadata-bbox`, single commit `2514dfb9`, on top of `c53c74a29`.

`write_metadata()` now derives the bbox from the current coordinate arrays and validity mask by default, with an opt-out:

```python
def write_metadata(self, surface, bbox=None, *, recompute_bbox=True):
    ...
    if bbox is not None:
        use_bbox = bbox                      # explicit override still wins
    elif recompute_bbox:
        use_bbox = compute_grid_bounds(      # default: derive from valid vertices
            surface._x, surface._y, surface._z, surface._valid_mask
        )
    else:
        use_bbox = surface.bbox              # opt-out: carry the stored value
```

`write()` passes an explicit bbox, so its behaviour is unchanged. `recompute_bbox=False` restores the old carry-verbatim behaviour for callers that deliberately preserve a stored box.

### Test

New `vesuvius/src/vesuvius/tifxyz/test_writer.py` (pytest style, next to the existing in-package `test_labels.py`), written failing-first:

- `test_write_metadata_recomputes_bbox_after_inplace_edit`: load, mutate coordinates, `write_metadata()`, meta.json must reflect the edit.
- `test_write_metadata_repairs_a_stale_stored_bbox`: a surface loaded with a wrong stored box is written out corrected. This is the #1272 repair path in miniature.
- `test_write_metadata_opt_out_keeps_stored_bbox`: `recompute_bbox=False` carries the loaded box verbatim.
- `test_explicit_bbox_argument_still_wins`: an explicit `bbox=` overrides both paths, covering the `write()` call site.

```
$ PYTHONPATH=src python -m pytest src/vesuvius/tifxyz/test_writer.py   # this branch
4 passed
$ PYTHONPATH=src python -m pytest src/vesuvius/tifxyz/test_writer.py   # same file on main
3 failed, 1 passed
$ PYTHONPATH=src python -m pytest src/vesuvius/tifxyz/                 # this branch, whole package
13 passed
```

Refs #1272 and companion to #1605: together they make every Python tifxyz metadata path derive its bbox from the vertices actually present. The C++ side was handled in #1285.

### DIFF
--- a/vesuvius/src/vesuvius/tifxyz/test_writer.py
+++ b/vesuvius/src/vesuvius/tifxyz/test_writer.py
@@ -1,0 +1,75 @@
+import json
+from pathlib import Path
+
+import numpy as np
+
+from vesuvius.tifxyz import read_tifxyz, write_tifxyz
+from vesuvius.tifxyz.types import Tifxyz
+from vesuvius.tifxyz.writer import TifxyzWriter
+
+
+def _fresh_surface() -> Tifxyz:
+    h, w = 4, 5
+    y_grid, x_grid = np.mgrid[:h, :w]
+    return Tifxyz(
+        _x=(10.0 + x_grid).astype(np.float32),
+        _y=(20.0 + y_grid).astype(np.float32),
+        _z=np.full((h, w), 30.0, dtype=np.float32),
+        uuid="segment",
+    )
+
+
+def _load_mutate(tmp_path: Path) -> Tifxyz:
+    """Write a surface, read it back, then edit its coordinates in place."""
+    write_tifxyz(tmp_path / "segment", _fresh_surface())
+    surface = read_tifxyz(tmp_path / "segment")
+    assert surface.bbox == (10.0, 20.0, 30.0, 14.0, 23.0, 30.0)
+    surface._x[:] += 100.0  # e.g. translate the surface after loading
+    return surface
+
+
+def test_write_metadata_recomputes_bbox_after_inplace_edit(tmp_path: Path) -> None:
+    surface = _load_mutate(tmp_path)
+
+    TifxyzWriter(tmp_path / "segment", overwrite=True).write_metadata(surface)
+
+    meta = json.loads((tmp_path / "segment" / "meta.json").read_text())
+    assert meta["bbox"] == [[110.0, 20.0, 30.0], [114.0, 23.0, 30.0]]
+
+
+def test_write_metadata_repairs_a_stale_stored_bbox(tmp_path: Path) -> None:
+    """A surface loaded with a wrong stored bbox is written out corrected."""
+    write_tifxyz(tmp_path / "segment", _fresh_surface())
+    meta_path = tmp_path / "segment" / "meta.json"
+    meta = json.loads(meta_path.read_text())
+    meta["bbox"] = [[-0.25, -0.25, -0.25], [14.0, 23.0, 30.0]]
+    meta_path.write_text(json.dumps(meta))
+
+    surface = read_tifxyz(tmp_path / "segment")
+    TifxyzWriter(tmp_path / "segment", overwrite=True).write_metadata(surface)
+
+    assert json.loads(meta_path.read_text())["bbox"] == [
+        [10.0, 20.0, 30.0], [14.0, 23.0, 30.0]
+    ]
+
+
+def test_write_metadata_opt_out_keeps_stored_bbox(tmp_path: Path) -> None:
+    surface = _load_mutate(tmp_path)
+
+    TifxyzWriter(tmp_path / "segment", overwrite=True).write_metadata(
+        surface, recompute_bbox=False
+    )
+
+    meta = json.loads((tmp_path / "segment" / "meta.json").read_text())
+    assert meta["bbox"] == [[10.0, 20.0, 30.0], [14.0, 23.0, 30.0]]
+
+
+def test_explicit_bbox_argument_still_wins(tmp_path: Path) -> None:
+    surface = _load_mutate(tmp_path)
+
+    TifxyzWriter(tmp_path / "segment", overwrite=True).write_metadata(
+        surface, bbox=(1.0, 2.0, 3.0, 4.0, 5.0, 6.0)
+    )
+
+    meta = json.loads((tmp_path / "segment" / "meta.json").read_text())
+    assert meta["bbox"] == [[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]]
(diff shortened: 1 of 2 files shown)


Nieuwlaar:

Bullo27:
Verified by running, on `main` at `c53c74a29`. The change is correct and the test discriminates. One question about the empty-surface case that this PR widens the reach of.

## What I checked

**The unpacking is right.** `compute_grid_bounds` returns a flat `(x_min, y_min, z_min, x_max, y_max, z_max)`, and `write_metadata` splits it as `[[0,1,2],[3,4,5]]` → `[[min_x, min_y, min_z], [max_x, max_y, max_z]]`. Valid-masked, x/y/z low-high. On a synthetic surface (40×31, ragged invalid border plus interior hole, 1124 valid vertices) that lands on `[[3001.8, 2502.0, 8005.4], [4797.5, 4499.6, 12997.6]]`, byte-identical to what `spiral-fitting/tifxyz.py:save_tifxyz` writes for the same points. So this now agrees with the convention #1604 names as canonical.

**It matches the path that already existed.** `write()` at `:295-302` already computed exactly this and handed it over as `bbox=`. This PR makes the standalone `write_metadata(surface)` call behave like the full write path instead of silently carrying a stale `surface.bbox` — a consistency fix rather than a new convention.

**Blast radius of flipping the default is nil inside the repo.** The only in-tree caller of `write_metadata` is `writer.py:302` itself, which passes an explicit `bbox=`, so it is unaffected by `recompute_bbox`'s default. The change is only observable to external callers, and `recompute_bbox=False` gives them the old behaviour verbatim.

**The tests discriminate.** Run against unpatched production code rather than only on the PR:

- on this PR: **4 passed**
- same test file against `main`'s `writer.py`: **3 failed, 1 passed**

## One question: what should an all-invalid surface write?

`compute_grid_bounds` returns `(0.0, 0.0, 0.0, 0.0, 0.0, 0.0)` when `not mask.any()`, so this writer emits:

| surface | `vesuvius` writer | `spiral-fitting/tifxyz.py` | `merge_concat_runs` (after #1605) |
|---|---|---|---|
| all-invalid | `[[0,0,0], [0,0,0]]` | `[[-1,-1,-1], [-1,-1,-1]]` | `[[-1,-1,-1], [-1,-1,-1]]` |

Measured, both before and after this PR — so it is pre-existing in `compute_grid_bounds` and **not** introduced here. I am raising it because this PR makes that branch reachable from `write_metadata(surface)` by default, where previously it only came through `write()`.

The reason it seems worth a decision rather than a shrug: a zero box is *plausible-looking* in exactly the way a sentinel is not. `[[-1,-1,-1],[-1,-1,-1]]` cannot be mistaken for geometry, but `[[0,0,0],[0,0,0]]` is a degenerate box at the origin, and the consumer this whole cluster is about — z-range filtering — would match it for any band starting at 0 rather than skipping it. Same class of silent-wrong-answer as the axis flip in #1604, just at the empty end.

I have not proposed a patch because the right answer depends on what you intend `bbox` to mean for an empty surface, and that is your call: leave it as-is, return `None` so the key is simply omitted, or align on the `-1` sentinel the two spiral writers use. Aligning would make all three writers agree on both the populated and the empty case, which is a nice property to be able to state.

## Limits

- **Synthetic surfaces only** — no real tifxyz corpus round-tripped through this writer.
- I did not check whether any consumer currently relies on `bbox` being carried verbatim across a load/save cycle; I only established that no in-tree caller hits the changed default.
- The empty-surface behaviour is measured, but I have not traced whether any real pipeline actually produces an all-invalid surface and writes it.

---
*Claude (AI agent) on behalf of @Bullo27. Measured against `main` at `c53c74a29` and PR head, Python 3.13 / numpy 2.4.2, on 2026-08-26.*

PR #1607, opened 2026-08-26
spiral: a dataset with tracks and no outer_shell/ should not crash the headless fit

Nieuwlaar:
**In one sentence:** a spiral dataset that ships tracks but no `outer_shell/` runs instead of dying during input loading.

**One real example:** Starting with a dataset root holding real public PHercParis4 inputs (`umbilicus.json`, two verified patches, a tracks DBM) and no `outer_shell/`, I ran `python fit_spiral.py --dataset <root>` with the shell losses off, and loading completed instead of raising `FileNotFoundError`.

**Before:** on `main` at `c53c74a29`, with `loss_weight_shell_outer = 0` so nothing asks for a shell:

```
$ FIT_SPIRAL_CONFIG_OVERRIDES='{"loss_weight_shell_outer": 0.0, ...}' python fit_spiral.py --dataset /tmp/ds
...
  File "spiral-fitting/fit_spiral.py", line 4611, in run
    self.load_host_inputs()
  File "spiral-fitting/fit_spiral.py", line 1527, in load_host_inputs
    shell_patch = load_tifxyz(self.shell_path)
  File "spiral-fitting/tifxyz.py", line 362, in load_tifxyz
    with open(f'{path}/meta.json', 'r') as meta_json:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/ds/outer_shell/meta.json'
```

There is no way around it short of patching the file: `outer_shell/` is never listed as an input for this run, no loss reads it, and the CLI has no flag that skips it while keeping the tracks. Setting `input_use_outer_shell: false` avoids the crash but is not a fix, since it is meant to disable shell supervision, not to work around a loader that opens a file it was told not to use.

**After this PR:** same dataset, same command, loading runs to the end:

```
skipping all verified/unverified patch loading
 loaded 0 patches
 loaded 0 unverified patches
fiber links: 0 resolved
pcls: 0 cross-patch, 0 unattached
loading tracks from /tmp/ds/tracks/2um_ds2_ps256_surf_v2.dbm
loaded 6 tracks within z-roi [4000, 17000)
fitting 0 patches
```

and with the two patches enabled as well:

```
loading tracks from /tmp/ds/tracks/2um_ds2_ps256_surf_v2.dbm
loaded 6 tracks within z-roi [4000, 17000)
fitting 2 patches
```

**Proof:** the two blocks are the same command against the same dataset root, one on `main` at `c53c74a29` and one on this branch (dataset root and checkout paths shortened in both transcripts). My box has no CUDA and no built `vc_spiral` extension, so the "after" run stops later at `Torch not compiled with CUDA enabled` in `build_device_state()`. That is my hardware, not the fix: on `main` it never gets that far, because host input loading dies first.

**Why / where this is useful:** anyone fitting a dataset assembled from tracks alone, which is the normal shape when you have a track store for a scroll but no outer shell for it yet. Today that layout simply cannot be run headless without editing `fit_spiral.py`.

## Details

Branch `fix-fit-spiral-optional-outer-shell`, single commit `f7c57789`, on top of `c53c74a29`.

`conventional_input_paths()` composes an `outer_shell` path from the dataset root and, as its docstring says, deliberately performs no existence probing. So `context.shell_path` is a non-empty string for every conventional dataset, whether or not the directory exists. `load_host_inputs()` then derived the track filter from that string alone (`spiral-fitting/fit_spiral.py:1519`):

```python
filter_tracks_by_shell = bool(self.tracks_dbm_path) and bool(self.shell_path)
shell_patch = None
if self.outer_shell_required() or filter_tracks_by_shell:
    ...
    shell_patch = load_tifxyz(self.shell_path)
```

`outer_shell_required()` is the fit-input catalog's answer to "does anything need the shell", and it is correct: with shell losses off and `dense_spacing_mode != "winding_model"` it returns False. `filter_tracks_by_shell` bypasses it, so the presence of tracks alone is enough to open the shell.

The change is to require the directory to be there:

```python
filter_tracks_by_shell = (
    bool(self.tracks_dbm_path) and bool(self.shell_path)
    and os.path.isdir(self.shell_path))
```

Nothing else moves. A shell that the losses or the winding model genuinely require is still loaded through `outer_shell_required()`, and still fails loudly if it is missing. Only the "we have tracks, so let us also open the shell to trim them" path becomes conditional on the shell existing. When it does exist, behaviour is identical.

### Test

New `spiral-fitting/tests/test_optional_outer_shell.py`, written failing-first, building a dataset root with a umbilicus, a one-track DBM and no `outer_shell/`:

- `test_absent_outer_shell_does_not_block_track_loading`: `load_host_inputs()` completes, `filter_tracks_by_shell` is False, `shell_patch` is None, and the track is loaded.
- `test_a_required_outer_shell_still_fails_when_it_is_missing`: with shell losses on, `outer_shell_required()` is True and loading still raises `FileNotFoundError`, so the guard cannot hide a shell someone actually needs.

```
$ python -m pytest tests/test_optional_outer_shell.py    # this branch
2 passed
$ python -m pytest tests/test_optional_outer_shell.py    # same file on main
1 failed, 1 passed

$ python -m pytest tests/test_input_toggles.py tests/test_input_catalog.py \
                   tests/test_config.py tests/test_spiral_helpers.py \
                   tests/test_optional_outer_shell.py
49 passed
```

The tests that exercise the surrounding input plumbing are the four above; the rest of the suite needs CUDA or the native `vc_spiral` extension, which this machine does not have, so I could not run it.

### DIFF
--- a/spiral-fitting/fit_spiral.py
+++ b/spiral-fitting/fit_spiral.py
@@ -1516,7 +1516,14 @@ class FitContext:
         # Patch loading and ROI filtering
         # ==========================================================================
 
-        filter_tracks_by_shell = bool(self.tracks_dbm_path) and bool(self.shell_path)
+        # The conventional layout resolves an outer_shell path without probing
+        # for it, so a dataset that ships tracks but no outer_shell/ still gets
+        # a shell path here. Only filter tracks by a shell that actually exists;
+        # a shell the losses genuinely require is still loaded (and still fails
+        # loudly) through outer_shell_required().
+        filter_tracks_by_shell = (
+            bool(self.tracks_dbm_path) and bool(self.shell_path)
+            and os.path.isdir(self.shell_path))
         shell_patch = None
         if self.outer_shell_required() or filter_tracks_by_shell:
             if not self.shell_path:
--- a/spiral-fitting/tests/test_optional_outer_shell.py
+++ b/spiral-fitting/tests/test_optional_outer_shell.py
@@ -1,0 +1,99 @@
+"""A dataset that ships tracks but no outer_shell/ must still load.
+
+The conventional layout resolves an outer_shell path without probing for it,
+so before this guard load_host_inputs() opened that path for track filtering
+even when no loss required the shell, and the fit died on the missing
+outer_shell/meta.json.
+"""
+import dbm
+import json
+import pickle
+import sys
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+SPIRAL_DIR = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(SPIRAL_DIR))
+
+from config import Config, FitConfig  # noqa: E402
+from fit_session import ScrollSpec, SpiralInputPaths  # noqa: E402
+from fit_spiral import FitContext  # noqa: E402
+
+Z_BEGIN, Z_END = 1000, 2000
+
+
+def tracks_only_dataset(root: Path) -> SpiralInputPaths:
+    """A dataset root with umbilicus + tracks and deliberately no outer_shell."""
+    root.mkdir(parents=True, exist_ok=True)
+    (root / "umbilicus.json").write_text(json.dumps({
+        "control_points": [
+            {"x": 3000, "y": 3000, "z": z, "score": 100}
+            for z in range(0, 4001, 500)
+        ]
+    }))
+
+    (root / "tracks").mkdir()
+    dbm_path = root / "tracks" / "surf.dbm"
+    track = np.stack([
+        np.arange(Z_BEGIN + 100, Z_BEGIN + 132),
+        np.full(32, 3200),
+        np.arange(3300, 3332),
+    ], axis=-1).astype(np.int32)
+    with dbm.open(str(dbm_path), "c") as db:
+        db[f"h:{Z_BEGIN}"] = pickle.dumps([track])
+
+    assert not (root / "outer_shell").exists()
+    return SpiralInputPaths(
+        dataset_root=str(root),
+        umbilicus=str(root / "umbilicus.json"),
+        tracks_dbm=str(dbm_path),
+        # exactly what conventional_input_paths() hands over: a path that was
+        # composed from the dataset root, never probed for
+        outer_shell=str(root / "outer_shell"),
+    )
+
+
+def tracks_only_config() -> dict:
+    return Config({
+        "z_begin": Z_BEGIN,
+        "z_end": Z_END,
+        "loss_weight_shell_outer": 0.0,
+        "input_use_verified_patches": False,
+        "input_use_unverified_patches": False,
+        "input_use_fibers": False,
+        "input_use_normals": False,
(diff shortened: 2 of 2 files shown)


Nieuwlaar:

Bullo27:
Verified by running, on `main` at `c53c74a29`. Change is correct, the test discriminates, and the widened `None` path is safe at every downstream use.

## The test fails on the bug it pins

- on this PR: **2 passed**
- the same test file against `main`'s `fit_spiral.py`: **1 failed, 1 passed**, with exactly the crash you describe:

```
E  FileNotFoundError: [Errno 2] No such file or directory:
   '.../ds/outer_shell/meta.json'
tifxyz.py:362: FileNotFoundError
```

So `test_absent_outer_shell_does_not_block_track_loading` genuinely reproduces the failure and the guard genuinely removes it.

## The guard only narrows, and `None` is already handled

Three things I checked rather than assumed, since adding a condition that skips a block is exactly where a latent `None` bites:

1. **`os` is available** — `import os` is line 1 of `fit_spiral.py`, and the file already uses `os.path` 21 times, so `os.path.isdir` needs no new import.
2. **`shell_patch = None` is guarded at every downstream site.** It is initialised `None` at `:1527`, and every consumer already tests for it — `:1912` (`if shell_patch is not None and filter_tracks_by_shell`), `:2710` (`shell_active = self.shell_patch is not None and ...`), `:2714` and `:3996`/`:4003` (`self.shell_patch is not None and ...`). Skipping the load leaves the object in a state the rest of the class already expects.
3. **When the shell does exist, nothing changes.** `os.path.isdir` is `True`, so `filter_tracks_by_shell` keeps its old value and the block runs exactly as before. The condition can only turn `True` into `False`, never the reverse.

And the `outer_shell_required()` half of the disjunction is untouched, so a shell the losses genuinely need is still loaded and still fails loudly — which is what your comment claims and what the second test pins.

## Note on the framing

The comment explaining *why* the path is non-empty for a dataset with no `outer_shell/` is the useful part of this diff — "the conventional layout resolves an outer_shell path without probing for it" is the non-obvious fact, and without it the `isdir` check reads like belt-and-braces rather than the actual fix. Worth keeping verbatim.

## Limits

- **Not run on a real spiral fit.** I exercised the guard through your test, not through an actual headless fit over a real dataset, so I have confirmed the crash is gone rather than that a full run now completes correctly.
- I did not check whether a dataset with tracks and no `outer_shell/` produces *good* results downstream — only that it no longer aborts on the missing directory.
- Environment note, not a review point: the test needs the full `fit_spiral` import chain (`wandb`, `kornia`, `rustworkx`, `pyro-ppl`, `torchdiffeq`, `trimesh`) before it will collect. That is the established convention here — nine existing tests in `spiral-fitting/tests/` already import `fit_spiral` — so this PR is consistent with its neighbours; it just meant a few installs on my side before I could run it.

---
*Claude (AI agent) on behalf of @Bullo27. Measured against `main` at `c53c74a29` and PR head, Python 3.13, on 2026-08-26.*

PR #1608, opened 2026-08-26
scripts: expand the aligned-21 contract into a runnable (optionally h…

khj1222:
**In one sentence:** Turn the published aligned-21 recipe into a config you can actually run,
and optionally hold a scroll or a segment out of it so the same recipe measures how well the
model transfers to data it never saw.

**One real example:** Starting with the released `ink_9um` labels and surface volumes, I ran

```bash
python scripts/make_holdout_config.py \
  --labels-root  /data/ink_9um/labels \
  --volumes-root /data/ink_9um/surface-volumes \
  --exclude-scroll 0139 \
  --seed 42 \
  --out configs/loso_no0139_s42.json \
  --run-dir runs/loso_no0139_s42
```

and it produced a 15-representation training config with quotas
`{'1667': 40, 'Paris4': 20, '0814': 4}` — byte-identical to the config I had previously
written by hand and trained on.

**Before:** `configs/aligned21_hybrid_3d2d.json` ships a single `datasets` entry whose paths
are `/path/to/aligned-21slice-data`, and the 29 representations it is meant to train on live
in a different file, `configs/aligned21_fixed_scroll_prior.json`. To run the recipe you join
the two by hand: 29 entries, each carrying its own `segments_path`, `surface_volume_paths`,
`sampling_physical_segment_keys` and `sampling_representation_keys`. To hold anything out you
also have to fix `fixed_scroll_prior.target_batch_counts` yourself. Drop a scroll without
renormalising and training stops at startup with
`fixed scroll quotas sum to 51, expected batch_size=64`, or, if the sums happen to work out,
with `fixed scroll quota keys must exactly match patch scrolls`. The sampler is right to
refuse; what is left to the user is finding a valid integer split that sums to `batch_size`,
once per arm.

**After this PR:** one command. Exclusions are `--exclude-scroll` / `--exclude-segment`, both
repeatable, and the quotas are renormalised over the survivors by largest remainder so they
still sum to `batch_size`.

**Proof:** the script regenerates the configs behind measurements I have already published,
from the contract rather than from my copies. Four arms, same seed, compared key by key
against the configs the checkpoints were actually trained with:

| arm | flags | representations | renormalised quotas | vs. the config I trained |
|---|---|---|---|---|
| leave-0139-out | `--exclude-scroll 0139` | 15 | `{1667: 40, Paris4: 20, 0814: 4}` | **identical** |
| leave-1667-out | `--exclude-scroll 1667` | 23 | `{0139: 44, Paris4: 17, 0814: 3}` | **identical** |
| leave-Paris4-out | `--exclude-scroll Paris4` | 21 | `{0139: 35, 1667: 27, 0814: 2}` | **identical** |
| two segments, both families | `--exclude-segment pherc0139-w035 --exclude-segment w035 --exclude-segment pherc0139-w039 --exclude-segment w039` | 25 | `{0139: 29, 1667: 22, Paris4: 11, 0814: 2}` | differs only in `dataloader_workers`, which I had lowered from 12 to 6 by hand |

With no exclusions it emits all 29 representations and reproduces the released quotas
`{0139: 29, 1667: 22, Paris4: 11, 0814: 2}` unchanged, which is the renormaliser's no-op case.

Failure modes are loud rather than silent: an unknown scroll prints the valid ones
(`error: unknown scroll(s) ['0247']; known: ['0139', '0814', '1667', 'Paris4']`), and a
representation whose labels or surface volume cannot be located is listed by name instead of
being written into a config that dies later in patch finding.

**Why / where this is useful:** anyone who wants to run the published recipe at all, and
anyone measuring cross-scroll generalisation — open problem #7. Holding a scroll out makes its
entire supervision mask honest held-out ground truth, so it can be scored directly. The three
leave-one-scroll-out arms above are what I used to get the first numbers on that problem, and
the point of upstreaming the generator is that the next person can re-run or disagree with
them without reconstructing the config by hand.

      stated data.

## Details

- Tested at `c61cc9f`, branched from `3ea17f5` (`merge-ink-pipelines` tip).
- Representations are located by exact directory name under the two roots, so a layout with
  `aligned/<segment>.zarr` and `native/<segment>/<one>.zarr` both work without flags. The walk
  does not descend into `.zarr` stores, so it stays fast on a full corpus (0.14 s over 29
  representations here).
- `--recipe` and `--contract` default to the two files in `configs/` next to the script and can
  be pointed elsewhere.
- Not covered: the script does not verify that a located surface volume is at the scale the
  recipe trains at. #1580 is adding exactly that check on the inference side, and it belongs
  there rather than here.
- No changes to any existing file.

**Why this matters to me:** I wanted to see whether the publicly released ink_9um model could
generalize to scrolls it had never seen before. The recipe was available, but it did not run
as-is, so before I could even get to that question, I had to manually define 29 dataset
entries. I then repeated the joins and quota calculations for each arm across three
leave-one-scroll-out splits and one two-segment holdout. At that point, it made more sense to
keep the setup in the repository than in my notes.

### DIFF
--- a/ink-detection/scripts/make_holdout_config.py
+++ b/ink-detection/scripts/make_holdout_config.py
@@ -12,9 +12,10 @@ This script does that join by locating each representation under the roots you
 give it. Holding scrolls or segments out turns the same recipe into a
 generalisation probe: nothing on the held-out scroll is ever sampled, so its
 whole supervision mask stays honest held-out ground truth. Per-scroll batch
-quotas are renormalised over the survivors, because ``FixedScrollPriorSampler``
-rejects a batch whose quotas do not sum to ``batch_size``, and rejects quota keys
-that do not exactly match the scrolls the patches came from.
+quotas are renormalised over the survivors, because
+``FixedScrollPriorStratifiedBatchSampler`` rejects a batch whose quotas do not sum
+to ``batch_size``, rejects a non-positive quota, and rejects quota keys that do not
+exactly match the scrolls the patches came from.
 """
 
 from __future__ import annotations
@@ -78,6 +79,12 @@ def renormalise(quotas: dict, keep: set, batch_size: int) -> dict:
     live = {scroll: value for scroll, value in quotas.items() if scroll in keep}
     if not live:
         sys.exit("error: every scroll was excluded")
+    if batch_size < len(live):
+        # Every survivor is floored to one slot below, because the sampler rejects a
+        # zero quota outright, so a smaller batch than the scroll count has no answer.
+        sys.exit(f"error: batch_size {batch_size} cannot cover {len(live)} scrolls; "
+                 "FixedScrollPriorStratifiedBatchSampler needs at least one slot per "
+                 "scroll")
     total = sum(live.values())
     exact = {scroll: batch_size * value / total for scroll, value in live.items()}
     out = {scroll: max(1, int(value)) for scroll, value in exact.items()}
@@ -174,6 +181,10 @@ def main() -> int:
     batch_size = args.batch_size or int(recipe["batch_size"])
     seed = args.seed if args.seed is not None else int(recipe["seed"])
     surviving = {entry["sampling_scroll"] for entry in groups.values()}
+    if not surviving:
+        # Only reachable under --allow-missing: without it the locate failure above exits.
+        sys.exit(f"error: none of the {len(kept)} representations could be located under "
+                 f"{args.labels_root} and {args.volumes_root}")
     quotas = renormalise(contract["target_batch_counts"], surviving, batch_size)
     held_out = sorted({rep["segment"] for rep in dropped})
 
--- a/ink-detection/scripts/make_holdout_config.py
+++ b/ink-detection/scripts/make_holdout_config.py
@@ -1,0 +1,211 @@
+#!/usr/bin/env python3
+"""Expand the aligned-21 sampling contract into a runnable training config.
+
+``configs/aligned21_hybrid_3d2d.json`` ships one ``datasets`` entry with
+``/path/to/...`` placeholders, and the 29 representations it is meant to train
+on live separately in ``configs/aligned21_fixed_scroll_prior.json``. Joining the
+two by hand means writing 29 entries, each with its own label directory and
+surface-volume path, and keeping ``fixed_scroll_prior.target_batch_counts``
+consistent with whatever subset you kept.
+
+This script does that join by locating each representation under the roots you
+give it. Holding scrolls or segments out turns the same recipe into a
+generalisation probe: nothing on the held-out scroll is ever sampled, so its
+whole supervision mask stays honest held-out ground truth. Per-scroll batch
+quotas are renormalised over the survivors, because ``FixedScrollPriorSampler``
+rejects a batch whose quotas do not sum to ``batch_size``, and rejects quota keys
+that do not exactly match the scrolls the patches came from.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+from collections import OrderedDict
+from pathlib import Path
+
+CONFIG_DIR = Path(__file__).resolve().parent.parent / "configs"
+DEFAULT_RECIPE = CONFIG_DIR / "aligned21_hybrid_3d2d.json"
+DEFAULT_CONTRACT = CONFIG_DIR / "aligned21_fixed_scroll_prior.json"
+
+
+def as_posix(path: Path) -> str:
+    return str(path.resolve()).replace("\\", "/")
+
+
+def walk_dirs(root: Path):
+    """Yield directories under root without descending into Zarr stores."""
+    stack = [root]
+    while stack:
+        current = stack.pop()
+        try:
+            entries = [entry for entry in current.iterdir() if entry.is_dir()]
+        except OSError:
+            continue
+        for entry in entries:
+            yield entry
+            if entry.suffix != ".zarr":
+                stack.append(entry)
+
+
+def find_label_dir(labels_root: Path, segment: str) -> Path | None:
+    """A segment's label directory is named after it and holds its inklabels."""
+    for candidate in walk_dirs(labels_root):
+        if candidate.name == segment and (candidate / f"{segment}_inklabels.zarr").exists():
+            return candidate
+    return None
+
+
+def find_volume(volumes_root: Path, segment: str) -> Path | None:
+    """Accept either <segment>.zarr or <segment>/<one>.zarr."""
+    holder = None
+    for candidate in walk_dirs(volumes_root):
+        if candidate.suffix == ".zarr" and candidate.stem == segment:
+            return candidate
+        if candidate.name == segment and candidate.suffix != ".zarr":
+            holder = candidate
+    if holder is None:
+        return None
+    stores = sorted(store for store in holder.iterdir() if store.suffix == ".zarr")
+    if len(stores) != 1:
+        sys.exit(f"error: expected exactly one *.zarr under {holder}, found {len(stores)}")
+    return stores[0]
+
+
+def renormalise(quotas: dict, keep: set, batch_size: int) -> dict:
+    """Spread the recipe's quotas over the surviving scrolls, summing to batch_size."""
+    live = {scroll: value for scroll, value in quotas.items() if scroll in keep}
+    if not live:
+        sys.exit("error: every scroll was excluded")
+    total = sum(live.values())
+    exact = {scroll: batch_size * value / total for scroll, value in live.items()}
+    out = {scroll: max(1, int(value)) for scroll, value in exact.items()}
+    # Largest-remainder top-up so the quotas land exactly on batch_size.
+    order = sorted(live, key=lambda s: exact[s] - int(exact[s]), reverse=True)
+    index = 0
+    while sum(out.values()) < batch_size:
+        out[order[index % len(order)]] += 1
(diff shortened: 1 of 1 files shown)


Bullo27:
Ran this against `merge-ink-pipelines` at `3ea17f54a`. Both of your quota claims hold exactly. One crash to fix: `renormalise` raises a raw `ValueError` whenever `--batch-size` is smaller than the number of surviving scrolls.

I have no aligned-21 corpus here, so I synthesized the input tree instead — 29 label directories and 29 `.zarr` stores named after the contract's representations, which is all `find_label_dir`/`find_volume` actually probe for. That exercises the real script end to end, but see Limits.

## Your claims, checked

**"With no exclusions it reproduces the released quotas unchanged"** — confirmed:

```
representations : 29 kept, 0 held out
quotas          : {'0139': 29, '1667': 22, 'Paris4': 11, '0814': 2} (batch 64)
dataset entries : 4
```

Identical to `aligned21_fixed_scroll_prior.json`'s `target_batch_counts`, all 29 representations placed across 4 dataset entries, and **no `/path/to/` placeholder survives anywhere in the output** — which is the actual point of the PR.

**Renormalisation over survivors** — every single-scroll holdout sums back to `batch_size`, with the held-out scroll absent from the quota keys:

| held out | quotas | sum |
|---|---|---|
| `0139` | `{'1667': 40, 'Paris4': 20, '0814': 4}` | 64 |
| `0814` | `{'0139': 30, '1667': 23, 'Paris4': 11}` | 64 |
| `1667` | `{'0139': 44, 'Paris4': 17, '0814': 3}` | 64 |
| `Paris4` | `{'0139': 35, '1667': 27, '0814': 2}` | 64 |

That is exactly the invariant `FixedScrollPriorSampler` enforces, so the renormalisation does its job.

## The crash: `batch_size < len(surviving scrolls)`

```
$ make_holdout_config.py ... --batch-size 3
  File ".../make_holdout_config.py", line 91, in renormalise
    victim = max((s for s in order if out[s] > 1), key=lambda s: out[s])
ValueError: max() iterable argument is empty
```

`out = {scroll: max(1, int(value)) ...}` floors every survivor to at least 1, so `sum(out) >= len(live)`. When `batch_size < len(live)` the top-down loop still has work to do, but every entry is already at 1, so the generator expression is empty and `max()` throws.

The boundary is exact — I swept it:

| surviving scrolls | works | crashes |
|---|---|---|
| 4 | batch ≥ 4 | 3, 2, 1 |
| 3 | batch ≥ 3 | 2, 1 |
| 2 | batch ≥ 2 | 1 |
| 1 | batch ≥ 1 | — |

So: crashes iff `batch_size < len(live)`. With the shipped recipe that means `--batch-size 1|2|3`, which is reachable through your own documented flag — a plausible thing to try for a smoke run on a small GPU.

**I don't think the arithmetic needs fixing** — the request is genuinely unsatisfiable. `FixedScrollPriorSampler` requires the quota keys to match the surviving scrolls exactly *and* sum to `batch_size`, so with fewer slots than scrolls there is no valid assignment. It is the diagnostics that are wrong, and the file already has a house style for precisely this:

```
$ make_holdout_config.py ... --exclude-scroll nope
error: unknown scroll(s) ['nope']; known: ['0139', '0814', '1667', 'Paris4']
```

Same class of user-input error, same file, clean one-line exit. Suggest matching it, right after `live` is built:

```python
if batch_size < len(live):
    sys.exit(f"error: batch_size {batch_size} cannot cover {len(live)} scrolls; "
             f"FixedScrollPriorSampler needs at least one slot per scroll")
```

`--batch-size 4` with 4 scrolls works fine (rc 0), so the guard only catches the genuinely impossible case.

## Limits

- **Synthetic input tree.** My directories are empty shells with the right names, so I exercised the join, grouping, exclusion and quota logic — not whether the located paths point at usable data. Your byte-identical comparison against the four configs you actually trained is yours; I could not repeat it, since those configs are not in the repo.
- My flat layout puts every segment directly under `--labels-root`, so `groups` keyed on `(labels.parent, scroll)` collapsed to one entry per scroll (4 entries). A real corpus with segments spread across different parent directories would produce more entries; I did not test that grouping.
- I did not run any training, so nothing here says the generated config trains correctly — only that it is internally consistent and placeholder-free.
- Not tested: `--exclude-segment`, `--allow-missing`, or the `find_volume` `<segment>/<one>.zarr` branch and its "expected exactly one *.zarr" exit.

One small process note, meant helpfully rather than as a nit: the template section below your write-up is still unfilled and the "I personally verified" box is unchecked, even though the prose above it already answers most of it (the four-arm comparison is exactly the "one real example" and "proof"). Might be worth lifting those three paragraphs into the template so a reviewer sees the verification claim where they expect it.

---
*Claude (AI agent) on behalf of @Bullo27. Measured against base `3ea17f54a` and PR head `c61cc9f14`, Python 3.13, on 2026-08-26.*

khj1222:
Thank you — the crash is real and it is fixed in `dc9edb6`.

I reproduced it exactly where you said. `--batch-size 3` with four surviving scrolls, on
`c61cc9f`:

```
  File "scripts/make_holdout_config.py", line 91, in renormalise
    victim = max((s for s in order if out[s] > 1), key=lambda s: out[s])
ValueError: max() arg is an empty sequence
```

and after the fix:

```
error: batch_size 3 cannot cover 4 scrolls; FixedScrollPriorStratifiedBatchSampler needs at least one slot per scroll
```

I used your wording with the class spelled out in full, because the reason the floor of 1
cannot simply be dropped is `samplers.py:84` — `fixed scroll quotas must all be positive`.
A zero quota is rejected there, so when `batch_size < len(live)` there is no valid config to
write and refusing is the whole answer. My own docstring had the class name wrong; that is
corrected too. The guard sits immediately after `live` is built, as you suggested.

Boundaries, on the real `ink_9um` tree: `--batch-size 4` with four scrolls gives
`{0139: 1, 1667: 1, Paris4: 1, 0814: 1}`, and `--batch-size 3 --exclude-scroll 0814` gives
`{0139: 1, 1667: 1, Paris4: 1}`. The three leave-one-scroll-out configs still regenerate
byte-identical to the ones I trained, so the guard does not touch the normal path.

On the branches you listed as untested — I have the released corpus on disk, so I ran them:

* `--exclude-segment`, four exclusions (`pherc0139-w035 w035 pherc0139-w039 w039`): 25
  representations kept, quotas `{0139: 29, 1667: 22, Paris4: 11, 0814: 2}`, and the output
  differs from the config I actually trained only in `dataloader_workers`, which I had
  lowered from 12 to 6 by hand for an unrelated Windows reason.
* zarr discovery: both branches run on every invocation here, because the released layout
  uses both forms — `surface-volumes/aligned9/<segment>.zarr` is the direct case and
  `surface-volumes/native9/<segment>/<one>.zarr` is the holder case.
* `--allow-missing`: this turned up a second wart, so thank you for pointing at it. With a
  labels root where nothing at all could be located, the script exited
  `error: every scroll was excluded` — which is untrue and sends you looking at your
  exclusion flags. It now names the two roots instead. The partial case behaves: with
  `--volumes-root` pointed at `aligned9` only, it keeps the 24 aligned representations and
  prints `WARNING: 5 input(s) not located`, and without the flag it refuses and lists them
  by name.

On the template — what you saw was GitHub's pre-fill. I pasted the prepared body over it a
few minutes after opening the PR, which was after your read; the filled template and the
checked verification box are in the body as it stands now. Sorry for the wasted look.

khj1222:
A question about targeting rather than about this diff.

This PR is based on `merge-ink-pipelines`, because `ink-detection/scripts/` only exists
there. That branch's tip is still `3ea17f5` from 2026-08-14, while `main` has been
committing daily since. I have four open PRs pointed at it, so before any of them go
further: **is `merge-ink-pipelines` still the branch ink-detection changes should target,
or has that work moved?** If it should go to `main` instead, I am happy to rewrite these
against whatever the current layout there is — the `main` copy of this pipeline has
diverged enough that it would be a rewrite rather than a rebase, and I would rather do that
once, deliberately, than guess.

Concretely, why it matters rather than being bookkeeping: #1663 patches a `.replace()`
inline in one script on this branch, but on `main` that same publish step is already
factored into `ink_detection/preprocessing/staged_write.py` with four callers, so the same
fix there would be one function covering all of them and a much smaller diff to read. #1662
is the same shape — the compile fallback it repairs lives in
`inference/inference_runtime.py` on `main`, shared, where I patched it inline here. I would
rather find that out before asking anyone to review the versions I have.

Asking now because #1535 was closed automatically yesterday under the repository's 14-day
inactivity policy, and a previously closed PR of mine (#1434) turned out not to be
reopenable, so I would rather ask while these are open than re-open the same work a third
time.

Status of this one, unchanged since 08-28: @Bullo27's review found a real crash —
`renormalise` returning all-1 quotas when `batch_size` is smaller than the number of
surviving scrolls, which made the trim loop call `max()` on an empty sequence — and that is
fixed in `dc9edb6` by refusing the configuration outright, since `samplers.py:84` rejects a
zero quota anyway so there is no valid config in that corner. Everything they flagged as
unverified was then run against the real corpus: `--exclude-segment` reproduces our three
LOSO configs byte-identically, both zarr-discovery branches fire on the real trees
(`aligned9` as `<seg>.zarr`, `native9` as `<seg>/<one>.zarr`), and `--allow-missing` turned
up a second wart, which is fixed too. CI is green; the one red check is Vercel asking a team
member to authorise a deploy for an outside contributor.

Also happy to shrink it if 222 lines is the obstacle — the `--exclude-*` flags and the
contract-expansion are separable.

PR #1612, opened 2026-08-26
sweep configs from instance / upload image and all metrics

bruniss:


Bullo27:
Ran this against `main` at `e1d7bf0a1`. The suite is sound — 86 passed here, and the same test file against unpatched `run_single.py` fails 10, so the new tests genuinely pin the new behaviour. One thing worth fixing before merge: the optional preview can take the required metrics down with it.

## A failed ink preview loses the numeric final metrics

In `log_seed_final_metrics`, the preview is built *inside* the try, before the log call:

```python
payload = {f"final/{key}": value for key, value in summary.items() if _numeric(value)}
if ink_dir is not None:
    preview = _build_rendered_ink_preview(ink_dir)     # raises on an empty dir
    payload["final/rendered_ink_full_lasagna"] = _wandb_image(preview, ...)
run.log(payload)                                        # never reached
```

`_build_rendered_ink_preview` raises `RuntimeError("no rendered ink strips found in ...")` when the directory holds no `.jpg`, and `FileNotFoundError` from `iterdir()` when it does not exist. Either lands in the outer `except Exception`, so `run.log(payload)` never runs and **the `final/*` scores and the new `final/satisfaction/*` values are silently dropped too** — metrics that logged fine before this PR.

Measured, using your own fake-run harness shape from `test_run_single.py`:

| `ink_dir` | `run.log` called? |
|---|---|
| valid directory with strips | ✅ metrics + image |
| `None` (behaviour before this PR) | ✅ metrics |
| **directory exists, no `.jpg`** | ❌ **nothing logged** |
| **directory missing** | ❌ **nothing logged** |

It also does not self-heal. `log_seed_final_metrics` returns `False`, so `_log_seed_final_metrics_once` never sets `wandb_final_logged`, and a resumed run retries the same path and fails the same way — that seed can never record final metrics.

The reason it slipped is visible in the test: `test_seed_final_wandb_logs_and_closes_rendered_ink` monkeypatches `_build_rendered_ink_preview` to a stub that always succeeds, so the failure path is never exercised.

Suggested shape — keep the image best-effort, keep the metrics mandatory:

```python
if ink_dir is not None:
    try:
        preview = _build_rendered_ink_preview(ink_dir)
        payload["final/rendered_ink_full_lasagna"] = _wandb_image(
            preview, caption="Full Lasagna-flattened rendered ink panorama")
    except Exception as exc:
        print(f"warning: rendered ink preview unavailable: {exc}", flush=True)
run.log(payload)
```

`run()` passes `fitted_dir / "ink"` unconditionally, so an absent ink stage is reachable, not hypothetical.

## Smaller things

- **Copilot's `boundary_total_patches` point is correct** — both `'total_patches'` and `'boundary_total_patches'` are `int(total_count)`, the same value under two keys. Confirming rather than re-litigating; worth resolving since these land in W&B where the two names imply different denominators.
- **Width truncates, height rounds.** `(max(1, int(width * scale)), max(1, round(height * scale)))` — mixing `int` and `round` means the panorama loses up to one pixel of width per strip while heights round to nearest. Harmless for a preview, and `sum(int(w*scale)) <= MAX_WIDTH` still holds, so the cap is safe; just noting the asymmetry looks unintended.
- **The satisfaction hand-off is pinned only by the fixture.** `_load_final_summary` reads `fitted_dir.parent.parent / "satisfaction_metrics_fitted.json"` while `satisfaction_metrics.py` writes `f'{out_path}/satisfaction_metrics_{suffix}.json'`. Those line up in your test, but the reader guards with `if satisfaction_path.is_file()` — so if `out_path` or `suffix` ever drift, the satisfaction metrics vanish from W&B with no warning at all. Might be worth a hard failure, or at least a printed note, when the file is absent but expected.

## Limits

- **No real run.** I exercised `log_seed_final_metrics` and `_build_rendered_ink_preview` directly with fakes; I did not run a fit, a render, or a real W&B upload.
- I did not check whether an actual 32768-pixel-wide panorama survives W&B's own image handling — only that the scaling maths stays under the cap.
- I did not review the sweep config values (`track_length_bin_weights`, `input_use_tracks`) on their merits; those are yours.
- `satisfaction_metrics.py`'s new summary block references several names bound earlier in a long function — I read them as in scope, but I did not execute that path, so treat that file as read rather than run.

---
*Claude (AI agent) on behalf of @Bullo27. Measured against `main` at `e1d7bf0a1` and PR head `4df507415`, Python 3.13, on 2026-08-26.*

Bullo27:
Merged as `a63dc9f96`, byte-identical to the head I reviewed (`git diff pr1612 a63dc9f96` → 0 files), so noting for the record that the preview issue is live on `main`. Re-ran the same checks against the merged commit:

| `ink_dir` | `run.log` called? |
|---|---|
| valid directory | ✅ |
| `None` | ✅ |
| empty directory | ❌ |
| missing directory | ❌ |

**One correction to my own wording above:** I said the metrics are dropped "silently", and that is not right. The pipeline does print a warning — the merged code produces:

```
WARNING: could not upload final W&B metrics for s: FileNotFoundError: [Errno 2] No such file or directory: '.../ink'
```

So an operator watching stderr does see something. What is actually lost without signal is the *distinction*: the message reads as "W&B upload failed" whether the panorama alone failed or the scores went with it, and there is no way to tell from the log which happened. Combined with `wandb_final_logged` never being set, a resumed seed retries and re-fails the same way.

Not asking you to reopen anything — the fix is the small try/except in my earlier comment, and it is yours to take or leave. If it would save you time I am happy to put it up as a PR against `main` with a regression test that pins the empty and missing directory cases; say the word and I will, otherwise I will leave it here.

---
*Claude (AI agent) on behalf of @Bullo27. Re-measured against merged `main` at `a63dc9f96`, Python 3.13, on 2026-08-26.*

PR #1613, opened 2026-08-26
spiral: updating hyperparameters following recent sweeps

pmh47:
SDT disabled, and more winding-model pairs


TAUIL-Abd-Elilah:
Quick question on the Aug 22 sweep: are the results for `defaults_no_band_seed_patches`, `patch_sampling_area_exponent_025`, and `_075` shareable (W&B link or exported metrics)?

I don't want to infer performance from non-promotion alone. I'm running a complementary PHercParis4 A/B prompted by the reported small-patch swamping issue: retain every `band-seed` patch, but cap that family's aggregate sampling mass at 25%, with the newly published `eval_fibers/` used only as held-out geometric-consistency evaluation. I'll share the exact config, family-stratified metrics, and logs once it finishes. Existing sweep outcomes would help avoid duplicate work and make the comparison fair.

PR #1614, opened 2026-08-26
VC photogrammetry end-to-end with documentation

giorgioangel:
# PR title

VC photogrammetry end-to-end with documentation

# PR body

**In one sentence:** This adds the PGS-Recon implementation and end-to-end instructions needed to reconstruct, physically scale, inspect and prepare real scroll meshes from within Villa.

**One real example:** The Docker context and pause, scale-review and resume workflow were exercised end to end on real scroll photographs before this source was prepared for Villa.

**Before:** The actively used PGS-Recon variant lived outside Villa, and reviewing the scale histogram before OpenMVS required a separate archive-and-patch handover.

**After this PR:** Villa contains the ready-to-build PGS-Recon source, pause/review/resume controls, the printable ArUco sheet and one chronological handover covering acquisition through case generation.

**Proof:** This PR contains the source, documentation and printable marker sheet only. Generated reconstruction outputs and handover result artifacts are intentionally excluded.

**Why / where this is useful:** Photogrammetry operators can reproduce the real-scroll workflow from a pinned Villa commit, stop before physical scaling, compare scale diagnostics and resume OpenMVS without repeating the expensive SfM stage.

## Details

- Adds the ready-to-build source under `foundation/pgs-recon` with AGPL provenance.
- Bases the snapshot on upstream revision `5ccaa10f4de821aaa5f98e60593fd8be69ecda3a` and records the global-scaler lineage from merge requests !39 and !41.
- Adds `--stop-after {sfm,autoscale}`, `--resume-sfm`, scaler histogram/debug forwarding and complete scaler command metadata.
- Adds `foundation/photogrammetry-docs/README_HANDOVER.md` and Seth Parker's printable `markers_0.tif`.
- Preserves the legacy one-command behavior when the new pause/resume options are absent.

Local PR checks:

- Python source compilation completed successfully.
- All included shell scripts passed syntax checking.
- Every relative link in the handover resolves locally.
- `markers_0.tif` was identified as a readable 2520×3564, 8-bit, LZW-compressed TIFF.
- The branch passes `git diff --check` and has a clean worktree.

The Docker context and pause/scale/resume route were built and exercised from a clean Ubuntu EC2 worker. No generated reconstruction or handover result artifacts are included or published by this PR.


Bullo27:
Ran the new `foundation/pgs-recon` snapshot and the handover document against each other on Linux. Most of what I have is **confirmation of the Codex review already sitting on this PR** — I want to be explicit up front that the two robustness findings below are its, not mine. What I add is a reproduction for each, and three smaller things it didn't flag.

One process note first, because it may be why these are still open: the merge landed at `19:25:04Z` and the Codex review at `19:32:00Z`. **The automated review arrived about seven minutes after the merge**, so none of it was visible pre-merge, and it is now five inline comments on a closed PR — easy to lose, and the code is in `main`.

## Confirming Codex's two P1s (its findings, my measurements)

### `global_scaler.cpp:739` — zero-width histogram

Reproduced. I sliced `WriteScaleHistogram` verbatim out of the committed file (lines 719–827, by brace-matching rather than retyping, so the harness can't drift from the repo), compiled it standalone under ASan/UBSan and ran it:

| samples | `minv` / `maxv` | `binw` | `(v-minv)/binw` | `int(…)` | result |
|---|---|---|---|---|---|
| 8 distinct (**control**) | 0.401 / 0.430 | 5.8e-4 | 0.0 | 0 | valid 5,616-byte SVG, exit 0 |
| 1 | 0.4137 / 0.4137 | 0 | `-nan` | **-2147483648** | **SEGV** |
| 3 identical | 0.25 / 0.25 | 0 | `-nan` | **-2147483648** | **SEGV** |

The faulting line is `:740` (`hist[b]++`), one below the comment anchor, with `b = INT_MIN`.

Then the trigger path end-to-end — same verbatim-slice treatment for `ComputeUmeyamaScaleStats` (`:584–677`), driven with a synthetic landmark set plus sub-millimetre jitter, calling the two in `main()`'s order:

| scene | samples | summary (true 0.2500) | result |
|---|---|---|---|
| 6 markers × 4 corners | 6 | 0.249874 | SVG written, exit 0 |
| **1 marker × 4 corners** | 1 | 0.249521 | **SEGV** |
| **1 marker × 3 corners** (the `:613` minimum) | 1 | 0.249517 | **SEGV** |

Two things this adds to the report:

1. **The scale estimate is correct in both crashing runs.** The computation succeeds; only the diagnostic write kills the process.
2. **The ordering makes it expensive.** The summary prints at `:1137`, the histogram is written at `:1146`, and `--output-scene` is not saved until `:1152–1156`. So the operator sees a plausible scale on stdout, then a segfault, and **no scaled scene** — the detection and triangulation work is lost, not just the histogram.

Worth flagging that my first control was invalid: I had built every marker as a geometrically perfect square, so all six estimates came out bit-identical and the control crashed too. The jitter is what makes it a control rather than a second copy of the bug.

### `RANSAC.hpp:64` — tied models keep the worse RMSE

Confirmed by differential. Two arms, identical but for that one comparison, fixed seed `0`, with `eval` giving every model identical fitness so the tie-break is the only thing that can decide, over RMSE candidates 0–9:

- committed (`>`): keeps **8.0**
- flipped (`<`): keeps **0.0**

One addition on where the fix belongs: **this is inherited, not introduced here.** I pulled `RANSAC.hpp` from the pinned upstream base `5ccaa10f` and the comparison is already `>` there, with the same include list — so it predates MR !39, and a fix probably wants to go upstream as well as into this snapshot.

A caveat on the harness, in case anyone re-runs it: `bestModel` is overwritten at `:81` by `fit(bestResult.inliers)`, so a probe whose `eval` returns a fixed inlier set cannot observe the loop's choice at all — both arms return the same value. My first attempt did exactly that and produced a false negative until the inlier set was made to carry the model.

I did not independently investigate the `:639` scale-1.0 fallback beyond reading it; the code does do what that comment describes.

## Three things not flagged

**1. `global_scaler.cpp:876` — `--no-ransac`'s help text describes the opposite of what the flag does.**

```cpp
("no-ransac", po::bool_switch(), "use RANSAC to make marker triangulation more resilient to false positive matches")
```

But `:1077` is `bool useRansac = not args["no-ransac"].as<bool>();` and `:1108` branches the same way, so passing `--no-ransac` *disables* RANSAC. An operator reading `--help` while trying to harden a marginal scale estimate is told that this flag adds the robustness they want. This matters a little more than a typo here because the handover doc (`README_HANDOVER.md`, "Before reconstruction") instructs the reader to retain and check the `--help` output, and separately states that RANSAC-enabled triangulation is the default — which is correct.

In mitigation, and I think this bounds the damage: `:1078–1079` print `Triangulating landmarks w/RANSAC` when it is actually on, so the console output does tell the truth at runtime.

**2. `reconstruct.py:209` cross-references a scaler flag that does not exist.**

`--autoscale-marker-pix`'s help says *"see the `pgs-global-scaler --min-marker-size` flag"*. There is no `--min-marker-size`: the scaler defines `--min-marker-pix` (`global_scaler.cpp:874`, default `32`). Zero occurrences of the string `min-marker-size` anywhere in the scaler. The other two such cross-references in that file (`--detection-method`, `--exclude-from`) both resolve, so it's one of three.

**3. `RANSAC.hpp` is not self-contained.**

It uses `std::sample` and `std::min` but includes only `<cmath>`, `<iterator>`, `<optional>`, `<random>`, `<type_traits>` and `educelab/core/utils/Math.hpp`. It compiles today only because the single translation unit that includes it, `global_scaler.cpp`, happens to include `<algorithm>` at `:1`, well before `RANSAC.hpp` at `:44`. Compiling the header first fails with `'sample' is not a member of 'std'` — that error is how I found it. Also inherited: the same include list is present at upstream `5ccaa10f`.

## What I checked and found clean

Since the handover document's whole purpose is to be followed after the author has gone, I checked it as an artifact in its own right rather than only reading the code:

- I harvested the **44 distinct `--flags`** the document tells the reader to type and checked the PGS-Recon ones against the real parsers (`add_argument` literals via AST, plus the `po::options_description` block). **Every PGS-Recon flag resolves.** In particular `--no-mvs-refine` **is** valid — `--mvs-refine` is a `BooleanOptionalAction` with `default=True` — which is not obvious from grepping. (The remainder are `docker`, `uv` and `sam2-photogrammetry` flags, which I did not check.)
- `--stop-after` really does accept exactly `{sfm, autoscale}`; the `--resume-sfm` incompatibility the doc describes is genuinely enforced (`reconstruct.py:289–299`); "the CLI always invokes texturing, with no top-level disable flag" is true (`:492`, unguarded).
- The scaler defaults the doc quotes are right: minimum marker size **32** px (`:874`), RANSAC on unless opted out, Umeyama samples requiring **≥3** triangulated corners (`:613`), and the summary weighted by corner count (`:628`).
- All **12** `console_scripts` entry points resolve to a real `main` in a real module; all 25 Python files parse.
- Every relative link in the document resolves, including `../sam2-photogrammetry/` and `../scrollcase/`.
- `pgs-global-scaler` is the `OUTPUT_NAME` of the `global_scaler` target and installs to `/usr/local/bin`, and `entrypoint.sh` activates the venv, so both documented `docker run … --help` smoke tests should resolve their binaries.

## Limits

- **I did not build the Docker image or run the real `pgs-global-scaler`.** The superbuild is OpenMVG + OpenMVS + CGAL + OpenCV from source; I did not have a worker for it. Every C++ result above comes from verbatim slices of the committed source compiled standalone, with stand-in typedefs (`Vec3` → `Eigen::Vector3d`) replacing the openMVG headers. That is enough for the arithmetic and the control flow shown, and it is *not* a substitute for an end-to-end container run.
- No real photographs and no SfM scene were involved — the landmark sets are synthetic.
- I have **not** checked whether the histogram issue also affects upstream after MR !41. On upstream `main` the call survives as `pgs::WriteScaleHistogram`, but the definition has moved out of `global_scaler.cpp` and I did not chase it down. Worth someone's five minutes given !41.
- I did not look at the two Codex P2s (the jpeg-turbo digest length, the hardcoded x86-64 AWS CLI).

Happy to open a PR against `main` with the two-line guards (`binw == 0` early-return, tie-break comparison) plus the `--no-ransac` help string and the `--min-marker-pix` cross-reference, with a small regression test pinning the one-marker and identical-scales cases — or to file a tracking issue instead, since these are now in `main` behind a closed PR. Just say which you'd prefer.

---
*Claude (AI agent) on behalf of @Bullo27. Measured on Ubuntu 24.04.4 / g++ 13.3.0 / ASan+UBSan against `6847063ff`, on 2026-08-26. `global_scaler.cpp`, `RANSAC.hpp`, `reconstruct.py` and `README_HANDOVER.md` are byte-identical (blob SHA) between the reviewed head `a29040c10` and the merge, so the line numbers hold for both. The two P1 findings above are Codex's; I contributed the reproductions, the upstream provenance check and the three unflagged items.*

PR #1615, opened 2026-08-26
config: read training YAML as UTF-8, not the locale codec

### LINKED ISSUE #1524: ConfigManager reads training YAML with the locale codec: UTF-8 BOM loads a silently wrong config on Windows, smart quotes crash it
## Summary

`ConfigManager.load_config` opens the training YAML with no encoding:

```python
# models/configuration/config_manager.py:28
with open(config_path, "r") as f:
    config = yaml.safe_load(f)
```

so the file is decoded with the **locale** codec, not UTF-8. On a Windows box
(cp1252) that gives three distinct failure modes, one of which is silent.

## Measured, on Python 3.14 / cp1252 locale

| config file | result |
|---|---|
| plain ASCII | OK |
| UTF-8, em-dash in a comment | OK (mojibake in the comment, harmless) |
| UTF-8, accented name in a comment | OK (same) |
| **UTF-8 + BOM, first line is a key** | **loads silently WRONG — `model_name` comes back `'Model'` instead of the configured value** |
| UTF-8 + BOM, first line is a comment | `yaml.ScannerError: mapping values are not allowed here` |
| UTF-8 with a byte undefined in cp1252 (e.g. a `”` smart quote) | `UnicodeDecodeError: character maps to <undefined>` |

The BOM case is the one that worries me. `\xef\xbb\xbf` decodes under cp1252 to
`ï»¿`, which gets glued onto the first key — `tr_setup` becomes `ï»¿tr_setup`,
that whole section is silently dropped, and training proceeds on defaults with
no warning. A UTF-8-with-BOM YAML is what several Windows editors and
`Set-Content -Encoding utf8` produce by default, so this is easy to hit
accidentally.

## How I hit it

Following the trainer's own suggestion to run `vesuvius.find_patches --config
<config.yaml>` for a Dataset059 run. I generated a variant config with
PowerShell, which wrote UTF-8-with-BOM, and got the `UnicodeDecodeError`
above. Chasing that is what turned up the silent-wrong-config case.

## Suggested fix

`encoding="utf-8-sig"` on the config read handles both plain UTF-8 and a BOM:

```python
with open(config_path, "r", encoding="utf-8-sig") as f:
```

There are ~25 other `open(..., "r")` calls on YAML/JSON/text in the package
with no encoding (`data/volume.py:66,558,1263`, `data/affine.py:22`,
`image_proc/.../chunks_filter.py:63`, etc.). Those are the same latent issue
for any non-ASCII path or file, though the config one is the only place I've
actually been bitten. Happy to send a PR for just `config_manager.py`, or a
sweep across all of them — say which you'd prefer.

Reproduction script (writes the six variants above and loads each) available
on request.

---

bkalita-git:
**In one sentence:** Training/config YAML is now read as UTF-8 (BOM-tolerant) instead of the machine's locale codec, so a config saved by a Windows editor can no longer silently lose its `tr_setup` section.

**One real example:** Starting with the repo's own real training config `models/configuration/single_task/ps128_guided_dinovol_ink.yaml`, saved exactly as PowerShell's `Set-Content -Encoding utf8` writes it (UTF-8 with BOM), I loaded it through `ConfigManager.load_config` on a machine whose locale codec is cp1252, and it produced a config whose `tr_setup` section was silently dropped — `model_name` came back as the default instead of `ps128_guided_dinovol_ink`. With this PR the same file loads intact.

**Before:** On a locale-codec machine (Windows cp1252 — the environment measured in #1524), the BOM decodes to three characters glued onto the first key: `tr_setup` becomes `ï»¿tr_setup`, the whole section is dropped, and training proceeds on defaults with no warning. A config containing a smart quote in a comment crashes with `UnicodeDecodeError`.

**After this PR:** The config read uses `encoding="utf-8-sig"`, which handles plain UTF-8 and UTF-8-with-BOM identically on every platform; both files load correctly. The config *write* side pins `encoding="utf-8"` (with `allow_unicode=True`) so a saved config round-trips on any locale.

**Proof:**

```
Python 3.14.7
================ BEFORE (main) ================
tree            : before-main
locale codec    : CP1252
tr_setup keys   : []
model_name      : None
RESULT          : *** tr_setup SILENTLY DROPPED - training would run under defaults ***
================ AFTER (fix branch) ===========
tree            : fix-branch
locale codec    : CP1252
tr_setup keys   : ['model_name']
model_name      : 'ps128_guided_dinovol_ink'
RESULT          : OK - config loaded intact
================ TESTS on fix branch ==========
21 passed, 2 warnings in 0.79s
```

(The `�` glyphs in the raw log are the em-dashes of the script's own output failing to encode under CP1252 — the locale genuinely cannot represent them, which is the point.)

Reproduced in a `python:3.14-slim` container with a generated `en_US.CP1252` locale (`LC_ALL=en_US.CP1252`) — the same class of environment as the Windows cp1252 box measured in #1524 — mounting `main` and this branch side by side and running the identical script against each. PyYAML strips a *single leading U+FEFF* from str input, which is why UTF-8-locale machines (Linux/macOS CI) never see this: the failure exists exactly and only where the locale codec is not UTF-8, matching the reporter's measurements.

**Why / where this is useful:** Anyone editing or generating training configs on Windows (several editors and PowerShell's default utf8 write BOM) currently risks training with silently wrong settings. This is the failure mode reported and measured in #1524.

Fixes #1524

## Details

- Tested commit: `main` at the merge-base of this branch; fix is two lines in `vesuvius/src/vesuvius/models/configuration/config_manager.py` (read: `utf-8-sig`, write: `utf-8` + `allow_unicode=True`).
- Three regression tests added to `vesuvius/tests/models/configuration/test_config_manager.py` (`TestConfigEncoding`): BOM before a key, BOM before a comment, non-cp1252-decodable smart quote in a comment. All 21 tests in the file pass (`PYTHONPATH=src pytest tests/models/configuration/test_config_manager.py`). Note: the BOM tests pass on UTF-8-locale CI regardless (PyYAML strips a lone U+FEFF); they pin the intended behavior and bite on any locale-codec runner.
- Scope kept to `ConfigManager` per the issue's suggested fix. #1524 also notes ~25 other unencoded `open(..., "r")` calls in the package; those are the same latent issue but are left for a separate pass to keep this reviewable.
- The write-side `allow_unicode=True` keeps non-ASCII characters (e.g. in comments or names) readable in the saved YAML instead of escaping them.

## From the submitter

Fixed cross platform config file reader issue by explicitly mentioning encoding. This issue was reported by a windows user in #1524. I went through the diff with Claude as my assistant and agree with the change. I ran the test suite myself, 21 passed.

*Disclosure: Claude (Anthropic) investigated the reported issue, prepared the fix and tests, and drafted this PR text under my direction. The container reproduction and test runs above were executed on my machine, and I reviewed the change and ran the tests myself before submitting.*

### DIFF
--- a/vesuvius/src/vesuvius/models/configuration/config_manager.py
+++ b/vesuvius/src/vesuvius/models/configuration/config_manager.py
@@ -25,7 +25,10 @@ class ConfigManager:
     def load_config(self, config_path):
         config_path = Path(config_path)
         self._config_path = config_path
-        with open(config_path, "r") as f:
+        # utf-8-sig: config files must not depend on the machine's locale
+        # codec, and Windows editors often emit a UTF-8 BOM that would
+        # otherwise be glued onto the first key, silently dropping it.
+        with open(config_path, "r", encoding="utf-8-sig") as f:
             config = yaml.safe_load(f)
 
         self.tr_info = config.get("tr_setup", {})
@@ -940,8 +943,8 @@ class ConfigManager:
         config_filename = f"{self.model_name}_config.yaml"
         config_path = model_ckpt_dir / config_filename
 
-        with config_path.open("w") as f:
-            yaml.safe_dump(combined_config, f, sort_keys=False)
+        with config_path.open("w", encoding="utf-8") as f:
+            yaml.safe_dump(combined_config, f, sort_keys=False, allow_unicode=True)
 
         print(f"Configuration saved to: {config_path}")
 
--- a/vesuvius/src/vesuvius/models/configuration/config_manager.py
+++ b/vesuvius/src/vesuvius/models/configuration/config_manager.py
@@ -28,8 +28,14 @@ class ConfigManager:
         # utf-8-sig: config files must not depend on the machine's locale
         # codec, and Windows editors often emit a UTF-8 BOM that would
         # otherwise be glued onto the first key, silently dropping it.
-        with open(config_path, "r", encoding="utf-8-sig") as f:
-            config = yaml.safe_load(f)
+        try:
+            with open(config_path, "r", encoding="utf-8-sig") as f:
+                config = yaml.safe_load(f)
+        except UnicodeDecodeError as e:
+            raise ValueError(
+                f"Config file {config_path} is not valid UTF-8. "
+                "Config files must be UTF-8 encoded; re-save the file as UTF-8."
+            ) from e
 
         self.tr_info = config.get("tr_setup", {})
         self.tr_configs = config.get("tr_config", {})
--- a/vesuvius/tests/models/configuration/test_config_manager.py
+++ b/vesuvius/tests/models/configuration/test_config_manager.py
@@ -1,7 +1,10 @@
 """Tests for ConfigManager target name validation."""
 
-from pathlib import Path
+import builtins
 import tempfile
+from pathlib import Path
+from unittest import mock
+
 import pytest
 import yaml
 
@@ -265,15 +268,38 @@ class TestTargetNameValidation:
 
 
 class TestConfigEncoding:
-    """Config files must load as UTF-8 regardless of the machine's locale codec (#1524)."""
+    """Config files must load as UTF-8 regardless of the machine's locale codec (#1524).
 
-    def _load_bytes(self, raw: bytes):
-        with tempfile.NamedTemporaryFile(suffix='.yaml', delete=False) as f:
+    The loads run with encoding-less text opens forced to cp1252 (the Windows
+    locale codec measured in #1524), so these tests fail on unpatched code on
+    every platform — on a UTF-8 runner PyYAML absorbs a lone leading U+FEFF,
+    which would otherwise mask the bug.
+    """
+
+    @staticmethod
+    def _as_locale_codec(codec):
+        """Make encoding-less text opens behave like a box whose locale codec
+        is `codec`, so the test measures the same thing on every platform."""
+        real_open = builtins.open
+        def fake_open(file, mode="r", buffering=-1, encoding=None, *a, **kw):
+            if encoding is None and "b" not in mode:
+                encoding = codec
+            return real_open(file, mode, buffering, encoding, *a, **kw)
+        return mock.patch.object(builtins, "open", fake_open)
+
+    @staticmethod
+    def _write_bytes(raw: bytes) -> str:
+        # written outside the locale patch: tempfile paths only, binary mode
+        with tempfile.NamedTemporaryFile(suffix=".yaml", delete=False) as f:
             f.write(raw)
-            temp_path = f.name
+            return f.name
+
+    def _load_on_cp1252_box(self, raw: bytes):
+        temp_path = self._write_bytes(raw)
         try:
             mgr = ConfigManager(verbose=False)
-            mgr.load_config(temp_path)
+            with self._as_locale_codec("cp1252"):
+                mgr.load_config(temp_path)
             return mgr
         finally:
             Path(temp_path).unlink()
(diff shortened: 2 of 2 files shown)


Bullo27:
Reproduced this independently on Linux and the diagnosis holds exactly as written — including the subtle part, that PyYAML absorbs a single leading `U+FEFF` so UTF-8 boxes never see the bug.

I simulated a cp1252 box by making encoding-less text opens default to cp1252 (`mock.patch` on `builtins.open`), so both arms run on the same machine under the same interpreter:

| config bytes | main | this PR |
|---|---|---|
| **UTF-8 + BOM** (the #1524 case) | `tr_setup={}` — **silently empty** | `tr_setup={'model_name': 'bom'}` |
| plain UTF-8, no BOM | OK | OK |

And the mechanism, confirmed directly: `yaml.safe_load("ï»¿tr_setup:\n …")` returns `{'ï»¿tr_setup': …}`, so `config.get("tr_setup", {})` yields `{}` and every downstream default silently applies. `utf-8-sig` also reads a BOM-less file correctly, so the common path is untouched.

So the fix is right. One substantive issue with the tests, and two smaller notes.

## ★ The three new tests pass against unpatched `main` on Linux

This is the thing I'd want changed before merge. I ran **your** new test class against **main's** `config_manager.py`, changing nothing else:

| locale codec | main | this PR |
|---|---|---|
| UTF-8 (what Linux CI uses) | **3 passed** | 3 passed |
| forced non-UTF-8 (`LC_ALL=C PYTHONUTF8=0`, codec `ANSI_X3.4-1968`) | 3 failed | 3 passed |

The whole file behaves the same way: **`21 passed` on main**, and `21 passed` on the branch. That `21 passed` in your proof block is real, but it is not evidence of the fix — it is the same 21 that pass without it.

That follows from your own explanation rather than contradicting it: PyYAML eats the single `U+FEFF`, so on a UTF-8 runner `open()` and `open(encoding="utf-8-sig")` are indistinguishable here. The consequence is that as written these tests will not fail if someone later drops the `encoding=` argument — on the platform CI runs, the guard is inert.

Making it locale-independent is enough to fix that; forcing the codec in-process reproduces the Windows condition anywhere. I ran this **under the normal UTF-8 locale, no env twiddling** — it fails on main and passes on your branch:

```python
def _as_locale_codec(codec):
    """Make encoding-less text opens behave like a box whose locale codec is
    `codec`, so the test measures the same thing on every platform."""
    real_open = builtins.open
    def fake_open(file, mode="r", buffering=-1, encoding=None, *a, **kw):
        if encoding is None and "b" not in mode:
            encoding = codec
        return real_open(file, mode, buffering, encoding, *a, **kw)
    return mock.patch.object(builtins, "open", fake_open)


def test_bom_not_dropped_on_a_cp1252_box(self):
    p = _write(b"\xef\xbb\xbftr_setup:\n  model_name: bom_model\n")
    try:
        mgr = ConfigManager(verbose=False)
        with _as_locale_codec("cp1252"):
            mgr.load_config(p)
        assert mgr.tr_info.get("model_name") == "bom_model"
    finally:
        Path(p).unlink()
```

Measured: `main` → `2 failed` (`UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d`), branch → `2 passed`. Write the temp file *outside* the patch — `tempfile` uses `io.open`, not `builtins.open`, but keeping the patch narrow avoids the question entirely.

Worth knowing either way: there were **0 check runs** on head `1167ebaf5` when I looked, so nothing was exercising these tests yet regardless.

## Two smaller notes

**1. This makes genuinely-cp1252 config files fail where they load today.** Measured on the simulated cp1252 box:

| | main | this PR |
|---|---|---|
| file actually encoded in cp1252 (`model_name: café`) | loads, `'café'` | `UnicodeDecodeError` |

I think that is the correct call — YAML mandates UTF-8, and failing loudly beats silent mojibake — but it is a real behaviour change for a Windows user who already has such a file, and it isn't in the PR description. A sentence in the body, or catching `UnicodeDecodeError` around the load to say "config must be UTF-8 encoded", would spare someone a confusing traceback.

**2. Completeness, as information rather than a request.** There are **8** encoding-less text opens on YAML in `vesuvius/src/vesuvius`; this PR fixes 2. The other 6:

- `vesuvius.py:23`, `vesuvius.py:34`, `data/volume.py:66`, `:558`, `:1263` — reads of `scrolls.yaml` / `cubes.yaml`
- `data/paths/local.py:192` (`save_yaml`) — the write side

**I checked before flagging these and the risk is genuinely lower, so I would not hold the PR for them.** Those five read package-installed or tool-generated configs rather than a file a user hand-edits. And I was wrong about the write side: I expected `save_yaml` to break on a cp1252 box with a non-ASCII folder name, and it does not — `yaml.dump` defaults to `allow_unicode=False`, so it ASCII-escapes everything and encodes cleanly in any codec. Measured with CJK and Polish folder names: all wrote fine. (Which is exactly why your `save_config` change needs `allow_unicode=True` *and* `encoding="utf-8"` together — that pair is right.)

## Limits

- I did not run on real Windows. The cp1252 behaviour is simulated by forcing the codec on encoding-less opens, plus a real `ANSI_X3.4-1968` locale arm; both are stand-ins for cp1252, not cp1252 itself.
- I did not exercise an actual training run — only `ConfigManager.load_config` / `save_config`.
- Both arms are the full `vesuvius/src` tree with only `config_manager.py` swapped between `main` and `1167ebaf5`; I verified by `diff -r` that nothing else differed.

---
*Claude (AI agent) on behalf of @Bullo27. Measured on Ubuntu 24.04.4, CPython 3.13.11, PyYAML 6.0.3, against `main` `6847063ff` and PR head `1167ebaf5`, on 2026-08-26.*

bkalita-git:
Thank you for the rigorous review — all three points addressed in `adade702`:

**★ Tests now bite on UTF-8 runners.** Adopted your `_as_locale_codec` pattern: the loads run with encoding-less text opens forced to cp1252, temp files written outside the patch. Measured the same A/B you did: `TestConfigEncoding` **4 failed** against `main`'s `config_manager.py`, **4 passed** with the fix, on a UTF-8 macOS runner (CPython 3.14.5, PyYAML 6.0.3). The full file is 22 passed on the branch.

**Behaviour change is now explicit, in code and in the description.** A config genuinely saved in a legacy codec now raises `ValueError: Config file <path> is not valid UTF-8. Config files must be UTF-8 encoded; re-save the file as UTF-8.` (chained from the `UnicodeDecodeError`) instead of a bare traceback — and there's a new test pinning that (`test_non_utf8_config_fails_loudly`). PR description updated to state the change.

**The 6 other encoding-less YAML opens**: agreed on leaving them out of this PR — noted in the description as a separate pass, with your read-side/write-side risk assessment linked.

*Claude (AI agent) on behalf of @bkalita-git; the A/B measurement above was run on our machine as described.*

PR #1619, opened 2026-08-27
ci(vc3d): restore smoke-tested runtime image publishing

### LINKED ISSUE #1588: container images are 3+ months stale: :edge AND :main both ship the 2026-05-13 build (1e3f4c02) — original crash here is already fixed on main
**In one sentence:** With a remote normal-grid store (`normal-grids-remote.json`), `vc_grow_seg_from_seed` still tries to open `<normal_grid_path>/metadata.json` from the local directory and core-dumps when it is not there — the documented streaming setup is missing one required local file, and the failure mode is an uncaught `std::runtime_error`.

**I was trying to:** Grow surface patches on PHerc0813 (which has no published segments) against its released m7 surface prediction, streaming the released normal grids instead of downloading them — the setup the remote-marker mechanism exists for: a `normal_grid_path` directory containing only `normal-grids-remote.json` with `{"url": "https://vesuvius-challenge-open-data.s3.amazonaws.com/PHerc0813/representations/predictions/surfaces/20250821151723-surface-20260413222639-surface-m7-L0-th0.2.normal-grids"}`.

**Using:** `ghcr.io/scrollprize/villa/volume-cartographer:edge` (2026-08-21), `vc_grow_seg_from_seed --volume <m7 surface zarr over https> --params seed.json --seed X Y Z` with `"normal_grid_path": "/work/ngrid"` in the params, on a RunPod Linux pod.

**What happened:** The tracer starts, opens the volume, resolves the seed — and then aborts:

```
seed location [3968, 3970, 14878] value is 103
terminate called after throwing an instance of 'std::runtime_error'
  what():  Cannot open: /work/ngrid/metadata.json
timeout: the monitored command dumped core
SEED_0_EXIT=134
```

`metadata.json` is fetched fine from the remote store's own prefix — the loader just never looks for it there. Everything else about the remote path works: after a one-line workaround (below), the same invocation streams the grids and grows to completion (8/8 seeds, exit 0).

**What I expected or needed:** Either the remote path fetches `metadata.json` from the same URL prefix it fetches grid chunks from, or the requirement for a local `metadata.json` beside `normal-grids-remote.json` is documented — and ideally the error is caught and reported as "normal_grid_path is missing metadata.json (fetch it from <url>/metadata.json when using a remote store)" instead of a core dump after the volume is already open.

**Evidence / reproduction:**

```bash
mkdir -p /work/ngrid
S=https://vesuvius-challenge-open-data.s3.amazonaws.com/PHerc0813/representations/predictions/surfaces
B=20250821151723-surface-20260413222639-surface-m7-L0-th0.2
printf '{"url":"%s/%s.normal-grids"}' "$S" "$B" > /work/ngrid/normal-grids-remote.json
vc_grow_seg_from_seed --volume "$S/$B.zarr" --target-dir /work/paths \
  --params seed.json --seed 3968 3970 14878     # aborts as above

# workaround — one file, then the identical invocation completes:
curl -sf "$S/$B.normal-grids/metadata.json" -o /work/ngrid/metadata.json
```

- [x] I personally encountered this while using the software.

## Details

The remote marker is read by `NormalGridVolume` (`normal-grids-remote.json`, `NormalGridVolume.hpp:18`), but the constructor's metadata read goes through the local path unconditionally, so the abort happens before any grid chunk is requested — after the (potentially slow) volume open, which makes it expensive to discover on a pod. The `xy/ xz/ yz/` chunk fetches themselves work over https once `metadata.json` is present locally, so the gap really is just this one file. Found while measuring why my grown patches were oriented across the scroll's wraps; the growth runs behind it are documented in a public repo (github.com/flummoxjr/gp13-ink-detectability, `hunt/grow_segments.md`).

*Disclosure: I used Claude (Anthropic) as a coding assistant to investigate this and draft the text. The pod runs and the captured failure are mine.*

TAUIL-Abd-Elilah:
**In one sentence:** After merge and a successful qualifying `main` run, users pulling the documented `:edge` or `:main` image will receive a current, smoke-tested VC3D/C++ runtime instead of the broken May 13 build.

**One real example:** The local amd64 evidence records one current-source generation with the public PHerc0813 m7 surface Zarr, remote-only normal-grid marker, and seed `(3968, 3970, 14878)`: it read the binary voxel as `255`, fetched the remote normal grids, generated `302.567159 vx²`, saved the TIFXYZ path, and exited 0.

**Before:** Both `:edge` and `:main` resolve to the same 2026-05-13 image (`1e3f4c02`), 285 commits behind the tested source. On this exact input that image read the voxel as `103` and crashed on remote metadata; the reporter's full eight-seed run also produced flat, zero-area surfaces.

**After this PR:** On each qualifying push to `main`, the workflow is configured to build native amd64 and arm64 runtime digests from the existing Dockerfile. Each digest must be pulled back and smoke-tested before the workflow can advance `main`, `edge`, and an immutable full-SHA tag.

**Proof:**

- Reporter failure and independent stale-image diagnosis: https://github.com/ScrollPrize/villa/issues/1588
- Reporter eight-seed before/after geometry: https://github.com/ScrollPrize/villa/issues/1588#issuecomment-5406053138
- Research-thread report that the current-source path unblocked surface growth: https://github.com/ScrollPrize/villa/issues/191#issuecomment-5406053252
- Local build/run and final implementation record: https://github.com/ScrollPrize/villa/issues/1588#issuecomment-5432544176
- Hash-bound local inputs, raw log, and two output metadata records: https://gist.github.com/TAUIL-Abd-Elilah/53c10e4d9aaf0967ed8ac19bc6426283 (revision `940863d17bedc712dfa4890efd3623c48ed5f580`)
- Independent smoke-gate replay and coverage review: https://github.com/ScrollPrize/villa/pull/1619#issuecomment-5432714620

```text
current-source build: 1244/1244 compile steps; 12/12 image steps; exit 0
amd64 image digest:    sha256:95cfa07a89d825fbba2f6a3064ca91c95b85d7d65ebd4f133b2328efe9860504
seed voxel:            255
remote normal grid:    loaded (marker was the only initial file)
generated surface:     302.567159 vx² / 0.000265 cm²
saved artifact:        auto_grown_20260826213017169
runtime exit:          0
```

**Why / where this is useful:** The README recommends this image because VC3D is expensive to build. The stale binary already cost a contributor three pod experiments and hid fixes needed for real PHerc0813 surface growth; restoring guarded publication prevents that failure from recurring for every later fix.

## Details

Fixes #1588.

Tested source: `6847063ffdb4da898ae8d1d494ebf7d71473f509`; workflow commit: `e3a318270`.

- Reuses `volume-cartographer/Dockerfile` and its existing `runtime` target; no application or image-content changes.
- Configures native `ubuntu-24.04` and `ubuntu-24.04-arm` runners, avoiding emulated multi-architecture builds.
- Pushes architecture images by digest, then re-pulls and tests that exact registry object.
- Checks the core tools plus the documented `vc_gen_normalgrids` and `flatboi`; scans every executable in `/usr/local/bin` for missing shared libraries; runs four representative CLI help paths; and requires VC3D to remain healthy for an eight-second offscreen startup.
- Publishes moving tags only after both architecture jobs pass. A failure may leave an untagged digest, but cannot update `main` or `edge`.
- Leaves the separate CI-builder workflow unchanged.

Scope is intentionally the existing C++ runtime. A broader prototype was rejected after verification showed that merely copying Spiral sources would omit its locked Python/CUDA/Lasagna environment and expose a partial service. Image slimming is also separate; the existing runtime target retains its build toolchain.

Local checks: `actionlint` 1.7.12 clean; Docker BuildKit `--check` clean; `git diff --check` clean. Arm64 execution is delegated to the workflow's native runner.

The eight-seed `68.1° → 20.5°` / `0 → 8.08–9.73 cm²` comparison is @flummoxjr's evidence; the local evidence reproduces only the one-generation constructor/runtime path above.

AI disclosure: Codex assisted with implementation and verification in my workspace. The local evidence records executed amd64 build and real-data commands; no arm64 workflow run is claimed. Limitations are stated above.

### DIFF
--- a/.github/workflows/vc3d-publish-runtime.yml
+++ b/.github/workflows/vc3d-publish-runtime.yml
@@ -1,0 +1,181 @@
+name: Publish VC3D runtime image (ghcr.io)
+
+# Keep the documented :main and :edge C++ runtime images aligned with the
+# source users get from the default branch. The builder-image workflow remains
+# separate so source-only changes do not republish the CI dependency image.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'volume-cartographer/**'
+      - '.github/workflows/vc3d-publish-runtime.yml'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: write
+
+env:
+  REGISTRY_IMAGE: ghcr.io/${{ github.repository_owner }}/villa/volume-cartographer
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  prepare:
+    name: Prepare image name
+    runs-on: ubuntu-24.04
+    outputs:
+      registry_image: ${{ steps.image.outputs.registry_image }}
+    steps:
+      - name: Require the default branch
+        if: github.ref != 'refs/heads/main'
+        shell: bash
+        run: |
+          echo "Runtime tags may only be published from main; got $GITHUB_REF" >&2
+          exit 1
+      - name: Normalize registry image name
+        id: image
+        shell: bash
+        run: echo "registry_image=${REGISTRY_IMAGE,,}" >> "$GITHUB_OUTPUT"
+
+  build:
+    name: Build runtime (${{ matrix.platform }})
+    needs: prepare
+    runs-on: ${{ matrix.runner }}
+    timeout-minutes: 90
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: linux/amd64
+            runner: ubuntu-24.04
+            artifact: linux-amd64
+          - platform: linux/arm64
+            runner: ubuntu-24.04-arm
+            artifact: linux-arm64
+    steps:
+      - uses: actions/checkout@v4
+      - uses: docker/setup-buildx-action@v3
+      - uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Generate image labels
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ needs.prepare.outputs.registry_image }}
+      - name: Build and push runtime by digest
+        id: build
+        uses: docker/build-push-action@v6
+        with:
+          context: ./volume-cartographer
+          file: ./volume-cartographer/Dockerfile
+          target: runtime
+          platforms: ${{ matrix.platform }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha,scope=vc3d-runtime-${{ matrix.artifact }}
+          cache-to: type=gha,mode=max,scope=vc3d-runtime-${{ matrix.artifact }}
+          provenance: true
+          outputs: type=image,name=${{ needs.prepare.outputs.registry_image }},push-by-digest=true,name-canonical=true,push=true
+      - name: Smoke-test published digest
+        env:
+          REGISTRY_IMAGE: ${{ needs.prepare.outputs.registry_image }}
+          DIGEST: ${{ steps.build.outputs.digest }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          docker pull "$REGISTRY_IMAGE@$DIGEST"
+          docker run --rm --entrypoint /bin/bash "$REGISTRY_IMAGE@$DIGEST" -lc '
+            set -euo pipefail
+            echo "architecture=$(uname -m)"
+            for binary in VC3D vc3d vc_render_tifxyz vc_grow_seg_from_seed \
+              vc_gen_normalgrids flatboi; do
+              test -x "/usr/local/bin/$binary"
+              echo "executable=$binary"
+            done
+
+            ldd_checked=0
+            while IFS= read -r -d "" binary; do
+              missing="$(ldd "$binary" 2>&1 | grep "not found" || true)"
+              if [[ -n "$missing" ]]; then
+                echo "missing shared libraries for $binary:" >&2
+                printf "%s\n" "$missing" >&2
+                exit 1
+              fi
+              ldd_checked=$((ldd_checked + 1))
+            done < <(find -L /usr/local/bin -maxdepth 1 -type f -perm /111 -print0)
+            echo "ldd_checked=$ldd_checked"
+
+            for binary in vc_render_tifxyz vc_grow_seg_from_seed \
+              vc_gen_normalgrids flatboi; do
+              "$binary" --help >/dev/null
+            done
+
+            set +e
+            QT_QPA_PLATFORM=offscreen HOME=/tmp timeout 8 VC3D \
+              >/tmp/vc3d-smoke.log 2>&1
+            rc=$?
+            set -e
+            tail -n 20 /tmp/vc3d-smoke.log
+            if [[ "$rc" -ne 124 ]]; then
+              echo "VC3D exited unexpectedly during the offscreen startup smoke (rc=$rc)" >&2
+              exit 1
+            fi
+            echo "runtime_smoke=ok"
+          '
+      - name: Export digest
+        shell: bash
+        run: |
+          mkdir -p "${{ runner.temp }}/digests"
+          digest="${{ steps.build.outputs.digest }}"
+          touch "${{ runner.temp }}/digests/${digest#sha256:}"
+      - name: Upload digest
+        uses: actions/upload-artifact@v4
+        with:
+          name: digests-${{ matrix.artifact }}
+          path: ${{ runner.temp }}/digests/*
+          if-no-files-found: error
+          retention-days: 1
+
+  publish:
+    name: Publish multi-platform runtime tags
+    needs: [prepare, build]
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    steps:
+      - uses: docker/setup-buildx-action@v3
+      - uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Download digests
+        uses: actions/download-artifact@v4
+        with:
+          path: ${{ runner.temp }}/digests
+          pattern: digests-*
+          merge-multiple: true
+      - name: Generate runtime tags
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ needs.prepare.outputs.registry_image }}
+          tags: |
+            type=raw,value=main
+            type=raw,value=edge
+            type=raw,value=sha-${{ github.sha }}
+      - name: Create and inspect manifest list
+        working-directory: ${{ runner.temp }}/digests
+        env:
+          REGISTRY_IMAGE: ${{ needs.prepare.outputs.registry_image }}
+        shell: bash
+        run: |
+          docker buildx imagetools create \
+            $(jq -cr '.tags | map("-t " + .) | join(" ")' <<< "$DOCKER_METADATA_OUTPUT_JSON") \
+            $(printf "$REGISTRY_IMAGE@sha256:%s " *)
+          docker buildx imagetools inspect "$REGISTRY_IMAGE:edge"


Bullo27:
Reproduced the premise from the registry and ran this workflow's own smoke assertions against real binaries. The gate works. One coverage gap is worth closing before merge, and one hypothesis of mine was wrong — both below.

## The premise, checked straight from ghcr

Pulled the manifests anonymously today. Both tags resolve to one image, exactly as stated:

| | `:edge` | `:main` |
|---|---|---|
| `created` | `2026-05-13T18:41:20Z` | `2026-05-13T18:41:20Z` |
| `image.revision` | `1e3f4c021f4e53be…` | `1e3f4c021f4e53be…` |
| `image.version` | `main` | `main` |

The package has **14 tags**, and the `builder-ubuntu-*` namespace the existing `vc3d-publish-ghcr.yml` writes is disjoint from `main` / `edge` / `sha-*` — no `sha-*` tag exists yet. So "leaves the separate CI-builder workflow unchanged" holds at the tag level. Worth knowing: a `volume-cartographer/Dockerfile` change now matches **both** workflows' path filters (yours is a superset), so that one commit triggers two full builds. Distinct tags, so it's cost, not corruption.

The live `:edge` index already lists `linux/amd64`, `linux/arm64` and two `unknown/unknown` entries — the shape BuildKit produces with attestations, which is what `provenance: true` + `imagetools create` here would also produce.

## The smoke test discriminates — measured, not assumed

A gate that can't fail is worse than no gate, so I ran each assertion against binaries built from `6847063ff` rather than reading them:

```
vc_render_tifxyz --help                                  -> rc=0    (5284B stdout)
vc_grow_seg_from_seed --help                             -> rc=0    (1040B stdout)
ldd {VC3D,vc_render_tifxyz,vc_grow_seg_from_seed}        -> 0 "not found"
QT_QPA_PLATFORM=offscreen HOME=<empty> timeout 8 VC3D    -> rc=124  <-- the assertion
```

`rc=124` is the one that mattered: under `set -euo pipefail`, a VC3D that exits cleanly on a headless start would fail the job and block every publish. It doesn't — it initialises the axis-aligned slice views and is still alive when `timeout` kills it. Fresh empty `HOME`, so no local settings carried in.

## Where I was wrong

I expected a silent capability regression, because the live May-13 image is a **different Dockerfile lineage**, not just an older commit of this one. Its recorded history builds libigl + SCOTCH + PaStiX + flatboi from source, installs a micromamba `py310` env, and sets `WORKDIR /src/build`. None of that is in the current 41-line Dockerfile.

Checked it, and the regression isn't there:

- `flatboi` — **present**, one of 47 binaries `--component vc_runtime` installs
- `vc_gen_normalgrids` — **present**
- `aws` and `jq`, both needed by `docker_s3_entrypoint.sh` — **both installed** by `install_build_deps.sh:24,29-32`
- `Cmd: ["/bin/bash"]`, no `ENTRYPOINT` — same as live, so the docs' `docker run … :edge` with no command argument still lands in a shell

Two real deltas remain, neither introduced by this PR: the `py310` env with `numpy`/`tqdm`/`wandb` disappears while `ENV WANDB_ENTITY` stays behind in the Dockerfile, and `WorkingDir` moves `/src/build` → `/src`.

## The gap I'd close: the smoke list covers 3 of 47

`--component vc_runtime` installs **47** binaries. The presence and `ldd` checks cover **3** of them (plus the `vc3d` wrapper the Dockerfile adds). Two of the misses are tools the docs hand to users directly:

- `vc_gen_normalgrids` — `scrollprize.org/docs/35_segmentation.md:99`, required input for the workflow this image exists to run
- `flatboi` — `scrollprize.org/docs/38_tutorial_spiral.md:167`, and `volume-cartographer/README.md:71` makes the image the *only* delivery path for it on Windows

So the failure this PR prevents for `vc_grow_seg_from_seed` can still ship silently for those two. I measured what widening actually costs before suggesting it:

| widening | result |
|---|---|
| `ldd` loop 3 → all 47 | **0 missing libs.** Free and safe. |
| `--help` loop 2 → all 47 | **36 exit 0, 11 do not.** Not safe as-is. |

The 11 that would break the job: `vc_create_segment_mask`, `vc_lasagna_line_probe`, `vc_obj_uv_lift`, `vc_rgb2tifxyz`, `vc_thinning`, `vc_tifxyz2rgb`, `vc_tifxyz_mmap_prepare`, `vc_tifxyz_trim`, `vc_tifxyz_winding`, `vc_volpkg_convert`, `vc_zarr_recompress`. Both documented tools above exit `0`, so the minimal version is two lines:

```diff
-for binary in VC3D vc3d vc_render_tifxyz vc_grow_seg_from_seed; do
+for binary in VC3D vc3d vc_render_tifxyz vc_grow_seg_from_seed vc_gen_normalgrids flatboi; do

-vc_render_tifxyz --help >/dev/null
-vc_grow_seg_from_seed --help >/dev/null
+for binary in vc_render_tifxyz vc_grow_seg_from_seed vc_gen_normalgrids flatboi; do
+  "$binary" --help >/dev/null
+done
```

Running the `ldd` loop over `/usr/local/bin/*` instead of a fixed list is the cheaper structural fix, since it stays correct as tools are added.

## By-catch, not this PR's business

`vc_tifxyz2rgb --help` aborts:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Cannot open: --help/meta.json
Aborted (core dumped)          # rc=134
```

It takes `--help` as a positional path. Same unguarded-`Json::parse_file` shape as the crash that started #1588. Happy to file it separately if useful.

## Limits

- Binaries built with a local preset, not `ci-release-gcc` inside `ubuntu:26.04`. The 47-binary list comes from CMake install rules, which don't vary by preset, but **I did not build the image** — @TAUIL-Abd-Elilah did that, and the one-generation run is theirs.
- **x86-64 only.** I have no arm64 machine, so `ubuntu-24.04-arm` is untested here, and whether a from-scratch build fits `timeout-minutes: 90` on a 4-core arm runner is an open question I did not measure.
- I did not exercise `push-by-digest` + `imagetools create` against a registry; the multi-arch shape is inferred from the live index, not from running this workflow.
- Fork PRs get no CI in this repo, so none of this can be validated by Actions until it is on `main`.

---
*Claude (AI agent) on behalf of @Bullo27. Manifests and configs fetched from ghcr.io on 2026-08-27; every exit code and count above is from a command actually run against binaries built at `6847063ff`.*

TAUIL-Abd-Elilah:
Thanks for measuring the gate rather than inferring it. I closed the coverage gap in `e3a318270`:

- executable presence and `--help` checks now include the documented `vc_gen_normalgrids` and `flatboi` tools;
- the shared-library check now scans every executable file under `/usr/local/bin`, so future installed tools cannot silently fall outside the fixed list; and
- I kept `--help` restricted to the four CLIs known to support it, rather than breaking on the 11 positional tools you identified.

`actionlint` 1.7.12 and `git diff --check` pass on the amended workflow. The arm64 and registry-publication limits in your review remain accurately stated: those paths cannot execute until the workflow reaches `main`.

Bullo27:
Ran your amended block at `e3a318270` rather than reading it. **It passes**, and the `find`-based version is better than what I suggested — a fixed list I proposed would have gone stale the next time a tool is added.

```
ldd_checked=49
help_ok
rc=0
```

Mock of `/usr/local/bin` = the 47 binaries `--component vc_runtime` installs, plus the `vc3d` wrapper the Dockerfile writes and the `chmod +x`'d `entrypoint.sh`.

## The thing your widening newly exposes, checked

`find -L … -type f -perm /111` now reaches files the old fixed list never touched. Two classes, both fine, but worth having on record since a failure here blocks every publish:

**Non-ELF executables.** `ldd` exits **1** on them and prints `not a dynamic executable` to *stderr* — which your `2>&1` now feeds into the grep. It doesn't match `not found`, and `|| true` absorbs the non-zero exit under `pipefail`, so `vc3d` and `entrypoint.sh` pass cleanly. That was the failure mode I most expected from this change and it isn't there.

**`install_build_deps.sh` puts three more symlinks in `/usr/local/bin`** that neither of us listed: `flang` → `/usr/bin/flang-21` (line 27), and `aws` + `aws_completer` from `/tmp/aws/install --update` (lines 31-33), whose default target is `/usr/local/bin`. `-L` resolves all three, so they now get scanned. I fetched the official `awscli-exe-linux-x86_64` bundle and checked the two it installs:

| | `ldd … \| grep "not found"` |
|---|---|
| `aws` | 0 lines |
| `aws_completer` | 0 lines |

So AWS CLI v2 is clean on amd64. **`flang` I could not test** — flang-21 isn't on this machine. It is an apt binary inside the same image that carries the toolchain, so I'd expect it to resolve, but I'm flagging it rather than asserting it, and it's the one file in the widened set that nobody has ldd'd.

## One cosmetic thing

`flatboi --help` writes its usage to **stderr** and exits `0`, so `>/dev/null` doesn't suppress it — the other three write to stdout:

```
vc_render_tifxyz       rc=0  stdout=5284B  stderr=   0B
vc_grow_seg_from_seed  rc=0  stdout=1040B  stderr=   0B
vc_gen_normalgrids     rc=0  stdout=2464B  stderr=   0B
flatboi                rc=0  stdout=   0B  stderr= 315B
```

A green run will therefore print what reads like an error into the job log. `>/dev/null 2>&1` in that loop makes a passing smoke silent. Purely cosmetic — the gate itself is correct.

Your restatement of the arm64 and registry-publication limits matches what I can and can't see from here; I've tested x86-64 only throughout, and neither of us can exercise the push path until this is on `main`.

---
*Claude (AI agent) on behalf of @Bullo27. Amended block executed verbatim against the mock tree on 2026-08-27; the AWS CLI figures come from the official bundle downloaded and inspected in place, not from an installed copy.*

TAUIL-Abd-Elilah:
Closing this to stay within the three-open-PR limit and focus the slot on a directly issue-linked tool fix with real before/after validation. The runtime-publishing branch and its evidence remain available if maintainers want to revisit it.

PR #1620, opened 2026-08-27
spiral: build the native helpers with MSVC

7jycwjmbfn-eng:
**In one sentence:** On Windows, installing spiral-fitting with `uv` can't build the vc_spiral native helpers under MSVC. It trips over OpenMP constructs the default mode rejects, a missing `M_PI`, and a POSIX-only mmap.

**One real example:** Windows 11, VS 2022 (MSVC 19.34), Python 3.14, RTX 4080 Laptop, villa @ c53c74a. I was setting up a fit of the PHercParis4 lasagna inputs on this machine when the install failed inside the native build:

```
track_crossings.cpp(1020,9): error C7660: '#pragma omp atomic update': requires '-openmp:llvm'
spiral_sampling.cpp(1027,77): error C2065: 'M_PI': undeclared identifier
track_store.cpp(16,10): fatal error C1083: Cannot open include file: 'sys/mman.h'
```

A run of knock-on const-initialization errors follows, like C2737 on `step`.

The other methods don't work either. `/openmp:llvm` would satisfy the pragmas, but it links a second OpenMP runtime alongside the libiomp5md.dll that torch's Windows wheels ship. Two copies of the runtime in one process, and init aborts with the familiar OMP Error #15.

On this branch the native build completes. All four modules come out, `vc_spiral.track_store` imports right next to torch with exit code 0, and the published PHercParis4 packed track store reads back 708,712 tracks / 24,329,815 points in ROI and 19,746,134 / 1,014,441,480 total. Its own metadata carries the same numbers.

**Before:** On a clean checkout, any uv install of spiral-fitting on Windows stops inside the native build with those compiler errors. The kernels never get built, so none of the spiral tools run on Windows/MSVC today.

**After this PR:**

* The kernels compile serial under MSVC. Its default OpenMP 2.0 mode doesn't know the atomic capture/update and min/max constructs these kernels use, and serial is the fallback the README already documents for toolchains without OpenMP, so CMake only adds a define and skips OpenMP here.
* CMake adds `_USE_MATH_DEFINES`; MSVC's `<cmath>` refuses to emit `M_PI` without it.
* `Mapping` in track_store.cpp gets a Windows branch (`CreateFileMappingW` / `MapViewOfFile`) with the same read-only, sequential-access semantics as the POSIX mmap path, and the same RAII ownership around it.

No changes anywhere else; every hunk sits behind `if(MSVC)` / `#ifdef _WIN32`.

**Proof:** Two terminal captures attached below.

- Before, a clean checkout: `python -m uv build --no-progress --wheel` stopping on the three error classes above. The C2737 lines are knock-ons of the missing `M_PI`; surface_index touches none of it and builds either way.
- After, this branch: the backend build finishing, `Built spiral` then `Prepared 1 package`, and both smoke steps passing with exit code 0. Imports sit right beside torch, and the packed-store load lands exactly on the metadata counts.

clean checkout, native build fails

this branch, builds, imports beside torch, loads the real store

**Why / where this is useful:** I do my spiral fits on a Windows gaming laptop. Going through WSL for that still feels too awkward, and what I actually want is to run things Windows native. That never worked before. Installing the repo hasn't succeeded here even once; `uv sync` stops in the native build before a single kernel runs. Recent Windows fixes each handled something else, like dropping posix-ipc, fixing an int64 `randint` crash, and gating CUDA torch for win32, yet installs went on failing because the C++ helpers themselves still didn't compile under MSVC. This branch takes care of that piece.

## Details

- Branch base: villa @ c53c74a. Only two files touched, `spiral-fitting/cpp/CMakeLists.txt` and `spiral-fitting/cpp/track_store.cpp`. Neither has changed upstream since Aug 21 (acbb61649), so this applies clean to current main.
- Environment: Windows 11, VS 2022 build tools (MSVC 19.34), Python 3.14, torch 2.13.0+cpu in the smoke environment.
- Build verification: scikit-build-core + nanobind against the VS toolchain, the `pyproject.toml` build backend driven without isolation. My machine's antivirus keeps breaking uv's isolated build envs, which is why isolation is off here; that part is local noise, and a plain `uv build` reproduces the compile errors in the before capture either way.
- Data: the published PHercParis4 volume inputs and its published packed track store, loaded through `vc_spiral.track_store.load()` with counts compared against the store's `metadata.json`.
- Scoped to the native build only. One known follow-up lives elsewhere: reading the published tracks DBM on Windows needs a small tracks.py path-handling fix, which I'll send separately.


Bullo27:
I can't test MSVC, but the claim a reviewer most needs assured here is the *other* one — that this is inert off Windows. I checked that three ways on Linux/GCC 13.3, base `6847063ff` vs head `086c0552c`, and it holds at the strongest level available.

## "No changes anywhere else", verified

| check | result |
|---|---|
| preprocessed `track_store.cpp` (`g++ -std=c++23 -E -P`) | **73,975 lines each, 0 diff** |
| `compile_commands.json` (both arms configured) | **15 TUs each, 0 diff** — the `if(MSVC)`/`elseif` restructure changes not one flag |
| all four modules rebuilt with `-g0` | **byte-identical** |

```
track_store        base=164688B sha=a84ec5d2604694d2  head=164688B sha=a84ec5d2604694d2
spiral_sampling    base=242544B sha=d9092c4cd82642e2  head=242544B sha=d9092c4cd82642e2
track_crossings    base=320624B sha=759e7d7ae77708b7  head=320624B sha=759e7d7ae77708b7
surface_index      base=193328B sha=446206c7a3f7d1cf  head=446206c7a3f7d1cf
```

Both arms configure, build and import; `vc_spiral.track_store` exports `['compact', 'inspect', 'load']` identically.

**One artifact difference worth pre-empting,** because anyone diffing built objects will hit it and wonder: in `RelWithDebInfo`, `track_store.so` is **168 bytes larger** on your branch. It is entirely debug metadata — `.debug_info` +154, `.debug_line` +42, `.debug_abbrev` −21, and **no allocatable section changes size**. The `#ifdef _WIN32` blocks push the POSIX code to higher line numbers, so the DWARF line table shifts. `-g0` collapses it to byte-identical, which is what the table above shows. (I also disassembled both: the only two differing lines were the file path in objdump's own header, not instructions.)

## Where I was wrong

I went looking for a link failure in your serial path. `track_store.cpp` and `track_crossings.cpp` call `omp_get_num_procs`, `omp_get_thread_num`, `omp_get_max_threads` and `omp_set_num_threads`, and with no OpenMP runtime linked those are unresolved externals — which would have made "the kernels build serial" impossible rather than merely slow.

It isn't there. Every one of those sites is behind `#ifdef _OPENMP`, including the `#include <omp.h>` itself (`track_store.cpp:24`, `track_crossings.cpp:19`), and `track_crossings.cpp:474` guards the `__gnu_parallel::sort` path with `defined(__GLIBCXX__) && defined(_OPENMP)` and falls back to `std::sort`. `effective_workers()` returns `1` in the `#else`. So the serial build is genuinely link-clean, and `spiral-fitting/README.md:78-79` says what you say it does:

> OpenMP is used when the toolchain provides it; the same modules build with serial kernels when it does not.

## Two Windows-side questions, not defects

I can't test either — flagging so they're decisions rather than accidents.

**`if(MSVC)` also captures clang-cl.** CMake sets `MSVC` true for any frontend simulating Visual C++, so clang-cl on Windows takes the serial branch too, even though it ships libomp and handles `/openmp:llvm` properly. If the reasoning is specifically the second-runtime clash with torch's `libiomp5md.dll`, then clang-cl has the same clash and serial is right — but as written the scope is wider than the comment describes.

**`WIN32_LEAN_AND_MEAN` and `NOMINMAX` are defined unconditionally.** If a toolchain or build system already defines either, that's a redefinition warning, and fatal under `/WX`. `#ifndef` around each is the usual idiom.

## Limits

- **No Windows or MSVC machine here.** The entire `#ifdef _WIN32` mapping path, the three compile errors you quote, and the OMP Error #15 behaviour are untested by me. The one part of this diff that actually does something is the part I cannot exercise.
- I could not locate the published PHercParis4 packed track store in the open-data bucket — `PHercParis4/representations` holds only `predictions` — so I did **not** check your 708,712 / 24,329,815 ROI or 19,746,134 / 1,014,441,480 totals. If you point me at the path I'll reproduce the load on Linux, which would at least confirm the counts are platform-independent.
- Fork PRs get no CI in this repo, so nothing here can be confirmed by Actions before merge.

---
*Claude (AI agent) on behalf of @Bullo27. Both arms configured, built and imported on 2026-08-27 with GCC 13.3.0 / nanobind 3.0.0 / Python 3.13; every hash and byte count above is from a command actually run.*

Pizzakira:
Hello,
Thank you for this. I ran into the same three build failures on windows and your branch covers them.

I have been running the spiral fit on a windows computer and hit the OpenMP side of this from a different angle, so the two notes below may be useful. Both were measured on my machine on 2026-08-27; the analysis was done with Claude Opus 5 under my supervision, the numbers are from my own runs.

First, a small correction to the CMakeLists comment. The OpenMP runtime clash does not abort at import. It aborts at the first parallel region. The module imports fine, and the process dies later, once a pragma actually runs. That matters in practice: a smoke test passes and a long fit still dies in the middle. Measured in both load orders, each in its own process, on MSVC 14.44.35207 and torch 2.11.0+cu128 :

  torch first, then the module  ->  module imports, torch runs, first parallel
                                    region of the module aborts, OMP Error #15, exit 3
  module first, then torch      ->  same abort, roles swapped

The clash is symmetric.

Second, there is a separate reason to avoid /openmp:llvm here, and it is not the runtime clash. Built with /openmp:llvm on MSVC 14.44.35207, track_crossings.cpp produces a binary that faults on a single thread, with no second runtime involved. I narrowed it to "#pragma omp atomic capture" when the captured lvalue goes through a function call. I tried nine access forms; GCC 11.4 (WSL, Ubuntu 22.04) accepts all nine and returns the right values. Hoisting anchor_cursor.data() into a local before the atomic region fixes it, three lines, portable. I have a standalone 78-line reproducer if it is useful.

The practical upshot for this PR: serial is not the only option. Plain /openmp (MSVC's own vcomp140) plus rewriting the kernels in OpenMP 2.0 builds, runs in parallel, and brings no second OpenMP runtime, so it does not clash with the libiomp5md that torch ships. I measure 10.2x on the hot function over the serial build.

Scope, so none of this reads wider than it is: MSVC 14.44.35207 on Windows 11 only, and the faulting path is not taken by the default config (it needs track_crossing_mode "count" with track_crossing_precompute_max 0). Latent defect, not something you hit in normal use.

Happy to open the reproducer as its own issue if you would rather keep this PR focused.

PR #1622, opened 2026-08-27
VC3D: line annotation stays live through re-optimization

skyward7187:
# VC3D: line annotation stays live through re-optimization

Placing a control point in the line-annotation dialog used to freeze the whole GUI, in
three bursts of up to several seconds each, on every placement. This PR removes all of
them: the click lands in single-digit milliseconds, the solve runs in the background,
panning stays live throughout, and the view no longer moves when an update lands.

## What was actually freezing

1. **The click itself** ran up to four Ceres solves plus a whole-line normal resample
   synchronously on the GUI thread, and while an async solve was running the dialog
   refused further edits behind an input-swallowing overlay.
2. **The felt multi-second bursts** (found with a stack sampler, not code reading):
   every async fiber-save completion re-parsed *every* fiber JSON in the package
   (~666 files) on the GUI thread via the atlas-search refresh — 3–5 s per save job,
   and one placement produces up to three save jobs.
3. **The tail**: per-chunk render kicks after a landing (~560 ms of back-to-back
   handlers), per-save fiber-panel rebuilds (span rows for every fiber), per-mouse-move
   shared-cache stats under contended mutexes, and the strip view drifting/snapping on
   every landing.

## The changes

**Async solve architecture**
- Control-point placement now uses a *geometric* prepare
  (`prepareGeometricControlPointEdit` + `lineModelFromPointsSplicing`): the dense line
  is spliced and normals resampled only in the replaced range; solved spans arrive
  asynchronously. Placement cost is microseconds.
- `OptimizationCoalescingQueue` (new, unit-tested): per-session epoch, single-flight
  solve, pending dirty-span union with control-index remapping. Edits during a solve
  bump the epoch (the in-flight result is discarded at `finishOptimization`), set the
  cancel flag, and dispatch one coalesced solve after a 150 ms quiet window. A
  superseded solve folds its dirty spans back into the pending union.
- Cooperative cancellation for line solves (`LineOptimizationConfig::cancelFlag`,
  SOLVER_ABORT from the iteration callback). No-op when the flag is unset, so batch
  callers are numerically unaffected.
- Solves run on a private 2-thread pool; fiber-mode solves show a passive badge instead
  of the blocking overlay (seed solves keep the overlay).
- Save integrity: `validateFiberSaveSnapshots` enforces the v3 loader contract (control
  points are an exact ordered subset of line points at 1e-8) on every save path;
  `applyOptimizationTaskResult` trusts solver-reported control indices only within the
  same tolerance, with a validated nearest-scan fallback.

**The root-cause fix**
- `fiberSnapshotsFromStorageWithPaths` now keeps a per-file (size, mtime) cache; only
  changed files are re-parsed. This alone removes the multi-second freeze bursts.

**Supporting fixes (each addressed a measured stall)**
- Debounces: session autosave 1.5 s (flushed on cross-fiber handlers, package switch,
  destructor), solve dispatch 150 ms, side-strip intersection queries 150 ms behind a
  cheap fingerprint, atlas-dock refresh coalesced at 50 ms.
- Caches (GUI-thread-only, generation/mtime-keyed, swept for vanished ids): per-fiber
  side-strip snapshots (`shared_ptr<const FiberPolyline>` + hash), control-span records
  (fiber summaries build 91 ms → 5 ms).
- Lazy work: fiber-tree span rows built on first expansion (panel emit 217 ms → ~0 for
  collapsed fibers), line-debug JSON serialized only when enabled, unused per-point
  z-slice planes skipped in view rebuilds (`LineViewConfig::buildLineZSlices`),
  `GeneratedViews` moved instead of copied.
- Chunk pipeline: chunk-ready render kicks coalesced to one queued submit per burst;
  blocking chunk reads jump the speculative prefetch backlog; lasagna channel sources
  yield the shared decoded budget to viewer tiles; bounded (10 s) follower wait with a
  duplicate-read fallback; OMP normal batches honor `omp_get_max_threads()` (was
  `hardware_concurrency` per calling thread) with `schedule(static)` per-slot writes,
  so results are width-independent; status-bar cache stats refresh at most every 250 ms.

**View stability on landing**
- The re-built strip surfaces are arc-length parameterized, so their surface-X ↔ fiber
  mapping drifts on every re-optimization. `QuadSurface::shiftSurfaceOrigin` (new) lets
  the dialog anchor the new strips' parameterization to the live strip cameras *before*
  registration (`anchorGeneratedStripSurfacesForUpdate`), so no frame is ever rendered
  un-anchored and the cameras never move — the strip only pans when the user pans.
- The current position is content-remapped across the line change
  (`remappedGeneratedLinePosition`: same 3D spot, new numbering) instead of reused
  numerically, and the old snap-to-placed-control-point recentering is removed.

## Numerics and determinism

Solver math is unchanged: the geometric prepare feeds the same solver the same spans;
cancellation only aborts work whose result is then discarded; chunk-priority and
eviction changes reorder I/O only; the OMP change fixes thread-pool sizing and keeps
per-slot writes with a static schedule, so outputs are independent of thread count.

## How verified

- 158/158 ctest targets pass (new unit tests: coalescing-queue state machine including
  dirty-span remap, geometric prepare, solve cancellation, `shiftSurfaceOrigin`
  mapping consistency, content-anchored position remap, lazy fiber-tree contract).
- Field-tested against a remote (S3) `.volpkg` with ~670 fibers loaded: placement loop
  measured at click ≈ 4 ms, apply ≈ 8 ms, save tail well under a frame; no stall-sampler
  hits in the editing loop (the diagnostic instrumentation used for the hunt was
  removed before this PR).
- Two full rounds of 5x independent Codex reviews (round 1: full-context, cold, Qt
  lifetime, concurrency, repo-rules/geometry; round 2, against the fixed tree:
  full-context, cold, solve-lifecycle state machine, side-strip async lifecycle,
  save/data integrity). All confirmed findings are fixed in the last five commits,
  covering: manual-reoptimization mode honored on every dispatch path (supersede,
  debounce, publish epilogue, quit-save) with manual edits recorded as pending;
  deletion escalates a debouncing request to whole-line; cancellation stops work
  promptly (post-`ceres::Solve` check, cancellable seed solves and span
  initialization, orphaned-session cancel+shutdown on package switch and pane
  teardown); the side-strip pipeline's token lattice (clears invalidate in-flight
  queries, the branch preview can't clobber a full result, worker deliveries queue
  on the application object, session snapshot caches die with their session);
  save integrity (snapshots validated before `_fibers` commits, quit-time supersede
  forces re-finalization, a finalized session's cancelled solve launches no
  successor, delete/rename drain queued save jobs, every persistent-artifact path —
  export, atlas, point collections, fiber open — flushes pending autosaves, edits
  flush the previous landing's autosave before mutating); rollback lifecycles
  (successful finalize clears them; edits invalidate mode-change rollbacks); off-line
  solver results rejected at apply; stale rapid-click positions re-derived from the
  3D point; and the anchor remap prefers index continuity so adjacent spiral wraps
  can't steal it.

## Accepted trade-offs

- Autosave debounce leaves a ≤1.5 s crash window and panel staleness (flush points
  cover every structural transition, plus fiber export and atlas creation).
- Structural edits (delete/split/merge/link/goal/mode) still refuse while a solve is
  running, by design; a partial solve over a stale pending set escalates to whole-line.
- When the decoded-chunk budget is fully occupied by viewer tiles, a solve's chunks
  evict among themselves (`decodedEvictionPreferSelf`) and may re-fetch across
  iterations — the deliberate side of "solves must not displace what the user is
  panning over".
- Cancellation is cooperative: a solve blocked inside a single chunk fetch or one
  native trace segment finishes that step before it can stop.
- Back-to-back landings within one frame can transiently mix overlay generations in a
  pane until its next frame adoption (self-healing via the geometry-epoch gates).
- `shiftSurfaceOrigin` is session-local by design (documented): it is not persisted
  and applies only to the in-memory strip surfaces.
- Pre-existing and unchanged: scaled collection exports do not re-validate the 1e-8
  control-point membership after scaling.

https://claude.ai/code/session_01SAftyRp3ECkXyJBWYPHayt


Bullo27:
Heads-up before this reaches a reviewer: **CI is red on `c46dc0b`, and the failure is a real test failure**, which sits awkwardly against the `158/158 ctest targets pass` line in the description.

## What failed

| job | step | annotation |
|---|---|---|
| `Linux base tests (Clang QuickBuild)` | **Build and test base shard** | `Process completed with exit code 8` |
| `Gate` | — | `This PR changes 33 files (> 20). Add at least 1 approving review to merge.` |

The `Gate` one is that workflow working as designed, not a defect — noting it only so it isn't mistaken for a second problem.

## Why exit code 8 means a test failed, rather than a build break

I measured ctest's exit codes instead of assuming them, on a throwaway project with one passing and one failing test (ctest 3.28.3):

| situation | exit code |
|---|---|
| selection matches no tests (`-R` or `-L`) | **0** |
| all selected tests pass | **0** |
| any selected test fails | **8** |

So an empty `vc-core` label set — the innocent explanation I checked first — would have exited 0, not 8. And compilation on this same commit is fine: `Linux VC3D compile and smoke`, `Linux Lasagna/fiber/GUI tests` and `Linux CLI compile coverage` all pass. That leaves a genuinely failing test in the base shard.

*Caveat I can't remove:* that table is ctest 3.28.3 on my machine. The CI container may ship a different ctest, and while I'd expect the codes to be stable, I haven't verified the version inside `vc3d-deps/linux`.

## The control

`Linux base tests (Clang QuickBuild)` is **`completed/success`** on main at `e1d7bf0a1`. It is a path-filtered job, so most main commits don't run it — that's the only recent main run I found — but it establishes the job can pass, so the red here isn't a standing breakage.

## Narrowing it

The job runs (`vc3d-ci.yml:177-178`):

```
ctest --test-dir build/ci-fast-core --output-on-failure --parallel 4 -L '^vc-core$'
```

On main locally that label selects **139 of 157** tests. This PR touches six test files, and two of the ones it edits are inside that selection:

- `core/test/test_fiber_widget.cpp` → `test_fiber_widget` (#14)
- `core/test/test_quadsurface_basics.cpp` → `test_quadsurface_basics` (#78)

Those are where I'd look first. `test_lasagna_line_optimizer` and `test_line_annotation_generated_views` are also edited here; the new `apps/VC3D/test/test_line_annotation_optimization_queue.cpp` is in a different target and wouldn't be selected by this label.

**I did not reproduce it.** The job log is 403 anonymously so I can't name the failing test, and I didn't build the branch: no clang on this box and a different preset than `ci-fast-core`, so a local pass would not have been evidence either way. Everything above is read off the checks API plus the workflow file. If it would help, I'm happy to build the branch under GCC and run the `vc-core` selection to name the test — say the word.

## One small thing for anyone reading the diff via the API

`GET /pulls/1622/files` pages at 30, so an unpaginated call reports **30 files / +2740 / −404**. The real totals are **33 files / +2992 / −420** — the three it drops are all test files (`test_lasagna_line_optimizer.cpp`, `test_line_annotation_generated_views.cpp`, `test_quadsurface_basics.cpp`), which is exactly the set that matters here. `Gate` counting 33 is the correct number.

Three checks are still in flight (`macOS compile`, `Compile (MSYS2 UCRT64)`, `PR scan (VC3D controllers)`), so this may not be the whole picture yet.

---
*Claude (AI agent) on behalf of @Bullo27. Check-run state and file list fetched 2026-08-27T13:5x UTC; the ctest exit-code table comes from a project I built and ran, not from documentation.*

Bullo27:
Following up on my own comment, because two things in it need correcting and the picture has changed.

## The base-tests job is green now

`Linux base tests (Clang QuickBuild)` was re-run at **13:53:02Z** and finished **success** at 13:55:55Z, on the same commit `c46dc0b`. `Gate` is the only remaining failure, and that one is by design.

So the red I flagged was **transient**. My measurement stands — exit code 8 does mean ctest saw a test fail, and it wasn't an empty selection or a compile break — but the conclusion a reader would draw from it, that this PR carries a broken test, is not supported. A run failed; the failure doesn't reproduce. I should have weighted "flaky infrastructure" more heavily before posting, and I'm sorry for the noise.

## I was wrong about which tests the label selects

I wrote that the new `test_line_annotation_optimization_queue.cpp` "is in a different target and wouldn't be selected by this label." That is **incorrect**, and it pointed away from exactly the right place. `apps/VC3D/test/CMakeLists.txt:293-297` calls `vc_register_ctest`, which applies `vc-core` (`CMakeLists.txt:894`), and ctest confirms it:

```
Test #158: line_annotation_optimization_queue
```

It is in the selection the failing job runs. Had the failure been real, that test — a state machine with a debounce window, newly added — would have been the first thing to look at, not the last.

## What I measured on the branch

Built `c46dc0b` in a clean worktree, GCC 13.3.0, `Release`, `VC_USE_PCH=OFF`:

| | main `6847063ff` | this branch |
|---|---|---|
| `ctest -N` total | 157 | **158** |
| `-L '^vc-core$'` | 139 | **140** |
| new targets | — | `line_annotation_optimization_queue` |
| removed | — | none |

**`158/158` in your description is confirmed** — exactly one added test, and the full suite passes. The precise CI command, `ctest --output-on-failure --parallel 4 -L '^vc-core$'`, passed **140/140**, and I then re-ran it four more times at `-j8` to try to shake out a timing-sensitive failure in the new async paths. Five runs, zero failures, no flake reproduced locally.

That doesn't prove the CI flake is benign — a different compiler, `-O0`, and a contended 4-way runner are exactly the conditions a debounce-sensitive test would be sensitive to, and none of them are what I ran. If it recurs, `line_annotation_optimization_queue` is where I'd start.

## Standing caveat

I built `VC_USE_PCH=OFF` only. PCH-on catches a different class of break — an identifier colliding with something the carrier header drags in, `emit` being the classic in Qt code — so my green run doesn't cover that configuration, and this PR adds a lot of Qt-adjacent code.

---
*Claude (AI agent) on behalf of @Bullo27. Built and run on 2026-08-27 in a detached worktree at `c46dc0b`; check-run timestamps read from the API after the re-run.*

hendrikschilling:
Ok most of these are not critical, however they can have performance impacts similar to what caused the indefinite render hangs some time back - P1 is that same pattern. Better way is to not postpone but run stuff at some low priority, display its result even if its outdated by now because that is still more up-to-date then the original stale data.
Also 3. should probably be changed to just wait on a shared request indefinitely and not duplicate work.

```
  1. P1: continuous edits can prevent optimization from ever publishing.
     Every control-point edit invalidates the running solve and requests cancellation (LineAnnotationController.cpp
     (https://github.com/ScrollPrize/villa/blob/c46dc0b7c6d10160f480cea0fe5e94d5e8d91578/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L7484-L7507)). Its result is then       
     discarded wholesale and a successor starts (LineAnnotationController.cpp
     (https://github.com/ScrollPrize/villa/blob/c46dc0b7c6d10160f480cea0fe5e94d5e8d91578/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L10537-L10595)). If edits arrive faster 
     than solves finish, optimized strip geometry and its autosave can be postponed indefinitely. This behavior is new because the previous UI blocked additional edits.

     The correct model for your requirement is one active operation plus one latest pending operation, without cancelling the active operation. Because a stale solve cannot simply         
     overwrite newer edits, publishing it requires span-versioned merging/rebasing, or another generation-aware application mechanism.

  2. P1: branch previews can create an unbounded queued-work tail.
     Every debounced request with branch links starts another global-thread-pool job (LineAnnotationController.cpp
     (https://github.com/ScrollPrize/villa/blob/c46dc0b7c6d10160f480cea0fe5e94d5e8d91578/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L13231-L13296)). The token check drops  
     obsolete results, but there is no active/pending bound. Continued interaction can queue one runnable every 150 ms; queued jobs retain copied strip and branch data, while
     already-running previews duplicate CPU work.

     Do not fix this by cancelling each preview. Remove the separate preview or make it one-active/one-latest-pending. The full query already publishes branch intersections as a partial   
     result, so the separate duplicate computation appears unnecessary.

  3. P1: the claimed 10-second read bound is not real.
     After waiting ten seconds for the shared request, every follower performs the same source read synchronously (ChannelSampler.cpp
     (https://github.com/ScrollPrize/villa/blob/c46dc0b7c6d10160f480cea0fe5e94d5e8d91578/volume-cartographer/core/src/lasagna/ChannelSampler.cpp#L906-L929)). That duplicate read has no    
     timeout or cancellation, so a GUI caller can still block indefinitely. Multiple timed-out followers can also fan out into duplicate remote reads.

     The source fetch itself needs a real deadline/cancellation mechanism. At most one elected takeover should be permitted if duplicate fallback remains.

  4. P2, pre-existing: side-strip full queries have the same refresh-starvation model.
     A newer request changes the shared token (LineAnnotationController.cpp
     (https://github.com/ScrollPrize/villa/blob/c46dc0b7c6d10160f480cea0fe5e94d5e8d91578/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L13221-L13224)); the running query      
     treats that as cancellation (LineAnnotationController.cpp
     (https://github.com/ScrollPrize/villa/blob/c46dc0b7c6d10160f480cea0fe5e94d5e8d91578/volume-cartographer/apps/VC3D/LineAnnotationController.cpp#L13609-L13617)). Its output is rejected 
     and only the latest pending query starts. This existed before the PR, but PR 1622 does not satisfy the requested invariant here either.
```

skyward7187:
Thanks — reworked all four areas onto the render scheduler's model in the four commits ending at b22b78ef0 (one active job never cancelled by input, one latest pending, publish even superseded results, cancellation = invalidation only):

1. **Solves**: edits no longer cancel the running solve. Its landing is published by span merge (`mergeSupersededSolveResult`, pure and unit-tested): every span the edits didn't touch adopts the solved geometry (endpoint-verified at the loader tolerance, metadata travelling with adopted spans, edited outer spans protecting their extrapolated tails), the edits' spans keep their provisional splice, and the pending union then dispatches from the merged baseline. When nothing is adoptable the landing still refreshes the views from the session's provisional geometry, so every landing makes visible progress. As a prerequisite, the manual-mode edit branch now runs the same geometric splice as auto (it used to leave the line without the moved control's vertex). Solve threads run at `QThread::LowPriority`.
2. **Branch previews**: removed — the full query computes branch links first and already publishes them as its first partial.
3. **Chunk followers**: wait on the shared in-flight read indefinitely again; the 10 s duplicate-read fallback is gone (it bounded nothing and fanned out duplicate source reads).
4. **Side-strip queries**: superseded queries complete and publish; pending requests are per-surface with FIFO start; the only cancellation is an invalidation watermark (pane close / package switch / intentional clears, scoped to their surface); a post-publish catch-up resolves A→B→A orderings.

Each step went through independent review passes (plan and implementation); the residual accepted trade-offs are listed in the updated PR description.

hendrikschilling:
Looks like its going the right direction!
2. could also be left as is - this just means we only save after some break in annotation.

```
  1. P1: control-point placement can still block on chunk I/O.
     LineAnnotationController.cpp:7328 calls lineModelFromPointsSplicing() synchronously on the GUI thread. That function calls sampleNormalBatch() at LineAnnotationController.cpp:14602,  
     which can prefetch and wait for Lasagna chunks. Remote or uncached Lasagna data can therefore still freeze placement. The comment claiming “without volume access” is incorrect.       
     Provisional geometry should reuse/interpolate normals and leave authoritative resampling to the background solve.

  2. P2: the claimed ≤1.5-second crash window remains unbounded during continuous edits.
     A superseded result is published as Incremental at LineAnnotationController.cpp:10620 and does not schedule autosave. Autosave is only scheduled by a complete publication
     around :9918, while flushSessionAutoSave() rejects non-Optimized state at :14973. If edits continually supersede solves, the latest work remains memory-only indefinitely. Normal      
     close/quit finalizes it, but a crash can lose arbitrarily old edits.

  3. P2: one side-strip A→B→A case still retains obsolete work.
     The cached-result shortcut at LineAnnotationController.cpp:13315 now runs while a same-surface B query is active, but it does not remove that surface’s pending request. B subsequently
     publishes and the obsolete pending request starts at :13843; only afterward does catch-up recover A. This causes visible A→B→pending→A oscillation and unnecessary full queries. At    
     minimum, a same-surface cache hit should retire the older pending intent.
```

skyward7187:
Round-2 items addressed in the three commits ending at 44770ab66:

1. **Click path is free of GUI-thread chunk sampling.** The splice no longer resamples: `spliceLineModelWithInterpolatedNormals` (pure, unit-tested) re-indexes the old replaced range's authoritative normals proportionally (full-range replacements included), blends hemisphere-aligned boundary normals only for pure insertions (geometric weights), and picks a display anchor that provably satisfies the view builder's tangent constraint. Authoritative resampling stays with the background solve, whose landing (or adopted merge spans) replaces the provisional normals; in manual mode they persist until an explicit solve — they're real pre-edit samples, and positions/solver inputs/saved geometry are untouched. The old sampling splice is deleted.
2. Left as is, per your note.
3. **Same-surface cache hits retire the older queued intent**, and the underlying scheduler got the two structural fixes this exposed: the reuse cache is per-surface (a single global slot let two panes' catch-ups evict each other into a self-sustaining re-query loop), and catch-up now runs on every successful publish rather than being suppressed by starting a queued other-surface request. A failed refresh keeps the freshest displayed set (restoring the older cache only when the failure itself painted partial markers).

Each fix went through plan review, implementation review, and a verification pass (several findings were confirmed by reviewers with reproduction harnesses before fixing).

skyward7187:
Field bug fixed in the four new commits: annotating with remote data could fail with "Line optimization failed: chunk invalidated" (and "Could not reopen fiber N: chunk invalidated"). Root cause: `getChunkBlocking` could wake to find its entry erased — decoded-budget eviction removes whole entries, and the `decodedEvictionPreferSelf` caches shed their own freshly decoded chunks first while a solve streams many of them. In the tight-budget regime the completion path even evicted the chunk *before notifying* the parked reader, so it failed deterministically.

Fix: parked blocking readers now pin their entry (`Entry::blockingWaiters`); both evictors (decoded-budget and metadata-capacity) skip pinned entries, the budget's victim selection skips them symmetrically so its enforce loop can't spin, and the reader re-enforces both limits when it releases the pin so nothing stays over its cap. Deliberate invalidation still clears pinned entries — a bounded retry loop refetches in that case. Overshoot from pins is bounded by the number of parked readers (one chunk each).

Three deterministic regression tests (mid-wait invalidation, sub-chunk byte budget, zero entry capacity), each verified failing before its commit. Three adversarial review rounds on the stack; findings fixed down to P3.

hendrikschilling:
close!

```
  - P2: invalidation can transfer a waiter pin to the wrong cache-entry incarnation. getChunkBlocking() increments blockingWaiters on the current entry at volume-cartographer/core/src/    
    render/ChunkCache.cpp:1679, but waitForResolvedLocked() waits only by chunk key at :5556. If invalidation removes the entry and one waiter creates a successor, other old waiters can   
    continue waiting on that successor without registering their own pins. They then decrement its count at :1716-1723.

    With multiple waiters and a tight budget, the first awakened thread can remove the only pin and trigger eviction. Remaining waiters then refetch sequentially; enough waiters can       
    exhaust the five-retry limit. The new tests cover only one blocking reader.

    Minimal fix: give each entry an incarnation/fetch serial, wait for that specific incarnation, and decrement only the entry the caller actually pinned. When the incarnation changes,    
    every waiter must loop and register a new pin before waiting again.
```

skyward7187:
Confirmed — the guarded decrement was single-waiter thinking; across an invalidation, old waiters could decrement a successor's pin they never registered, and the first-awakened thread could strip the only real pin. Fixed in the four commits ending at 4157940d4.

I started from your incarnation suggestion, but review of that plan found a residual gap: re-registering on wake is not guaranteed to happen before the successor resolves, fully unpins, and gets evicted — a descheduled waiter can still lose its data and refetch, so one invalidation could cascade into up to N sequential refetches even with correct incarnation accounting. The implemented design goes one step further in the same direction you pointed: **the protection is keyed to the reader, not the entry**. `getChunkBlocking` registers a per-key count in `State` (`blockingReadersByKey_`) on entry and releases it exactly once on every exit; both evictors and the budget's victim probe skip keys with parked readers. Since no entry lifecycle event can destroy or transfer a registration, a successor is protected from the moment it exists no matter which thread created it, waiters have no re-registration step to be late for, and a misattributed release has no mechanism. `Entry::blockingWaiters` is gone; `waitForResolvedLocked` inlined into the reader's wait (whose predicate also marks `backgroundDemand`, so demand pruning can't erase a successor a parked reader has observed).

Hardening the reviews surfaced along the way: queue-setup throws no longer orphan an InFlight entry that later readers would park on forever (both in `getChunkBlocking` and in `refreshFetchers`' restart/requeue loops, where one throwing fetcher also used to skip the reader-waking notify and starve the remaining restarts); listener callbacks on scheduler worker threads are now exception-isolated (an escape terminated the jthread, and inside `runSourceTransfer` a throwing start-listener turned a healthy fetch into IoError); invalidation stop events reach every listener even when one throws.

Accepted bounds, documented in the commits: the count lets the shared budget overshoot by at most one chunk per parked reader (re-enforced on release); a reader that loses its entry to five successive deliberate invalidations mid-read still errors (livelock backstop — each retry makes progress otherwise, and only that reader is affected).

Tests: `Stats` gains a `blockingReaders` observability seam. The single-reader trio remains the deterministic discriminators; new contract tests cover the successor-created-by-another-party case (exactly two fetches in every interleaving) and your two-reader scenario (both registrations observed before the fetch resolves; the first exiting reader cannot strip the other's protection; budget drains only after both exit), plus a six-reader invalidation stress. Two plan review rounds (5 lenses each) and a 5-lens implementation round + convergence passes; 158/158 suite, chunk-cache suite 20x consecutive clean runs.

hendrikschilling (approved):

PR #1624, opened 2026-08-27
Fix public S3 access with stale AWS credentials

hendrikschilling:
**In one sentence:** VC3D can attach public S3 Lasagna manifests even when ambient AWS credentials are stale or invalid, while retaining authenticated access for private data.

**One real example:** Starting with the public PHerc0139 Lasagna manifest at `https://vesuvius-challenge-open-data.s3.amazonaws.com/PHerc0139/representations/predictions/fibers/20260102150214-fibers-20260801084232-L1/PHerc0139-20260102150214-las-sd1-92481a4c.lasagna.json`, I loaded it with invalid AWS credentials present, and both the manifest and its `presence` Zarr metadata opened successfully.

**Before:** Recognized S3 URLs were always signed when AWS credentials were present. Invalid or expired credentials caused public objects to fail with errors such as `HTTP 400 InvalidToken`, without attempting anonymous access.

**After this PR:** Explicit AWS authentication failures retry anonymously. Successful public access becomes sticky for subsequent manifest, metadata, and chunk requests, while valid authenticated access to private objects remains unchanged.

**Proof:** The real-data network test opens the original PHerc0139 manifest and verifies the `presence` array shape as `(9620, 3314, 3314)`. The focused test suite passes, including repeated concurrent fallback tests.

**Why / where this is useful:** Users can attach public Lasagna datasets from S3 without clearing credentials belonging to another AWS account or refreshing an unrelated expired session.

## Details

- Adds a shared, thread-safe S3 authentication fallback policy.
- Retries only HTTP 401/403 and explicit AWS credential errors such as `InvalidToken`, `ExpiredToken`, and `SignatureDoesNotMatch`.
- Does not retry unrelated HTTP 400, 404, 5xx, or transport failures anonymously.
- Coordinates concurrent fallback attempts so only one anonymous probe is made per failure wave.
- Applies the policy to remote manifest downloads, Lasagna Zarr access, and the existing generic Zarr fetcher.
- Does not change project or manifest serialization.

Verification:

```bash
cmake --build volume-cartographer/build --parallel 32 \
  --target test_http_fetch_errors test_remote_file_cache \
  test_lasagna_manifest test_lasagna_project_volumes \
  test_remote_url VC3D

ctest --test-dir volume-cartographer/build --output-on-failure \
  -R "^(test_http_fetch_errors|test_remote_file_cache|test_lasagna_manifest|test_remote_url|test_lasagna_project_volumes)$"

ctest --test-dir volume-cartographer/build --output-on-failure \
  --repeat until-fail:20 -R "^test_http_fetch_errors$"

VC_TEST_REQUIRE_NETWORK=1 \
  volume-cartographer/build/bin/test_lasagna_manifest
```

### DIFF
--- a/volume-cartographer/core/CMakeLists.txt
+++ b/volume-cartographer/core/CMakeLists.txt
@@ -29,6 +29,7 @@ add_library(vc_core
         src/RemoteUrl.cpp
         src/RemoteFileCache.cpp
         src/HttpFetch.cpp
+        src/S3AuthFallback.cpp
         src/PostProcess.cpp
         src/Geometry.cpp
         src/PointCollections.cpp
--- a/volume-cartographer/core/include/vc/core/render/ZarrChunkFetcher.hpp
+++ b/volume-cartographer/core/include/vc/core/render/ZarrChunkFetcher.hpp
@@ -58,16 +58,17 @@ OpenedChunkedZarr openHttpZarrPyramid(
     std::optional<int> baseScaleLevel = std::nullopt);
 
 // Production remote-volume open policy shared by VC3D and standalone tools:
-// resolve the locator, discover credentials when requested, and retry public
-// S3 data anonymously when stale credentials cause an authentication error.
+// resolve the locator, try recognized S3 sources anonymously, then retry with
+// discovered credentials only when anonymous access is denied.
 OpenedRemoteChunkedZarr openRemoteZarrPyramid(
     const std::string& url,
     RemoteZarrOpenOptions options = {});
 
 // S3 deliberately returns AccessDenied rather than NotFound for a missing key
 // when the caller lacks ListBucket. Optional discovery probes must therefore
-// accept a generic 403 while preserving explicit credential failures so the
-// anonymous retry policy can still run. Exposed for deterministic tests.
+// accept a generic 403 while preserving explicit credential failures. The
+// opener remembers suppressed denials and retries authenticated if no required
+// array metadata can ultimately be found.
 bool isOptionalRemoteMetadataMiss(
     long status,
     std::string_view key,
--- a/volume-cartographer/core/include/vc/core/util/S3AuthFallback.hpp
+++ b/volume-cartographer/core/include/vc/core/util/S3AuthFallback.hpp
@@ -1,0 +1,50 @@
+#pragma once
+
+#include <utils/http_fetch.hpp>
+
+#include <condition_variable>
+#include <cstdint>
+#include <functional>
+#include <mutex>
+#include <optional>
+#include <string_view>
+
+namespace vc
+{
+
+[[nodiscard]] bool hasExplicitAwsCredentialError(std::string_view detail);
+
+[[nodiscard]] bool isAwsAuthenticationFailure(long status, std::string_view detail);
+
+class S3AuthFallback final
+{
+public:
+    using Attempt = std::function<utils::HttpResponse(bool anonymous)>;
+
+    struct Result {
+        utils::HttpResponse response;
+        std::optional<utils::HttpResponse> authenticatedFailure;
+        bool usedAnonymous = false;
+    };
+
+    S3AuthFallback(bool isS3, bool credentialsLoaded);
+
+    [[nodiscard]] Result request(const Attempt& attempt) const;
+    [[nodiscard]] bool usesAnonymous() const;
+
+private:
+    enum class Mode {
+        Disabled,
+        Authenticated,
+        ProbingAnonymous,
+        Anonymous,
+        AuthenticatedOnly,
+    };
+
+    mutable std::mutex mutex_;
+    mutable std::condition_variable probeFinished_;
+    mutable Mode mode_;
+    mutable std::uint64_t probeGeneration_ = 0;
+};
+
+}  // namespace vc
--- a/volume-cartographer/core/include/vc/core/util/S3AuthFallback.hpp
+++ b/volume-cartographer/core/include/vc/core/util/S3AuthFallback.hpp
@@ -3,7 +3,6 @@
 #include <utils/http_fetch.hpp>
 
 #include <condition_variable>
-#include <cstdint>
 #include <functional>
 #include <mutex>
 #include <optional>
@@ -16,6 +15,8 @@ namespace vc
 
 [[nodiscard]] bool isAwsAuthenticationFailure(long status, std::string_view detail);
 
+// Selects anonymous S3 access on the first successful request and retains
+// authenticated access only after anonymous access is denied.
 class S3AuthFallback final
 {
 public:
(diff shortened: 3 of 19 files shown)


Bullo27:
Reproduced the premise against live S3 and drove `S3AuthFallback` directly. **The headline claim holds and the concurrency design is sound** — I tried to break it and failed. Two narrower things look like real gaps, and one of my own findings turned out to be a measurement error; all below.

## Independently confirmed

I fetched the manifest without any credentials at all, and resolved the array myself rather than taking the number:

| | |
|---|---|
| manifest, unsigned anonymous GET | **HTTP 200**, 699 B |
| `groups.presence.zarr` | `…_presence.ome.zarr/3`, `scaledown: 3` |
| `.zarray` shape | **[9620, 3314, 3314]** — matches the PR body exactly |
| internal check | `base_shape_zyx` = [76954, 26512, 26512]; `ceil(base / 2³)` = **[9620, 3314, 3314]** ✓ |

The premise reproduces too, and the two credential shapes differ in a way that matters below:

| credential state | GET result |
|---|---|
| none (anonymous) | **200** |
| stale **session token** | **400**, body `<Code>InvalidToken</Code>` |
| bogus long-term key | **403**, `<Code>InvalidAccessKeyId</Code>` |

`InvalidAccessKeyId` isn't in the three examples the PR body names, but it *is* in the 11-marker list in the code, and it's a 403 anyway — so it's covered twice.

I compiled `S3AuthFallback.cpp` standalone (g++ 13.3, `-std=c++20`) and drove the real state machine with synthetic responses. Against the bodies S3 actually returned to me: 400+`InvalidToken` → retry, 403+`InvalidAccessKeyId` → retry, 400+`InvalidRequest` → no retry, 404/500 → no retry. All correct.

**The concurrency claim holds.** Counting anonymous attempts that *start before the first one finishes*:

```
threads=2   authenticated_attempts=1  anonymous_total=2   started_before_probe_finished=1
threads=8   authenticated_attempts=1  anonymous_total=8   started_before_probe_finished=1
threads=32  authenticated_attempts=1  anonymous_total=32  started_before_probe_finished=1
```

One probe per wave, and only **one** authenticated attempt across all 32 threads — better than advertised. Stickiness holds too: after the first wave, five further requests made zero authenticated attempts.

## Gap 1 — the credential-marker scan runs on *successful* responses

`request()` applies the guard to the authenticated response regardless of status:

```cpp
auto authenticated = attempt(false);
if (mode != Mode::Authenticated || !isAwsAuthenticationFailure(authenticated.status_code, authenticated.body_string())) {
```

`isAwsAuthenticationFailure` returns true if the body contains any marker — so an **HTTP 200 whose body merely mentions one** is treated as an auth failure. Driving it with a private object (authenticated 200, anonymous 403):

```
200, ordinary manifest body      anon_probes=0  returned=200  -> caller gets the data     (control)
200, body mentions AccessDenied  anon_probes=1  returned=403  -> caller gets an ERROR
200, body mentions InvalidToken  anon_probes=1  returned=403  -> caller gets an ERROR
```

The successful authenticated response is discarded and the failed anonymous one is returned. Row 1 is the control, so this isn't my harness misbehaving.

Worth flagging because this PR **adds `AccessDenied` to the list** — the previous `ZarrChunkFetcher` list had 10 markers and did not include it. `AccessDenied` is an ordinary string to find inside stored content: IAM policy documents, CloudTrail exports, error logs, or a provenance blob quoting a past failure. For public data the retry is harmless (same bytes). For private data it turns a good 200 into a thrown error, which is the case the PR promises leaves "valid authenticated access to private objects unchanged."

Gating the marker scan on `!response.ok()` would keep every behaviour the new tests assert while removing this.

Your own `preserves private authenticated mode` test passes today only because its 200 body is the literal `"private"`. I ran it with that one string changed and nothing else:

```
body="private"               ok()=PASS(200)  validSigned==1 PASS  validAnonymous==0 PASS
body="private AccessDenied"  ok()=FAIL(403)  validSigned==1 PASS  validAnonymous==0 FAIL (was 1)
```

## Gap 2 — `exists()` uses HEAD, and HEAD has no body to scan

`Dataset::exists()` resolves through the fallback with `head()`. A HEAD response carries no body by definition, so the marker scan has nothing to read, and only the 401/403 status rule can fire. But a stale **session token** — the PR's own headline scenario — answers **400**:

```
GET  400 + InvalidToken body   isAwsAuthenticationFailure=true   anon_probes=1  final=200  -> exists()=true
HEAD 400 + empty body          isAwsAuthenticationFailure=false  anon_probes=0  final=400  -> exists()=FALSE
HEAD 403 + empty body          isAwsAuthenticationFailure=true   anon_probes=1  final=200  -> exists()=true   (control)
```

Rows 1 and 3 both fire, so row 2's silence is a property of HEAD rather than of my harness. `exists()` returns `result.response.ok()`, so the key is reported **absent** rather than retried anonymously — a silent wrong answer instead of an error.

This is order-dependent and I want to be precise about the blast radius: `get_if_exists()` uses GET, so once any GET has established anonymous mode on the shared member, later `exists()` calls ride it. The exposure is an `exists()` that runs *before* any GET has flipped the policy.

Your test suite has `CHECK(vc::isAwsAuthenticationFailure(403, {}))` — an empty body with 403 — but no 400-with-empty-body case, which is why this sits just outside it.

## A smaller question

`RemoteFileCache.cpp:253` constructs the policy as a **per-call local**, while `Dataset.cpp:510` holds it as a member. So "successful public access becomes sticky for subsequent manifest, metadata, and chunk requests" is true on the Lasagna path but not in `defaultFetch`, where every fetch repeats the authenticated failure first. Intentional (one-shot fetches), or worth hoisting?

## A finding of mine that was wrong

My first harness reported 2/8/32 anonymous probes for 2/8/32 threads and I briefly had it written up as "the one-probe-per-wave claim does not hold." **It was my instrument.** I counted every anonymous *request*, but threads that correctly wait for the probe must still fetch their own object afterwards — that traffic is the design working, not duplicated probing. Measuring attempts that begin before the probe completes gives 1 in every case, as shown above. Recording it so the wrong number doesn't end up quoted from a half-read comment.

## Limits

- I did **not** build VC3D or run your ctest targets; the probe compiles `S3AuthFallback.cpp` standalone against `utils/http_fetch.hpp`, so it exercises the policy object, not the libcurl path.
- Live-S3 signing was done with boto3, not VC3D's client. The HEAD/GET body asymmetry is an HTTP property, but I have not observed it through `utils::HttpClient` itself.
- Both gaps are reachability arguments from the state machine plus synthetic responses. I have not produced either against a real private bucket — I don't have one.
- File count cross-checked independently (`git diff --name-only main..pr1624` = 17, matching the API page, which truncates at 30).

---
*Claude (AI agent) working on behalf of @Bullo27. Every status code, shape and count above came from a command run today against `e1902132` and live `vesuvius-challenge-open-data`; the probe was compiled and executed, not sketched.*

hendrikschilling:
@codex review

PR #1625, opened 2026-08-27
Spiral gap expander fix

bruniss:


Bullo27:
Verified the math and the test evidence here. **Everything I could check holds** — including the hand-written Triton gradient, which was the part I expected to find something in. One dead assignment and one coverage gap below.

The description is empty, so in case it helps @pmh47 / @giorgioangel: as I read it this is two independent fixes plus a migration. **Parameterisation** — the gap map changes from the unbounded `dr · exp(200·L)` to a lower-bounded softplus, `min_gap + (dr − min_gap) · softplus(bias + S·L) / softplus(bias)`. **Capacity** — `model_gap_expander_num_windings` was serving as both the physical winding estimate and the allocated lattice extent; the allocation half moves to a new `model_gap_expander_capacity_windings` (default 144). `checkpoint_migrations.py` carries old checkpoints across both.

## The parameterisation claims

Imported `gap_parameterization.py` straight from the branch and drove it, rather than reading the docstrings:

| claim | result |
|---|---|
| "a zero latent maps exactly to `dr_per_winding`" | **exact**, abs err `0.0e+00` across 6 (dr, min_gap, bias) combinations incl. bias 0 and −2 |
| calibration preserves the legacy slope `200·dr` at latent 0 | **holds**, rel err ≤ `2.1e-16` (float64 autograd) in all 6 |
| gap ≥ `min_gap`, always | holds over latents `−1e6 … +1e6`, all finite, min hits exactly `min_gap` |
| removes the exponential's overflow | legacy is `7.1e18` at L=0.2 and **`inf` by L=0.5**; new map is `1.2e4` at L=2 |
| documented `ValueError`s | all four raise (`min_gap` ≥ dr, = 0, < 0; `initial_dr == min_gap`) |

The calibration check has a control: perturbing `softplus_scale` by 1% moves the measured slope to a rel err of `1.0e-02`, so the test can fail.

A detail I think is an unadvertised improvement: the new `initial_dr_logit` → `lower_bounded_dr` round-trip is **exact**, where main's `softplus(logit · 12)` with `logit = initial_dr/12` only approximated it. At `model_initial_dr_per_winding = 5.0`, main recovers `5.006715` (err `6.7e-03`); the new path recovers `5.0` exactly. It converges for larger dr, so this matters at small winding spacings.

## The Triton path

`gap_triton.py` reimplements the same map independently and hand-derives the backward pass. I checked that derivative against the analytic gradient of the Python reference:

```
gap_logit_derivative = (dr - min_gap) * sigmoid(bias + S·L) * S / softplus(bias)
```

which is exactly `d/dL [ min_gap + (dr − min_gap)·softplus(bias + S·L)/softplus(bias) ]`. The `truncate_frac` variants agree too — `dr_derivative = 1 + tf·(ratio − 1)` and the `tf`-scaled logit gradient are the right derivatives of `lerp(dr, gap, tf)`. **No error found.**

**The gap I'd flag:** no test references `gap_bracketing_radii` or `gap_search_radii` — grepping the branch, those symbols appear only in `gap_triton.py` and `transforms.py`. So the formula now lives in two independent implementations plus a hand-written gradient, with nothing pinning them together. A parity test (same logits through the Triton path and `lower_bounded_gap`, forward and via `gradcheck`) would be cheap insurance against silent drift, and it's the one place here where drift would be hard to notice — a wrong gradient degrades fits rather than raising.

## On the existing review comment

@copilot-pull-request-reviewer got here first and raised one thing — that `_capacity_geometry()` concatenates an int64 `cumsum` with a float32 `torch.zeros(1)`, and that "`torch.cat` requires matching dtypes, so this will raise at runtime when expanding checkpoint capacity."

**I don't think that one lands.** `torch.cat` type-promotes rather than requiring matching dtypes:

```
num_by_winding dtype = torch.int64
torch.cat([zeros(1) float32, cumsum(int64)]) -> no raise, result dtype = torch.float32
                                                values = [0.0, 6.0, 16.0, 31.0, 50.0, 73.0, 100.0, 131.0]
```

Control: a genuinely invalid `cat` (mismatched non-concatenated dim) does raise `RuntimeError` in the same harness, so it isn't blind to cat failures. Calling `_capacity_geometry(...)` directly returns `(int64, float32)` without raising, and `tests/test_gap_parameterization.py` — which exercises the expansion path and asserts on the expanded config — passes 4/4 on this head.

The float32 intermediate is also harmless downstream, because `expand_gap_checkpoint_capacity` immediately normalises it:

```python
new_indices = new_indices_cpu.to(dtype=old_indices.dtype, device=old_indices.device)
```

and takes `new_width` through an `int()`. Measured on torch 2.10.0+cu128 — if the intent is to support an older torch with stricter `cat`, that would change the answer, so worth a word if so.

## Test evidence

Full `spiral-fitting/tests/` suite, same machine, same interpreter:

```
main       41 failed, 620 passed, 16 skipped, 1 error, 42 subtests passed
PR head    41 failed, 626 passed, 16 skipped, 1 error, 42 subtests passed
```

**+6 passed, and the failing set is byte-identical** — 42 failing/erroring nodes on each, with 0 unique to either side. Those 42 are all in the native-extension modules (`test_spiral_sampling`, `test_surface_index`, `test_packed_patch_satisfaction`, …) and are **my environment**, not this branch: `RuntimeError: Patch sampling requires vc_spiral.spiral_sampling.PatchSamplingAtlas`. They fail identically on main, so they say nothing about this PR either way.

Two of the changed tests genuinely discriminate — run against **main's** production code they fail, and pass on the branch:

- `test_config.py::test_gap_capacity_and_numerical_floor_are_explicit_and_validated` (main 11 passed/1 failed → PR 12 passed)
- `test_spiral_helpers.py::ResolveOuterWindingIdxWiringTests::test_gap_expander_control_also_runs_without_a_shell` (main 20/1 → PR 21 passed)

I also confirmed the constraint the new config description states is actually enforced, rather than only documented: `config.py:589` floors capacity at 3, and `spiral_helpers.py:1072` enforces the real `shell_outer_winding_idx + 3` — with `shell_outer_winding_idx = 130`, that's the `>= 133` your test asserts.

## One dead assignment

`transforms.py:495` still sets `self.dr_per_winding_scale = 12.  # larger value increases effective learning rate`, but `get_dr_per_winding()` now calls `lower_bounded_dr()`, which uses `DR_PARAMETER_SCALE` from `gap_parameterization.py`. Grepping the branch, `dr_per_winding_scale` is assigned at :495 and never read again.

The two constants agree at `12.0` today, so there's no behavioural change — I checked that specifically, since a silent shift in the dr learning rate would have been easy to ship here. But the comment now sits on an inert variable, and someone tuning that number would see no effect.

## Notes on my own method

One thing I initially recorded as discrimination was an artifact: running `test_phase_spacing.py` from a temp directory fails on main with `ModuleNotFoundError: No module named 'test_sdt_losses'` — a sibling-import failure caused by *where I ran it*, not by main's code. It passes in place on both. Likewise `test_gap_parameterization.py` "failing" on main only means the module is new. Neither is evidence; the two tests listed above are.

## Limits

- **I did not execute the Triton kernels.** The gradient check above is symbolic — I compared the kernel's expression to the analytic derivative by reading. Triton 3.6.0 and CUDA are available here, but the surrounding tests need the native extensions I don't have built, so the GPU path is unexercised on my side.
- No real-scroll fit was run. Everything above is unit-level: the parameterisation functions, the config validators, and the test suite.
- I have not evaluated the checkpoint migration against a real legacy checkpoint — `checkpoint_migrations.py` (+256) is the largest new file here and the part I have the least evidence about.

---
*Claude (AI agent) working on behalf of @Bullo27. Every number above is from a command run today against `4d7988c8` and `6847063f`; the parameterisation probe was executed, not sketched.*

Issue #1621, opened 2026-08-27
spiral satisfaction metric cannot detect a sheet switch: a patch displaced by any whole number of windings scores identically to a correctly placed one


jonmarrs:
**Summary.** `get_patch_satisfied_areas` decides whether a patch is satisfied by comparing it
against a target it derives from the patch's own position: it snaps the patch's median
shifted-radius to the nearest integer winding, then checks two residuals against that self-derived
target. Because the target moves with the patch, a patch sitting on the wrong wrap is scored against
the wrong wrap and passes. For a patch lying on a winding, displacing it by a whole winding changes
the satisfied-quad fraction by exactly zero.

This is the failure mode the 2026 open problems page lists fourth in its bottleneck table, "Meshes
can jump from one wrap to another", where the "what would help" column asks for conservative failure
detection.

**The blindness is periodic, not a one-winding special case.** What decides acceptance is the
displacement's distance from the nearest integer winding, not its magnitude. Measured against
villa's own unmodified function on an analytic spiral, at `dr = 12.81` voxels:

| displacement (windings) | distance from nearest winding | verdict |
|---|---|---|
| 0.5 | 0.5000 | rejected |
| 1.0 | 0.0000 | satisfied |
| 5.5 | 0.5000 | rejected |
| 23.8006 | 0.1994 | **satisfied** |

A patch displaced by 23.8006 windings, roughly 305 voxels at that spacing, is scored satisfied,
while one displaced by half a winding is rejected. The acceptance edge is bracketed at
`(0.44, 0.46]` and re-resolves identically five windings out, so it does not move with magnitude.
`0.45` is `metrics_config['satisfaction_radius_tolerance']`, so the acceptance half-width is the
tolerance itself in units of `dr_per_winding`.

**It holds under the configuration that gates the output mesh, which is the more permissive of the
two.** All of the above is under the default `metrics_config`. The second call site, inside
`save_mesh`, overrides all three thresholds to `0.495` / `12.0` / `0.90` and decides which patches
are spliced into the output mesh. Under that configuration the whole-winding delta is still exactly
zero, and the acceptance edge moves outward to `(0.49, 0.499]`. So the path with downstream
consequences rejects a narrower strip around the midpoint than the reporting path does.

It also holds across the theta = 0 seam, where the branch-offset unwrap is exercised, and under
radial warps interpolated from the inter-winding spacings actually measured along rays in the
published PHercParis4 `winding_model` export.

[...]

PR #1626, opened 2026-08-27
spiral: report direct absolute-winding disagreements

TAUIL-Abd-Elilah:
## Summary

Add a report-only direct absolute-winding check beside the existing periodic patch-satisfaction metrics, and make `find_inconsistent_windings --pcl` honor its explicit diagnostic inputs even when the source checkpoint disabled those inputs during fitting.

This addresses #1621 conservatively: it reports evidence of an integer-winding disagreement without changing loss, satisfaction decisions, or mesh selection.

This covers directly anchored patch cells only. It does **not** add the complementary track check: `get_track_satisfied_counts_in_chunks` still discards `mode_winding_per_track`, so that remaining path needs separate treatment.

## What changes

- compare each directly attached `metadata.winding_is_absolute` annotation with the native self-derived winding target at the same retained quad cell;
- handle both directions of the theta-zero seam using the same reference-step convention as `ThetaCrossingMap`;
- emit `absolute_winding_diagnostic_<suffix>.json` plus one summary log line;
- keep relative/fiber propagation out of this report rather than promoting indirect evidence to an absolute claim;
- honor explicit PCL/fiber arguments in `find_inconsistent_windings` even when a supervision-free checkpoint has those inputs disabled;
- preserve all existing satisfaction, splicing, and optimizer behavior.

## Real before/after validation

On two real PHercParis4 5,000-step checkpoints trained with PCL supervision disabled:

- before: explicit `--pcl` paths were silently filtered; `0 absolute`, `0 direct votes`;
- after: 6 absolute collections loaded and the one in-window direct anchor was compared;
- baseline: annotation-derived 37 versus native 45, difference +8;
- second fit: annotation-derived 37 versus native 43, difference +6.

Both anchor quads and both patches were already native-strict-unsatisfied, so this example validates the real diagnostic path but is not claimed as a previously accepted false negative. With one anchor, it also cannot distinguish a global integer gauge offset from a local switch.

The predeclared follow-up across all six existing checkpoints reproduced a
nonzero direct-cell disagreement every time: baseline `[+8, +7, +7]` and
cap-0.75 `[+6, +6, +5]` for seeds 17/23/101. Every whole-patch strict verdict
was false; the seed-101 cap-0.75 anchor quad was locally strict-true despite a
+5 disagreement. This remains one spatial anchor and is not a prevalence or
treatment-effect claim.

The predeclared dense-anchor follow-up then compared 49/50 in-window anchors
across four distinct patches on a fresh 5,000-step seed-17 fit. The raw
difference mode was +12 windings (46 anchors); after removing it, three anchors
on three patches had residuals -12, +1, and +3. None was native-strict-satisfied
at either its anchor quad or whole patch. This validates multi-anchor coverage
and finds localized candidates, but **does not** demonstrate the strongest case
of a silently accepted whole-winding error. Forty-six comparisons also come
from one collection on one patch, and the singleton second-collection residual
can reflect a collection convention difference.
Exact outputs, hashes, failed attempts, and limitations:
https://gist.github.com/TAUIL-Abd-Elilah/bceb2ae8923aa07d309e81dabe55ec31

Replication protocol and four additional reports:
https://gist.github.com/TAUIL-Abd-Elilah/3c53376761897c67695f75bfa663732e

## Verification

- 50 related tests passed in independent audit, including the actual CUDA path on an RTX 3090;
- whole-winding shifts of +1, +2, and +23 remain native-satisfied in the synthetic control but are separated by the report;
- ordinary 2-voxel radial scatter produces no disagreement;
- both theta-seam directions are covered;
- the loaded Windows native binding is pinned by SHA-256
  `32d9269d...617a327` (this PR does not change native code);
- `git diff --check` passes;
- the clean PR commit has the same tree as the audited and real-run evidence commit.

AI disclosure: Codex assisted with implementation and verification. Every real-data number above comes from retained command output, and the negative/limited findings are stated explicitly.

### DIFF
--- a/spiral-fitting/find_inconsistent_windings.py
+++ b/spiral-fitting/find_inconsistent_windings.py
@@ -180,14 +180,31 @@ def build_fit_inputs(checkpoint, patches_dir, pcl_paths, filter_z_begin, filter_
     Returns (fit_config, scroll, paths, model_z_begin, model_z_end); the model
     z-range shapes the transform's flow field independently of the filtering
     z-range."""
-    from fit_session import PclInputSpec, ScrollSpec, SpiralInputPaths
+    from fit_session import (
+        PclInputSpec,
+        ScrollSpec,
+        SpiralInputPaths,
+        effective_pcl_role,
+    )
 
     cfg = fs.Config().as_dict()
     cfg.update(checkpoint['cfg'])
+    pcl_specs = tuple(PclInputSpec(path=str(path), role=None)
+                      for path in pcl_paths)
     # This tool IS the patch graph — a checkpoint trained supervision-free
     # (disable_patches, e.g. the 2026-07-17 normals-only baseline) must not
     # stop the loaders from reading the patches it wants to analyse.
     cfg['input_disable_patches'] = False
+    cfg['input_use_verified_patches'] = True
+    # --pcl and --fibers are explicit diagnostic inputs rather than requests
+    # to reproduce a checkpoint's training-input ablation. Enable only the
+    # supplied roles, leaving every unrelated checkpoint input unchanged.
+    # Role-less paths retain the historical basename convention.
+    for spec in pcl_specs:
+        role = effective_pcl_role(spec.role, spec.path)
+        cfg[f'input_use_pcl_{role.value}'] = True
+    if fibers_path:
+        cfg['input_use_fibers'] = True
     # We compute no shell losses; zero their weights so the context's
     # shell_losses_enabled() gate stops load_host_inputs() from loading the
     # shell (replacing the old fs.shell_losses_enabled = lambda: False
@@ -214,8 +231,7 @@ def build_fit_inputs(checkpoint, patches_dir, pcl_paths, filter_z_begin, filter_
     # unless a fibers dir is passed.
     paths = SpiralInputPaths(
         umbilicus=umbilicus_path,
-        pcls=tuple(PclInputSpec(path=str(path), role=None)
-                   for path in pcl_paths),
+        pcls=pcl_specs,
         fibers=fibers_path or '',
         verified_patches=patches_dir,
     )
--- a/spiral-fitting/fit_spiral.py
+++ b/spiral-fitting/fit_spiral.py
@@ -4665,6 +4665,7 @@ class FitContext:
                 patches_dict=self.verified_patches,
                 patch_atlas=self.patch_atlas,
                 unattached_pcl_strips=self.unattached_pcl_strips,
+                cross_patch_pcls=self.cross_patch_pcls,
                 tracks=self.tracks,
                 unverified_patches_list=self.unverified_patches_list,
                 unverified_patches_dict=self.unverified_patches,
--- a/spiral-fitting/satisfaction_metrics.py
+++ b/spiral-fitting/satisfaction_metrics.py
@@ -7,6 +7,7 @@ import numpy as np
 import torch
 
 from sample_spiral import (
+    get_theta,
     get_theta_and_radii,
     get_theta_crossing_step_adjustments,
     radius_from_unwrapped_shifted,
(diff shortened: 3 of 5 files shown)


Bullo27:
Ran this on Linux/CPU as an independent check of the RTX 3090 audit. **The tests pass, the regression evidence is clean, and the `--pcl` fix genuinely discriminates.** One scope observation and one thing I could not check.

To make the check meaningful I first built the Spiral native extensions locally (`cmake -S spiral-fitting/cpp`, nanobind, g++ 13.3, Python 3.13) — without them `report_absolute_winding_diagnostic` cannot run at all, since packed satisfaction raises `RuntimeError: Packed satisfaction requires vc_spiral.spiral_sampling.PatchSatisfactionAtlas`. Worth stating because a reviewer without those built will see all five of your new tests fail for reasons that have nothing to do with this PR — that was my first run, and it was entirely my environment.

## Regression evidence

Full `spiral-fitting/tests/`, same machine and interpreter, native extensions present:

```
main       12 failed, 663 passed, 11 skipped, 42 subtests passed
PR head    12 failed, 669 passed, 11 skipped, 42 subtests passed
```

**+6 passed, and the failing set is identical** — 12 nodes on each side, 0 unique to either. So nothing here regresses.

Those 12 are all `test_vram_reductions.py::CpuTrackStorageTests` and they fail the same way on clean `main`, so they are not this PR's business. I have **not** determined whether they are a genuine main-branch break or an artefact of my locally-built `track_store`, so I am flagging rather than attributing them — worth someone checking separately.

## Your new tests, on different hardware

```
tests/test_absolute_winding_diagnostic.py        5 passed
tests/test_find_inconsistent_windings_inputs.py  1 passed
```

Including `test_direct_anchor_diagnostic_indexes_cpu_profiles_from_cuda_evaluation`, which passes here on the CPU path — so the mixed CPU/CUDA tensor indexing isn't relying on the 3090.

Both discriminate against `main`, which is the part that matters:

- `test_find_inconsistent_windings_inputs.py` fails on main with a plain `assert False is True` — a **logic** failure, not an import error. That's the `--pcl` filtering bug reproducing exactly as described: on main, `build_fit_inputs` builds `PclInputSpec(role=None)` and never sets `cfg['input_use_pcl_<role>']`, so a checkpoint carrying `input_use_pcl_absolute: false` silently drops paths the operator passed explicitly on the command line. Silent filtering of an explicit CLI argument is the failure class worth fixing on its own, independent of the diagnostic.
- `test_absolute_winding_diagnostic.py` errors on main only because `report_absolute_winding_diagnostic` doesn't exist there, so that one is trivially true rather than evidence.

## A scope note

Your docstring states the mechanism precisely — patch satisfaction derives its target from the evaluated patch, so it cannot distinguish a patch on the intended winding from the same patch translated by a whole winding. That is the patch side of #1621, and this covers it.

The **track** side isn't covered here, and I don't think it should be silently assumed to be. In #1621 @jonmarrs measured that `get_track_satisfied_counts` already computes `mode_winding_per_track` and returns it as its fifth value, while `get_track_satisfied_counts_in_chunks` — the entry point `satisfaction_metrics.py` imports — unpacks five and returns two, discarding it. His table also showed the two signals are complementary rather than redundant: the half-winding control drives the satisfied count to zero while showing **zero** winding disagreement, and the whole-winding cases pass the count while showing nonzero disagreement. Neither catches the other.

So a reader of this PR might reasonably conclude #1621 is now covered. It's covered for directly-anchored patch cells; the track path still discards the quantity a symmetric check would need. Not a request to widen this PR — the conservative scoping is the right call — just worth saying explicitly in the PR or the issue so the remaining half doesn't get lost.

## What I could not check

- **None of your real-data numbers.** I have no PHercParis4 5,000-step checkpoints, so the `+8/+7/+7` and `+6/+6/+5` tables, the six absolute collections, and the `col5 / point 97 / distance 0.317` anchor are unverified by me. Your gists document them; I'm simply not evidence either way.
- **No CUDA run.** I have an RTX 3060 but the suite above ran CPU-only, so your 3090 regression is not independently reproduced — only the CPU path of the same test.
- I did not build an independent synthetic fixture for the diagnostic; the whole-winding separation claim is verified here **through your tests**, not through a construction of my own. That's weaker evidence than an independent harness would be, and I'd rather say so than imply otherwise.
- Fork PRs get no CI in this repo, so none of this is confirmable by Actions until it's on a branch that runs them.

For what it's worth, the conservative framing reads right to me: report-only, direct anchors only, explicit skip counters instead of guessed comparisons, and the honest statement that one anchor cannot separate a global integer gauge offset from a local switch. That last caveat is the one I'd have worried about had you not made it yourself.

---
*Claude (AI agent) working on behalf of @Bullo27. Every count above is from a command run today against `1c124d53` and `6847063f` on Linux/CPU; where I had no data to run, I've said so rather than inferred.*

TAUIL-Abd-Elilah:
Thank you for the independent Linux/CPU run and for separating the patch and track scopes so clearly. I agree with that boundary and have left this PR unchanged: #1626 covers direct patch anchors only; the chunked track path remains separate.

I have now completed the preregistered dense-anchor follow-up mentioned in the PR evidence. This was a fresh 5,000-step PHercParis4 fit on `[15000,16000)`, seed 17, with every PCL/fiber/track input disabled during fitting, followed by the frozen evaluator on the public absolute annotations.

Results:

- 50 in-window candidates, 49 compared, 4 distinct patches;
- raw native-minus-annotation differences `{0: 1, +12: 46, +13: 1, +15: 1}`;
- unique point-weighted global mode `+12`;
- residuals after removing that mode `{-12: 1, 0: 46, +1: 1, +3: 1}`;
- 3 nonzero-residual anchors on 3 patches;
- **0** nonzero residuals on native-strict-satisfied anchor quads or whole patches.

So the frozen interpretation is `localized_candidate_only_on_native_strict_failed_patches`. This validates the multi-anchor report and finds off-mode candidates, but it does **not** demonstrate the strongest case of a whole-winding error silently accepted by the native strict metric. I will not present it as that.

The dependence caveat matters: 46/49 comparisons come from PCL 1 on one `0000_top_band` patch. The lone PCL-2 point accounts for the `-12` residual after a PCL-1-dominated global mode and can reflect a collection convention difference. The two off-mode PCL-1 rows are `+1` and `+3`, both on native-strict-failed patches.

Full result, all 49 rows, manifest, hashes, and the earlier four reports are in the existing evidence gist:
https://gist.github.com/TAUIL-Abd-Elilah/3c53376761897c67695f75bfa663732e

Checkpoint SHA-256: `068d877e3d87eb7c0fd2507dac8c2bb132c4a03ab23953cb1bd75d02a33e6a42`. Source stayed clean at `1c124d53`; the checkpoint validator confirmed 5,000 completed iterations and every disabled-input gate. Analyzer/validator tests: 13 passed.

For the separate track-side plumbing you identified, I prepared a current-main branch that preserves the already-computed modes through `get_track_satisfied_counts_in_chunks` behind an opt-in kwarg, leaving the existing two-value caller unchanged. Its synthetic regression exercises three chunks and confirms the complementary half/whole-winding behavior (3 passed):
https://github.com/ScrollPrize/villa/compare/main...TAUIL-Abd-Elilah:track-absolute-winding-diagnostic

AI disclosure: Codex assisted with the implementation, frozen evaluation, and this report. Numeric claims above are copied from the retained outputs; the negative strongest-case result and dependence limits are included explicitly.

pmh47:
The metric adds almost nothing versus the existing pcl-satisfaction; I also don't trust it's handling of theta=0 crossing case. There might be some value to a PR that fixed the find_inconsistent... thing, and another that changed pcl-satisfaction metric slightly to report errors in abs w-n separately to relative.

aviad12g:
@TAUIL-Abd-Elilah, could you share one of the existing PHercParis4 checkpoints used in your `find_inconsistent_windings --pcl` before/after, plus the matching umbilicus and a minimal patch/PCL input set (or pinned public links)?

Your reports and hashes are available, but I could not find downloadable model weights. The earlier `[10500,11500)` two-patch example with seed patch `0000_mid_band_final` would be enough. Please include the Villa revision, checkpoint hash, diagnostic command, and exact scan/coordinate scale for those inputs.

We are independently checking the explicit-input loading issue credited in this PR. Local native-loader comparisons pass on a synthetic fixture, but we have not reproduced your real-data run. The aim is a narrowly scoped reproduction of input loading.

Disclosure: this request was drafted by Codex for the project owner.

TAUIL-Abd-Elilah:
@aviad12g — everything for the `[10500,11500)` two-patch case with seed patch `0000_mid_band_final` is here: https://github.com/TAUIL-Abd-Elilah/pherc-paris4-absolute-winding-repro

- **Checkpoints** (release `v1`, 474.7 MB each): `sealed-baseline-seed17.ckpt` sha256 `341048c2…31e9c`, `sealed-cap075-seed17.ckpt` sha256 `26bd5754…60d68` (full hashes in the README).
- **Inputs** in the repo: umbilicus, both patches, and the three PCL files the run loaded, each hash-pinned. Source is `dl.ash2txt.org/datasets/spiral_datasets/PHercParis4`.
- **Villa revision:** `1c124d53de222f882b389a1d7287eff5e48392ff`
- **Frame:** PHercParis4, `voxel_size_um 9.6`, umbilicus `coordinate_scale 1.0`, `lasagna_scale 4`, `CW`; native z `[10500,11500)`; seed zyx `[10936.452, 2740.049, 4185.588]`; `max_hops 0`.
- **Expected:** baseline annotation 37 vs native 45 (+8); cap 0.75 37 vs 43 (+6). Recorded baseline JSON and stdout are in `outputs/`.

One thing to flag up front: `spiral-scroll.json` is **not** served upstream — its own `provenance_note` says it was materialised locally from villa's historical production defaults. If your loader derives the frame differently, that is the first place I'd look.

Line endings are pinned (`* -text`) so the hashes verify on Windows clones too. Happy to run anything on my side if a number disagrees.

aviad12g:
Thanks for supplying the bundle. We verified the baseline checkpoint SHA-256 and ran the unchanged native host loader on the real inputs at Villa `757f70c0140a4cfbbbd44975ef09558444b96980`.

The original loads zero annotation collections. With the narrow input-selection correction below it loads 460 source collections, retains the expected 6 absolute / 10 relative / 9 neither collections, and attaches col5 point 97 to `0000_mid_band_final`. The coordinates, fractional IJ, and attachment distance match your exported record exactly. Both patch arrays and validity masks are unchanged. This confirms your reported input-suppression issue; we are not claiming its discovery.

You offered to run checks on your side. If your existing CUDA environment supports this pinned revision, could you run the full diagnostic once unmodified and once with the patch below? Our completed comparison only covers host loading; we have not restored the transform or reproduced the GPU result. No training or second checkpoint is needed. If this requires a new setup or fails at checkpoint migration, a log of that limitation is enough.

Use two clean copies at the pinned revision, applying this diff only to the candidate. Both commands below use the same downloaded baseline and your provided inputs. Please retain JSON and stdout/stderr from each; stop each process after 10 minutes if it has not finished.

```diff
--- a/spiral-fitting/find_inconsistent_windings.py
+++ b/spiral-fitting/find_inconsistent_windings.py
@@ -180,14 +180,23 @@
     Returns (fit_config, scroll, paths, model_z_begin, model_z_end); the model
     z-range shapes the transform's flow field independently of the filtering
     z-range."""
-    from fit_session import PclInputSpec, ScrollSpec, SpiralInputPaths
+    from fit_session import (PclInputSpec, ScrollSpec, SpiralInputPaths,
+                             effective_pcl_role, pcl_role_toggle_key)
 
     cfg = fs.Config().as_dict()
     cfg.update(checkpoint['cfg'])
+    pcl_specs = tuple(PclInputSpec(path=str(path), role=None) for path in pcl_paths)
     # This tool IS the patch graph — a checkpoint trained supervision-free
     # (disable_patches, e.g. the 2026-07-17 normals-only baseline) must not
     # stop the loaders from reading the patches it wants to analyse.
     cfg['input_disable_patches'] = False
+    # Explicit diagnostic inputs are independent of the checkpoint's training
+    # selection. Absolute PCLs also require the selected verified patches.
+    cfg['input_use_verified_patches'] = True
+    for spec in pcl_specs:
+        cfg[pcl_role_toggle_key(effective_pcl_role(spec.role, spec.path))] = True
+    if fibers_path:
+        cfg['input_use_fibers'] = True
     # We compute no shell losses; zero their weights so the context's
     # shell_losses_enabled() gate stops load_host_inputs() from loading the
     # shell (replacing the old fs.shell_losses_enabled = lambda: False
@@ -214,8 +223,7 @@
     # unless a fibers dir is passed.
     paths = SpiralInputPaths(
         umbilicus=umbilicus_path,
-        pcls=tuple(PclInputSpec(path=str(path), role=None)
-                   for path in pcl_paths),
+        pcls=pcl_specs,
         fibers=fibers_path or '',
         verified_patches=patches_dir,
     )
```

In an already configured Python/CUDA environment, set the three paths and run this small comparison driver:

```python
import os
from pathlib import Path
import subprocess
import sys

original = Path("/path/to/original/spiral-fitting").resolve()
candidate = Path("/path/to/candidate/spiral-fitting").resolve()
bundle = Path("/path/to/pherc-paris4-absolute-winding-repro").resolve()
results = Path("villa-pcl-gpu-comparison").resolve()
results.mkdir(exist_ok=False)
env = dict(os.environ, AGENTS_AGENT_MODE="1", PYTHONDONTWRITEBYTECODE="1",
           FIT_SPIRAL_COMPILE="0", FIT_SPIRAL_TRITON="0",
           FIT_SPIRAL_NUM_THREADS="2", OMP_NUM_THREADS="2", MKL_NUM_THREADS="2",
           FIT_SPIRAL_PATCH_LOAD_WORKERS="1", FIT_SPIRAL_PATCH_LOAD_IO_THREADS="1",
           WANDB_MODE="disabled", MPLCONFIGDIR=str(results / "matplotlib-cache"))
for side, source in [("original", original), ("candidate", candidate)]:
    command = [sys.executable, str(source / "find_inconsistent_windings.py"),
               "--checkpoint", str(bundle / "sealed-baseline-seed17.ckpt"),
               "--patches-dir", str(bundle / "inputs/patches"),
               "--umbilicus", str(bundle / "inputs/umbilicus.json"),
               "--patch-id", "0000_mid_band_final"]
    for name in ["abs_winding.json", "relative_windings.json", "same_windings.json"]:
        command += ["--pcl", str(bundle / "inputs/pcl" / name)]
    command += ["--z-range", "10500,11500", "--max-hops", "0",
                "--step-size", "1.0", "--medial-weight", "4.0",
                "--detect-loops", "--no-min-fix", "--no-plot",
                "--output", str(results / (side + ".json"))]
    with (results / (side + ".log")).open("x") as log:
        subprocess.run(command, env=env, cwd=results, stdout=log,
                       stderr=subprocess.STDOUT, timeout=600, check=True)
```

Expected input effect: original 0 absolute / 0 direct votes; candidate 6 absolute / 1 direct vote from col5 point 97, annotation 37. Please report the actual seed winding, inferred seed vote, and dr rather than forcing your older numbers to match. Your exported seed values 46/39 are different from the direct-anchor-cell 45/37 comparison. We retain your warning that this is one anchor and that the scroll metadata was reconstructed from historical defaults.

If it completes, would you also describe whether restoring this warning changes any actual annotation or model-review decision for you? An unfavorable or no-benefit finding is useful too. The software reproduction itself does not establish anatomical correctness.

TAUIL-Abd-Elilah:
Ran it on GPU at `757f70c0`, both sides exit 0. Logs and JSON: https://github.com/TAUIL-Abd-Elilah/pherc-paris4-absolute-winding-repro/tree/master/outputs/757f70c0

| | abs / rel / neither PCLs | votes | seed winding (raw → rounded) | inferred seed vote | `dr_per_winding` |
|---|---|---|---|---|---|
| original | 0 / 0 / 0 | 0 | 46.347 → 46 | — | 15.8060 |
| candidate | 6 / 10 / 9 | 1 direct | 46.347 → 46 | 39 | 15.8060 |

Candidate: `abs pcl 5 ('col5') point 97 on patch 0000_mid_band_final: annotation=37 -> expected seed winding=39.000`. That is the input effect you predicted.

Actual values, not forced to match: the continuous seed winding is **46.347** here versus **46.473** in my August record at `1c124d53`. Rounded value, vote and `dr` are unchanged. I haven't looked into which change between the two revisions moves the raw value.

Two deviations from your driver, both in the committed script: `check=False`, so one side failing can't suppress the other's log; and the MSVC-built `vc_spiral` `.pyd` files placed inside each tree's own `spiral-fitting/vc_spiral/`. A `PYTHONPATH` entry doesn't work because the source `vc_spiral/` next to the script shadows it — my first attempt failed with `PatchSamplingAtlas` missing for exactly that reason. The patch touches only a `.py`, so one native build serves both sides. Your patch applied cleanly; a failure I first hit was my own encoding bug when extracting it from the comment.

On your question — **no, it has not changed an annotation or model-review decision for me.** The one anchor it exposes sits on a patch the native metric already rejects (13.15% strict satisfied against the 95% threshold), so the extra vote added nothing decision-relevant there. What it fixes is that the diagnostic silently reported `0 absolute` for supervision-free checkpoints: a correctness fix for the tool rather than a result that moved a decision. One anchor still can't separate a global gauge offset from a local sheet switch.

PR #1629, opened 2026-08-27
spiral-fitting: document the default full-scroll fit window and its o…

Drfiya:
**In one sentence:** Records what actually happens if you run `fit_spiral.py` without setting `z_begin`/`z_end`, which is not the failure the tutorial prepares you for.

**One real example:** On PHerc0826, under a reduced config (patches off, `dense_spacing_mode: grad_mag`, shell-loss weights zeroed), we left the z-range at its default and the run reached real optimization, accelerated past 0.9 it/s, and then died around 300 iterations — no traceback, no error, ETA by then over 8 hours. GPU memory never went above 8 GiB of a 32 GiB card.

**Before:** The README says nothing about the default window. The tutorial says to consider a small range because fitting all of Scroll 1 needs "around 60 GB" of GPU memory — which reads as "you're fine if you have enough VRAM." We had enough VRAM, so we ran it, and got a silent death instead of an OOM. A silent death is much harder to diagnose than the error you were told to expect.

**After this PR:** One paragraph noting that `config.py` defaults to `z_begin = 4000`, `z_end = 17000` (the range inherited from the PHercParis4 production dataset, not something derived from the scroll you're fitting), what we saw at that scale, and the honest state of the diagnosis — we don't know what kills the process; a host-RAM kill during the ~8.4M-track preparation is our leading candidate. The advice is to set a smaller window regardless of your card's VRAM.

**Proof:**

Deliberate run with no window override, PHerc0826, this project's reduced config (patches disabled, `dense_spacing_mode: grad_mag`, zeroed shell-loss/dense-spacing weights), real box, captured 2026-08-27 (`logs/d41_default_window_oom.log`):

```
scaled per-step counts by 1.368 for the 13000-slice z-range [4000, 17000) (reference 9500 slices):
  sample_count_tracks_per_step=65684
  ...
PROGRESS Optimizing — 1/30,000 iterations (0.0%) — 0.0 it/s — elapsed 2m 13s — ETA 1106h 48m
PROGRESS Optimizing — 113/30,000 iterations (0.4%) — 0.4 it/s — elapsed 4m 44s — ETA 20h 51m
PROGRESS Optimizing — 233/30,000 iterations (0.8%) — 0.7 it/s — elapsed 5m 15s — ETA 11h 09m
PROGRESS Optimizing — 313/30,000 iterations (1.0%) — 0.9 it/s — elapsed 5m 35s — ETA 8h 49m
/root/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/lib/python3.14/multiprocessing/resource_tracker.py:297: UserWarning: resource_tracker: There appear to be 2 leaked semaphore objects to clean up at shutdown: {'/mp-n4x9qydx', '/mp-0dh30lg4'}
  warnings.warn(
```
(Log ends there — no further output, no Python traceback, process gone.)

No `dmesg`/kernel-log access exists inside a RunPod pod to confirm the exact cause; a host-RAM kill during preparation of 8,384,681 tracks for the full 13,000-slice window is the leading candidate, not confirmed — the resource-tracker warning immediately before the log stops is consistent with an external process kill, not a Python exception.

**Why / where this is useful:** The existing guidance frames the risk as GPU memory, so someone on a big card reasonably skips it. One README paragraph turns a day of "why did my job vanish" into a non-event.

- [x] I personally verified that the example and proof above were produced by this PR on the stated data. (The GPU-memory figure is observed-live from watching nvidia-smi during the run, not from a saved capture — see Limitations.)

## Details

**Method:** villa commit `6847063ffdb4da898ae8d1d494ebf7d71473f509` (2026-08-26 21:25 CEST). RunPod EU-RO-1, RTX PRO 4500 32GB (Blackwell, sm_120). `SCROLL=PHerc0826`, no `Z_BEGIN`/`Z_END` override (villa's own `config.py` default). Command: `fit_spiral.py` with our other config overrides (patches disabled, `dense_spacing_mode: grad_mag`, zeroed shell/spacing weights) but `z_begin`/`z_end` left unset.

**The added paragraph** (in `spiral-fitting/README.md`, placed right after the intro — see Files changed).

**Comparisons:** N/A (documentation-only change).

**Limitations:** The exact cause of the silent death (host-RAM OOM vs. something else) is not confirmed — this PR documents the observed symptom and its real cost (multi-hour ETA, no clean error) honestly as unconfirmed, not as a diagnosed root cause. Only tested against PHerc0826 and this project's specific reduced config; behavior under villa's own full default config (patches enabled, `dense_spacing_mode: phase`, nonzero shell weights) was not tested and may differ. The GPU-memory figure is an observed-live fact, not backed by a saved terminal capture — happy to re-run the deliberate default-window test with memory logging if that would help.

### DIFF
--- a/spiral-fitting/README.md
+++ b/spiral-fitting/README.md
@@ -4,6 +4,26 @@ Code and helpers to fit a canonical Archimedean spiral to deformed scrolls.
 `spiral_service.py` hosts one persistent interactive fit session over HTTP for
 the VC3D Spiral workspace; `fit_spiral.py` is the underlying fitter.
 
+## A note on the default fit window
+
+If you don't set `z_begin`/`z_end`, `fit_spiral.py` defaults to the entire
+written region of the scroll (`config.py`'s `z_begin = 4000`, `z_end =
+17000` — the range inherited from the PHercParis4 production dataset, not
+something derived from the scroll you're fitting). This is not what the
+tutorial's first-run example uses, and for good reason: under a reduced
+config (patches disabled, `dense_spacing_mode: grad_mag`, zeroed
+shell-loss weights — a tutorial-style setup), a
+full-range PHerc0826 run reached real optimization, accelerated past
+0.9 it/s, and then died silently around 300 iterations with no traceback
+and an ETA that had climbed past 8 hours — not the GPU out-of-memory error
+a "~60 GB, will OOM" figure might lead you to expect (actual GPU memory
+use stayed under 8 GiB of a 32 GiB card throughout). Whatever kills the
+process at this scale is not confirmed (a host-RAM kill during the
+~8.4M-track preparation for the full window is the leading candidate). Set
+`z_begin`/`z_end` to a smaller window (the tutorial's own ~1,000-slice
+recommendation is a reasonable first run) regardless of your GPU's VRAM
+headroom.
+
 ## Sweep runner output
 
 `runners/run_sweep.py` prefixes each active fit's live `PROGRESS` and


Bullo27:
Checked the factual spine of this from the source. **Every checkable claim holds**, and I think the framing argument is stronger than you put it.

**The defaults are exactly what you say.** `spiral-fitting/config.py`:

```
265:  self.z_begin = 4000
266:  self.z_end   = 17000
```

**And the tutorial confirms your "inherited from Scroll 1" point in its own words.** `scrollprize.org/docs/38_tutorial_spiral.md:124`:

> "Consider starting with a small range: the whole written region of Scroll 1 is roughly z 4,000–17,000, and fitting all of it needs a lot of GPU memory (around 60 GB). A ~1,000-slice range is a good first run on a smaller GPU."

So the default window is precisely "the whole written region of Scroll 1", and the guidance attached to it is framed twice over as a GPU-memory concern — "needs a lot of GPU memory", "on a smaller GPU". Someone on a 32 GB card reading that reasonably concludes the advice isn't aimed at them. That's your argument, and the doc supports it verbatim.

**Your log line is genuine and the mechanism is real.** `spiral_helpers.py:102` sets `REFERENCE_Z_RANGE_NUM_SLICES = 9500`, and `fit_spiral.py:4836` prints the scaling. `13000 / 9500 = 1.368`, matching the `scaled per-step counts by 1.368` in your capture exactly.

That last point is worth putting in the paragraph, because it sharpens the trap. The same tutorial sentence ends:

> "Per-step sample counts are scaled automatically to the size of the z-range, so hyperparameters don't need retuning when you change it."

That reads as reassurance — and it is, for *quality*. But it's also the mechanism that multiplies your per-step sample counts by 1.368 at the default window. So the two halves of that sentence pull in opposite directions for someone deciding whether the default is safe: the first half says the risk is VRAM, the second says you needn't retune. Neither prepares you for a host-side death. A reader who has enough VRAM and trusts the auto-scaling has been told, twice, that they're fine.

**On the diagnosis:** I think you're right to leave it unconfirmed. `sample_count_tracks_per_step=65684` with 8.4M tracks prepared, no traceback, and a resource-tracker leaked-semaphore warning immediately before the log ends is consistent with an external kill, and stating a host-RAM OOM as *the* cause without `dmesg` would be overclaiming. Documenting the symptom and the cost, with the cause marked unknown, is the honest version and still solves the reader's problem — they need to know to set a window, not why the process died.

**Limits on my side:** I ran none of this. No PHerc0826, no 32 GB card, and I did not attempt a default-window fit, so the ~300-iteration death, the 0.9 it/s, the 8h ETA and the sub-8 GiB VRAM are your observations and not independently reproduced here. What I verified is that the defaults, the reference-slice constant, the 1.368 scaling and the tutorial's framing are all exactly as you describe.

Your offer to re-run with memory logging seems worth taking up if a maintainer wants the cause pinned — but I'd merge the paragraph regardless, since its value doesn't depend on the root cause.

---
*Claude (AI agent) working on behalf of @Bullo27. The config values, the constant, the tutorial quote and the arithmetic above are from commands run today against `6847063ff`.*

pmh47:
Closing this since the better solution will be to default the z-range to None and have that pick up the actual full size from spiral-scroll.json (or other metadata)

PR #1630, opened 2026-08-28
vc_obj2tifxyz: fail instead of writing an all-sentinel tifxyz for normalized UVs

### LINKED ISSUE #1320: vc_obj2tifxyz: 'Successfully converted' and exit 0 after rasterizing 0 points

```
UV-metric mode: grid 2 x 2
Valid grid points: 0 / 4 (0%)
```

success.

what to pass.

AnayGarodia:
Fixes #1320

## Motivation

I was converting some of the path segments to feed into the render + ink steps when, with default args, the tool wrote full -1 output and still said "Successfully converted" with exit 0. This fix makes it fail loudly and say which flag to pass instead.

## What changed

- Exit non-zero when zero grid points were rasterized, before any output is written (previously: warning to stderr, then "Successfully converted", exit 0, and an all-sentinel `x/y/z.tif` + `bbox [[-1,-1,-1],[-1,-1,-1]]`).
- When the UV range looks normalized (<= 1.5 per axis), say so and name the flags that make the conversion work (explicit `stretch_factor` or `--uv-to-obj`).
- Fix the `--help` example labeled "(legacy behavior)" — the plain invocation actually runs in the default UV-metric mode.

## Evidence on real scroll data

Segment `PHercParis4.volpkg/paths/20230503225234` (normalized-UV OBJ, dl.ash2txt.org); control: `PHercParis4/segments/20260602204401-5753_-7` (metric-UV OBJ, s3 open-data bucket). Built and run on macOS arm64 (`scripts/build_macos.sh`).

**Before (main):** default args -> `Valid grid points: 0 / 4 (0%)` -> "Successfully converted", exit 0, output dir contains only -1 sentinels.

**After (this branch):**
- default args -> error naming the problem and the flags, exit 1, no output written
- `--uv-non-metric` (issue case 2) -> same error, exit 1
- explicit `stretch_factor 800` -> `Valid grid points: 503348 / 641601 (78.45%)`, real bbox, exit 0
- metric-UV control -> `26585600 / 45470061 (58.47%)`, converts exactly as before, exit 0

https://claude.ai/code/session_0189NTjHVfExtaqTtNSpXYck

### DIFF
--- a/volume-cartographer/apps/src/vc_obj2tifxyz.cpp
+++ b/volume-cartographer/apps/src/vc_obj2tifxyz.cpp
@@ -63,6 +63,7 @@ private:
     // When set, the output grid is sized to the source sampling density so the
     // flattened tifxyz keeps the input's scale instead of metric mode's 1.0.
     cv::Vec2f src_scale = cv::Vec2f(0.f, 0.f);  // <=0 = unset
+    int last_valid_count = -1;  // rasterized grid points from the last createQuadSurface()
 
 public:
     void setUVMetric(bool v) { uv_is_metric = v; }
@@ -70,6 +71,10 @@ public:
     void setUVDownsample(float f) { uv_downsample = std::max(1.0f, f); }
     void setGridCapPixels(uint64_t cap) { grid_cap_pixels = cap; }
     void setSrcScale(const cv::Vec2f& s) { src_scale = s; }
+    int validCount() const { return last_valid_count; }
+    cv::Vec2f uvRange() const { return uv_max - uv_min; }
+    int gridWidth() const { return grid_size[0]; }
+    int gridHeight() const { return grid_size[1]; }
     void setSrcApproval(const cv::Mat& m) { src_approval = m; }
     bool hasApproval() const { return !src_approval.empty() && !grid_uv.empty(); }
     const cv::Mat_<cv::Vec3b>& outApproval() const { return out_approval; }
@@ -312,6 +317,7 @@ public:
         
         std::cout << "Valid grid points: " << valid_count << " / " << (grid_size[0] * grid_size[1]) 
                   << " (" << (100.0f * valid_count / (grid_size[0] * grid_size[1])) << "%)" << std::endl;
+        last_valid_count = valid_count;
         
         // Scale is currently in OBJ units. Convert to micrometers now.
         if (valid_count == 0) {
@@ -539,7 +545,7 @@ int main(int argc, char *argv[])
         std::cout << std::endl;
         std::cout << "Note: Scale factors are automatically calculated from the mesh grid structure." << std::endl;
         std::cout << "Examples:" << std::endl;
-        std::cout << "  " << argv[0] << " mesh.obj outdir                       (legacy behavior)" << std::endl;
+        std::cout << "  " << argv[0] << " mesh.obj outdir                       (UV-metric mode, default)" << std::endl;
         std::cout << "  " << argv[0] << " mesh.obj outdir 800 1.0 --uv-metric  (UV is metric, OBJ units == UV units)" << std::endl;
         std::cout << "  " << argv[0] << " mesh.obj outdir --uv-metric --uv-to-obj=0.001" << std::endl;
         return EXIT_SUCCESS;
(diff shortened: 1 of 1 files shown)


Bullo27:
Reproduced #1320 from a synthetic mesh — no scroll data needed — and ran your fix against it on Linux x86-64. **The fix does exactly what it says.** One adjacent case survives it, which I think is worth a decision rather than a patch.

## The bug reproduces without any real segment

A four-vertex quad whose UVs form a diamond — so the UV bounding box corners all fall *outside* the covered region — is enough. Built from `main`:

```
Valid grid points: 0 / 4 (0%)
Successfully converted to tifxyz format          <- exit 0
files written: meta.json, x.tif, y.tif, z.tif
```

Same mesh against your branch:

```
Valid grid points: 0 / 4 (0%)
Error: 0 grid points rasterized; not writing an all-sentinel tifxyz.
UV range is 1 x 1 (normalized UVs?), which yields only a 2 x 2 grid.
Pass a stretch_factor (pixels per UV unit, e.g. 800) or --uv-to-obj=<OBJ units per UV unit>.
                                                  <- exit 1, 0 files written
```

A second fixture (a single triangle with UVs spanning 0.1) also hits `0 / 4` and produces the same refusal, with the range reported as `0.1 x 0.1` — so the `<= 1.5` normalized-UV hint fires on a genuinely different mesh, not just the one I built first.

**Controls, so the guard isn't just refusing everything:**

| fixture | main | this branch |
|---|---|---|
| metric-UV quad (UV span 800) | `640000 / 641601`, exit 0 | **identical**, exit 0 |
| normalized quad + explicit `800` | `640000 / 641601`, exit 0 | **identical**, exit 0 |

So the documented escape hatch still works and metric meshes are untouched.

The `--help` correction checks out too — `main` prints `mesh.obj outdir (legacy behavior)`, your branch prints `(UV-metric mode, default)`.

## The case that still gets through

While building a fixture I accidentally found one. A plain quad with UVs `[0,1]²` — the *same* normalized-UV mistake — doesn't rasterize zero points. It rasterizes **one**:

```
Valid grid points: 1 / 4 (25%)
Successfully converted to tifxyz format          <- exit 0, 4 files
```

Identical on `main` and on your branch, because the guard is `validCount() == 0` exactly. So the same root cause, on a mesh whose UV layout happens to cover one grid corner, still writes a 2×2 tifxyz that is 75% sentinel and still reports success. I spent a while trying to get my quad to produce a zero before realising it never would — which is roughly the position a user is in when their segment lands on 1/4 instead of 0/4.

I'm raising this as a question rather than proposing a threshold, because where the line belongs is a judgement call and I don't have the corpus to say what a legitimate low-coverage conversion looks like. But the two obvious candidates seem worth considering: refusing when the grid itself is degenerate (a 2×2 grid is never a useful tifxyz regardless of how many points landed in it), or refusing below some valid-fraction floor. The grid dimensions are already available via your new `gridWidth()`/`gridHeight()`, so the first is nearly free.

Not a blocker — this PR strictly improves on `main`, and the exact-zero case is what #1320 reported.

## Limits

- Synthetic meshes only. I don't have `20230503225234` or the metric-UV control segment, so your real before/after screenshots are your evidence, not reproduced here. What I verified is that the same *behaviour* reproduces on meshes I constructed.
- **Linux x86-64**, built into `build/local-verify` at `61a62c445`; you tested macOS arm64. Both builds compiled clean (exit 0, zero errors).
- I did not check the approval-mask path or `--uv-to-obj`; only the default and explicit-stretch_factor routes.
- Fork PRs get no CI here, so Actions won't confirm any of this.

The `delete surf;` before the early return is the right call given the raw `new` above it.

---
*Claude (AI agent) working on behalf of @Bullo27. Every exit code, count and file listing above is from a command run today against `61a62c445` and `f02042d7`.*

AnayGarodia:
Thanks for the verification and especially for the 1/4 quad — that's a real leak and your framing was right: the line belongs at the grid, not the count. Pushed 44de6fa: a 2x2 output grid is now refused regardless of valid count, since it's the clamped minimum and means the UV range collapsed below one output pixel. Your 1/4 quad now exits 1 with no output written; the 0/4 case, the explicit stretch_factor escape hatch (99.75% valid at 800), and the metric-UV control are unchanged.

I went with the degenerate-grid rule rather than a valid-fraction floor for the reason you gave — I don't have the corpus to say what a legitimate low-coverage conversion looks like either, and the 2x2 case is unambiguous.

<sub>Claude (AI agent) working on behalf of @AnayGarodia; the verification runs above were executed today against 44de6fa.</sub>

Bullo27:
Ran `44de6fa` against the same four fixtures, in place from the same Release build tree:

| case | exit | files written | key line |
|---|---|---|---|
| diamond quad, 0/4 valid (the #1320 bug) | **1** | 0 | `Warning: no valid grid points were rasterized.` |
| unit quad, 1/4 valid (the leak) | **1** | 0 | `Error: degenerate 2 x 2 output grid; not writing a sentinel-filled tifxyz.` |
| metric-UV control | 0 | 4 | `Valid grid points: 640000 / 641601 (99.7505%)` |
| normalized + `stretch_factor 800` escape | 0 | 4 | `Valid grid points: 640000 / 641601 (99.7505%)` |

That closes the exact-zero blind spot without touching any legitimate path I can construct — the two healthy cases are byte-for-byte the same counts as before the change. The degenerate-grid rule also reads better than a fraction floor would have, for the reason you gave: it needs no opinion about what legitimate low coverage looks like. Nothing further from me.

<sub>Verification by Claude (AI agent) operating on behalf of @Bullo27; runs executed today against `44de6fa` on Linux x86-64.</sub>

bruniss:
instead of recommending the user use `--uv-to-obj` just throw an error for an invalid tifxyz , a 2x2 will still fail even with that recommendation

AnayGarodia:
Good catch, dropped it. The message now just reports the invalid tifxyz and exits, no flag suggestion.

You're right that the suggestion was misleading. A 2x2 grid is the clamped minimum, so it comes out degenerate whatever the user passes next, and pointing them at a flag implies the output would have been fine with different args.

Diff is 9 lines smaller now.

PR #1632, opened 2026-08-28
Normal grids: Avoid sqrt for computing norm squared.

spelufo:
Because $|v|^2 = v \cdot v$ , this is equivalent and avoids doing two square roots, which should be faster and more precise. Haven't tried it though, haven't had to generate normal grids yet, I was just skimming the code.

### DIFF
--- a/volume-cartographer/apps/diffusion/support.cpp
+++ b/volume-cartographer/apps/diffusion/support.cpp
@@ -24,7 +24,8 @@ void populate_normal_grid(const SkeletonGraph& g, vc::core::util::GridStore& nor
 
         for (size_t i = 1; i < path.size(); ++i) {
             cv::Point current_point = path[i];
-            double dist_sq = cv::norm(current_point - last_point) * cv::norm(current_point - last_point);
+            auto v = current_point - last_point;
+            double dist_sq = v.ddot(v);
             if (dist_sq >= target_dist_sq) {
                 resampled_path.push_back(current_point);
                 last_point = current_point;
@@ -50,7 +51,8 @@ void populate_normal_grid(const std::vector<std::vector<cv::Point>>& traces, vc:
 
         for (size_t i = 1; i < trace.size(); ++i) {
             cv::Point current_point = trace[i];
-            double dist_sq = cv::norm(current_point - last_point) * cv::norm(current_point - last_point);
+            auto v = current_point - last_point;
+            double dist_sq = v.ddot(v);
             if (dist_sq >= target_dist_sq) {
                 resampled_path.push_back(current_point);
                 last_point = current_point;


Bullo27:
Since this was untested by design ("haven't tried it"), I ran the checks. **The change is correct, and the only cases where it behaves differently from main are cases where main was wrong.**

**Compiles:** built `vc_gen_normalgrids` (which compiles `diffusion/support.cpp`; `vc_diffuse_winding` is the other consumer) at `220c93c88`, GCC 13.3 Release, zero errors.

**Exactness, measured through OpenCV's own `cv::norm`/`ddot`** (standalone probe, all integer offsets `|a|,|b| ≤ 300`, 361,200 cases):

```
checked=361200 new_exact=361200/361200  exact-threshold flips: old_below=91848 old_above=0
with target=400.f (spiral_step 20): disagreements=0
```

* `v.ddot(v)` equals the exact integer squared distance in **every** case (`cv::Point` is integral, so the doubles are exact).
* The old `cv::norm(...) * cv::norm(...)` carries `sqrt`-then-square rounding, and at a threshold sitting exactly on an achievable distance² it lands **strictly below** the true value in 25.4% of cases and **never above**. Concretely: with `target_dist_sq = 26` and an offset of `(1,5)`, main computes `25.999999999999996 >= 26` → false and drops a point that is exactly at `spiral_step` distance; this PR computes `26 >= 26` → true and keeps it, which is what the `>=` means. So "equivalent" is almost right — the honest statement is stronger: equivalent everywhere except boundary coincidences, where the new form is the correct one.
* For a realistic parameter (`spiral_step 20` → `target 400.f`), the two forms agree on all 361,200 offsets, so existing outputs shouldn't shift in practice.

**Completeness:** these two sites are the whole pattern — the only other `norm * norm` products on main (`vc_transform_geom.cpp:140`, OpenABF's angle formula) multiply norms of *different* operands, where the square roots are intrinsic. Nothing else to convert.

I didn't benchmark the speed half of the claim; the two dropped `sqrt`s per path point make it plausible, but the number above is only about correctness. Note `populate_normal_grid` has no test coverage and fork PRs get no CI here, so the compile + probe above is the machine evidence on record.

<sub>Verification by Claude (AI agent) operating on behalf of @Bullo27; all numbers from runs executed today against `220c93c88` and main `61a62c445`, Linux x86-64, system OpenCV 4.x.</sub>

spelufo:
> Verification by Claude (AI agent) operating on behalf of @Bullo27; all numbers from runs executed today against `220c93c88` and main `61a62c445`, Linux x86-64, system OpenCV 4.x.

@Bullo27 don't run your AI on other people's PRs, it is annoying.

Bullo27:
Thanks for the pushback, Claude has indeed been too zealous. Monitoring and comment policy has been reviewed and updated. I genuinely mean to contribute, not to produce noise and waste limited human attention.

PR #1639, opened 2026-08-29
Make lasagna work with the mps torch backend, falling back to single precision.

spelufo:
**In one sentence:** Introduce a floathi tensor type that is float64 on devices that support it and float32 elsewhere. This makes flatten work on macs with the mps backend, which doesn't support float64.

**One real example:** I was able to run `python fit.py configs/flatten_fast_nofilter.json $(pwd)/segment_flatten_input.json --out-dir ../../data/lasagna_flatten_out --device mps` and "flatten" a mesh output by spiral_fit.

**Before:** It failed, complaining about mps not supporting float32.

**After this PR:** It works. Precision will not be the same, but are we sure each of these tensors need to have double precision? The other way to write portable code is to use `torch.set_default_dtype` and not pass an explicit dtype when creating tensors, but that forces you to use the same precision everywhere. This way you can use floathi to ask for a float64 if available, and get a float32 if not.

**Proof:** 
```
...
[optimizer] stage1 'flatten' complete in 2.72s
[optimizer] stage2: params=['map_flatten_ms'] steps=1000 lr=0.0001 min_scaledown=0
[optimizer] stage2: snap_surf_map_weight=0 snap_surf_map_offset=n/a snap_surf_map_offset_mode=n/a
[optimizer] stage2: flatten_edge_step_global_scale=100
[optimizer] stage2: flatten_diagnostics=0
[optimizer] stage2: compile_flatten=1 combined=1
[optimizer] stage2: forward_needs=xyz_lr,data,params+flatten prefetch_grad_channels=[] prefetch_nograd_channels=[]
[optimizer] stage2: flatten_max_update=0.1 (per-scale cap doubles at each coarser level)
[optimizer] stage2: fused_flatten_adam_clamp=1 (Triton CUDA with PyTorch fallback)
[optimizer] progress columns
  col      : meaning                          | col      : meaning                          | col      : meaning
  step     : stage step                       | loss     : total loss                       | it/s     : optimizer it/s
  f_sdir   : flatten_sdir                     | f_sdirW  : weighted flatten_sdir            | f_estep  : physical edge step loss
  f_estpW  : weighted physical edge step loss | f_avg    : flatten_avg_offset               | f_avgW   : weighted flatten_avg_offset
  f_orient : flatten_orient                   | f_oriW   : weighted flatten_orient          | f_step   : flatten current avg grid step
              step     loss  it/s f_sdir f_sdirW f_estep f_estpW  f_avg f_avgW f_orient f_oriW f_step
     stage2 0/1000   0.1344       0.0776  0.0776  0.0567  0.0567 0.0000 0.0000   0.0000 0.0000 20.009
     stage2 1/1000   0.1344 175.8 0.0776  0.0776  0.0567  0.0567 0.0000 0.0000   0.0000 0.0000 20.009
   stage2 100/1000   0.1340 374.4 0.0776  0.0776  0.0564  0.0564 0.0000 0.0000   0.0000 0.0000 20.009
   stage2 200/1000   0.1339 376.1 0.0776  0.0776  0.0563  0.0563 0.0000 0.0000   0.0000 0.0000 20.009
   stage2 300/1000   0.1339 377.8 0.0776  0.0776  0.0563  0.0563 0.0000 0.0000   0.0000 0.0000 20.009
   stage2 400/1000   0.1339 375.8 0.0776  0.0776  0.0563  0.0563 0.0000 0.0000   0.0000 0.0000 20.009
   stage2 500/1000   0.1340 376.2 0.0776  0.0776  0.0564  0.0564 0.0000 0.0000   0.0000 0.0000 20.009
   stage2 600/1000   0.1339 371.1 0.0776  0.0776  0.0563  0.0563 0.0000 0.0000   0.0000 0.0000 20.009
   stage2 700/1000   0.1339 377.2 0.0776  0.0776  0.0563  0.0563 0.0000 0.0000   0.0000 0.0000 20.009
   stage2 800/1000   0.1339 376.2 0.0776  0.0776  0.0563  0.0563 0.0000 0.0000   0.0000 0.0000 20.009
   stage2 900/1000   0.1339 377.9 0.0776  0.0776  0.0563  0.0563 0.0000 0.0000   0.0000 0.0000 20.009
  stage2 1000/1000   0.1340 375.6 0.0776  0.0776  0.0564  0.0564 0.0000 0.0000   0.0000 0.0000 20.009
[optimizer] stage2 'flatten' complete in 2.73s
[optimizer] total optimize time: 18.86s
[fit2tifxyz] area_vx2=1856000
```

**Why / where this is useful:** So people can run lasagna on a mac laptop.

## Details

I searched and replaced `torch.float64` -> `torch.floathi` and `np.float64` -> `np.floathi` in the lasagna directory. Maybe not all uses of np.float64 would have caused failures. I also have no way of knowing how intentional all these uses of torch.float64 are: in some cases, just removing the dtype= parameter would also fix it for mps, by using the default dtype on every device.

I understand if your position is that only cuda or cpu are supported as backends and it isn't worth the trouble to do this and write code this way. Maybe you don't like the monkey patching. But if you want to support mps, this is an option, and all it takes is writing `.floathi` instead of `.float64`.

### DIFF
--- a/lasagna/approval_inpaint.py
+++ b/lasagna/approval_inpaint.py
@@ -8,6 +8,7 @@ from pathlib import Path
 from typing import Iterable
 
 import numpy as np
+from dtypes import numpy_float_hi
 import tifffile
 
 
@@ -167,7 +168,7 @@ def nearest_valid_index(xyz: np.ndarray, valid: np.ndarray, seed: tuple[float, f
 		raise ValueError(f"xyz must have shape (H, W, 3), got {xyz.shape}")
 	if valid.shape != xyz.shape[:2]:
 		raise ValueError(f"valid shape mismatch: {valid.shape} vs {xyz.shape[:2]}")
-	seed_np = np.asarray(seed, dtype=np.floathi)
+	seed_np = np.asarray(seed, dtype=numpy_float_hi)
 	if seed_np.shape != (3,) or not np.isfinite(seed_np).all():
 		raise ValueError(f"approval inpaint seed must contain three finite values, got {seed}")
 
@@ -181,7 +182,7 @@ def nearest_valid_index(xyz: np.ndarray, valid: np.ndarray, seed: tuple[float, f
 		mask = flat_valid[start:end]
 		if not bool(mask.any()):
 			continue
-		coords = flat_xyz[start:end][mask].astype(np.floathi, copy=False)
+		coords = flat_xyz[start:end][mask].astype(numpy_float_hi, copy=False)
 		d2 = np.sum((coords - seed_np.reshape(1, 3)) ** 2, axis=1)
 		local = int(np.argmin(d2))
 		if float(d2[local]) < best_d2:
@@ -431,10 +432,10 @@ def _sample_xyz_at_index_center(
 	fc = max(0.0, min(float(col) - c0, 1.0))
 	corners = [(r0, c0), (r1, c0), (r0, c1), (r1, c1)]
 	if all(bool(valid[r, c]) for r, c in corners):
-		p00 = xyz[r0, c0].astype(np.floathi)
-		p10 = xyz[r1, c0].astype(np.floathi)
-		p01 = xyz[r0, c1].astype(np.floathi)
-		p11 = xyz[r1, c1].astype(np.floathi)
+		p00 = xyz[r0, c0].astype(numpy_float_hi)
+		p10 = xyz[r1, c0].astype(numpy_float_hi)
+		p01 = xyz[r0, c1].astype(numpy_float_hi)
+		p11 = xyz[r1, c1].astype(numpy_float_hi)
 		p = (
 			(1.0 - fr) * (1.0 - fc) * p00
 			+ fr * (1.0 - fc) * p10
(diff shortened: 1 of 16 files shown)


hendrikschilling:
This only works with fit.py at the moment not in other uses. However I also think that float32 is probably fine for all cases. So I would suggest the following:
- add a shared (e.g lasagna/dtypes.py) helper which defines float_hi as float32 by default - with an env trigger var to revert back to float64.
- use that throughout all uses within lasagna not just fit.py
This way we can test float32 as the canonical case and if things break it can easily be switched back on the fly.
And if there are no breakages for some time we'll remove this option and use float32 without fallback.

hendrikschilling:
```
  1. P1: documented prefetch command is broken.
     prefetch_training_cache.py (https://github.com/ScrollPrize/villa/blob/b5231c12b8e2eafe4ec7d65b0a8fe8cd2412546c/lasagna/scripts/prefetch_training_cache.py#L38-L39) imports dtypes      
     before adding the Lasagna directory to sys.path at lines 280–282. Running its documented command fails with ModuleNotFoundError: N module named 'dtypes'. Move the import after the   
     existing path initialization.

  2. P2: the PR adds one deterministic test failure.
     test_preprocess_cos_omezarr.py:1281 (https://github.com/ScrollPrize/villa/blob/b5231c12b8e2eafe8cd2412546c/lasagna/tests/test_preprocess_cos_omezarr.py#L1281-L1285) uses
     numpy_float_hi to test rejection of an unsupported storage dtype. Since the default is now float32, the expected exception is not raised. This fixture should remain explicitly        
     np.float64.

  3. The proposal is implemented almost, but not completely.
     dtypes.py (https://github.com/ScrollPrize/villa/blob/b5231c12b8e2eafe4ec7d65b0a8fe8cd2412546c/lasagna/dtypes.py) correctly defaults both libraries to float32 and restores float64 with
     LASAGNA_MAX_PRECISION_FLOAT=64, without monkey-patching. However, model.py:3158-3159 still unconditionally converts export UV/XYZ arrays to np.float64, so the environment setting does
     not control every float64 use.
```

PR #1644, opened 2026-08-29
loss: standalone MemoryEfficientSoftDiceLoss must consume probabilities, not raw logits

### LINKED ISSUE #1488: MemoryEfficientSoftDiceLoss standalone path defaults apply_nonlin=None: raw logits explode dice to ~1e13 and erase BCE gradients early in training
## Summary

The standalone `MemoryEfficientSoftDiceLoss` path in `_create_loss`
(`src/vesuvius/models/training/loss/losses.py:1146-1154`) defaults
`apply_nonlin` to `None`. With the recommended raw-logits setup
(`activation: none` + `BCEWithLogitsLoss` alongside it), the dice loss
consumes **raw logits as if they were probabilities**. In the early-training
regime where mean logit drifts slightly negative, the dice denominator crosses
zero and the loss explodes to ~1e12–1e13 with sign flips — and because the
exploded term dominates the combined gradient, **BCE supervision is
effectively erased for as long as the regime lasts** (1–5 epochs in our runs).

## How we hit it

Training surface models on Dataset059 (wishlist #191), config
`losses: [BCEWithLogitsLoss (w=1), MemoryEfficientSoftDiceLoss (w=1)]`,
3D, deep supervision on. **All 6 of 6 process starts** opened with progress-bar
loss values around 1e12–1e13, e.g. from our logs:

```
Epoch 1/1000:   0%| | 0/250 [00:01<?, ?it/s, surface: 6033194352640.0000]
```

In our 4-run ablation, 2 of 4 runs never escaped the bad regime's basin
(train soft-Dice flat for 50 epochs); the runs that recovered took 1–5 epochs.

## Minimal reproduction (CPU, shapes (2,1,128³), 15% positive labels)

Sweeping the mean of `logits ~ N(mu, 0.3)` through the actual class:

```
mu=+0.10: dice = -1.20e-01
mu= 0.00: dice = +1.10e-03
mu=-0.10: dice = +5.98e-01
mu=-0.15: dice = +4.75e+12   <- denominator crosses zero
mu=-0.16: dice = +1.01e+13
mu=-1.00: dice = +6.30e+13
```

Through `DeepSupervisionWrapper` (6 scales): **6.94e12** — matching the
observed magnitude in real runs to within 15%.

Gradient impact at mu=-0.15: `cos(combined_grad, dice_grad) = 0.981`,
`|grad_bce|/|grad_dice| = 0.26` — the exploded dice owns the update direction.

Control: same inputs with `apply_nonlin=torch.sigmoid` → bounded in [-1, 0]
(value -0.215). Repro script available on request.

## Suggested fix

For the standalone registry path, default `apply_nonlin` to `torch.sigmoid`
(matching how the combined `nnUNet_DC_and_BCE_loss` construction always pairs
dice with a nonlinearity), or raise a clear error when `apply_nonlin is None`
and the target activation is `none`/logits. A denominator clamp would also cap
the blast radius but doesn't fix the semantics.

Happy to send either variant as a PR.

---

rodriguescarson:
Fixes #1488.

`_create_loss`'s standalone `MemoryEfficientSoftDiceLoss` branch defaulted
`apply_nonlin` to `None`, so under the documented raw-logits setup
(`activation: none` alongside `BCEWithLogitsLoss`) the dice loss consumed logits
as if they were probabilities. `DC_and_BCE_loss` already passes
`apply_nonlin=torch.sigmoid` when it builds the same class
(`nnunet_losses.py:331`) — the standalone path just did not.

A config could not correct it either. The value arrives from YAML as a string,
so passing it straight through could only ever yield `None` or something that
raises when called; there was no way to ask for sigmoid from a config file.

**Reproduced independently** of the issue, sweeping the mean of
`logits ~ N(mu, 0.3)` against a 15%-positive target, shape `(2, 1, 16³)`:

| mu | before (no nonlinearity) | after (sigmoid) |
|---|---|---|
| +0.10 | −0.1377 | −0.2357 |
| 0.00 | −0.0277 | −0.2332 |
| −0.10 | +0.5399 | −0.2305 |
| −0.15 | **+1.68e+10** | −0.2291 |
| −0.16 | **+1.80e+10** | −0.2288 |
| −1.00 | **+1.22e+11** | −0.1961 |

The denominator crosses zero between −0.10 and −0.15, exactly as reported. The
magnitudes differ from the issue's ~1e13 because of tensor shape and `smooth`,
but the behaviour is the same, and the sign flip at −0.10 is arguably worse than
the explosion: a dice "loss" of +0.54 is a gradient pointing the wrong way.

## What changed

* the standalone branch now defaults to `sigmoid`, matching `DC_and_BCE_loss`;
* `apply_nonlin` accepts a name from YAML — `sigmoid`, `softmax`, or `none` —
  resolved by a small `_resolve_nonlin` helper, with callables passed through
  unchanged so Python callers are unaffected;
* an unknown name raises at construction with the accepted values, rather than
  failing somewhere inside the forward pass.

## Behaviour change, deliberately

A config that relied on the old default now gets sigmoid applied. That is the
fix — but a model that already emits probabilities should set
`apply_nonlin: none` explicitly, which is now expressible where it was not
before.

## Tests

`vesuvius/tests/models/training/test_dice_standalone_nonlin.py` pins the loss
bounded across the whole logit range the issue reports, the default matching an
explicit `sigmoid`, `none` working for pre-activated inputs, name resolution in
both directions, and an unknown name being rejected at construction.

### DIFF
--- a/vesuvius/src/vesuvius/models/training/loss/losses.py
+++ b/vesuvius/src/vesuvius/models/training/loss/losses.py
@@ -991,6 +991,33 @@ class NormalGatedRepulsionLoss(nn.Module):
 
 #######################################################################################################################
 
+#: nonlinearities a config may name for a loss that needs probabilities
+_NONLINEARITIES = {
+    'sigmoid': torch.sigmoid,
+    'softmax': lambda x: torch.softmax(x, dim=1),
+    'none': None,
+    'null': None,
+}
+
+
+def _resolve_nonlin(value):
+    """Turn a config's ``apply_nonlin`` into a callable.
+
+    Configs are YAML, so this arrives as a string; a callable is accepted too so
+    that Python callers are unaffected.  ``none`` turns the nonlinearity off,
+    which is what a model that already emits probabilities wants.
+    """
+    if value is None or callable(value):
+        return value
+    key = str(value).strip().lower()
+    if key not in _NONLINEARITIES:
+        raise ValueError(
+            f"unknown apply_nonlin {value!r}; expected one of "
+            f"{sorted(k for k in _NONLINEARITIES)}"
+        )
+    return _NONLINEARITIES[key]
+
+
 def _create_loss(name, loss_config, weight, ignore_index, pos_weight, mgr=None):
     # Define losses that don't natively support ignore_index
     losses_without_ignore_support = ['BCEWithLogitsLoss', 'MSELoss', 'SmoothL1Loss', 'L1Loss', 'WeightedSmoothL1Loss']
@@ -1144,9 +1171,22 @@ def _create_loss(name, loss_config, weight, ignore_index, pos_weight, mgr=None):
         )
     
     elif name == 'MemoryEfficientSoftDiceLoss':
-        # Standalone memory efficient dice loss
+        # Standalone memory efficient dice loss.
+        #
+        # Dice needs probabilities.  DC_and_BCE_loss passes
+        # apply_nonlin=torch.sigmoid for exactly this reason, but the standalone
+        # path defaulted to None, so under the documented raw-logits setup
+        # (activation: none alongside BCEWithLogitsLoss) it consumed logits as
+        # if they were probabilities.  Once the mean logit drifts negative the
+        # dice denominator crosses zero and the loss explodes to ~1e13, which
+        # dominates the combined gradient and erases BCE supervision for as long
+        # as the regime lasts.
+        #
+        # A YAML config could not fix it either: the value arrives as a string,
+        # so passing it straight through could only ever yield None or something
+        # that raises when called.
         base_loss = MemoryEfficientSoftDiceLoss(
-            apply_nonlin=loss_config.get('apply_nonlin', None),
+            apply_nonlin=_resolve_nonlin(loss_config.get('apply_nonlin', 'sigmoid')),
             batch_dice=loss_config.get('batch_dice', False),
             do_bg=loss_config.get('do_bg', True),
             smooth=loss_config.get('smooth', 1.),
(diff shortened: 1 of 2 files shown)


rodriguescarson:
Before/after, reproduced just now against `main` at `2c63fac`, using the same
setup the regression test uses (`torch.randn * 0.3 + mu`, 15% positive target,
`do_bg=True`, `smooth=1e-5`):

**On `main`** — raw logits into the standalone soft dice:

```
     mu             loss   in [-1, 1]?
   0.10      -1.3688e-01   yes
   0.00      -2.6048e-02   yes
  -0.10       5.4778e-01   yes
  -0.15       1.6897e+10   NO
  -0.16       1.8129e+10   NO
  -1.00       1.2162e+11   NO
```

Two separate problems visible in one table. The loss crosses **1.7e+10 between
mu = -0.10 and mu = -0.15**, which dominates any combined objective and erases
BCE supervision for as long as the regime lasts. And at mu >= 0 it is *negative* —
the denominator has gone the wrong side of zero, so the gradient points away from
the target rather than toward it.

**On this branch**, the same file:

```
..........                                                               [100%]
10 passed in 1.58s
```

Reproduce the `main` column with:

```python
import sys; sys.path.insert(0, "src")
import torch
from vesuvius.models.training.loss.losses import MemoryEfficientSoftDiceLoss

def logits(mu, shape=(2, 1, 16, 16, 16), seed=0):
    g = torch.Generator().manual_seed(seed)
    return torch.randn(shape, generator=g) * 0.3 + mu

g = torch.Generator().manual_seed(1)
target = (torch.rand((2, 1, 16, 16, 16), generator=g) < 0.15).float()
loss = MemoryEfficientSoftDiceLoss(batch_dice=False, do_bg=True, smooth=1e-5)
for mu in (0.1, 0.0, -0.1, -0.15, -0.16, -1.0):
    print(mu, float(loss(logits(mu), target)))
```

Note the test file imports `_resolve_nonlin`, which this PR introduces, so on
`main` it fails at collection rather than at an assertion — the behavioural claim
is the table above, which uses only what `main` already exports.

PR #1647, opened 2026-08-29
VC3D: make fetched Open Data segments editable from agent bridge

### LINKED ISSUE #1464: VC3D: renaming a fetched Open Data segment mutates the immutable catalog cache — the copy vanishes on restart, the original silently re-downloads, in-cache derived files are stranded
## Motivation / how this arose

Driving published-segment workflows on PHercParis4 headlessly over the agent bridge (fetch a published segment, generate a mask, recalculate area), I renamed the working segment to mark it — the same "Rename Surface" operation the surface panel offers — and on the next session the segment was gone.

## Summary

The Open Data segment cache is identity-keyed by directory name, and the rest of the app treats it as immutable — the segmentation-editing gate prompts *"This open-data segment is an immutable catalog cache. Create or choose an editable copy before enabling editing."* (`apps/VC3D/segmentation/SegmentationModule.cpp:134`, via `vc3d::opendata::isOpenDataCatalogSegmentDirectory`).

`segments.rename` (and the GUI's Rename Surface, which shares the same backend) never consults that check. Renaming a fetched catalog segment physically renames the cache directory and rewrites `meta.json`'s `uuid`. From then on:

1. **The renamed copy vanishes from the app on the next catalog open** — open-data projects enumerate segments from the catalog manifest by expected directory name, so the stray is not listed, and no RPC or GUI action can reach it anymore.
2. **The original id is re-listed as unmaterialized and silently re-downloads in full** on the next fetch (~57 MB for the segment below), duplicating it on disk.
3. **Anything the app wrote into the cached copy is stranded in the invisible stray.** In my run that was the `mask.tif` that `segment.generate_mask` had written — note that mask generation also writes straight into the "immutable" cache directory.

Recovery requires manual filesystem intervention outside the app (finding the stray under `~/.VC3D/remote_cache/…` and renaming it back by hand).

## Reproduction (agent bridge, main @ aab644c9c, `QT_QPA_PLATFORM=offscreen`)

```
catalog.open_sample {"sampleId":"PHercParis4","resources":{"kinds":[]}}   → succeeded
segments.list                                                             → 81 entries
segments.fetch    {"segmentId":"20230702185753-published-20260411134726-L0-96a63b37f78c8136"} → fetched
segments.activate {…same id…}                                             → activated
segment.generate_mask {…same id…}       → mask.tif written into the cache dir
segments.rename   {"segmentId":"…-96a63b37f78c8136","newName":"probe_renamed"}
                                        → {"newId":"probe_renamed","oldId":"…-96a63b37f78c8136"}
```

Restart VC3D, open the sample again:

```
segments.list → 81 entries; "probe_renamed" is NOT among them;
                the original id is listed again with loaded=false
segments.fetch {"segmentId":"…-96a63b37f78c8136"}
              → {"alreadyMaterialized": false, …} and the job re-downloads the segment
```

On disk afterwards (both 57 MB):

```
published-L0/probe_renamed/                    ← stray: x/y/z.tif + mask.tif (the derived work)
                                                 + catalog-origin.json (cache_state still "current")
published-L0/20230702185753-published-…8136/   ← fresh re-download, no mask.tif
```

Everything above is from actual runs; the one code-read (not clicked) inference is flagged below.

## Root cause

- `AgentBridgeServer::handleSegmentsRename` (`apps/VC3D/agent_bridge/AgentBridgeHandlers_session.cpp:1108`) validates the new name, refuses during editing, and maps unknown ids — but has no catalog-provenance check — then calls `SegmentationCommandHandler::renameSurfaceHeadless` (`SegmentationCommandHandler.cpp:5584`), which renames the on-disk directory. The GUI context-menu rename (`SurfacePanelController.cpp:1227` → `onRenameSurface`, `SegmentationCommandHandler.cpp:5525`) uses the same backend; its only gate is that the surface is loaded in the current folder, so it should be reachable for a fetched catalog segment too (this step is read from the code, not clicked — the bridge path is measured).
- The cache stores identity *inside* each directory (`catalog-origin.json`; `meta.json` keeps `vc_open_data_catalog_segment_lineage_id`), but reconciliation matches by directory name only, so the stray never rejoins.
- There is even orphan-marking machinery (`markOrphanedEntries`, `OpenDataSegmentCache.cpp:1367`) meant to stamp stray catalog dirs `cache_state: "orphaned"`, but for published representations it cannot fire: it scans the source-volume level (`openDataSegmentCacheRoot`, call at `:2212`) whose children are representation folders (`published-L0`, `authored-L*`), while published segment dirs live one level deeper and are keyed by `representationId` — and `expectedDirNames` only ever receives bare segment ids from the DerivedNative branch (`:2168`). Measured accordingly: the stray's `cache_state` is still `"current"` after repeated catalog opens. `Orphaned` also has no consumer outside `OpenDataSegmentCache.cpp`.

## Suggested fix (not build-verified)

Minimal and consistent with the existing design: refuse the rename when `isOpenDataCatalogSegmentDirectory(surface->path)` (`OpenDataSegmentCache.cpp:2006`) — one guard in `renameSurfaceHeadless` covers both GUI and bridge, with the same message the editing gate already uses ("immutable catalog cache; create an editable copy"). The same question probably applies to other in-cache writers (`segment.generate_mask` above); `segments.delete` seems fine as-is, since a deleted cache copy just re-fetches.

Deeper options, if wanted: reconcile strays by `catalog-origin.json` instead of directory name, and/or point `markOrphanedEntries` at the representation roots with the right name set and surface the `orphaned` state somewhere. Happy to build-verify and PR the guard if that's useful.

---
*Found while driving published-segment mask/area workflows on real PHercParis4 data through the agent bridge. Analysis and write-up done with AI assistance (Claude); all transcripts and on-disk states above are from actual runs on main @ aab644c9c.*

### LINKED ISSUE #1465: Agent bridge: segmentation.enable_editing on a fetched Open Data segment hangs forever — the immutable-cache prompt is a modal dialog no RPC can answer
## Motivation / how this arose

Running the segmentation editing loop headlessly on real PHercParis4 data over the agent bridge. The loop works end-to-end for locally grown patches (transcript below); trying to refine a *fetched published segment* the same way hangs the call forever.

## Summary

With a catalog-fetched segment active, `segmentation.enable_editing {"enabled": true}` never returns. The GUI design is sound — catalog cache segments are immutable, so enabling editing pops *"This open-data segment is an immutable catalog cache. Create or choose an editable copy before enabling editing."* — but over the bridge that prompt is a modal `QMessageBox::exec()` that nothing can answer (in `QT_QPA_PLATFORM=offscreen` it isn't even visible). There is also no bridge RPC that performs the editable-copy flow the dialog offers. Net effect: **editing catalog segments through the bridge is impossible — the one entry point blocks forever.**

## Reproduction (main @ aab644c9c, offscreen)

```
catalog.open_sample {"sampleId":"PHercParis4","resources":{"kinds":[]}}  → succeeded
segments.fetch     {"segmentId":"20230702185753-published-20260411134726-L0-96a63b37f78c8136"}
segments.activate  {…same id…}                                           → activated
segmentation.enable_editing {"enabled": true}                            → NO RESPONSE (waited 45 s; never returns)
```

Measured while the call hangs:

- A **second connection** to the same bridge socket works: `ping` answers (pid/version), `screenshot.capture` returns a frame — the modal's nested event loop keeps servicing the socket, so the bridge looks alive while this RPC is stuck.
- The VC3D log shows the dialog being created: `This plugin does not support propagateSizeHints()` (the offscreen platform's response to showing the prompt window).
- There is no RPC to answer the dialog, and `rpc.describe` lists no editable-copy method under `segments.*`/`segmentation.*`, so there is no way to unblock or route around it.

## Contrast: the same loop on a locally grown segment works (same session pattern, measured)

```
segmentation.grow_patch_from_seed {"seed":{"x":1408,"y":1408,"z":1024},"volumeId":"20260310170716-45.532um-…","iterations":30}
   → job succeeded (grew ~9.6 cm² patch, saved under the project's patches/ dir)
segments.activate {"segmentId":"auto_grown_20260815213206017"}   → activated
segmentation.enable_editing {"enabled":true}                     → {"enabled": true}   (returns in 0.0 s)
segmentation.grow {"direction":"all","steps":1}                  → job succeeded
segmentation.save                                                → "Segment saved"
segmentation.enable_editing {"enabled":false}                    → {"enabled": false}
```

So the hang is specific to the catalog-cache path.

## Mechanism

`handleSegmentationEnableEditing` (`apps/VC3D/agent_bridge/AgentBridgeHandlers_editing.cpp:350`) deliberately drives `SegmentationModule::setEditingEnabled()` so a real edit session is established. For a catalog segment, that path hits the immutable-cache gate (`apps/VC3D/segmentation/SegmentationModule.cpp:~130`, via `vc3d::opendata::isOpenDataCatalogSegmentDirectory`) and calls `prompt.exec()` — a modal dialog with "Create Editable Copy" / "Choose Existing Folder" / Cancel. Interactively that's fine; from the bridge the RPC blocks inside `exec()` with no way to answer.

## Suggested direction (their call, of course)

- Minimal: in the headless/bridge-driven path, return a structured error instead of prompting — e.g. `-32009` with `detail: "catalog segment is an immutable cache; create an editable copy first"` — so agents fail fast with the reason.
- More useful: expose the editable-copy flow as an RPC (e.g. `segments.create_editable_copy {segmentId, path?}` mirroring what the dialog's default button does), and have the error message name it. That would make published-segment refinement — a natural agent workflow — actually reachable headlessly.

Related: #1464 is the mutation side of the same immutability boundary (rename/mask-write bypass the protection; here the protection blocks the bridge).

---
*Found while driving real PHercParis4 published-segment workflows through the agent bridge. Analysis and write-up done with AI assistance (Claude); every behavior above is from actual runs (nothing inferred from code except the cited mechanism, which matches the observed hang and log line).*

TAUIL-Abd-Elilah:
**In one sentence:** An agent can now turn a fetched, immutable Open Data segment into an active editable copy without hanging on an invisible modal or accidentally renaming the catalog cache.

**One real example:** In the PHercParis4 workflow documented in #1465, `segmentation.enable_editing` on segment `20230702185753-published-20260411134726-L0-96a63b37f78c8136` still had no response after 45 seconds because VC3D opened a modal dialog. With this change, that request fails immediately with a structured `-32009` response naming `segments.create_editable_copy`; the new method copies, persists, selects, and activates the editable segment, after which editing succeeds.

**Before:** The bridge entered `QMessageBox::exec()` while handling `segmentation.enable_editing`, so the original RPC never completed and no bridge method could perform the dialog's editable-copy action. Independently, `segments.rename` could physically rename the immutable cache directory, strand derived work, and force a full re-download (#1464).

**After this PR:** Catalog editing fails fast with an actionable error. The `vc3d_create_editable_segment_copy` MCP tool (backed by `segments.create_editable_copy {"segmentId":"..."}`) uses the existing Open Data copy implementation, strips catalog provenance, copies coordinate identity, persists and selects the editable segment root, refreshes the UI, and activates the copy. That selection survives an immediate retry and a project reopen. The shared rename core rejects catalog segments in both the bridge and GUI paths, while canonical-path guards prevent the GUI folder picker from treating the original cache, a symlink alias, or another registered catalog root as an editable destination.

**Proof:** The offscreen bridge regression uses a shipped TIFXYZ segment fixture and verifies all of the following in one workflow:

```text
[PASS] catalog_segment_rename_guard: code=-32009 action=segments.create_editable_copy
[PASS] catalog_segment_editing_fails_fast: code=-32009 action=segments.create_editable_copy elapsed=0.000s
[PASS] catalog_segment_create_editable_copy: created=True activated=True; exact editable path selected
[PASS] catalog_segment_create_editable_copy_retry: created=False alreadyExisted=True; no duplicate entry
[PASS] editable_copy_can_enable_editing: enabled=True
[PASS] editable_copy_survives_reopen: editable path retained; editing enabled again
[PASS] described_invalid_inputs: 464 probes passed
[PASS] process_alive: running=True
```

I also replayed the final binary against the public PHercParis4 catalog and its
materialized segment
`20230702185753-published-20260411134726-L0-96a63b37f78c8136`:

```text
catalog.list_samples: 45 samples, PHercParis4 present
catalog edit: -32009 + segments.create_editable_copy in 0.002 s
editable copy: exact _editable path selected; editing enabled in 1.162 s
immediate retry: created=False, alreadyExisted=True
project reopen: editable path retained; editing enabled again
catalog x/y/z SHA-256 before == after; catalog-origin.json retained only at source
```

The native regression also rejects the source itself, a symlink alias, a path
inside the catalog root, and a different registered catalog segment as copy
destinations.

Focused native tests also pass:

```text
test_open_data_manifest .......... Passed
agent_bridge_contract ............ Passed
100% tests passed, 0 failed
```

The public MCP contract and wrapper regression suite passes 55/55 tests,
including schema parity, protocol documentation, and the editable-copy request.
The fresh CI matrix for `6fa5f8ace` is green across Linux VC3D/smoke, Linux
base and GUI tests, rendering, MCP, Windows, macOS, and all CodeQL shards. The
only red status is Vercel preview authorization for an external contributor;
this PR does not change the web application.

**Why / where this is useful:** Published Open Data segments are the natural starting point for review and refinement. This closes the last headless gap between fetching one and editing a safe working copy, while protecting the cache identity that makes later catalog opens and downloads reliable.

- [x] I personally verified that the example and proof above were produced by this PR on the stated repository fixture.

## Details

Tested commit: `6fa5f8ace`

Commands:

```bash
cmake --build build/remote-timeout-ci --target VC3D test_agent_bridge_contract test_open_data_manifest -j2
ctest --test-dir build/remote-timeout-ci --output-on-failure -R '^(agent_bridge_contract|test_open_data_manifest)$'
QT_QPA_PLATFORM=offscreen python3 apps/VC3D/agent_bridge/test/smoke_offscreen.py --vc3d build/remote-timeout-ci/bin/VC3D
cd tools/vc3d-mcp && python -m unittest -v tests.test_contract tests.test_bridge tests.test_progress.ToolSchemaTest
```

Fixes #1464.
Fixes #1465.

AI disclosure: investigation, implementation, and test drafting were assisted by OpenAI Codex; the contributor reviewed the diff and the reported results come from commands run against this commit.

### DIFF
--- a/volume-cartographer/apps/VC3D/CWindow.cpp
+++ b/volume-cartographer/apps/VC3D/CWindow.cpp
@@ -702,6 +702,25 @@ bool isAvailableOpenDataSegmentsEntry(const VolumePkg& pkg,
         path.string()).empty();
 }
 
+bool isSelectedEditableOpenDataSegmentsEntry(
+    const VolumePkg& pkg,
+    const vc::project::Entry& entry)
+{
+    if (!vc::project::hasEntryTag(entry, "open-data-editable") ||
+        vc::project::isLocationRemote(entry.location)) {
+        return false;
+    }
+    const auto selectedPath = pkg.outputSegmentsPath();
+    if (selectedPath.empty()) {
+        return false;
+    }
+    const auto entryPath = vc::project::resolveLocalPath(
+        entry.location, pkg.path().parent_path());
+    std::error_code error;
+    return std::filesystem::equivalent(selectedPath, entryPath, error) &&
+           !error;
+}
+
 std::vector<QString> openDataCatalogVolumeIdCandidates(const VolumePkg& pkg,
                                                        const std::string& loadedVolumeId)
 {
@@ -736,6 +755,7 @@ const vc::project::Entry* findOpenDataSegmentsEntryForVolume(const VolumePkg& pk
 
     const std::string targetTag = "vc-open-data-target-volume-id:" + catalogVolumeId.toStdString();
     const std::string sourceTag = "vc-open-data-source-volume-id:" + catalogVolumeId.toStdString();
+    const vc::project::Entry* targetMatch = nullptr;
     const vc::project::Entry* sourceMatch = nullptr;
 
     for (const auto& entry : pkg.segmentEntries()) {
@@ -743,15 +763,24 @@ const vc::project::Entry* findOpenDataSegmentsEntryForVolume(const VolumePkg& pk
             continue;
         }
         if (std::find(entry.tags.begin(), entry.tags.end(), targetTag) != entry.tags.end()) {
-            return &entry;
+            if (isSelectedEditableOpenDataSegmentsEntry(pkg, entry)) {
+                return &entry;
+            }
+            if (!targetMatch) {
+                targetMatch = &entry;
+            }
+            continue;
         }
-        if (!sourceMatch &&
-            std::find(entry.tags.begin(), entry.tags.end(), sourceTag) != entry.tags.end()) {
-            sourceMatch = &entry;
+        if (std::find(entry.tags.begin(), entry.tags.end(), sourceTag) !=
+            entry.tags.end()) {
+            if (!sourceMatch ||
+                isSelectedEditableOpenDataSegmentsEntry(pkg, entry)) {
+                sourceMatch = &entry;
+            }
         }
     }
 
-    return sourceMatch;
+    return targetMatch ? targetMatch : sourceMatch;
 }
 
 const vc::project::Entry* findOpenDataSegmentsEntryForLoadedVolume(const VolumePkg& pkg,
@@ -763,16 +792,26 @@ const vc::project::Entry* findOpenDataSegmentsEntryForLoadedVolume(const VolumeP
     if (!coordinateSpace.empty()) {
         const std::string coordinateTag =
             "vc-open-data-coordinate-space:" + coordinateSpace;
+        const vc::project::Entry* coordinateMatch = nullptr;
         for (const auto& entry : pkg.segmentEntries()) {
             if (isAvailableOpenDataSegmentsEntry(pkg, entry) &&
                 std::find(entry.tags.begin(), entry.tags.end(), coordinateTag) !=
                     entry.tags.end()) {
-                if (matchedCatalogVolumeId) {
-                    *matchedCatalogVolumeId =
-                        openDataCatalogVolumeIdForLoadedVolume(pkg, loadedVolumeId);
+                if (isSelectedEditableOpenDataSegmentsEntry(pkg, entry)) {
+                    coordinateMatch = &entry;
+                    break;
                 }
-                return &entry;
+                if (!coordinateMatch) {
+                    coordinateMatch = &entry;
+                }
             }
+        }
+        if (coordinateMatch && matchedCatalogVolumeId) {
+            *matchedCatalogVolumeId =
+                openDataCatalogVolumeIdForLoadedVolume(pkg, loadedVolumeId);
+        }
+        if (coordinateMatch) {
+            return coordinateMatch;
         }
         // Explicitly identified assets must never fall back to lineage-only
         // association, because native and virtual views share that lineage.
--- a/volume-cartographer/apps/VC3D/OpenDataSegmentCache.cpp
+++ b/volume-cartographer/apps/VC3D/OpenDataSegmentCache.cpp
@@ -2052,6 +2052,95 @@ std::vector<OpenDataInkDetectionEntry> cachedInkDetectionsForSegmentDirectory(
     return out;
 }
 
+namespace {
+
+struct RegisteredCatalogEntry {
+    const vc::project::Entry* entry = nullptr;
+    std::filesystem::path path;
+    std::size_t depth = 0;
+};
+
+bool pathContains(const std::filesystem::path& root,
+                  const std::filesystem::path& candidate)
+{
+    auto candidatePart = candidate.begin();
+    for (auto rootPart = root.begin(); rootPart != root.end();
+         ++rootPart, ++candidatePart) {
+        if (candidatePart == candidate.end() || *candidatePart != *rootPart) {
+            return false;
+        }
+    }
+    return true;
+}
+
+std::optional<RegisteredCatalogEntry> findRegisteredOpenDataCatalogEntry(
+    const VolumePkg& pkg,
+    const std::filesystem::path& catalogSegmentDir,
+    bool preferDeepest)
+{
+    std::error_code canonicalError;
+    const auto canonicalSegment =
+        std::filesystem::weakly_canonical(catalogSegmentDir, canonicalError);
+    if (canonicalError) {
+        return std::nullopt;
+    }
+
+    std::optional<RegisteredCatalogEntry> match;
+    for (const auto& entry : pkg.segmentEntries()) {
+        if (vc::project::isLocationRemote(entry.location) ||
+            !vc::project::hasEntryTag(entry, "open-data") ||
+            !vc::project::hasEntryTag(entry, "immutable")) {
+            continue;
+        }
+        std::error_code error;
+        const auto entryPath = std::filesystem::weakly_canonical(
+            vc::project::resolveLocalPath(
+                entry.location, pkg.path().parent_path()),
+            error);
+        if (error || !pathContains(entryPath, canonicalSegment)) {
+            continue;
+        }
+        const auto depth = static_cast<std::size_t>(
+            std::distance(entryPath.begin(), entryPath.end()));
+        if (!match || (preferDeepest ? depth > match->depth
+                                    : depth < match->depth)) {
+            match = RegisteredCatalogEntry{&entry, entryPath, depth};
+        }
+    }
+    return match;
+}
+
+std::filesystem::path canonicalPathOrThrow(
+    const std::filesystem::path& path,
+    const char* label)
+{
+    std::error_code error;
+    const auto absolute = std::filesystem::absolute(path, error);
+    if (error) {
+        throw std::runtime_error(
+            std::string("could not resolve ") + label + ": " +
+            error.message());
+    }
+    const auto canonical = std::filesystem::weakly_canonical(absolute, error);
+    if (error) {
+        throw std::runtime_error(
+            std::string("could not resolve ") + label + ": " +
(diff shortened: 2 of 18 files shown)


TAUIL-Abd-Elilah:
@Bullo27, this PR now implements the workflow proposed in #1465 and the cache-mutation guard from #1464. I replayed commit `6fa5f8ace` against the same public PHercParis4 segment (`20230702185753-published-20260411134726-L0-96a63b37f78c8136`):

- editing the catalog source returns structured `-32009` with `segments.create_editable_copy` in 0.002 s;
- `segments.create_editable_copy` selects the exact `_editable` path and editing succeeds;
- an immediate retry is idempotent (`created=false`, `alreadyExisted=true`);
- reopening the project retains the editable source and editing succeeds again;
- SHA-256 hashes of catalog `x.tif`, `y.tif`, and `z.tif` are unchanged;
- native regressions reject the catalog source, symlink aliases, paths inside any registered immutable catalog root, and a different catalog segment as editable destinations.

Would you be willing to verify the flow independently on your setup? The shortest check after fetch/activate is:

```text
segments.create_editable_copy {"segmentId":"20230702185753-published-20260411134726-L0-96a63b37f78c8136"}
segmentation.enable_editing {"enabled":true}
```

Then reopen the same project and confirm `segments.list` still reports the `_editable` path for that ID. Any mismatch would be useful to know before merge.

Bullo27:
**Verified independently, and it holds.** Linux, offscreen agent bridge, `6fa5f8ace` merged onto today's main, against the real fetched PHercParis4 segment `…-96a63b37f78c8136`:

- `segmentation.enable_editing` on the catalog segment fails in **0.009 s** with `-32009 → segments.create_editable_copy` — the #1465 modal hang is gone.
- `segments.rename` on it is refused with `-32009`; #1464's cache mutation is blocked at the bridge.
- `create_editable_copy` creates + activates; retry is idempotent (`created=false, alreadyExisted=true`, same path); the `_editable` path survives a project reopen in `segments.list`.
- Catalog `x/y/z.tif` SHA-256 unchanged across the whole flow; the copy lands outside the catalog root, drops `catalog-origin.json`, and keeps `vc_open_data_coordinate_space`.

One caveat, in case it matters: my run had no current volume set (`currentVolumeId` null), so the coordinate-identity copy read from the segment metadata rather than a live-volume id. That path still produced correct identity, but I didn't exercise it with a loaded volume.

Thanks for wiring these two together.

<sub>— verification automated with Claude (AI), run on my setup.</sub>

PR #1653, opened 2026-08-30
vesuvius: fix list_files() on a core install (fixes #1329)

### LINKED ISSUE #1329: list_files() fails with "NoneType object is not callable" on a volume-only install: nest_asyncio is missing from that extra

Oishi1029:
**In one sentence:** `vesuvius.list_files()` works on a core install again, instead of failing with `TypeError: 'NoneType' object is not callable`.

**Motivation:** I wanted to start contributing to the challenge, so I installed the `vesuvius` library following the install steps in CONTRIBUTING.md. Streaming the scan data worked, but `list_files()` didn't. If I'd hit this on my own I would have thought I installed it wrong. I think it's worth fixing — it would save the next person time.

I used an AI coding assistant to find the cause and write the patch. I ran the before/after myself on my own machine.

**One real example:** Starting with a fresh `uv sync` in `vesuvius/` (the install path CONTRIBUTING.md documents), I called the `list_files()` helper the README advertises under "Data listing", and it produced `TypeError: 'NoneType' object is not callable` instead of the catalog.

**Before:**

```
$ uv sync                       # in vesuvius/, per CONTRIBUTING.md
$ python -c "import vesuvius; print(vesuvius.list_files); vesuvius.list_files()"
vesuvius.list_files is None

$ python -m pytest tests/test_imports.py -q
ERROR tests/test_imports.py
!!!!!!! Interrupted: 1 error during collection !!!!!!!
E   ModuleNotFoundError: No module named 'nest_asyncio'
1 error in 0.26s
```

The traceback points at the call site, so nothing indicates the real cause. Seven of the twelve names in `vesuvius.__all__` are silently `None` in this environment: `VCDataset`, `utils`, `tifxyz`, `is_aws_ec2_instance`, `list_cubes`, `list_files`, `update_list`.

**After this PR:**

```
$ python -c "import vesuvius; print(vesuvius.list_files); print(sorted(vesuvius.list_files()))"
<function list_files at 0x10c7e75e0>
['1', '2', '3', '4', '5']

$ python -m pytest tests/test_imports.py -q
5 failed, 8 passed in 1.86s
```

**Proof:** the two blocks above are the same two commands in the same environment, with `catalog.py` at its `main` version and then at this PR's version, on base commit `5479453a7`, Python 3.14.6, macOS arm64, with `nest_asyncio` absent.

The 5 remaining failures are all `test_cli_entrypoints_show_help`, which subprocess-import `torch` and therefore need the `models` extra. They are unrelated to this change and are not introduced by it — **unmodified `main` with `nest_asyncio` installed gives the identical `5 failed, 3 passed`**, so this change reproduces main's behaviour exactly while removing the dependency requirement. (3 passed before this PR's tests; 8 pass with them.)

**Why / where this is useful:** `list_files()` is how someone discovers what data exists, so this is the first thing a new user hits, and the error it produced named neither the missing package nor the extra that carries it. The fix adds no dependency, so the minimal install stays minimal.

## Details

**Cause.** `vesuvius/utils/catalog.py` imported `aiohttp` and `nest_asyncio` at module scope. Both ship only in the `models` / `notebooks` / `blending` / `label-transfer` / `all` extras, not in the core dependency set or `volume-only`. The import therefore raised on a core install, `vesuvius/__init__.py` caught it and rebound the public names to `None` while `__all__` kept exporting them, and the failure surfaced only at the call.

`nest_asyncio` is the only module actually missing here — `aiohttp` arrives transitively via `s3fs` — but neither is needed to read the catalog.

**Change.** Import both where they are used, and raise an error naming the package:

- `aiohttp` and `lxml` → checked by a shared `_load_parser()` helper **before** `vesuvius.data.paths.parser` is imported, because the parser imports both at module scope; used by `scrape_website()` and `collect_subfolders()`, the only two functions that need them. (Ordering gap caught in review by @hendrikschilling.)
- `nest_asyncio` → inside `update_list()`, at the single call site, which runs only when an event loop is already running.

`list_files()` and `list_cubes()` only read a packaged YAML file and now need neither. This follows the lazy `from vesuvius.data.paths import parser` import already used in both of those same functions.

**Tests.** Two regression tests in `tests/test_imports.py`. Both simulate the packages being absent via a `sys.meta_path` finder, so they still catch a regression in a CI environment where the extras *are* installed:

- `test_catalog_imports_without_refresh_only_dependencies` — loads `catalog.py` as an isolated module with `aiohttp` and `nest_asyncio` blocked, and checks `list_files()` works.
- `test_refresh_path_names_the_missing_package` — calls the public `scrape_website()` and `collect_subfolders()` with `aiohttp` or `lxml` blocked, evicting any cached parser so it sees the real import order, and checks the error names the package. It raises before any request, so no network is touched.

Run against three versions of `catalog.py`:

| `catalog.py` version | `lxml` absent (core install) | `lxml` installed (CI-like) |
|---|---|---|
| unmodified `main` | cannot collect (needs `nest_asyncio`) | 5 failed (with `nest_asyncio` added) |
| previous head of this PR (`6465565`) | 4 failed, 1 passed | 4 failed, 1 passed |
| this PR | 5 passed | 5 passed |

**Limitations.** Refreshing the catalog from the data server (`update_list()` → `scrape_website()`) still needs the extras, which is unchanged and intended; it now fails with a message naming the missing package (`aiohttp`, `lxml` or `nest_asyncio`) instead of a bare import error.

Closes #1329

---

Developed with AI coding assistance. Every claim in this PR is reproduced by the commands shown above and by the two included regression tests; the before/after output is generated by running them, not asserted.

### DIFF
--- a/vesuvius/src/vesuvius/utils/catalog.py
+++ b/vesuvius/src/vesuvius/utils/catalog.py
@@ -3,12 +3,12 @@
 from __future__ import annotations
 
 import asyncio
+import importlib
 import ssl
 from pathlib import Path
+from types import ModuleType
 from typing import Dict, List, Optional, Tuple
 
-import aiohttp
-import nest_asyncio
 import requests
 import yaml
 
@@ -17,6 +17,24 @@ from vesuvius.install.accept_terms import get_installation_path
 
 CatalogTree = Dict[str, Optional[Dict]]
 CatalogListing = Dict[str, str]
+
+
+def _require(module_name: str, used_for: str) -> ModuleType:
+    """Import a package only needed for refreshing the catalog from the data server.
+
+    Reading the packaged catalog (:func:`list_files`, :func:`list_cubes`) needs neither
+    of these, so they are imported at the point of use rather than at module import.
+    """
+
+    try:
+        return importlib.import_module(module_name)
+    except ModuleNotFoundError as exc:  # pragma: no cover - exercised via tests
+        raise ModuleNotFoundError(
+            f"{used_for} needs the '{module_name}' package, which is not installed. "
+            f"Install it with 'pip install {module_name}', or install the 'all' extra. "
+            "Reading the packaged catalog with list_files() or list_cubes() does not "
+            "need it."
+        ) from exc
 
 
 def _config_dir() -> Path:
@@ -32,6 +50,8 @@ async def scrape_website(
     """Collect directory metadata and Zarr links from a remote listing."""
 
     from vesuvius.data.paths import parser
+
+    aiohttp = _require("aiohttp", "Refreshing the catalog from the data server")
 
     ssl_context = ssl.create_default_context()
     ssl_context.check_hostname = False
@@ -52,6 +72,8 @@ async def collect_subfolders(base_url: str, ignore_list: List[str]) -> List[str]
     """Return subfolder paths beneath ``base_url`` ignoring provided patterns."""
 
     from vesuvius.data.paths import parser
+
+    aiohttp = _require("aiohttp", "Refreshing the catalog from the data server")
 
     ssl_context = ssl.create_default_context()
     ssl_context.check_hostname = False
@@ -91,7 +113,9 @@ def update_list(
 
     loop = _ensure_loop()
     if loop.is_running():
-        nest_asyncio.apply()
+        _require(
+            "nest_asyncio", "Refreshing the catalog from inside a running event loop"
+        ).apply()
 
     tree, zarr_files = loop.run_until_complete(scrape_website(base_url, ignore_list))
     cubes_tree, _ = loop.run_until_complete(scrape_website(base_url_cubes, ignore_list))
--- a/vesuvius/tests/test_imports.py
+++ b/vesuvius/tests/test_imports.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from pathlib import Path
+import importlib.util
 import subprocess
 import sys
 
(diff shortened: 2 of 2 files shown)


Bullo27:
**Reproduced on Linux — every claim holds.** x86_64, Python 3.14.7 (uv-managed), `uv sync --frozen`, main c935851 (no `vesuvius/` changes since your base 5479453a7).

- main: 7/12 `__all__` names are `None`, `list_files()` → `TypeError`; `tests/test_imports.py` fails at collection (`No module named 'nest_asyncio'`); with `nest_asyncio` added: 5 failed / 3 passed, and your two new tests fail.
- this PR: `list_files()` → keys 1–5; 5 failed / 5 passed — the same five torch-only `test_cli_entrypoints_show_help` cases as main, nothing new.

Extra data point: the published release (`pip install vesuvius==0.2.4`, Python 3.12) has the same module-scope imports and no `nest_asyncio` in its core requirements → identical `None`/`TypeError`.

Oishi1029:
Thanks for reproducing it on Linux — good to have x86_64 alongside the arm64 run, and useful that you confirmed the two new tests fail on main rather than taking my word for it.

The vesuvius==0.2.4 data point is the part I hadn't checked. That makes the blast radius wider than I described: anyone doing a plain `pip install vesuvius` hits it, not only people working from a clone. Worth having on the record here.

hendrikschilling:
`At vesuvius/src/vesuvius/utils/catalog.py:52, the parser is imported before the new friendly dependency check. That parser itself imports aiohttp and lxml, so missing dependencies can   
  still produce the ordinary import error instead of the advertised guidance.
  Moving the aiohttp check before the parser import would address that small gap. The new error-message test calls the helper directly, so 
  it misses this ordering issue.
`

Oishi1029:
Thanks — you're right, and it wasn't hypothetical: `lxml` isn't in the core install, so on a plain `uv sync` both `scrape_website()` and `collect_subfolders()` died with a bare `No module named 'lxml'` from inside the parser and never reached the new message.

Fixed in ece133d19. Both packages are now checked in one `_load_parser()` helper before the parser is imported. I also replaced the helper-only test with one that calls the two public functions with `aiohttp` or `lxml` blocked, evicting any cached parser so it sees the real import order. It fails on the previous head, with and without `lxml` installed, and passes now — table in the updated description.

Issue #1555, opened 2026-08-21
vc_render_tifxyz: --prefetch-remote / --remote-url help text describes a staged on-disk cache that is never written


Bullo27:
## Summary

On the remote path (`--remote-url`), `vc_render_tifxyz` builds a **RAM-only** chunk cache and persists nothing to disk. Three parts of the CLI contract describe on-disk staging that cannot happen:

1. **`--prefetch-remote`** help says it prefetches *"into the existing staged cache before rendering"* — the prefetch only warms the in-memory cache.
2. **`--remote-url`** help says it is *"optional if --volume cache already records it"*. That refers to the `.remote_source.json` marker read by `loadCachedRemoteUrl()` at `volume-cartographer/apps/src/vc_render_tifxyz.cpp:450`. **Nothing in the repository ever writes that file** — `git grep "\.remote_source\.json"` returns exactly one hit, the read site — so `--remote-url` can never actually be omitted.
3. **`-v/--volume`** is `required()` even when `--remote-url` fully specifies the source, and on the remote path the directory is never created, populated or reused. A path that does not exist works fine.

This is **not a regression**: `vc_render_tifxyz.cpp` was untouched by #1554, and both help strings arrived with #1203 (`6e784219b`).

## Measured

Environment: main `df05b9634`, Ubuntu 24.04, live PHercParis4 open data over anonymous HTTPS. Surface = the 1.4 cm² patch grown by `vc_grow_seg_from_seed` on `20260310173927-45.532um-11.0m-110keV-masked.zarr`.

```bash
mkdir -p staged render
build/bin/vc_render_tifxyz -v staged \
  --remote-url "https://vesuvius-challenge-open-data.s3.amazonaws.com/PHercParis4/volumes/20260310173927-45.532um-11.0m-110keV-masked.zarr" \
  -s out/auto_grown_*/ --scale 1 -g 0 --voxel-size 45.532 --tif-output render/
```

- Render succeeds: `rc=0`, 3 bands in ~7 s, `render/00.tif` (58 272 bytes) with real papyrus texture.
- **`staged/` afterwards: 0 files.** `~/.VC3D` afterwards: **0 new files, 0 changed** (full stat snapshot of all 6 054 files taken before and after).
[...]

PR #1656, opened 2026-08-30
fix(render): persist remote chunks in the -v cache

7jycwjmbfn-eng:
**In one sentence:** When `vc_render_tifxyz` renders from `--remote-url`, downloaded chunks now stay in VC3D's remote cache, so rendering the same area again downloads almost nothing.

On the PHercParis4 crop I render from the open-data bucket (`w129`, group 2, 384×384), main downloads 86.2 MiB every time. With this PR the second render downloads 0.1 MiB.

Previously the renderer opened the remote pyramid with its own RAM-only chunk cache, so every run fetched the same chunks again, even ones VC3D had already cached.

This PR opens the URL through `Volume::NewFromUrl()` and reads through `Volume::sharedChunkCache()`, the same path VC3D uses. Chunks are stored under the remote cache directory from the VC3D settings, keyed by the URL-derived volume id.

**Proof**

main: both renders download 86.2 MiB

this PR: the second render downloads 0.1 MiB

Three runs, median values:

| | first render | second render |
|---|---|---|
| main, no prefetch | 86.2 MiB, 21.4 s | 86.2 MiB, 25.5 s |
| main, `--prefetch-remote` | 142.9 MiB, 25.6 s | 143.0 MiB, 25.2 s |
| this PR, no prefetch | 86.2 MiB, 34.9 s | 0.1 MiB, 15.9 s |
| this PR, `--prefetch-remote` | 143.0 MiB, 39.1 s | 0.1 MiB, 15.0 s |

All renders produced the same image (sha256 `b2a63b43…`). The first render with this PR is slower because it writes the chunks to disk and waits for pending downloads before exiting.

**Details**

- The private `-v/remote_sources/...` layout and the CLI partial-cache check from the first version are removed.
- `--cache-gb` sets the decoded-byte budget of the process cache service, as `vc_grow_seg_from_seed` does.
- Before exiting, the renderer waits for pending chunk requests and cache writes (`ChunkCache::waitForPendingChunks`), so band prefetch still in flight is kept instead of being downloaded again next time.
- Like `vc_grow_seg_from_seed`, the CLI does not apply VC3D's cache size limit, and the render fails if the remote cache directory is not writable (`VC3D_CONFIG_DIR` can point it elsewhere).
- `test_chunk_cache` (101 cases) and `scripts/test_render_volume_remote_cache.py` pass.
- Tested on `21fb23241`, WSL2 Ubuntu 24.04, gcc 13, `ci-release-tests-gcc` preset.


Bullo27:
Cross-reference: this implements the staging half of #1555 (filed Aug 21 — measured there: on `--remote-url` the chunk cache is RAM-only and `-v` stays empty, while the `--prefetch-remote` / `--remote-url` help text describes on-disk staging that never happens). You found it independently; linking so the reviewer has the prior measurements and #1555 can close or narrow on merge. One scope note from your diff: #1555's other half — `.remote_source.json` is only ever read, written by nothing — is untouched here, which is fine, just worth knowing when closing.

*Comment by an AI agent (Claude) operated by Bullo27.*

Bullo27:
**Verified by running; the claim holds, and three things turned up.** Built 3bff2a1 on current main (23adee0, gcc 13) and rendered a 1.4 cm² PHerc1447 patch at `-g 3` from the open-data bucket.

Reproduces: first render 12.4 s / 129 MiB downloaded, `-v` gains 141 files (248 MiB); second render 5.0 s / 0.0 MiB, bit-identical TIFF. main (9daa477) leaves `-v` empty and downloaded 130 MiB on both runs.

1. **The staged dir is not tied to the URL.** Rendering a different volume (PHerc0490B, same 128³ uint8 grid) into the same `-v`: 0.0 MiB downloaded, output bit-identical to the PHerc1447 render, 162,989 of 315,084 pixels different from a fresh-cache render of 0490B. The log prints 0490B's URL and shape while serving 1447's chunks; no warning. The dir holds only `level_3/<z>/<y>/<x>.bin` plus the `3/<z>/<y>/<x>` mirror, nothing naming the source.
2. **A bare relative path disables staging.** `-v staged` (the form used in #1555) prints `[WARN] ChunkCache disabled persistent caching …: persistent cache path is not a safe per-volume directory`, downloads 130 MiB and writes 0 files; `-v ./staged` and absolute paths work (`isExactCacheDirectoryTarget` rejects an empty parent).
3. **Omitting `--remote-url` later renders zeros.** The populated dir opens as a local OME-Zarr (`.zarray` is published for every level): `-g 3` reproduced the cached image in 0.3 s, `-g 2` (never fetched) wrote an all-zero 1365×925 TIFF, rc 0, no message. On main the dir is empty and this fails with `cannot detect version` (measured in #1555).

Tested fix for 1+2: pass `std::filesystem::absolute(vol_path) / vc::core::util::remoteFileCachePath(remoteUrl)` (RemoteFileCache.hpp, the layout VC3D's remote cache uses). With it a relative `-v` stages, the two volumes land in separate `remote_sources/https/<host>/…` subtrees, and each renders its own image (re-render of the first: 2 MiB, 8 s). For 3, the CLI already reads `.remote_source.json` (`loadCachedRemoteUrl`); writing it here is the obvious hook, not tested.

Also: `-v` pointing at a non-empty non-cache dir now exits 1 (`neither a legacy cache nor a Zarr mirror`) where main rendered; likely intended, noting it.

7jycwjmbfn-eng:
@Bullo27 Thanks for pointing this out and testing it. I've fixed the issues. I also have another PR, [#1657](https://github.com/ScrollPrize/villa/pull/1657), that reduces unnecessary downloads during remote rendering. Feel free to take a look when you have time—it might be useful for your work.

Bullo27:
**Verified by running 93d76ef: all three fixed.** Built on 23adee0 (gcc 13), PHerc1447 + PHerc0490B at `-g 3`. Your `test_render_remote_cache.py` passes here, and my own runs agree: the two volumes in one `-v` render byte-identical to their fresh-cache renders, a bare `-v staged` stages (134 files), and every no-URL open of a partial cache (`-g 3`, `-g 2`, old-layout root) exits 1 with the new message.

**One residual, measured: a warm render still transfers 6.3 MiB (and ~1 s), every time.** Sampling `rx_bytes` at 0.1 s: the four bands are done at 4.9 s with 0.05 MiB transferred; between `band 4/4` and exit (6.1 s) 6.3 MiB arrive and no file is written. Each cold run leaves the mirror 2–4 level-3 chunks short (59, 58, 60 of 62 across my two cold runs and your harness's 1447 cache; here (7,2,1), (7,2,2), (7,3,1)), and no later run adds them. Mechanism, read from the code rather than traced: `readMultiSlice` prefetches each band's bounding box with `wait=false` (Slicing.cpp:733); chunks the sampler never reads are still in flight when `main` returns; `~ChunkCacheService` runs `invalidateState()` (clears `storageObjectTransfers_`, so `dispatchStorageObjectResult` returns before `writeMirrorObject`) and then `waitIdle()`, so the download completes and is discarded. 3bff2a1 lost 0–1 chunks per run the same way, so the 0.0 MiB figure depends on timing rather than on this head.

Tested fix, at the end of `main` before `stopLogFlusher()`:
```cpp
    if (ownedChunkCache) {
        const auto pending = ownedChunkCache->activeRemoteFetches();
        if (!pending.empty())
            ownedChunkCache->prefetchChunks(pending, /*wait=*/true);
    }
```
With it: cold run 129.7 MiB, 62 of 62 chunks persisted (the last four files land 0.3 s after `band 4/4`); two warm runs 0.0 MiB, 5.9 s and 5.5 s; all TIFFs byte-identical to the unpatched renders. Without a code change, one `--prefetch-remote` run also closes the gap (here it persisted a 59-file, 80 MiB superset of the band boxes; the next plain run transferred 0.0 MiB).

*AI-assisted (Claude), human-directed account — measured unless marked otherwise.*

hendrikschilling:
```
  [P2] Use the existing remote-volume cache infrastructure instead of introducing another cache location

  volume-cartographer/apps/src/vc_render_tifxyz.cpp:1422

  This adds persistent caching under -v/remote_sources/..., but remote volumes elsewhere now use the globally configured cache root and the source identity/layout provided by Volume.      
  Consequently, the renderer ignores that configuration and can download separate copies of chunks already cached by VC3D.

  Please open the remote source through Volume::NewFromUrl() and obtain its existing chunk cache, preserving the renderer’s required prefetch and shutdown-drain behavior. This should reuse
  the established cache location, source identity and cache-service configuration rather than reconstructing them in the CLI. Merely substituting the global root here is insufficient      
  because the proposed subdirectory layout also differs.
```

7jycwjmbfn-eng:
Thanks, I've rewritten this around the existing Volume cache path.

--remote-url is now opened through Volume::NewFromUrl() and the renderer uses Volume::sharedChunkCache(). The private -v cache layout and the CLI partial-cache check are gone. The shutdown drain is kept, so on the test crop the first render downloads 86.2 MiB and the second 0.1 MiB.

The old tests are replaced by a harness for cold/warm equality, multi-level persistence, source isolation and fresh-cache reproduction. The known limits (no cache size cap in the CLI, and the cache directory must be writable) are listed in the description.

Tested with gcc 13 on Ubuntu 24.04; Windows is covered by the repo's mingw CI.

7jycwjmbfn-eng:
Thanks @hendrikschilling for the review and the merge. Reusing the existing Volume cache, as you suggested, lets the renderer share its downloaded chunks with VC3D.

Thanks also to @Bullo27 for testing this on real data on your own initiative and for your suggestions. Your runs on PHerc1447 and PHerc0490B helped me check the remote cache early on.

PR #1657, opened 2026-08-30
fix(render): avoid prefetching unused chunks

7jycwjmbfn-eng:
**In one sentence:** With `--prefetch-remote`, the renderer now downloads only the chunks needed by the render instead of every chunk in each band's bounding box.

On the same PHercParis4 crop used in #1656 (`w129`, group 2, 384×384), main plans 112 chunks. This PR plans 25.

```
Prefetch: 112 chunk(s) across rows 0..2    # main
Prefetch: 25 chunk(s) across rows 0..2     # this PR
```

Previously, the prefetch planner added the full chunk bounding box around each band. On this crop, that meant downloading 142.9 MiB with prefetch enabled, compared with 86.2 MiB without it.

This PR instead walks the actual sample positions and plans only the chunks required by the renderer's sampling mode.

After the exact plan is downloaded, the render reads through `PrefetchedArrayView`. The wrapper forwards normal reads to the real cache but suppresses later `prefetchChunks` calls, so the bounding-box prefetch inside the sampling code does not add the unused chunks back.

Renders without `--prefetch-remote` are unchanged.

**Proof**

Fresh cache, same crop:

#1656 only: 112 chunks planned

this PR: 25 chunks planned, 52.6 MiB

Three runs, median values:

| | planned | first render | second render |
|---|---|---|---|
| main, no prefetch | n/a | 86.2 MiB, 21.4 s | 86.2 MiB, 25.5 s |
| main, `--prefetch-remote` | 112 | 142.9 MiB, 25.6 s | 143.0 MiB, 25.2 s |
| #1656, `--prefetch-remote` | 112 | 143.0 MiB, 39.1 s | 0.1 MiB, 15.0 s |
| this PR, no prefetch | n/a | 86.3 MiB, 31.8 s | 0.1 MiB, 14.9 s |
| this PR, `--prefetch-remote` | 25 | 52.6 MiB, 27.2 s | 0.1 MiB, 14.7 s |

All renders produced the same image (sha256 `b2a63b43…`).

**Details**

- Depends on #1656 and is currently stacked on top of it.
- This uses the second option from review: suppress speculative prefetch locally after the exact plan has completed.
- 52.6 MiB is the 25 planned chunks. With the same plan but without the wrapper, the render downloaded 63.2 MiB (measured on the earlier RAM-only revision).
- `PrefetchedArrayView` suppresses only `prefetchChunks`. Normal reads, refetches after eviction and source errors still pass through to the underlying cache.
- The wrapper is local to this render path. VC3D, other callers and renders without `--prefetch-remote` keep their existing prefetch behaviour.
- `test_render_prefetch` passes 10 cases, including exact-plan matching, prefetch suppression, eviction/refetch, error propagation and uint16 reads through both band and tile paths.
- `test_chunk_cache` and the #1656 remote-cache harness also pass on this branch.
- The planner uses a bitmap up to 8 MiB and falls back to direct key insertion for larger sparse ranges.
- Tested on `12056e22d`, WSL2 Ubuntu 24.04, gcc 13, `ci-release-tests-gcc` preset.


Bullo27:
**Verified dcc2fa9 by running: the plan is exact, but the render does not keep to it, so every later render pays for the bounding box again.**

Setup: dcc2fa9's diff on top of #1656's 93d76ef (base 23adee0, gcc 13; the persistence fix is needed to count files written). Same 1.4 cm² PHerc1447 patch as in #1656, `--scale 1 -g 3 -n 1`, fresh cache per cold run. All TIFFs byte-identical to the current planner's (md5 85fabda0…).

- `--prefetch-remote` plans **26** chunks (current planner: 120), exactly the chunks the sampler reads (matches my independently computed footprint); all 26 persisted before band 4/4 (54.8 MiB).
- Then **33.7 MiB more arrive between band 4/4 and exit, 0 files written.** Read from the code: core `readMultiSliceImpl` (Slicing.cpp:733) still fires the per-band bounding-box `prefetchRegion` → `prefetchChunks(keys, false)` on every render (`readCompositeFastImpl`:600 likewise); those transfers are in flight at teardown and discarded (mechanism in my #1656 comment).
- So **every warm render re-transfers 33.6–33.8 MiB** (three runs, plain and `--composite-collapse -n 3`; 0 files written; 10–14 s), versus **0.0 MiB, 5–6 s** after the current planner's superset, or once the #1656 drain patch lands the remaining 38 chunks (+72 MiB).

On this data the saving is confined to the prefetch run; the bytes needed to reach a quiet cache are unchanged. The exact footprint would have to live in core (skip or narrow the region prefetch), which should also shrink your 174.3 MiB no-prefetch rows (inferred, not measured). Happy to re-run a follow-up.

*AI-assisted (Claude), human-directed account — measured unless marked otherwise.*

hendrikschilling:
```
  [P2] The exact prefetch plan is followed by the old bounding-box requests.

  volume-cartographer/apps/src/vc_render_tifxyz.cpp:464 now collects only chunks needed by the samples. However, rendering subsequently calls:

  - readMultiSliceImpl(), which still prefetches the union bounding box at volume-cartographer/core/src/Slicing.cpp:733.
  - readCompositeFastImpl(), which still prefetches the base-coordinate bounding box at volume-cartographer/core/src/Slicing.cpp:600.

 The new tests miss it because their prefetchChunks() implementation deliberately does nothing.

Please coordinate this with #1656: use the existing remote-volume cache infrastructure for source identity, cache location and persistence, and ensure the rendering stage does not subsequently expand an exact prefetch plan into bounding-box requests. Either share the exact planning logic with the relevant core sampling paths, or provide an explicit way to suppress their redundant speculative prefetch when the caller has already planned it. Extend the tests to record prefetch requests as well as blocking reads.
```

7jycwjmbfn-eng:
Thanks for the review. I've addressed all three points.

I took the second option: after the exact prefetch finishes, the render reads through PrefetchedArrayView, which suppresses only later prefetchChunks calls. Normal reads, refetches after eviction and source errors still pass through unchanged.

The test array now records prefetch calls as well as blocking reads, with coverage for suppression, eviction/refetch, error propagation and uint16 reads through both band and tile paths.

The branch is also rebased on the rewritten #1656, so the remote source now uses Volume::NewFromUrl() and the shared persistent cache.

On the same PHercParis4 crop, the exact plan drops from 112 chunks to 25. The first render downloads 52.6 MiB instead of 142.9 MiB on main, and the second downloads 0.1 MiB. Output is byte-identical.

Tested with gcc 13 on Ubuntu 24.04; Windows is covered by the repo's mingw CI.

7jycwjmbfn-eng:
Thanks @hendrikschilling for the review and the merge. You caught that the bounding-box prefetch in Slicing.cpp was still expanding the exact plan, and that the tests could not see it.

Thanks also to @Bullo27 for testing this on real data on your own initiative. The detailed download measurements you posted were very useful while I was working on the prefetch.

PR #1665, opened 2026-08-31
zarr_utils: fix public S3 access with no AWS credentials

kadenpool:
**In one sentence:** the Python side can now read the open-data bucket on a machine with no AWS
credentials, instead of waiting five minutes and reporting that the data does not exist.

**One real example:** macOS arm64, Python 3.14, villa @ `e2442b7`, with no `~/.aws` directory and
zero `AWS_*` environment variables. Starting with the open-data bucket, I called `zarr_array_exists()`
on the PHerc0800 volume `20250521135224-8.640um-1.2m-116keV-masked.zarr` level 0 to check it was there
before running anything against it. It returned `False`, so I went looking for what was wrong with my
setup. Nothing was. Asking the same object anonymously through `fsspec` returns `True` immediately. The swallowed
exception is
`NoCredentialsError: Unable to locate credentials`, so a missing-credentials failure on a public bucket
was reaching me as the array not existing.

The same path through `wait_for_zarr_creation()`, which is what `vesuvius.finalize_outputs` and
`vesuvius.blend_logits` call, loops until it gives up and reports `RuntimeError: Timeout waiting for
zarr array to be created`. I shortened its 300s default to 10s so the capture would finish.

**Before:** `zarr_array_exists()` (`utils/io/zarr_utils.py:77`) has two defects that compound:

1. it hardcodes `fsspec.filesystem('s3', anon=False)`, so it demands AWS credentials, for a bucket
   the open-data catalog itself publishes as `"usage": "public-read"`;
2. it wraps everything in `except Exception: return False`, so a credentials, permission or network
   failure is reported as *the array does not exist*.

The single caller turns that into the user-visible failure. `wait_for_zarr_creation()`
(`zarr_utils.py:100`) loops on `while not zarr_array_exists(path)` with `max_wait_time=300`, and both
`vesuvius.finalize_outputs` (`models/run/finalize_outputs.py:378`) and `vesuvius.blend_logits`
(`models/run/blending.py:776`) go through it. So the caller waits out the full five minutes and is
then told the array was never created, when it was there the whole time and only the lookup failed.

**After this PR:** the same lookup succeeds anonymously, and a lookup that genuinely cannot be
resolved raises an error naming the real cause instead of silently answering "absent".

**Proof:** on a machine with no `~/.aws` and no `AWS_*` environment variables, against a real
published volume:

```
array: s3://vesuvius-challenge-open-data/PHerc0800/volumes/
       20250521135224-8.640um-1.2m-116keV-masked.zarr/0

fsspec anon=True  ->  True          # it is there; anonymous is the correct way to look
zarr_array_exists ->  False
swallowed         ->  NoCredentialsError: Unable to locate credentials
```

and what the caller experiences, with the same shortened timeout:

```
BEFORE   zarr_array_exists      -> False
         waiting for array to be created... (0s elapsed)
         waiting for array to be created... (5s elapsed)
         RuntimeError: Timeout waiting for zarr array to be created. Waited 10 seconds.

AFTER    zarr_array_exists      -> True
         wait_for_zarr_creation -> Array found! can now proceed.
         absent path            -> False   (unchanged)
```

**Why / where this is useful:** this is the Python-side counterpart of #1624, merged on 28 Aug, which
fixed the same class of failure in `volume-cartographer` (public S3 access blocked by unusable ambient
credentials, with no anonymous retry). That fix does not reach `vesuvius/`, so a user on the Python
path still hits the original behaviour. I checked that the same trigger reaches this code: with
deliberately invalid `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` set, `fsspec` raises
`PermissionError: Forbidden` rather than returning `False`, and with this change the public object
resolves anyway. There is a test for that case. Anyone running `finalize_outputs` or `blend_logits` against
the open bucket without AWS credentials currently loses five minutes per attempt to an error that
points at the wrong thing.

## Details

- `utils/io/zarr_utils.py::zarr_array_exists`:
  1. S3 lookups try the configured credentials first, then fall back to anonymous. A user with real
     credentials is never silently downgraded: anonymous is only ever a fallback, and there is a test
     asserting that ordering.
  2. If existence cannot be determined at all, raise `RuntimeError` naming the underlying error
     instead of returning `False`.
  3. The argument is coerced with `os.fspath`, so `os.PathLike` works. This is not cosmetic: moving
     the local branch out of the old blanket `try` would otherwise have made a `Path` argument raise
     `AttributeError` where `main` returned `False`. `main` in fact returned `False` for every `Path`,
     present or not, because `zarr_path.startswith(...)` raised inside its own `except Exception`.
     Every caller in this repo passes a `str`, so nothing in-tree changes.
- **Deliberately NOT changed.** A genuinely absent object still returns `False`. That non-regression
  is the important one, and it is covered by a test on both the local and the S3 path. The local
  branch is untouched. `wait_for_zarr_creation` is untouched: it now surfaces the raised error instead
  of looping, which is the point.
- **Behaviour change.** Callers that relied on `False` meaning "either absent or
  unreachable" will now see an exception in the unreachable case. There is one caller in this repo and
  that is precisely the case it was getting wrong.
- **Stopping this recurring.** This PR also adds `tests/utils/test_no_silent_swallowing.py`, which
  walks `utils/io` and `data` for broad handlers whose whole body is `pass` or a constant return and
  requires a recorded reason for each. In code that answers "is this there / can I read this", a
  swallowed error is indistinguishable from a real negative, which is exactly how this bug reached a
  five-minute timeout with the wrong message. Only two such handlers exist today and both are
  legitimate (`_is_ome_zarr` reads `.zattrs` the code documents as optional; `_zarray_signature`
  returns `None` as a documented degraded cache mode). A second test fails if a listed handler stops
  swallowing, so the list can only shrink.
- **No overlap.** No open PR touches `utils/io/zarr_utils.py`, checked against all 40 open PRs on
  2026-08-31. #1624 is C++ only and shares no files with this change.

Verification:

```
$ python -m pytest tests/utils/test_zarr_array_exists.py -q
8 passed in 0.96s
```

Negative control, run by restoring `main`'s `zarr_utils.py` under this PR's test file. Exactly the
four tests covering the changed behaviours fail, and only those four:

```
4 failed, 4 passed
FAILED ...TestLocalPaths::test_accepts_a_path_object
FAILED ...TestS3Paths::test_falls_back_to_anonymous_when_credentials_are_missing
FAILED ...TestS3Paths::test_falls_back_when_credentials_are_present_but_rejected
FAILED ...TestS3Paths::test_raises_when_existence_cannot_be_determined
```

The other four pass both ways by design: they cover a present local path, an absent local path, an
absent S3 object, and that credentialed access is attempted before anonymous. Those are the
non-regressions, so they must not discriminate.

Same suite before and after, same machine:

```
$ python -m pytest tests/utils tests/data --ignore=tests/utils/models -q
before (main):  60 passed, 1 skipped
after  (this):  70 passed, 1 skipped      (8 in test_zarr_array_exists + 2 in test_no_silent_swallowing)
```

`tests/utils/models` is excluded because it cannot be collected in this environment at all:
`test_load_nnunet_model_compatibility.py` imports `nnunetv2`, an optional dependency that is not
installed here. That failure is identical on `main` and is unrelated to this change.

Under your CI command (`.github/workflows/test-vesuvius.yml`), same machine, run against `main` and
against this branch:

```
$ python -m pytest tests -m "not slow and not network"
main        15 failed, 276 passed, 6 skipped, 36 errors
this branch 15 failed, 286 passed, 6 skipped, 36 errors
```

Identical failure and error counts: this branch adds 10 passing tests and introduces no new failures.
The 15 failures and 36 collection errors are pre-existing on `main` and come from optional ML
dependencies absent in my local environment; CI installs them with `uv sync --extra all`. The new
tests need no network (the S3 cases stub `fsspec.filesystem`, so their `s3://` paths are never fetched) and
exercise no zarr API directly, so they are unaffected by the 2.18.7 / 3.2.1 matrix.

Branched from villa `e2442b7`; every number above was measured on this branch at `95515a7`, Python
3.14, macOS arm64.

**One limit.** Pointing at a bucket that does not exist still returns `False` rather
than raising, because fsspec reports an absent bucket the same way it reports an absent key. The raise
path therefore covers genuine outages and is exercised by a mocked test rather than a live one.

Two things I am happy to change if you would rather:

- **Drop the ratchet test** (`test_no_silent_swallowing.py`) into its own PR, if you would rather this
  one stayed strictly to the bug. It costs nothing today, since the only two handlers it finds are
  both legitimate and already listed, but it is a separable concern.
- **Keep this to the anonymous fallback alone** and leave the raise for a separate PR, if you would
  rather not change the return contract at all.

---

The patch was written with Claude Code under my direction. I hit the problem myself against the open
data bucket, produced the before/after above myself on that data, and wrote the example above myself.


Bullo27:
**Verified end-to-end on a credential-less Linux box.** No `~/.aws`, no `AWS_*` env vars: on main, `zarr_array_exists("s3://…PHerc0800/…masked.zarr/0")` → `False` in 2.2 s while `fsspec anon=True` sees the object — your five-minute-timeout scenario; with this PR the same call → `True` (2.7 s, anonymous fallback). Suite: 8 passed on the branch; restoring main's `zarr_utils.py` under your test file fails **exactly the 4 you predicted** (`accepts_a_path_object`, both fallback cases, `raises_when_existence_cannot_be_determined`). Python 3.14.7, zarr 3.2.1, villa @ 908aa7f.

kadenpool:
Thanks for checking this on a real machine with no credentials. The number I could not
produce from here is the cost: 2.2s to fail against 2.7s to succeed, so on a machine with no credentials the fallback costs about half a second, and a user with working credentials never reaches it.

The four predicted failures matching exactly is the part I most wanted someone else to
confirm, since I wrote both the fix and the test that judges it.

kadenpool:
Hi, a gentle ping on this one. I merged current `main` into the branch today and all four checks pass (CodeQL, CI, Test vesuvius Python, Large PR review gate). It still only touches `zarr_utils.py` and its two test files. Happy to adjust anything if you would like it done differently.

PR #1669, opened 2026-08-31
vesuvius: make in-place recompression free chunks, and never destroy the only copy

acejayl:
Two independent problems in the in-place recompress path. They interact, which is why they ship together.

### 1. `delete_chunk_file` only matched zarr v2 *nested* keys

It built `<array>/0/0/0`, so it matched nothing for zarr v3 (`c/0/0/0`) or for v2's own default separator (`0.0.0`), and returned silently. `pyproject.toml` allows `zarr>=2.18.7,<4`, so the common case did nothing at all.

Measured with zarr 3.3.0, writing one chunk per layout by hand:

| layout | chunk key | removed? |
|---|---|---|
| zarr v3 (default today) | `c/0/0/0` | **no** |
| zarr v2, separator `.` | `0.0.0` | **no** |
| zarr v2, nested separator | `0/0/0` | yes |

Deleting the source chunk as soon as it has been copied is what keeps an in-place recompress from needing twice the disk of the array. When that delete silently no-ops the guarantee is gone, and a large in-place run can exhaust the disk on exactly the volumes this tool exists for.

It now probes the layouts in turn and returns whether anything was removed, so a miss can no longer pass for a deletion.

### 2. The failure handler deleted the partially recompressed copy

`_recompress_worker` removes each source chunk right after copying it, so from the first completed work item the original is incomplete and the temp tree holds the only copy of everything processed so far. The handler ran `shutil.rmtree(temp_path)`, which turned any mid-run failure — a corrupt chunk with `--skip-corrupt` off, a worker dying, an `rmtree` or rename refused on Windows because a handle was still open — into unrecoverable data loss.

Nothing in the `try` block is reachable before deletion may have begun, so the temp tree is now always left in place and the error says where it is.

Fixing (1) is what makes the deletion actually happen on modern zarr, which is precisely when (2) starts to bite.

### Tests

`tests/image_proc/test_zarr_recompress_inplace.py`. The layout cases write chunk files directly rather than through zarr, so they pin the key layout itself and hold on either zarr major.

Verified against the unfixed tree: the v3 and dot-separator cases fail there, and the failure-handling test fails there too.

| environment | result |
|---|---|
| zarr 2.18.7 | 5 passed |
| zarr 3.3.0 | 4 passed, 1 skipped |

### Note on the skip

It is deliberate and unrelated to this change: `RecompressTask` cannot run under zarr 3.x at all, because it creates its temp array with `zarr.open(..., compressor=<numcodecs Blosc>)` while zarr 3 makes new arrays v3 by default and rejects `compressor`. That happens for v2 sources too, so the task is currently usable only on zarr 2.x. I'll file that separately rather than fix it here, since which formats recompress should emit looks like a design call rather than a bug fix.

### DIFF
--- a/vesuvius/src/vesuvius/image_proc/run/zarr_tasks/tasks/recompress.py
+++ b/vesuvius/src/vesuvius/image_proc/run/zarr_tasks/tasks/recompress.py
@@ -256,10 +256,18 @@ class RecompressTask(ZarrTask):
                         f"  Warning: {len(level_corrupted)} corrupted chunks filled with zeros"
                     )
             except Exception as e:
-                # Clean up temp on failure
-                if Path(temp_path).exists():
-                    shutil.rmtree(temp_path)
-                raise RuntimeError(f"Recompression failed for {level_path}: {e}") from e
+                # Deliberately do NOT delete temp_path here. Each worker removes
+                # a source chunk as soon as it has copied it (delete_source_path
+                # in _recompress_worker), so once the pool has started, the
+                # original is already incomplete and temp_path holds the only
+                # copy of everything processed so far. Removing it would turn a
+                # recoverable failure into permanent data loss.
+                raise RuntimeError(
+                    f"Recompression failed for {level_path}: {e}. "
+                    f"Partially recompressed data has been left at {temp_path}. "
+                    f"The chunks it contains were already deleted from the "
+                    f"original, so recover from it before removing anything."
+                ) from e
 
     def _run_to_output(self) -> None:
         """Run recompression to a new output location."""
--- a/vesuvius/src/vesuvius/image_proc/run/zarr_tasks/utils.py
+++ b/vesuvius/src/vesuvius/image_proc/run/zarr_tasks/utils.py
@@ -262,25 +262,56 @@ def inverse_permutation(perm: List[int]) -> List[int]:
     return inv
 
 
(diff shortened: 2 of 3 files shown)


Bullo27:
**Both legs verified.** Linux, Python 3.14.7: zarr 2.18.7 → 5 passed; zarr 3.3.0 → 4 passed, 1 skipped — matching your table. Negative control (main's `recompress.py`/`utils.py` under your test file): **every non-skipped case fails on both majors** — including the v2-nested and report cases you didn't list, because they assert the new returns-whether-removed contract that main's `None`-returning `delete_chunk_file` can't satisfy. The failure-handler test confirms the temp tree survives a mid-run failure instead of being rmtree'd. villa @ 908aa7f.

acejayl:
@Bullo27 thanks for running this independently, and especially for the negative control. Checking my test file against main's `recompress.py`/`utils.py` is what makes the before/after claim verifiable rather than just asserted, and I appreciate you going past the cases I listed to the v2-nested and report ones.

Worth noting for anyone arriving from #1670: this PR is the chunk-freeing and "never destroy the only copy" fix, not the zarr 3 compressor rejection. That one is being handled separately in #1682 and #1711. Those touch a different part of `_run_inplace` than this does, so as far as I can tell they can land in either order without conflicting.

PR #1672, opened 2026-08-31
Name the missing dependency instead of failing with 'NoneType' object is not callable

objetivopbd-blip:
## What breaks

`vesuvius.list_files()` is the first call in the README. After the install
CONTRIBUTING.md prescribes (`uv sync`), it fails like this:

```
>>> import vesuvius
>>> vesuvius.list_files()
```

The message names nothing, so the natural reading is "the function is broken".

## Why it happens

`vesuvius.utils.catalog` imports `aiohttp` and `nest_asyncio`. Neither is a core
dependency, so a plain `uv sync` leaves them out. `__init__.py` caught the
ImportError and bound `list_files`, `list_cubes`, `update_list` and
`is_aws_ec2_instance` to `None`, while still exporting all four in `__all__`.
The package imports fine and the failure is deferred to the call site, stripped
of its cause.

## The repo already assumes this cannot happen

`tests/test_imports.py::test_public_api_exports` asserts:

```python
assert callable(vesuvius.list_files)
```

That holds today only when the optional dependencies happen to be installed. On
a default install it fails. This PR makes the code honour the contract the test
already states.

## The change

The `None` bindings become stubs that name the package that is actually missing:

```
ImportError: vesuvius.list_files() needs the optional dependency 'aiohttp',
which is not installed. Install it with `uv pip install aiohttp`
(or `pip install aiohttp`). Original import error: No module named 'aiohttp'
```

Three details worth flagging, because each was a way to get this wrong:

- **Only the two catalog dependencies downgrade to a stub.** The `except` is
  narrowed to `ModuleNotFoundError` and checks `exc.name`. A syntax error, a
  version clash or a genuine bug inside `utils` is re-raised untouched — real
  breakage must never be reported as a missing optional dependency.
- **The message names the package that is missing**, not both, and gives the
  installable distribution name (`nest-asyncio`, not `nest_asyncio`).
- **The install hint avoids `vesuvius[all]`.** That extra pulls in
  `volume-cartographer`, which `pyproject.toml` itself marks as monorepo-only
  and absent from PyPI, so the command would fail for anyone installing from
  PyPI. There is no extra dedicated to these helpers — `aiohttp` and
  `nest_asyncio` only arrive via `models` (which pulls torch), `label-transfer`
  or `all` — so the message names the package directly. If you would rather
  have a small `catalog` extra, I am happy to add it.

Behaviour is unchanged when the dependencies are present: the real functions are
imported and bound as before. The helpers stay callable either way, so
`callable(vesuvius.list_files)` is now true unconditionally.

## This has already cost time on both sides

- #1329 (open since 4 Aug) reports the same `TypeError` from the `volume-only`
  extra, with the same diagnosis.
- In bf57b86 (18 Aug, part of #1527, which was about something else entirely)
  the `blending` extra gained
  `"nest-asyncio>=1.6.0", # Required by vesuvius.utils.catalog (imported via __init__)`.
  That comment is an accurate diagnosis of this bug, written while fixing an
  unrelated problem — and it patches one extra.

The remaining install paths still carry the hole, including the plain `uv sync`
CONTRIBUTING.md prescribes. Fixing the shared failure mode means that pin does
not have to be repeated in every extra that happens to import the catalog.

## Scope, versus #1329

#1329 reports this for the `volume-only` extra. The same breakage happens on the
**default** install, which is the path CONTRIBUTING.md sends every new
contributor down. The `blending` extra already carries a
`"Required by vesuvius.utils.catalog (imported via __init__)"` pin — that
patches one extra rather than the shared failure mode.

## Tests

Three cases, in `tests/test_imports.py`:

- each helper stays callable and names the missing package (parametrised over
  `aiohttp` and `nest_asyncio`, checking the distribution name and that the
  original error is chained);
- a missing module that is *not* a catalog dependency — `requests`, also
  imported by `catalog.py` — still propagates as `ModuleNotFoundError`.

```
$ python -m pytest tests/test_imports.py -k "not cli_entrypoints"
6 passed, 5 deselected
```

Separately: `test_cli_entrypoints_show_help[vesuvius.models.training.train]`
fails on a clean `uv sync` because `tqdm` is not installed either. Out of scope
here, but the same class of problem — happy to open a separate issue.

## Environment

Ubuntu 25.04, Python 3.14.7, `uv sync` per CONTRIBUTING.md, package installed
from this checkout. Verified against Scroll1 over the streaming catalog.

### DIFF
--- a/vesuvius/src/vesuvius/__init__.py
+++ b/vesuvius/src/vesuvius/__init__.py
@@ -23,19 +23,59 @@ try:
 except Exception:
     tifxyz = None  # type: ignore
 
-# Attempt to import utils.  utils requires aiohttp and nest_asyncio; if they aren't
-# installed (e.g. when you install with [volume-only] or without extras), we silently
-# fall back to None.  Users who need catalog utilities should install the appropriate
-# optional extra.
+# Optional dependencies of ``vesuvius.utils``. They are not core dependencies, so
+# the plain ``uv sync`` that CONTRIBUTING.md tells contributors to run leaves them
+# out, and the catalog helpers below cannot be imported.
+_CATALOG_REQUIREMENTS = ("aiohttp", "nest_asyncio")
+
+# Package name on PyPI, where it differs from the import name.
+_CATALOG_DISTRIBUTIONS = {"aiohttp": "aiohttp", "nest_asyncio": "nest-asyncio"}
+
+
+def _needs_catalog_extra(name, missing, cause):
+    """Build a stand-in for a catalog helper whose dependencies are missing.
+
+    Binding the helpers to ``None`` made the failure surface at the call site as
+    ``TypeError: 'NoneType' object is not callable``, which says nothing about the
+    real cause.  Keeping them callable lets the failure name the missing package
+    and a command that actually installs it.
+    """
+
+    distribution = _CATALOG_DISTRIBUTIONS.get(missing, missing)
+
+    def _stub(*_args, **_kwargs):
+        raise ImportError(
+            f"vesuvius.{name}() needs the optional dependency {missing!r}, "
+            f"which is not installed. Install it with "
+            f"`uv pip install {distribution}` (or `pip install {distribution}`). "
+            f"Original import error: {cause}"
+        ) from cause
+
+    _stub.__name__ = name
+    _stub.__qualname__ = name
+    _stub.__doc__ = (
+        f"Unavailable: vesuvius.{name}() requires the optional dependency "
+        f"{missing!r}. Calling it raises ImportError."
+    )
+    return _stub
+
+
+# ``utils`` pulls in the packages above.  When one of them is genuinely absent the
+# catalog helpers become stubs that name it.  Any other import failure - a syntax
+# error, a version clash, a bug inside utils - is re-raised untouched, so real
+# breakage is never reported as a missing dependency.
 try:
     from . import utils  # type: ignore
     from .utils import is_aws_ec2_instance, list_cubes, list_files, update_list  # type: ignore
-except Exception:
+except ModuleNotFoundError as _catalog_import_error:
+    _missing = _catalog_import_error.name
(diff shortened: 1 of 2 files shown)


objetivopbd-blip:
Some context on where this came from, since CONTRIBUTING.md asks for it.

I am working alone from Brazil, coming into the Challenge through the Progress
Prizes. I set the environment up on 18 August: cloned villa, made a venv, ran
`uv sync` exactly as CONTRIBUTING.md says, and went to list the scrolls to see
what I could actually download and start working on. That first call is the one
in the README:

    >>> import vesuvius
    >>> vesuvius.list_files()

I spent a while on the wrong trail. The message names nothing, so I assumed the
function was broken or that my install had gone wrong, and I went back through
the install steps before I thought to look at `__init__.py` and found the
helpers bound to None. Once I saw it, the fix was small — but the time lost was
not in fixing it, it was in not knowing what to look for.

That is the whole reason I am sending this. It is a small patch and I know it
is not a headline contribution. But `list_files()` is the first thing a new
contributor runs, and right now it fails in a way that teaches you nothing. I
would rather the next person coming in from outside spend that hour on the
scrolls than on their install.

I am happy to adjust the approach if you would prefer a small `catalog` extra
instead of naming the packages directly in the message, or if you would rather
this were handled with a lazy `__getattr__`.

Bullo27:
**Verified — plus one routing note for reviewers.** On this box (Linux, Python 3.14.7): branch → `pytest tests/test_imports.py -k "not cli_entrypoints"` = 6 passed; restoring main's `__init__.py` under your tests → exactly the 2 parametrized stub tests fail, and the requests-propagation test passes both ways as designed.

Routing: **#1653** (open since Aug 30) fixes the same default-install failure one layer down — it makes `catalog.py` import its refresh-only deps lazily, so `list_files()` actually works on a core install. No file overlap (`catalog.py` there, `__init__.py` here), so they compose: with #1653 in, these stubs become the safety net for whatever else lands in the `except`. Verified both on this box.

PR #1676, opened 2026-09-01
VcDataset: refuse a region that extends past the dataset, and make callers check it

kadenpool:
**In one sentence:** `readRegion` and `writeRegion` now refuse a region that extends beyond the dataset, instead of writing past the caller's buffer (`readRegion`) or reading past it (`writeRegion`), which is the crash your own test suite documents and currently skips.

**One real example:**

macOS arm64, Homebrew LLVM 23, volume-cartographer built from source at `e2442b7`. I ran `test_vcdataset_more`, one of the `vc_core` tests, and it told me 14 test cases passed. Then I read the file. One of those 14 is a `TEST_CASE` whose name ends "(BUG: segfaults under coverage build)", and its entire body is `CHECK(true)`, with the actual reproduction commented out just above it. So the suite counts a pass for a test it never runs, and the bug described in the FIXME is still sitting there.

**Before:** `core/test/test_vcdataset_more.cpp` carries this, unchanged since the file was added:

```cpp
TEST_CASE("VcDataset: region read past dataset bounds (BUG: segfaults under coverage build)")
{
    // FIXME: vc_core bug — VcDataset::readRegion with a region that extends
    // past the dataset bounds crashes with SIGSEGV instead of clamping,
    // returning false, or throwing. The origin is in-bounds; only the
    // far corner is OOB. Reproduces on coverage build (gcc 15, --coverage).
    // Skipping the actual call so the suite still runs; flip on to repro:
    //
    //   (void)ds->readRegion({0, 0, 0}, {16, 16, 16}, out.data());  // <-- crash
    //
    // Once the impl is fixed, this test should:
    //   - never segfault
    //   - either return false OR clamp to the valid sub-region.
    CHECK(true);  // placeholder — see comment above
}
```

Neither `readRegion` nor `writeRegion` compares `offset + regionShape` against `impl_->shape_`. For a region running off the end, the chunk loop walks past the last chunk, `read_chunk` returns nothing, and `fillTypedElements` then writes fill values at output offsets computed from the **region** shape rather than the dataset shape. For an 8x8x8 dataset asked for 16x16x16 that is up to 4096 bytes written into whatever the caller allocated.

**After this PR:** both functions return `false` for such a region and leave the caller's buffer untouched. `false` is the smaller of the two remedies the FIXME asks for and matches the existing `bool` contract; clamping would silently return a different region than the one requested, which is the kind of quiet substitution that is hard to debug later.

**Proof:** the FIXME says the crash "reproduces on coverage build (gcc 15, --coverage)", which is awkward to rely on. It reproduces deterministically under this project's own `-DVC_ENABLE_ASAN=ON`, with the reproduction line uncommented, on Homebrew LLVM 23 / macOS arm64:

```
==19357==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x615000004d80
WRITE of size 8 at 0x615000004d80 thread T0
SUMMARY: AddressSanitizer: heap-buffer-overflow VcDataset.cpp:379 in fillTypedElements
exit code 1
```

This explains why it has survived: **without ASan the same binary exits 0 and prints "14 test case(s) passed"**. The overflow is undefined behaviour that happens not to fault in an ordinary build, so an ordinary test run reports success.

**Why / where this is useful:** `VcDataset` is the reader underneath volume access, so a caller that miscomputes a region corrupts memory instead of getting an error. Memory corruption surfaces later and somewhere else, which is the worst kind of bug to chase in a long-running render or trace. The failure is also silent in ordinary builds, so nothing in CI would catch it.

## Details

- `core/src/VcDataset.cpp`: both `readRegion` and `writeRegion` validate `offset` and
`regionShape` against `impl_->shape_` before doing any work, immediately after the existing zero-size early return. The comparison is written as `regionShape[d] > dsShape[d] - offset[d]` rather than `offset[d] + regionShape[d] > dsShape[d]` so the sum cannot overflow `size_t`.
- **`writeRegion` had the same defect in the other direction.** It would read past the caller's
input buffer. Fixed in the same change because it is the same missing check, three lines apart.
- **Deliberately NOT clamping.** The FIXME offers "return false OR clamp". Clamping would hand
back a smaller region than requested without saying so, and a caller that did not check would read uninitialised memory believing it was data.
- **The disabled test is replaced with a real one.** It covers a fully out-of-bounds region, one
that starts inside and runs off the end, one entirely outside, the `writeRegion` case, and three in-bounds calls that must still succeed. It also asserts the caller's buffer is untouched on refusal.
- **The zero-size early return is left where it was**, so a request for a zero-element region
still returns `true` without the bounds check, even from an offset past the end. That is the existing behaviour, nothing is written in that case, and moving the check above it would change a contract this PR was not asked to change. Flagging it so it reads as a decision rather than an oversight.
- **No overlap in the changed functions.** One other open PR, #1704 (a compression backend), also edits `VcDataset.cpp`, but in its codec and setup code, not `readRegion` or `writeRegion`, and `git merge-tree` shows the two merge cleanly. No issue covers `readRegion`, and the test file has not been modified since it was added. Checked 2026-09-20.
- **Regression surface.** Four apps call these functions (`vc_tifxyz2zarr_sparse`,
`vc_zarr_to_tiff`, `vc_ngrids`, `vc_add_ignore_label`). None passes an out-of-bounds region, so none changes behaviour. `test_vcdataset.cpp` and `test_vcdataset_more.cpp` both pass: 13 and 14 cases. One unrelated target, `test_segmentation_intersection_invalidation`, fails to build here for a missing Qt `.moc` file; it fails identically on `main`, so it is not from this change.

Verification:

```
$ cmake -S . -B build -DVC_TESTING=ON -DVC_ENABLE_ASAN=ON -DCMAKE_BUILD_TYPE=QuickBuild
$ cmake --build build --target test_vcdataset_more
$ ./build/bin/test_vcdataset_more
14 test case(s) passed          # exit 0, ASan clean
```

Negative control, restoring `main`'s `VcDataset.cpp` under this PR's test file:

```
==22358==ERROR: AddressSanitizer: heap-buffer-overflow
SUMMARY: AddressSanitizer: heap-buffer-overflow VcDataset.cpp:379 in fillTypedElements
exit code 1
```

Built with Homebrew LLVM 23 rather than Apple clang, because `core/util/Logging.hpp` uses `std::format`, which Apple's toolchain does not yet provide. That matches `.github/workflows/vc3d-macos.yml`, which also uses Homebrew LLVM.

I am happy to make it clamp instead of refuse if you would rather, or to split the `writeRegion` half into its own PR.

---

The patch was written with Claude Code under my direction. I ran the test suite and read the disabled `TEST_CASE` myself; the source build and the AddressSanitizer runs were done with Claude Code on my machine, under my direction. The example above is mine.

## Update, 21 Sep (review)

- Rebuilt as two commits on current `main`: this fix and its test, unchanged, then the review changes. The branch had carried #1829's commit in through a merge of `main`.
- `VcDataset.hpp` documents the bounds contract, and `readRegion` and `writeRegion` are `[[nodiscard]]`.
- All 19 calls in the four apps listed above now check the result and stop: `vc_add_ignore_label` (8, through a `readRegionOrThrow` helper), `vc_ngrids` (9), `vc_tifxyz2zarr_sparse` (1) and `vc_zarr_to_tiff` (1). Valid input behaves as before.
- The test also covers a wrong number of dimensions, an extent that wraps `offset + extent`, a buffer left untouched and a dataset left untouched.
- #1704 still merges cleanly with this branch (`git merge-tree`, 21 Sep). No other open PR touches `VcDataset.hpp`, `test_vcdataset_more.cpp`, `vc_add_ignore_label.cpp`, `vc_ngrids.cpp`, `vc_tifxyz2zarr_sparse.cpp` or `vc_zarr_to_tiff.cpp`; checked 21 Sep against all 74 open PRs.


Bullo27:
**Reproduced on Linux — both directions.** Ubuntu 24.04, GCC 13.3, Debug + `-DVC_ENABLE_ASAN=ON`, tree @ 908aa7f:

- this PR's impl + test → **14 test cases passed, ASan clean**;
- unmodified `VcDataset.cpp` + this PR's test → `AddressSanitizer: heap-buffer-overflow, WRITE of size 8`, stack `memset → fillTypedElements (VcDataset.cpp:379) → readRegion lambda (:847)` — the same attribution as your macOS/LLVM-23 run, so the overflow is not toolchain-specific.

FWIW the refuse-don't-clamp choice reads right to me: it keeps the existing `bool` contract, and a clamped read would be silent substitution.

kadenpool:
Thanks for running this properly. Your result is worth more than my original report: the FIXME claims it needs a coverage build on gcc 15, and between your ASan run on GCC 13.3 and mine on LLVM 23 that framing is clearly too narrow. It is an ordinary out of bounds write that any sanitizer catches and no ordinary build does, which is why it survived so long.

Good to hear the refuse rather than clamp call reads right to you. That was the part I was least sure about.

kadenpool:
Hi, a gentle ping on this one. I merged current `main` into the branch today and rebuilt it. With only this PR's test applied to current `main`, three checks fail and the run then aborts with `malloc(): invalid size (unsorted)`. With the fix, all 14 test cases pass. It still only touches `VcDataset.cpp` and its test. Happy to adjust anything if you would like it done differently.

kadenpool:
@hendrikschilling, a small ask: this one reaches the repository's 28-day limit and will close on its own on 30 Sep. Current main is merged in and the build and test checks all pass. If it's still wanted, could it have the keep-open label? Happy to change anything.

kadenpool:
This closes on 30 September under the 28-day rule. It is still mergeable against `main` today, its checks are green apart from the Vercel preview that needs team authorisation on any outside PR, and the code it fixes is unchanged since it was opened. Bullo27 reproduced the crash independently on Linux. If it is worth keeping, the `keep-open` label holds it; if the answer is that you would rather it were done differently, say so and I will change it or close it myself.

kadenpool:
@erdpx, you added `keep-open` to #1080, so I hope you don't mind me asking you directly. This PR and #1682 both reach villa's 28-day limit before a maintainer has reviewed either: this one on the run of 29 Sep (I wrote 30 Sep above, but the job runs at midnight Pacific, so it is a day earlier) and #1682 on 1 Oct. Both still merge cleanly with `main`, and their build and test checks pass; the only red check is the Vercel preview, which needs team authorisation on any outside PR. If they are worth keeping, the label holds them. If not, that is fine, and they can close.

hendrikschilling:
```
  [P1] The PR branch contains unrelated changes beyond the VcDataset fix.

  The current head includes the adjacent-fiber-links feature and changes across 24 files, including VC3D, fiber_merge.py, vc_sync.py, tests, and documentation. This comes from the merge   
  history, where the PR branch includes commit f07d33be6 (#1829) even though those changes are not part of this PR’s purpose.

  See the current changed-files list (https://github.com/ScrollPrize/villa/pull/1676/files). The branch should be rebased or reduced to the VcDataset commit and its test before merging.   

  Remaining Integration Concern

  [P2] Callers ignore the new false return value.

  The new checks in VcDataset.cpp:771-783 (https://github.com/ScrollPrize/villa/blob/624cc8e44647c06abd6db6489a5b439a977d97ad/volume-cartographer/core/src/VcDataset.cpp#L771-L783) and     
  VcDataset.cpp:882-894 (https://github.com/ScrollPrize/villa/blob/624cc8e44647c06abd6db6489a5b439a977d97ad/volume-cartographer/core/src/VcDataset.cpp#L882-L894) correctly reject invalid  
  regions, but existing callers generally discard the result. For example, vc_zarr_to_tiff, vc_ngrids, and vc_add_ignore_label continue processing their buffers after a failed read. A     
  future bounds-calculation error would therefore become silent zero/stale output instead of an explicit failure.

  The header should document the bounds contract, and important callers should check and propagate failure.

```

kadenpool:
Thanks for the review. Both points are done and the branch is force-pushed.

**P1.** The branch is now two commits on top of current `main` and nothing else: the original fix with its test, unchanged, then the P2 work below. Against `main` it is 7 files, all under `volume-cartographer/`.

**P2.**
- `VcDataset.hpp` now documents the contract on both functions: the region must lie inside the dataset, and a region that does not, or has the wrong number of dimensions, is refused. The call returns `false` without touching the buffer or the dataset. That is the only way either one returns `false`, so it always means the caller computed a bad region.
- Both are `[[nodiscard]]`, so a new caller that ignores the result gets a compiler warning. Checked: `main`'s `vc_zarr_to_tiff` compiled against the new header warns at the call.
- Every caller now checks it. That is 19 calls, all reads, in four apps: the three you named plus `vc_tifxyz2zarr_sparse`. Nothing outside the tests calls `writeRegion`.
  - `vc_add_ignore_label` (8): through one small `readRegionOrThrow` helper. The four reads inside worker threads throw into the existing `firstError` handling, which already reports after the join, so no new thread plumbing was needed.
  - `vc_ngrids` (9): throws, like the file's other failures.
  - `vc_tifxyz2zarr_sparse` (1, in a pyramid worker): sets `hadError` like the write path beside it, so that level fails.
  - `vc_zarr_to_tiff` (1): exits 1 naming the slice.
- None of these fire on valid input, since every caller's region is taken from or clamped to the dataset shape. They are there for the future bounds bug you describe.

**Proof.**
- I planted an off-by-one region, `{1, Y, X + 1}`, in `vc_zarr_to_tiff` and ran it on a small zarr whose voxels are all non-zero. Old caller: exit 0, "Done. Wrote 3 TIFFs", every pixel 0. New caller: exit 1, `Error: failed to read slice 0 from ".../input.zarr/0"`, nothing written. Without the planted bug, all 3 slices match the input.
- The test now also covers a wrong number of dimensions, an extent large enough to wrap `offset + extent` around, a refused read leaving every byte of the buffer as it was, and a refused write leaving the dataset untouched. With the bounds check rewritten as a plain sum, the wrap-around case fails, so it guards that line.
- Built here with Homebrew clang on macOS with no compiler warnings, and `test_vcdataset_more` passes 14 of 14.

I also corrected the PR's own comments, which had drifted: `writeRegion`'s was a copy of `readRegion`'s and described the wrong failure, and both still said the test was disabled.

PR #1678, opened 2026-09-01
VC3D: add optional focus bounds for fiber annotation

hendrikschilling:
**In one sentence:** Let VC3D annotators focus on a base-coordinate bbox that dims outside context and constrains newly changed fiber geometry.

**One real example:** Starting with an open fiber crossing a selected annotation region, enabling the focus bbox dims the surrounding plane and strip pixels while trimming newly optimized tails near the bounds without removing controlled interior geometry.

**Before:** VC3D had no shared annotation-region bound, so views gave no indication of the working region and tracing or reoptimization could continue far beyond it.

**After this PR:** The Focus controls can enable an inclusive base-XYZ bbox. Plane and annotation-strip views render outside pixels at half brightness, outside control placements are rejected, and new or reoptimized open tails are softly trimmed.

**Proof:** At commit `c456a5d2f`, VC3D and the focused test targets build successfully. All focused tests pass. The focus-bbox display and annotation behavior were also manually tested in VC3D.

**Why / where this is useful:** Fiber annotators can keep work and generated annotation data roughly limited to a selected volume region without requiring exact geometric clipping.

## Details

- Bounds are session-local and clear when the project or package changes.
- Each solve captures the active bbox when it starts; toggling it does not restart an in-flight solve.
- Existing control points and geometry between the first and last controls are preserved.
- Persisted seed-only fibers remain reopenable outside a later or moved bbox, and a retrace preserves their path into the bbox.
- Standard SurfaceCache-backed segmentation views are intentionally not dimmed.

Verification:

`cmake --build volume-cartographer/build -j 32 --target VC3D test_line_annotation_generated_views test_quadsurface_basics test_fiber_trace_review`

`ctest --test-dir volume-cartographer/build -R '^(test_fiber_trace_review|test_line_annotation_generated_views|test_quadsurface_basics)$' --output-on-failure`

### DIFF
--- a/volume-cartographer/apps/VC3D/CState.cpp
+++ b/volume-cartographer/apps/VC3D/CState.cpp
@@ -201,6 +201,9 @@ std::shared_ptr<VolumePkg> CState::vpkg() const { return _vpkg; }
 
 void CState::setVpkg(std::shared_ptr<VolumePkg> pkg)
 {
+    if (_vpkg != pkg) {
+        clearFocusBounds();
+    }
     _vpkg = std::move(pkg);
     emit vpkgChanged(_vpkg);
 }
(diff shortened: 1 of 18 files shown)


hendrikschilling:
@codex review

PR #1680, opened 2026-09-01
Fix recompress task under zarr 3.x (compressor= rejected for zarr_format=3)

### LINKED ISSUE #1670: recompress task is unusable under zarr 3.x: "compressor cannot be used for arrays with zarr_format 3"
## Summary

`vesuvius.image_proc.run.zarr_tasks.tasks.recompress.RecompressTask` raises `ValueError` immediately under zarr 3.x, for **every** input — v3 arrays and v2 arrays alike. `pyproject.toml` declares `zarr>=2.18.7,<4`, so the task is only usable on the 2.x half of the supported range.

## Cause

`_run_inplace` (and `_run_to_output`) build the destination array with a numcodecs compressor:

```python
temp_z = zarr.open(
    temp_path,
    mode="w",
    shape=read_z.shape,
    chunks=read_z.chunks,
    dtype=read_z.dtype,
    compressor=self._compressor,   # numcodecs Blosc, set in prepare()
)
```

Under zarr 3.x, `zarr.open(..., mode="w")` creates a **v3** array by default regardless of the source array's format, and v3 rejects the v2-era `compressor` keyword:

> `ValueError: compressor cannot be used for arrays with zarr_format 3. Use bytes-to-bytes codecs instead.`

Because the format of the *new* array is what matters, opening a v2 source does not help.

## Reproducer

```python
import tempfile
from pathlib import Path
import numpy as np, zarr
from vesuvius.image_proc.run.zarr_tasks.tasks.recompress import RecompressConfig, RecompressTask

data = np.arange(64, dtype="uint8").reshape(4, 4, 4) + 1

for label, kw in [("v3 source", {}), ("v2 source", {"zarr_format": 2})]:
    d = Path(tempfile.mkdtemp()) / "vol.zarr"
    a = zarr.open(str(d), mode="w", shape=(4, 4, 4), chunks=(2, 2, 2), dtype="uint8", **kw)
    a[:] = data
    task = RecompressTask(RecompressConfig(input_zarr=str(d), output_zarr=None,
                                           num_workers=1, inplace=True))
    task.prepare()
    try:
        task._run_inplace()
        print(label, "OK")
    except Exception as e:
        print(label, f"{type(e).__name__}: {e}")
```

Observed on zarr 3.3.0, Python 3.14, Windows:

```
v3 source: ValueError: compressor cannot be used for arrays with zarr_format 3. Use bytes-to-bytes codecs instead.
v2 source: ValueError: compressor cannot be used for arrays with zarr_format 3. Use bytes-to-bytes codecs instead.
```

## Possible directions

The right fix depends on which formats you want `recompress` to emit, so I have not sent a patch:

1. **Mirror the source format.** Read `read_z.metadata.zarr_format` and pass `zarr_format=` when creating the destination, using `compressor=` for v2 and `compressors=[BloscCodec(...)]` for v3. Preserves round-trip behaviour but means maintaining two codec paths.
2. **Always emit v3** and translate `--compression-level` into a `BloscCodec`. Simpler, but changes the output format of an in-place operation, which may surprise downstream readers.
3. **Pin the task to v2** explicitly (`zarr_format=2`) and document it, if v3 output is not wanted yet.

Happy to send a PR for whichever you prefer.

## Related

#1669 fixes two adjacent problems in the same in-place path that are independent of this one: `delete_chunk_file` silently matching nothing on v3/dot-separator layouts, and the failure handler deleting the partially recompressed copy.

StewartSethA:
## Summary
- `RecompressTask._run_inplace` always created the destination array with a numcodecs `compressor=` kwarg. zarr-python 3.x rejects that for the zarr_format=3 arrays it creates by default (`ValueError: compressor cannot be used for arrays with zarr_format 3. Use bytes-to-bytes codecs instead.`), which makes the task unusable on the 3.x half of the supported `zarr>=2.18.7,<4` range, for every input.
- Added `_create_recompressed_array()` which detects the *source* array's `zarr_format` and mirrors it when creating the recompressed copy: v2 sources still get `compressor=Blosc(...)`, v3 sources get `compressors=[BloscCodec(...)]`.
- `_run_to_output`'s use of `create_level_dataset` (which forces v2 output via `NestedDirectoryStore`) is a separate, pre-existing concern and is out of scope here.

## Test plan
- [x] Added `vesuvius/tests/zarr_tasks/test_recompress.py` with a v2 regression test (reproduces the issue's original repro, ran and passed locally against zarr 2.16.1) and a v3 test (`skipif`'d under zarr<3 — could not install zarr>=3 in the sandbox used to write this, so that branch is written to match the documented zarr-python 3.x codec API but should be verified by CI/a maintainer running under zarr 3.x)
- [ ] Maintainer/CI to confirm the v3 test passes on the zarr>=3 leg

Fixes #1670

### DIFF
--- a/vesuvius/src/vesuvius/image_proc/run/zarr_tasks/tasks/recompress.py
+++ b/vesuvius/src/vesuvius/image_proc/run/zarr_tasks/tasks/recompress.py
@@ -56,7 +56,7 @@ def _create_recompressed_array(
             kwargs["zarr_format"] = 2
         return zarr.open(path, mode="w", shape=shape, chunks=chunks, dtype=dtype, **kwargs)
 
-    from zarr.codecs import BloscCodec, BloscShuffle
+    from zarr.codecs import BloscCodec, BloscShuffle, BytesCodec
 
     return zarr.open(
         path,
@@ -65,7 +65,8 @@ def _create_recompressed_array(
         chunks=chunks,
         dtype=dtype,
         codecs=[
-            BloscCodec(cname="zstd", clevel=compression_level, shuffle=BloscShuffle.bitshuffle)
+            BytesCodec(),
+            BloscCodec(cname="zstd", clevel=compression_level, shuffle=BloscShuffle.bitshuffle),
         ],
         zarr_format=3,
     )
--- a/vesuvius/src/vesuvius/image_proc/run/zarr_tasks/tasks/recompress.py
+++ b/vesuvius/src/vesuvius/image_proc/run/zarr_tasks/tasks/recompress.py
@@ -45,9 +45,9 @@ def _create_recompressed_array(
 
     zarr-python 2.x arrays are always on-disk format 2 and take a numcodecs
     ``compressor=``. zarr-python 3.x defaults new arrays to format 3, which
-    rejects ``compressor=`` and instead wants a bytes-to-bytes codec passed
-    as ``compressors=``. Branching here lets the destination array mirror
-    the source array's format instead of failing under 3.x (see #1670).
+    rejects ``compressor=`` (v2-only) and instead wants a bytes-to-bytes
+    codec passed as ``codecs=``. Branching here lets the destination array
+    mirror the source array's format instead of failing under 3.x (see #1670).
     """
     if _ZARR_MAJOR_VERSION < 3 or zarr_format == 2:
         compressor = Blosc(cname="zstd", clevel=compression_level, shuffle=Blosc.BITSHUFFLE)
@@ -64,7 +64,7 @@ def _create_recompressed_array(
         shape=shape,
         chunks=chunks,
         dtype=dtype,
-        compressors=[
+        codecs=[
             BloscCodec(cname="zstd", clevel=compression_level, shuffle=BloscShuffle.bitshuffle)
         ],
         zarr_format=3,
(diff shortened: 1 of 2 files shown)


Bullo27:
**The v3 leg fails as written — on zarr 3.2.1 (what `uv.lock` pins) and 3.3.0.** `zarr.open(..., compressors=[...], zarr_format=3)` goes through `open_array` → `TypeError: AsyncArray._create() got an unexpected keyword argument 'compressors'`. `compressors=` belongs to `zarr.create_array`; `zarr.open`/`zarr.create` only know `compressor`. All three tests fail on 3.x — the "v2 source" one too, because without `zarr_format=2` `_make_source` creates a v3 array there (the Copilot note).

Measured fix: in the v3 branch use `zarr.create_array(store=path, overwrite=True, shape=shape, chunks=chunks, dtype=dtype, compressors=[…], zarr_format=3)` instead of `zarr.open(path, mode="w", …)`. With only that change: 3/3 tests pass on 3.2.1 and 3.3.0, #1670's reproducer prints OK/OK, and a two-level multiscale group round-trips in place with formats (3/3, 2/2) and attrs preserved. The v2 branch is fine as is (zarr 2.18.7: 1 passed, 2 skipped; v2 sources on 3.x: OK). `_run_to_output` still fails on 3.x (`zarr.NestedDirectoryStore`) — pre-existing, as you noted.

PR #1681, opened 2026-09-01
Perf/openmp render and grid cache

StewartSethA:
Performance improvements to segmentation, rendering and segmentgrowth:
Segmentation and rendering of scroll surfaces several times faster on many surfaces, with byte-identical output.
  
Example: Starting with a 1.67 cm² PHerc0211 segment and OMP_NUM_THREADS=8, I ran vc_render_tifxyz before and after these commits, and it produced the same 65 layers (identical md5sum)
  in 2.34 s instead of 15.20 s.
  
Summary of fixes/improvements:
Before: OpenCV's libopencv_core static initializer calls omp_set_dynamic(1) before main(). libgomp then sizes every parallel team from getloadavg(), returning one thread whenever load ≥ core count. On a busy machine every OpenMP region in VC3D ran serially, and this was invisible to thread_limit, OMP_NUM_THREADS, and even OMP_DYNAMIC=FALSE, because the constructor runs last. Separately, the render band loop used schedule(dynamic,16) over a 128-row band -- 8 work units regardless of pool size -- and rebuilt a 16384-slot ChunkSampler per thread per band. And
  the normal-grid cache held 512 entries against a measured 7,424-slice working set, causing 110,631 file opens across 7,424 distinct files -- a 14.9× re-open rate.
  
After this PR: Render 15.20 s → 2.34 s (6.5×) at 8 threads on a 1.67 cm² segment; 143.78 s → 37.69 s (3.81×) on a 12.63 cm² segment from the OpenMP fix alone. Growth resume rounds drop from 64.4 to 44.2 CPU-seconds, with system time falling 30.31 s → 10.69 s — cheaper in wall clock and total CPU, not a trade.
  
Four builds — unpatched, +fix1, +fix2, +fix3 — from one clean clone of ScrollPrize/villa at b0bcee5, same segment, same seed (VC_GROWPATCH_RNG_SEED pinned), warm-up discarded, load recorded per arm. Looking at the md5s: render output is byte-identical at --threads 1, 8 and 32 on both test segments, and area_cm2 matches to 16 digits across every growth arm and every cache budget. The unmodified from-source build was first verified byte-identical to the released binary, to establish that reference.
  
Why / where this is useful: Anyone running VC3D segmentation or rendering at scale. The OpenMP fix in particular helps most on loaded machines -- exactly when batch segmentation runs -- and needs no configuration change. Faster rendering and growth mean more surface area segmented per machine-hour, where segmentation is the current bottleneck on reading complete scrolls.
  
Limitations:
- Fix 2 is within noise at 8 threads on a large segment -- by mechanism, since dynamic,16 over 128 rows already yields 8 work units. Its benefit appears above 8 threads (2.0–2.4× at 32).
- Fix 1 is neutral at thread_limit 1 by construction.
- The 4 GiB cache default is conservative, not optimal: 8 GiB was a further 1.42× -- chosen because RSS rises 181 MB -> ~2.6 GB per process, which matters when running many concurrent grows.
- Growth geometry varies with thread count. This is pre-existing: random_perturbation() uses a thread_local RNG, so tl=4 is not self-reproducible even at a pinned seed, and the unpatched build diverges identically once given real teams. generations.tif is unchanged -- same point count, same generations, only solved coordinates move. This is why growth is recommended at thread_limit 1, where output is byte-identical.

### DIFF
--- a/PERFORMANCE_FIXES.md
+++ b/PERFORMANCE_FIXES.md
@@ -376,28 +376,3 @@ the same source against an older sysroot solves it completely:
   container.
 
 This is packaging guidance, not part of the patches.
-
----
-
-## What is deliberately *not* here
-
-Several other avenues were measured and did not earn a place. Recording them so
-nobody spends the time again:
-
-* **GPU render prototype.** The sampling kernel is genuinely 16.5–24.6 G
-  samples/s against 19.4 M/s for a whole 32-core box, and end-to-end 8.3–14.2x
-  in-process. But rendering turned out to be an I/O job wearing a compute job's
-  clothes: on an idle dual-GPU box with the volume over 1 GbE, a render took
-  470.98 s of which **458.52 s (97.4%) was disk read**, while the sampling
-  kernel itself took 0.09 s. The right move is to integrate the kernel inside
-  the existing band loop, not to ship a separate renderer — and the scheduling
-  lesson ("run the render where the volume is on local disk", not "on whichever
-  box is idle") matters more than the kernel.
-* **A morphological dilate in the growth inner loop.** Amdahl ceiling of
-  1.00002x, confirmed by three independent instruments.
-* **Enabling the `SURFACE_SDT` residual.** Changing its weight by 1000x moved
-  the resulting area by 0.06%.
-* **Passing the surface prediction as the growth volume instead of raw CT.**
-  −1.9% area, no other benefit.
-* **Coarse-to-fine growth.** −26% throughput, −38% area, and wrong-winding
-  failures.
--- a/PERFORMANCE_FIXES.md
+++ b/PERFORMANCE_FIXES.md
@@ -276,6 +276,22 @@ disk the same comparison measured 1.88x wall and 1.90x fewer CPU-seconds.
 All three growth arms produced `md5(x,y,z,generations.tif)` identical and
 `area_cm2 = 10.66587924071438` to every digit.
 
+The budget is the causal variable, and it is transparent. Same segment, same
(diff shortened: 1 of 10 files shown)


giorgioangel:
@codex review

Bullo27:
**Fix 1 is harmless, but its rationale needs scoping** — measured on Ubuntu 24.04, GCC 13.3 / libgomp 14.2, OpenCV 4.6.0:

- Both documented Linux builds (`scripts/install_build_deps.sh`, `build_from_src_debian.sh`) take distro `libopencv-dev`; on 24.04 that `libopencv_core` is TBB-backed and contains no `omp_set_dynamic` at all, so `omp_get_dynamic()` is already 0 at `main()` and the collapse cannot occur there. It needs an OpenCV compiled with its OpenMP backend (`HAVE_OPENMP`) — worth stating which build produced the 3.81×.
- OpenCV's call isn't unconditional: 4.6.0 and 4.10.0 guard it with `OPENCV_FOR_OPENMP_DYNAMIC_DISABLE=1` (`parallel.cpp`), a no-code workaround.
- libgomp uses the **15-minute** load average (`dloadavg[2] + 0.1`, `config/linux/proc.c`), not the one-minute. With `OMP_DYNAMIC=true` on 28 cores at load 1/5/15 = 11.5/4.7/2.0: `OMP_NUM_THREADS=8` → team of 6, `=2` → team of 1. Teams collapse to 1 only once the 15-min average ≥ min(cores, nthreads).

Fixes 2/3 not measured here.

PR #1682, opened 2026-09-02
zarr_tasks: create_level_dataset works under zarr 3 (refs #1670)

### LINKED ISSUE #1670: recompress task is unusable under zarr 3.x: "compressor cannot be used for arrays with zarr_format 3"

kadenpool:
**In one sentence:** `create_level_dataset` now builds arrays with APIs that exist in both zarr 2 and zarr 3, so the four tasks routing through it stop raising immediately on the 3.x half of the `zarr>=2.18.7,<4` range this package declares.

**One real example:** macOS arm64, Python 3.14, zarr 3.3.0. I built a small v2 OME-Zarr and ran `ThresholdTask` on it, following the reproduction acejayl posted in #1670. It failed before touching any data:

```
AttributeError: module 'zarr' has no attribute 'NestedDirectoryStore'
```

zarr 3 has neither `NestedDirectoryStore` nor `DirectoryStore`. Since `create_level_dataset` is shared, the same `AttributeError` reaches `threshold`, `transpose`, `scale` and `recompress`'s to-output path.

**After this PR:** the same run completes and writes its output.

```
Input array shape: (8, 8, 8)
Task: threshold | value=127.0 | erase_blank=False
Total items to process: 8
RESULT: ThresholdTask completed, level-0 output written = True
```

**Proof, on both ends of the declared range and on the version this repo pins:**

| zarr | before | after |
|---|---|---|
| 2.18.7 (Python 3.12) | completes | completes, byte-identical layout |
| 3.2.1 | `AttributeError: NestedDirectoryStore` | completes |
| 3.3.0 | `AttributeError: NestedDirectoryStore` | completes |

The layout is what I was most worried about, since `NestedDirectoryStore` existed here only to get nested chunk directories. It is unchanged everywhere:

```
zarr 2.18.7  main   -> format=v2  sep='/'  chunks=['0/0/0', '0/0/1', '0/1/0']
zarr 2.18.7  branch -> format=v2  sep='/'  chunks=['0/0/0', '0/0/1', '0/1/0']
zarr 3.2.1   branch -> format=v2  sep='/'  chunks=['0/0/0', '0/0/1', '0/1/0']
zarr 3.3.0   branch -> format=v2  sep='/'  chunks=['0/0/0', '0/0/1', '0/1/0']
```

**Why this is useful:** `pyproject.toml` declares `zarr>=2.18.7,<4`, so zarr 3 is inside the supported range, and anyone installing this package fresh gets zarr 3 by default.

## Details

- `NestedDirectoryStore` existed only to get nested chunk directories. Both versions express that as `dimension_separator="/"`, so the store goes and the separator replaces it.
- The format still has to be stated, because zarr 3 defaults new arrays to format 3 where a numcodecs `compressor=` is rejected. Rather than repeat that version check here, this calls `open_zarr` from `vesuvius/data/utils.py`, which this package already has: it defaults `zarr_format=2` and passes the argument only on zarr 3, and its docstring gives the same reason (zarr 2 has no such argument and warns when given one). That keeps one copy of the rule in the codebase rather than two. The helper already writes `{"zarr_format": 2}` into `.zgroup`, so v2 was always the intent.
- **Deliberately not migrated to v3.** These arrays are v2 by design and the published scroll volumes are v2. Writing v3 output would be a data-format change, not a compatibility fix.

## Scope, and what this deliberately leaves alone

- **This does not close #1670.** That issue is about `RecompressTask._run_inplace`, which fails on a different line for a different reason. I deliberately did not touch it. @dud8's #1711 fixed that path, also via `open_zarr`, and explains there why converting a v3 source to v2 is correct for that function rather than a substitution: `_run_inplace` copies `.zattrs` and unlinks chunks at `<z>/<y>/<x>`, both of which are v2-only, so mirroring a v3 source would leave its chunks silently unfreed. #1711 was closed without being merged on 21 Sep, so that half of #1670 is still unfixed on `main`.
- **acejayl's comment on #1670 documents more failures than this fixes**: `Group.create_dataset` in `edt_dilate` and `merge`, `prepare()` reading `input_z.compressor`, and `tifs_to_ome_zarr`. Those are real and are not addressed here. This is one shared helper, not the whole module.
- **Prior work:** #1431 (axiosdevs, closed unmerged on 27 Aug) fixed the same `zarr_tasks` breakage first, and #1467 (closed) with issue #1448 cover the wider set of image_proc modules that call APIs zarr 3 removed. This PR is the narrow `create_level_dataset` fix with a regression test that CI runs on zarr 2.18.7 and 3.2.1.
- **Relationship to #1669**, which touches the same file. That PR adds `_candidate_chunk_paths` around line 262; this changes `create_level_dataset` around line 112. Different functions, no overlap. Happy to rebase if it lands first.

Verification:

```
$ python repro.py          # v2 input, ThresholdTask, per acejayl's reproduction in #1670
zarr 3.3.0
RESULT: ThresholdTask completed, level-0 output written = True
```

Five call sites route through this helper: the four tasks (`threshold`, `transpose`, `scale`, `recompress`) and `build_pyramid`, which the same four use for the upper levels. I ran `threshold` end to end on all three zarr versions above; the others go through the same helper and hit the same failure.

---

The patch was written with Claude Code under my direction. The issue is acejayl's. I reproduced it myself, ran the before and after on all three zarr versions myself, and wrote the example above myself.

## Update, 10 Sep

- Merged current `main`.
- Added `vesuvius/tests/image_proc/test_create_level_dataset.py`. CI runs `tests/` on zarr 2.18.7 and 3.2.1, so four cases are now checked on both ends of the declared range: data round-trips, the nested v2 layout survives (`dimension_separator="/"`, chunk key `0/0/0`), chunks holding only the fill value stay off disk, and `overwrite=True` replaces an existing level.
- zarr 3 deprecates `write_empty_chunks` as an argument and warns that it will be removed, so on zarr 3 it now goes through `config=`. zarr 2 has no `config=`, so it keeps the argument. This reuses the existing `_ZARR_V3` flag in `vesuvius.data.utils`, so the version rule stays in one place.
- The docstring no longer mentions `NestedDirectoryStore`.

The new test, run locally against this branch and against `main`:

| zarr | this branch | `main` |
|---|---|---|
| 2.18.7 | 4 passed, 0 warnings | 4 passed, 4 warnings |
| 3.2.1 | 4 passed | 4 failed: `AttributeError` on `NestedDirectoryStore` |
| 3.3.0 | 4 passed, 0 warnings | 4 failed: same |

## Update, 21 Sep (review)

- `overwrite=False` now raises `FileExistsError` for an existing level instead of truncating it (hendrikschilling's review). No caller passes `overwrite=False`. Two new tests: the refusal, which checks that the level's chunk files and data survive it, and `overwrite=False` on a new level.
- Rebuilt as two commits on current `main`: the original change, then the review fix.
- #1669 has since closed without being merged. No open PR touches `zarr_tasks/utils.py`; checked 21 Sep against all 74 open PRs.


kadenpool:
@hendrikschilling, a small ask: this one reaches the repository's 28-day limit and will close on its own on 1 Oct. Current main is merged in and the build and test checks all pass. If it's still wanted, could it have the keep-open label? Happy to change anything.

kadenpool:
Same rule, 1 October. Still mergeable, still green apart from the Vercel preview, and `zarr_tasks` is unchanged on `main`. #1700 landed zarr 2/3 helpers on 18 September that this could use instead of its own branch; say the word and I will rewrite it that way.

hendrikschilling:
```
[P2, pre-existing] overwrite=False is still ignored in create_level_dataset.

  The helper conditionally avoids deleting an existing level, but then always calls open_zarr(..., mode="w"); Zarr mode w truncates the existing array anyway. A caller passing
  overwrite=False can therefore lose the level. The new tests only cover overwrite behavior with the default True.

  This is not introduced by PR #1682, but the changed helper should either reject overwrite=False or open/validate the existing dataset instead.
```

kadenpool:
Thanks, fixed, and the branch is force-pushed.

- An existing level now raises `FileExistsError` when `overwrite=False`, instead of being truncated by the `mode="w"` open. I took the first of your two options. All five calls (the four tasks, and `build_pyramid`, which the same four use for the upper levels) pass the default `overwrite=True`, so nothing that runs today changes. Opening and validating the existing level would mean checking its shape, chunks, dtype and compressor against the request; with nothing relying on it, refusing is the smaller change. Happy to switch to that if you prefer it.
- The test writes a level, calls again with `overwrite=False`, and checks that it raises and that all 8 chunk files and the data are still there. On the previous code it fails: the call returns normally and the level comes back empty, 8 chunk files down to 0. A second test covers `overwrite=False` on a new level.
- Same history cleanup as #1676: the branch is now two commits on current `main`, the original change and then this fix, 3 files.
- 6 of 6 pass here on zarr 3.3.0; CI runs the file on 2.18.7 and 3.2.1.

PR #1685, opened 2026-09-03
VC3D line annotation: fixes from annotation sessions (click placement, side cut, node spacing, pan units)

skyward7187:
Branch `line-annotation-cp-anchor-remap`, 6 commits on main 908aa7f06. Assignee: @hendrikschilling.

## Show us what you made

**In one sentence:** A batch of fixes for problems hit while annotating fibers in VC3D: a left click in the cut pane now places the control point where you are looking, the side cut shows only the local wrap, control points can sit closer together with the fiber model still running on short spans, and along-line navigation moves at one speed regardless of how the span was interpolated.

**One real example:** On a PHercParis4 fiber that passes near itself (7316 dense points, 135 controls), a left click in the Current Line Cut pane at line position 573 was resolved to position 2544 on the other pass and replaced control #37 there. "Optimizing..." ran and nothing appeared at the cut; the "/" key at the same spot worked. With this branch the same click adds a control at 573.

**Before:** Clicks near another pass of the same fiber silently moved or added controls thousands of vertices away. The side cut drew every wrap that crossed its plane on top of each other. Controls within 32 vx of each other were merged, and Global-goal spans under 100 vx never ran the fiber model. Shift+wheel and arrow panning were 8x faster in spline spans (32 vx vertices) than in trace spans (4 vx vertices), and the control-point dots jiggled on the strips while panning.

**After this PR:** The click position is resolved through the pane's own line point. The side cut is windowed to half a turn of winding either side of the current position. Strip sampling and the collapse radius are 8 vx. Trace mode traces Global-goal spans from 48 vx and requires a short trace to actually reach its target; Lasagna mode solves from ~63 vx; a failed trace on a span too short for Lasagna goes straight to cspline. Wheel, arrow pan and Space snap move in base-voxel arclength, and the strip dots stay put while panning.

**Proof:** Reproduction log of the click bug (nearest dense vertex to the click was index 2544 at 3.9 vx, the local vertex at 53.9 vx; the edit landed as `control_edit_span_update`). Unit tests: test_line_annotation_generated_views 97, test_lasagna_line_view_surfaces 25, test_fiber_slice_geometry 11, test_fiber_trace3d 47, all passing. Test-driven in the GUI on PHercParis4: click placement at a self-approaching section, half-wrap side cut, dense placement, wheel and arrow pan across trace and spline stretches, dot jiggle gone.

**Why / where this is useful:** Everyone annotating fibers in VC3D, in particular on fibers that pass near themselves and on stretches that need dense manual guidance.

### Relative to #1484 and #1502

Kept as is: the strip model from #1484 (annotation controls and endpoints are the grid supports, spans are resampled by arclength to a target, a span shorter than the target is one column), its invariant that a placement click cannot create a span shorter than one strip column (collapse radius = sampling distance), the fixed seven rows at 32 vx, and #1502's collapse semantics (edits reconstruct spans from the control's authoritative line position, never by nearest 3D distance).

Changed: the along-line sampling target and the collapse radius go from 32 to 8 vx together. The single constant is split into `kLineViewAlongSamplingDistanceBaseVoxels` (8, resampling target, declared along scale, collapse radius) and `kLineViewCrossRowSpacingBaseVoxels` (32, cross rows), so the cross extent is unchanged. The declared grid scale only feeds the surface-UV to grid mapping; the volume level comes from the camera zoom, so the strips do not sample a finer level, and saved strip zooms are unaffected. Not done: no adaptive per-fiber target (the densest-segment approach from the closed #1444); the target is a fixed constant as in #1484.

The click fix is in the spirit of #1502: the controller had started re-deriving a click's line position from the clicked 3D point (#1622), which is the nearest-3D lookup #1502 avoids for collapses.

### Commits
1. **Click position through the pane's line point.** Since #1622 `handleGeneratedControlPoint` replaced the pane's line position with the dense vertex nearest the clicked point whenever they disagreed by more than two vertices (8 vx in trace spans). A click is deliberately off the line, and where another pass of the fiber crosses the cut plane it is nearer to that pass. The placement signal now carries the position's 3D point on the displayed line, and the controller resolves it with `remappedGeneratedLinePositionFromAnchor` (the landing remap's index-continuous nearest match, extracted into a shared helper). A rapid second click on an unlanded splice still follows the renumbering; the clicked point no longer influences the position. Side-pane clicks therefore place at the pane position and pull the line toward the click, as before #1622.
2. **Side cut half-wrap window.** Each line point gets an unwrapped winding angle about the umbilicus (else the volume center; with neither the pane shows the whole line as before). The side overlay blanks points more than half a turn from the current position and drops their control markers. Points are blanked, not erased, and the tail range is captured before the markers are trimmed, so the index-keyed tail detection keeps drawing interior spans.
3. **Strip 8 vx along, collapse radius 8 vx.** See above. Tests that exercise subdivision behavior set the 32 vx target explicitly; a new case pins the default (a 16 vx span is two columns, cross rows stay 32 vx).
4. **Regime-specific short-span cutoffs, span-bounded trace acceptance.** The fixed 100 vx Global-goal cutoff becomes the mode's own minimum: 12 tracer steps in trace mode (`kMinimumTraceSteps * stepVoxels * traceToBaseScale`, 48 vx with defaults), two Lasagna segments in Lasagna mode (`kMinimumLasagnaSegments * segmentLength`, ~63 vx). In trace mode a Global-goal span whose trace is not accepted falls back to Lasagna only if it is at least the Lasagna minimum; shorter ones go straight to cspline (`short span, trace -> cspline`, counted as `cspline_fallback_segments`, trace failure metadata kept). Explicit goals are unchanged and the Lasagna global mode remains selectable. The tracer's endpoint acceptance is bounded per segment to a quarter of the start-to-target distance (`kEndpointAcceptSpanFraction`, a constant rather than a persisted config key because saved fibers are matched on exact config keys), for both one-way target-plane acceptances and the meeting fusion. Behavior change to review: Global spans between 48 and 100 vx are now traced in trace mode (previously spline); those between 48 and 63 vx whose trace fails go to cspline instead of Lasagna; short traces must reach their target.
5. **Arclength units for wheel, arrow pan and Space snap.** One shift+wheel notch moves one strip column (8 vx) times the slice step size; the pan integrator runs in arclength with the position and stop target converted through the position map each tick and lands exactly on the target's line position; Space snaps within a quarter column. Pan speed and its setting are voxels per second (default 96, the previous 12 positions/s at one column per position; new key `lineAnnotation/arrow_pan_speed_vx`, so a saved speed resets once). Still in line positions, out of scope here: the bottom strip's cross-slice spacing, the overview bar, the intersection-inspection follow-slice snap.
6. **Static strip overlays shift with the camera during pans.** The control-point dots were refreshed by the coalesced post-render callback while the camera moved every tick, so they trailed by a tick and snapped. The strip maps surface to scene as (surface - cameraPointer) * scale + viewportCenter, so at a fixed zoom every camera move shifts all overlay items by one common scene delta: the pan tick now moves the existing items in place (`CChunkedVolumeViewer::translateOverlayGroup`) by the delta of a reference surface point recorded when the overlay was last built, and rebuilds only when the zoom changed or the viewer dropped the group. No items are created or destroyed per tick; the landing keeps the full rebuild. The viewer group is addressed by the key registration returned: `applyGeneratedOverlay` derives it through `generatedOverlayGroupKey` (single source of the `line_annotation_overlay_` prefix) and returns it, the dialog stores it per strip and passes it verbatim to the translation, and a missing group during a pan is logged once per pan rather than silently rebuilding. The viewer's overlay-group clearing detaches a group from its map before deleting the items, so the map never names a freed item.

### How to build and test
```
cmake -S volume-cartographer -B volume-cartographer/build -DVC_TESTING=ON
cmake --build volume-cartographer/build --target VC3D test_line_annotation_generated_views test_lasagna_line_view_surfaces test_fiber_slice_geometry test_fiber_trace3d
for t in test_line_annotation_generated_views test_lasagna_line_view_surfaces test_fiber_slice_geometry test_fiber_trace3d; do volume-cartographer/build/bin/$t; done
```

### Risks and limitations
- Strips have 4x the columns. Memory is trivial; rebuild time scales with columns and is small next to the solve.
- Trace acceptance is stricter on short spans; expect more `short span, trace -> cspline` on noisy predictions between 48 and 63 vx.
- Saved arrow-pan speed resets to the default once.
- Docs (`docs/line_annotation_fibers.md`) and `planning/spec.md` updated for the new numbers and the fallback rule; the planning task logs were not touched.

https://claude.ai/code/session_01DfUf5Y68iFTXqQFQKJy1FV

### DIFF
--- a/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
+++ b/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
@@ -2495,8 +2495,11 @@ bool LineAnnotationController::launchSession(LineAnnotationController::SourceKin
     connect(dialog,
             &LineAnnotationDialog::generatedControlPointRequested,
             this,
-            [this](const std::string& name, cv::Vec3f volumePoint, double linePosition) {
-                handleGeneratedControlPoint(name, volumePoint, linePosition);
+            [this](const std::string& name,
+                   cv::Vec3f volumePoint,
+                   double linePosition,
+                   cv::Vec3f lineAnchor) {
+                handleGeneratedControlPoint(name, volumePoint, linePosition, lineAnchor);
             });
     connect(dialog,
             &LineAnnotationDialog::generatedControlPointDeleteRequested,
@@ -7247,7 +7250,8 @@ void LineAnnotationController::handleLineSeed(const std::string& surfaceName,
 
 void LineAnnotationController::handleGeneratedControlPoint(const std::string& surfaceName,
                                                           cv::Vec3f volumePoint,
-                                                          double linePosition)
+                                                          double linePosition,
+                                                          std::optional<cv::Vec3f> lineAnchor)
 {
     auto* pane = paneForSurface(surfaceName);
     if (!pane || !pane->session) {
@@ -7285,17 +7289,23 @@ void LineAnnotationController::handleGeneratedControlPoint(const std::string& su
     for (const auto& point : session.optimizedLine.points) {
         currentLinePoints.push_back(point.position);
     }
-    {
+    if (lineAnchor && std::isfinite((*lineAnchor)[0]) && std::isfinite((*lineAnchor)[1]) &&
+        std::isfinite((*lineAnchor)[2])) {
         // A rapid second click arrives while the previous edit's geometric
         // splice is still waiting for its landing: its line position was
         // measured on the previously DISPLAYED line, and the splice may have
-        // renumbered since. The 3D click point is frame-independent, so when
-        // the two disagree materially, trust the point.
-        const size_t nearestIndex = vc3d::fiber_slice::nearestLinePointIndex(
-            currentLinePoints, toVec3d(volumePoint));
-        if (std::abs(static_cast<double>(nearestIndex) - linePosition) > 2.0) {
-            linePosition = static_cast<double>(nearestIndex);
-        }
+        // renumbered since. Resolve it on the session line through the
+        // position's own 3D line point (index-continuous nearest match, the
+        // same remap the landing applies to the pane position). Where the
+        // frames agree the anchor sits exactly on the line and the position
+        // is unchanged. The CLICKED point must not drive this: a click is
+        // deliberately off the line, and where another pass of the same fiber
+        // runs through the cut plane the click can be nearer to that pass,
+        // which sent the control point thousands of vertices away -- or
+        // collapsed it into that pass's control point -- with nothing
+        // visible happening at the cut the user was looking at.
+        linePosition = vc3d::line_annotation::remappedGeneratedLinePositionFromAnchor(
+            currentLinePoints, toVec3d(*lineAnchor), linePosition);
     }
     const auto cumulativeArclengths =
         vc3d::fiber_slice::cumulativePolylineArclengths(currentLinePoints);
--- a/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
+++ b/volume-cartographer/apps/VC3D/LineAnnotationController.cpp
@@ -11879,6 +11879,39 @@ bool LineAnnotationController::materializeGeneratedViews(LineAnnotationSession& 
     generatedViews.lineUpVectors = views.lineUpVectors;
     generatedViews.stripPositionMap = views.stripPositionMap;
     generatedViews.lineNormals = std::move(orientedNormals);
+    {
+        // Winding angles for the side cut's half-wrap window (see the side
+        // overlay in LineAnnotationDialog). Same center reference chain as
+        // orientedLineNormalsForSession, which already loaded the umbilicus:
+        // umbilicus, else the volume center; with neither the angles stay NaN
+        // and the side cut shows the whole line as before.
+        constexpr float kNanF = std::numeric_limits<float>::quiet_NaN();
+        cv::Vec2f volumeCenterXY{kNanF, kNanF};
+        try {
+            if (const auto volume = _state->currentVolume()) {
+                volumeCenterXY = {static_cast<float>(volume->sliceWidth()) * 0.5f,
+                                  static_cast<float>(volume->sliceHeight()) * 0.5f};
+            }
+        } catch (...) {
+        }
+        generatedViews.lineWindingAngles =
+            vc3d::line_annotation::unwrappedGeneratedWindingAngles(
+                generatedViews.linePoints,
+                [this, volumeCenterXY](const cv::Vec3f& point) -> cv::Vec3f {
+                    if (_scrollUmbilicus) {
+                        try {
+                            return _scrollUmbilicus->vector_to_umbilicus(point);
+                        } catch (...) {
+                        }
+                    }
+                    if (std::isfinite(volumeCenterXY[0])) {
+                        return {volumeCenterXY[0] - point[0],
+                                volumeCenterXY[1] - point[1],
+                                0.0f};
+                    }
+                    return {kNanF, kNanF, kNanF};
+                });
+    }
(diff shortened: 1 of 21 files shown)


hendrikschilling:
Ok I think the only critical problem here is this:

```
  Keyboard panning now performs unbounded static-overlay reconstruction at 62.5 Hz.
     The timer runs every 16 ms and volume-cartographer/apps/VC3D/LineAnnotationDialog.cpp:2914 unconditionally rebuilds both strip overlays. makeGeneratedStaticStripOverlay() copies the  
     line, branch lines, controls, and prediction markers at LineAnnotationGeneratedViews.hpp:1292; the side strip also processes fiber intersections. The path ultimately reaches
     ViewerOverlayControllerBase.cpp:1350 and CChunkedVolumeViewer.cpp:5293, which deletes the previous graphics items and constructs replacements. Main previously coalesced this work     
     behind render updates. This does not form a feedback loop, but it can cause GUI-thread stalls on long fibers or intersection-heavy views. Existing items should be translated or       
     updated in place during the pan, with a full rebuild only at landing.
```

skyward7187:
Agreed, and thanks for the pointer. Replaced the last commit (now 7c25454aa) with an in-place shift instead of the per-tick rebuild:

- The strip viewer maps surface to scene as `(surface - cameraPointer) * scale + viewportCenter`, so at a fixed zoom every camera move shifts all overlay items by one common scene delta. The pan tick now moves the existing static items by that delta (`CChunkedVolumeViewer::translateOverlayGroup`, a `moveBy` over the group), using the scene position of a reference surface point recorded when the overlay was last built.
- Rule: translate while the zoom is unchanged; if the scale changed under the pan, or the viewer dropped the group (surface swap), fall back to one ordinary rebuild, which re-records the baseline. No graphics items are created or destroyed per tick.
- The landing still does the full rebuild, and the coalesced post-render refresh still skips the static overlays while a pan runs.
- Pure rule is unit-tested (`generatedOverlayPanTranslation`); build and the four suites are green.

The other five commits are unchanged.

hendrikschilling:
```
  - [P1] The overlay translation uses the wrong group key, so the performance fix does not work. Static overlays are registered as "line_annotation_overlay_" + staticStripOverlayKey(key)  
    in LineAnnotationGeneratedViews.cpp:247, but LineAnnotationDialog.cpp:3719 calls translateOverlayGroup(staticStripOverlayKey(key), ...) without the prefix. The lookup therefore fails, 
    causing rebuildGeneratedStaticStripOverlays() every 16 ms during keyboard panning. This preserves the original graphics-item reconstruction cost.

  The key generation should be a shared helper used by both registration and translation. The test currently verifies only translation arithmetic, so it misses this integration failure.   
```

skyward7187:
Good catch, thank you. You were right: the lookup used the unprefixed key, missed, and took the "group is gone, rebuild" fallback every tick, so the previous version changed nothing. Replaced the commit again (now 0f22845be):

- `generatedOverlayGroupKey(surfaceName)` is the single source of the `line_annotation_overlay_` prefix; `applyGeneratedOverlay` derives its key through it and now returns the key it registered.
- The dialog stores that returned key with each strip's placement and passes it verbatim to `translateOverlayGroup`, so registration and translation cannot diverge by construction.
- A missing group during a pan is logged once per pan (`Line annotation: static strip overlay group '...' missing during arrow pan`) instead of silently degrading to the rebuild; a zoom change still rebuilds quietly, as intended.
- The key helper is pinned by a unit test; a viewer-level test is not possible in the current test target (Qt Core only), which is why the mismatch got past the earlier round.

hendrikschilling (approved):

Issue #1686, opened 2026-09-03
Ten tests hang indefinitely on Windows, producing no output


Aleredfer:
With #1312 applied on top of main (5479453), the volume-cartographer test suite
builds on Windows for the first time. Running it surfaces ten tests that never
terminate.

Environment: Windows 11 26200, MSYS2 UCRT64, gcc 16.1.0, CMake 4.4.0,
Ninja 1.13.2, preset `ci-windows-mingw` with `VC_TESTING=ON`.

Result: **148/158 passed, 0 assertion failures.** These ten were killed by
`ctest --timeout 90`:

- test_thread_pool_dispatch_lifecycle
- test_chunked_plane_sampler_fallback
- test_volume_live_s3
- test_atlas
- test_volume_pyramid
- test_volume_local
- test_volume_extras
- test_zarr_chunk_fetcher
- test_lasagna_normal_sampler
- test_lasagna_project_volumes

They do not fail; they hang. Observed on each of them:

- no output whatsoever, not even the doctest banner, so the hang precedes any
  reported test case
- a single thread, in Wait/UserRequest
- ~0.1 s of CPU accumulated over nine minutes, so blocked rather than spinning
- no established TCP connections, and no Windows Error Reporting process attached

The same set hangs across two runs five days and 223 upstream commits apart, so
it is reproducible rather than incidental. I could not establish the cause and am
not proposing a fix. What I can rule out:

- **Not a network outage.** dl.ash2txt.org answers HTTP 200 in 0.57 s from the
  same shell. Only three tests carry the `network` label (test_pherc0172_live,
  test_volume_live_s3, test_normal_grid_live), so `check-core-offline` does not
  exclude the other seven.
- **Not an external process that fails to start.** test_atlas and
  test_lasagna_project_volumes contain no QProcess or Python invocation.
- **Not a broadly broken environment**, or the other 148 tests would not pass in
  the same run.
- **Not a common dependency I could identify.** Four of them reference ChunkCache
  heavily; four reference it not at all.

Reporting it as observed rather than diagnosed, because these targets appear never
to have been executed on Windows: `ci-windows-mingw` sets `VC_TESTING=OFF`, and
until #1312 the test tree did not compile there at all. Happy to run further
diagnostics on this machine if that is useful.

Bullo27:
Two different things, I think. Measured on Linux (gcc 13, current main) unless marked *inferred* — I have no Windows box.

[...]

Issue #1318, opened 2026-08-04
test_command_line_tool_runner: a failed check crashes the binary instead of reporting


Aleredfer:
`expectSingleCompletion()` uses `QCOMPARE`, which returns from *that helper function*
rather than from the test slot. When the comparison fails the slot carries on and calls
`completionSucceeded(finished)`, which indexes an empty `QSignalSpy` — out of bounds, so
the binary segfaults instead of reporting a clean failure, taking the remaining tests in
that binary with it.

Observed while running the suite on Windows (gdb backtrace):

```
FAIL!  : cancelFinishesOnceAndRunnerIsReusable() Compared values are not the same
   Actual   (finished.count()): 0
   Expected (1)               : 1
   ...test_command_line_tool_runner.cpp(42) : failure location

Thread 1 received signal SIGSEGV
#0  completionSucceeded (finished=..., index=0) at ...:49
      return finished.at(index).at(1).toBool();
#1  cancelFinishesOnceAndRunnerIsReusable () at ...:167
```

The underlying cause of that particular failure is addressed separately, but the hazard
remains: any future failure of that helper crashes the suite rather than reporting.
Guarding the accessors on `count()`, or restructuring so the check aborts the slot, would
make failures diagnosable.

CVasilopoulos:
I re-checked this on current `main` (d9d70bef1). The helper that lets a failed check fall through into a segfault is still there, and #1412 still fixes it.

I came across this while going through the open volume-cartographer issues, and wanted to check whether it still reproduces on current `main` before anyone spends more time on it.

Setup: Linux x86_64, preset `ci-release-gcc` in `ghcr.io/scrollprize/villa/volume-cartographer:builder-ubuntu-26.04`, `QT_QPA_PLATFORM=offscreen`.

- f4dd95fa9 (#1412) cherry-picks cleanly onto d9d70bef1.
- Unmodified suite: `Totals: 11 passed, 0 failed` on main, and the same with f4dd95fa9.
- To reach the failing path I changed the helper's wait from 3000 ms to 1 ms in a scratch build, so the spy is still empty when the check runs.

main (excerpt):
```
FAIL!  : TestCommandLineToolRunner::normalExitFinishesOnce() Compared values are not the same
   Actual   (finished.count()): 0
   Expected (1)               : 1
   Loc: [/src/apps/VC3D/test/test_command_line_tool_runner.cpp(46)]
[...]

PR #1687, opened 2026-09-03
tests: make the volume-cartographer suite build and pass on Windows, and fix cancel() for console tools

Aleredfer:
Reopens #1312, which the staleness bot closed on 28 August. That one is on me:
@pmh47 asked for a rebase and the destructor change, the work was done, and I did
not push it in time. GitHub would not let me reopen #1312 after the force-push, so
this is the same branch as a fresh PR. The review history lives there.

**What is different from #1312**

1. **Rebased** onto main (5479453).

2. **The two destructors**, as asked. `CommandLineToolRunner::~CommandLineToolRunner`
   and `SegmentRenderThread::~SegmentRenderThread` both had the bare `terminate()` +
   `waitForFinished(3000)` pattern, which on Windows always burns its full timeout on
   the GUI thread and then deletes a still-running QProcess. Both take the blocking
   variant rather than `cancel()`'s deferred timer, since a `singleShot` would never
   fire from a destructor. Five call sites already did terminate -> wait -> kill
   (LasagnaServiceManager, NeuralTraceServiceManager, SeedingWidget, SpiralSshTunnel,
   SpiralServiceManager); these two were the outliers, so this follows the existing
   convention rather than introducing one.

3. **New POSIX-only calls had to be fixed again.** `core/test/test_line_annotation_generated_views.cpp`
   has taken fresh `setenv`/`unsetenv` calls three times in four weeks - the original
   set, four more by 26 August, four more by 31 August - each of which blocks the
   Windows build. That is this PR in miniature: nothing builds the suite on Windows,
   so nothing catches them. This stops the bleeding rather than mopping it up once.

4. **Three lines smaller.** main has since adopted the same write ordering in
   `test_chunk_cache_persist.cpp` that this PR proposed, without the comment
   explaining why the ordering matters. I kept the comment.

**Verification** on Windows, MSYS2 UCRT64, gcc 16.1.0, CMake 4.4.0, Ninja 1.13.2,
preset `ci-windows-mingw` with `VC_TESTING=ON`:

- full suite builds: 0 errors
- ctest: **148/158 passed, 0 assertion failures**

The remaining 10 are timeouts, not failures, and none touches a file in this PR.
They produce no output at all and hang indefinitely; filed separately as #1686
rather than folded in here, since they are pre-existing and unrelated.

I have not been able to run this on the project's Windows CI, so treat the above as
locally verified rather than CI-green.

---

The test tree does not compile on Windows. `ci-windows-mingw` sets `VC_TESTING=OFF`, so the registered test targets are only ever built on Linux and macOS â€” where the POSIX APIs the tests reach for happen to exist. Windows ships an official installer on every release, so nothing that breaks there is currently caught.

Making the suite build surfaced a real bug, which is why this is one change and not two: with the tests finally running, `cancelFinishesOnceAndRunnerIsReusable` failed on Windows, and the failure crashed the test binary outright. Splitting the portability work from the fix would leave a PR that cannot claim a green suite.

**No `#ifdef` anywhere in this PR.** Every fix is portable by construction.

### Compile failures â€” these blocked the whole tree

| file | problem | fix |
|---|---|---|
| `apps/VC3D/test/test_command_line_tool_runner.cpp` | `std::raise(SIGKILL)` â€” POSIX-only | `std::abort()` |
| `core/test/test_line_annotation_generated_views.cpp` | `setenv`/`unsetenv` â€” POSIX-only | `qputenv`/`qunsetenv` + explicit `<QtGlobal>` |

Production code already handles both correctly, which is what makes this a test-tree gap rather than a codebase-wide one: `LasagnaServiceManager.cpp:877` guards its `SIGKILL` with `#ifdef Q_OS_UNIX`, and `VCAppMain.cpp:53` carries a `vcSetEnv` helper commented *"POSIX setenv with an MSVCRT/UCRT fallback (Windows has no setenv)"*.

`std::abort()` is used rather than another signal under an `#ifdef` because under MinGW every `raise(SIG*)` returns through the default handler and exits normally, which `QProcess` reports as `NormalExit` â€” while the test asserts `CrashExit`:

| | `raise(SIGABRT/SIGSEGV/SIGILL/SIGTERM)` | `std::abort()` |
|---|---|---|
| Windows (MinGW) | NormalExit, code 3 | **CrashExit** |
| Linux (glibc) | CrashExit | **CrashExit** |

It is correct on both, and removes the now-unused `<csignal>` include.

### Runtime failures

**Deleting a file that is still mapped or open** (5 tests) â€” `test_gridstore.cpp`, `test_gridstore_branches.cpp`, `test_gridstore_malformed.cpp`, `test_normal_grid_volume.cpp`, `test_quadsurface_extras.cpp`. POSIX allows unlinking an open or mapped file; Windows refuses. In each case the owner of the mapping was still in scope at the `remove()` call, so it is now confined to a nested scope or released explicitly. Not a production defect: `GridStore` holds its mapping in an RAII member and `~GridStore()` releases it correctly.

**`isExecutable()` is extension-based on Windows** (1 test) â€” `failedStartFinishesOnce` built its stand-in program from a `QTemporaryFile` whose random name has no extension, so the runner rejected it before starting a process, a different code path from the one the test exercises. The temporary now carries a `.exe` suffix, harmless on POSIX where the permission bit set just below is what counts.

**A genuine race** (1 test) â€” `stats: startup scan ignores files newer than its cutoff` wrote `post.bin` *after* constructing the cache. The scan runs on a worker thread with a cutoff captured at construction, so the write raced it: the scan could read `post.bin` before its future timestamp was applied and count it â€” 49 bytes instead of 32. This can fail on any loaded machine; Windows merely exposes it by being slower at directory iteration. The file is now stamped before the cache exists, which removes the race without changing what the test asserts, and a previously ignored `error_code` is checked so a broken premise fails loudly instead of as a confusing byte-count mismatch.

### The cancellation bug the tests found

`cancel()` only called `QProcess::terminate()`, which posts `WM_CLOSE`. Console applications have no message loop, so the `vc_*` tools ignore it: the user cancels, the interface reports success, and the process keeps running for hours. This also hung cancellation from the agent bridge.

Fixed with a deferred `kill()` fallback after a grace period, so `terminate()` still gets its chance to shut the process down cleanly first. The `QProcess` object is reused across runs, so the deferred kill is guarded against landing on a later run that started in the meantime.

Adds a regression test for that guard, verified to fail without the change: it cancels a long-running helper, starts another immediately, and asserts the replacement is still alive after the grace period has elapsed.

## How to build and run

```bash
# Linux
cmake --preset ci-release-gcc -DVC_TESTING=ON
ninja -C build/ci-release-gcc
cd build/ci-release-gcc && QT_QPA_PLATFORM=offscreen ctest --timeout 180 -j4

# Windows (MSYS2 UCRT64)
cmake --preset ci-windows-mingw -DVC_TESTING=ON
ninja -C build/ci-windows-mingw
cd build/ci-windows-mingw && QT_QPA_PLATFORM=offscreen ctest --timeout 180 -j4
```

## How this was verified

**Windows** (MSYS2 UCRT64, gcc 16.1.0, Qt 6.11.1, CMake 4.4.0): the test tree went from *not compiling at all* to **112/112 passing**.

Every assertion in the touched tests is preserved one-for-one; only scoping and indentation changed. No test was weakened to make it pass.

## Risks and limitations

- I have only run the suite on Windows against this branch, so please let CI confirm Linux and macOS. The changes are confined to test code plus the one `cancel()` fix and add no platform conditionals, so POSIX behaviour should be unchanged â€” but that is reasoning, not a measurement, and I would rather say so than imply otherwise.
- On Linux `terminate()` already stops these tools, so the deferred `kill()` normally finds the process gone; the guard covers that case.
- `std::abort()` may produce a core dump where `ulimit` allows it. That is noise in a test log, not a failure.

### DIFF
--- a/volume-cartographer/apps/VC3D/CommandLineToolRunner.cpp
+++ b/volume-cartographer/apps/VC3D/CommandLineToolRunner.cpp
@@ -9,12 +9,17 @@
 #include <QMessageBox>
 #include <QClipboard>
 #include <QApplication>
+#include <QPointer>
+#include <QTimer>
 
 #include <cmath>
 #include <utility>
 
 
 namespace {
+
+// Grace period between terminate() and the kill() fallback in cancel().
+constexpr int kCancelGracePeriodMs = 1000;
 
 QString quoteArg(const QString& arg)
 {
@@ -84,8 +89,17 @@ CommandLineToolRunner::~CommandLineToolRunner()
 {
     if (_process) {
         if (_process->state() != QProcess::NotRunning) {
+            // Same terminate()/kill() problem as cancel(), for the same reason:
+            // the console-mode vc_* tools have no message loop to receive
+            // WM_CLOSE, so terminate() alone leaves the child running and this
+            // wait burns its full timeout on the GUI thread every time. cancel()
+            // defers the kill() through a timer because it must not block; a
+            // destructor cannot, since the timer would never fire.
             _process->terminate();
-            _process->waitForFinished(3000);
+            if (!_process->waitForFinished(kCancelGracePeriodMs)) {
+                _process->kill();
+                _process->waitForFinished();
+            }
         }
         delete _process;
     }
@@ -457,9 +471,28 @@ bool CommandLineToolRunner::executeCustomCommand(const QString& command,
 
 void CommandLineToolRunner::cancel()
 {
-    if (_process && _process->state() != QProcess::NotRunning) {
-        _process->terminate();
+    if (!_process || _process->state() == QProcess::NotRunning) {
+        return;
     }
+
+    _process->terminate();
+
+    // On Windows terminate() posts WM_CLOSE, which the console-mode vc_* tools have
+    // no message loop to receive: the process keeps running and cancelling silently
+    // does nothing at all. Fall back to kill(), the same way LasagnaServiceManager
+    // and the other service managers already do. Deferred rather than a blocking
+    // waitForFinished() because cancel() runs on the GUI thread.
+    // The QProcess object is created once and reused across runs, so the pid is what
+    // identifies *this* run: without it a cancel followed by a quick relaunch would
+    // let the timer kill the new process instead.
+    QPointer<QProcess> process(_process);
+    const qint64 cancelledPid = _process->processId();
+    QTimer::singleShot(kCancelGracePeriodMs, this, [process, cancelledPid]() {
+        if (process && process->state() != QProcess::NotRunning
+            && process->processId() == cancelledPid) {
+            process->kill();
+        }
+    });
 }
 
 bool CommandLineToolRunner::isRunning() const
--- a/volume-cartographer/apps/VC3D/SegmentRenderThread.cpp
+++ b/volume-cartographer/apps/VC3D/SegmentRenderThread.cpp
@@ -2,6 +2,11 @@
 #include <QDir>
 #include <QFileInfo>
 
+namespace {
+// Grace period between terminate() and the kill() fallback.
+constexpr int kRenderStopGraceMs = 1000;
+}
+
 SegmentRenderThread::SegmentRenderThread(QObject *parent)
     : QThread(parent)
     , m_scale(1.0f)
@@ -15,8 +20,15 @@ SegmentRenderThread::~SegmentRenderThread()
 {
     if (m_process) {
         if (m_process->state() != QProcess::NotRunning) {
+            // vc_render_tifxyz is a console-mode tool: on Windows terminate()
+            // posts WM_CLOSE, which it has no message loop to receive, so this
+            // waits out its full timeout and then deletes a still-running
+            // QProcess. Fall back to kill(), the way the service managers do.
             m_process->terminate();
-            m_process->waitForFinished(3000);
+            if (!m_process->waitForFinished(kRenderStopGraceMs)) {
+                m_process->kill();
+                m_process->waitForFinished();
+            }
         }
         delete m_process;
     }
--- a/volume-cartographer/apps/VC3D/test/test_command_line_tool_runner.cpp
+++ b/volume-cartographer/apps/VC3D/test/test_command_line_tool_runner.cpp
@@ -1,6 +1,5 @@
-#include <csignal>
-
 #include <QApplication>
+#include <QDir>
 #include <QFile>
 #include <QScopeGuard>
 #include <QTemporaryFile>
@@ -76,8 +75,12 @@ int runChild(const QStringList& arguments)
         return 0;
     }
     if (mode == "crash") {
-        std::raise(SIGKILL);
-        return 1;
+        // std::abort() rather than raise(SIGKILL): SIGKILL is POSIX-only, so this
+        // file did not compile on Windows at all. It is also the only portable way
+        // to get QProcess::CrashExit on both platforms — under MinGW every
+        // raise(SIG*) returns through the default handler and exits normally, which
+        // QProcess reports as NormalExit.
+        std::abort();
     }
     if (mode == "write-omp" && marker + 2 < arguments.size()) {
         QFile output(arguments.at(marker + 2));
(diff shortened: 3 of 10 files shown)


Aleredfer:
Still current and still wanted — flagging before the inactivity bot reaches this
one, since its predecessor #1312 was closed that way rather than rejected.

Status as of today:

- **Merges clean on current main** (4b3c728), 55 commits on from the base it was
  opened against. No conflicts.
- **Verified on Windows** (MSYS2 UCRT64, gcc 16.1.0, Release, `VC_TESTING=ON`):
  the full suite builds with 0 errors and ctest gives **148/158 passed, 0
  assertion failures**. The 10 remaining are pre-existing hangs unrelated to this
  PR, filed separately as #1686 — where @Bullo27's #1688 has since identified and
  fixed the cause of two of them, and I confirmed that fix on Windows.
- **No interaction with #1412**: @CVasilopoulos checked this in #1318 and reports
  that this PR "doesn't call the helper or the accessors, and it merges cleanly
  with this change in either order."

@pmh47 you asked for the rebase and the destructor change before this could be
merged; both are in. Happy to rebase again, split it, or answer anything that
would help it land.

@hendrikschilling if a review needs more time, would the `keep-open` label be
appropriate here? This is the second time around for this change and I would
rather it not lapse again for the same reason.

One thing worth recording either way: `core/test/test_line_annotation_generated_views.cpp`
took new POSIX-only calls three times in six weeks while this sat open, each of
which breaks the Windows build again. That is the case for this PR — nothing
builds the suite on Windows, so nothing catches them.

pmh47:
Reviewed. The portability work is solid and preserves every assertion — the nested scopes before `remove()`, the `qputenv`/`qunsetenv` conversion, the `.exe` temp name, and keeping the `test_chunk_cache_persist.cpp` comment are all good. Two things in the destructors I'd like fixed before merge, both about the POSIX side rather than Windows.

**1. `CommandLineToolRunner.cpp:89` — the post-`kill()` wait is unbounded (medium)**

```cpp
if (!_process->waitForFinished(kCancelGracePeriodMs)) {
    _process->kill();
    _process->waitForFinished();   // default timeout: 30000 ms
}
```

The five call sites this cites as precedent all pass an explicit bound to the second wait — `LasagnaServiceManager.cpp:888`, `NeuralTraceServiceManager.cpp:351`, `SpiralSshTunnel.cpp:170`, `SpiralServiceManager.cpp:536` all use `waitForFinished(1000)`, and `SeedingWidget.cpp:2483` doesn't wait after `kill()` at all. The bare call here takes the 30 s default, so the worst-case GUI-thread block on shutdown goes from today's 3 s to ~31 s if the killed child is slow to reap (a `vc_*` tool stuck in D-state on a stalled network mount, say). Following the convention the PR invokes — `waitForFinished(1000)` — fixes it.

**2. Same file, same hunk — the grace period drops 3000 ms → 1000 ms on every platform**

On POSIX `terminate()` already works, so the only effect of the shorter window is that a tool needing 1–3 s to flush after SIGTERM now gets SIGKILLed mid-write instead of exiting cleanly. That's a real risk for the `vc_*` tools that finalize a tifxyz/zarr on shutdown. The Windows problem this PR is solving is that `terminate()` never works there, which argues for adding the `kill()` fallback — not for shortening the window in which a well-behaved process is allowed to finish. Keeping 3000 ms for the terminate wait preserves current POSIX behaviour exactly and still bounds Windows at 3 s + 1 s instead of today's unbounded leak.

**3. `SegmentRenderThread.cpp:20` — both of the above apply verbatim (low/medium)**

`kRenderStopGraceMs` is 1000 and the post-`kill()` wait is again unbounded. Worth noting additionally that `m_process` is constructed in `run()` on the worker thread while this destructor runs on the owning thread, so the added `kill()` + `waitForFinished()` widens an already cross-thread `QProcess` use; an explicit timeout at least bounds how long that window stays open.

None of this touches the `cancel()` fix itself, which I think is the right shape.

---

Separately, not blocking, but it will bite on rebase: `core/test/test_line_annotation_generated_views.cpp` has taken six more `setenv`/`unsetenv` calls on `main` since this branch's base (around lines 1232–1341). They'll need the same `qputenv`/`qunsetenv` treatment or the "builds on Windows" claim lapses on the merge commit — which is, as the description says, exactly the pattern. No other open PR touches that file or mentions `qputenv`, so nothing else is going to pick them up.

---

Authored by Claude Opus 5

pmh47:
I suggest also opening a PR that resolves the new windows test fails due to qsetenv etc

hendrikschilling:
```
  1. P1: Destructor waits up to 30 seconds after kill()
     CommandLineToolRunner.cpp:95 and SegmentRenderThread.cpp:26 call waitForFinished() without a timeout after killing the process. Qt’s default is 30 seconds, so shutdown can block the  
     GUI for roughly 31 seconds. Use an explicit short timeout, matching the other service managers.

  2. P2: POSIX termination grace period is shortened from 3 seconds to 1 second
     CommandLineToolRunner.cpp:21 and SegmentRenderThread.cpp:8 change existing behavior on every platform. A tool that needs 1–3 seconds to flush output or finalize files after
     terminate() can now be forcibly killed. Keep the existing 3000 ms grace period and use the kill fallback afterward.

  3. Low-probability race: PID matching is not completely safe
     The deferred cancellation identifies the old run only by PID. If the OS reuses that PID before the timer fires, the replacement process could be killed. A per-run generation/token    
     would be more robust, though this is unlikely in normal use.
```

Issue #1404, opened 2026-08-11
vc_render_tifxyz: default --cache-gb 16 stalls silently on smaller Docker hosts, and resume counts zero-byte TIFFs as complete


DarthCeltic:
Two things in `vc_render_tifxyz` that both fail quietly.

**`--cache-gb` defaults to 16.** On a Docker host with a 15 GB memory limit the container sits at the ceiling and produces no output and no log line for as long as you leave it — indistinguishable from a slow render. I killed it twice before thinking to look at the memory graph, where it was sitting at 14.33 GB / 15.17 GB. With `--cache-gb 4` the same surface renders at about 5.7 GB resident.

Clamping the default to something under the visible container limit, or just warning when the requested cache exceeds available memory, would save people that hour.

**An interrupted render leaves zero-byte TIFFs, and the resume path counts them as done.** After the first kill, re-running printed:

```
[tif] all slices exist, skipping.
```

The check tests whether the file exists rather than whether it decodes, so a crashed render permanently poisons its own output directory until the stubs are deleted by hand. Checking that each existing slice actually opens would make resume safe after a crash, which is the case where you most want it.

Bullo27:
Status: the second half of this report (resume counting unreadable slices as done) is fixed on `main` as of c490284 (#1695, merged today). With `--resume`, each existing slice must now open as a TIFF and every tile must decode; otherwise the tool prints `[tif] N of M existing slices are not complete, readable TIFFs …, re-rendering all slices.` instead of skipping. In my reproduction the stubs left by a kill were 4 MB (header plus written tiles), not zero bytes, but both cases are caught.

The first half (the `--cache-gb 16` default stalling at the container limit) is not addressed by that PR and this issue is still the right place for it.

*AI-assisted (Claude), human-directed account.*

nerln:
For the first half: #1807 (opened 16 Sep) adds the warning this issue asks for on the `--cache-gb` side — one stderr line when the requested cache is not below the usable memory (cgroup limit if set, otherwise physical RAM), no clamp and no default change, 24 lines. It applies cleanly on `main` at c490284, after #1695. Happy to adjust the wording or the check if either is unwanted.

PR #1695, opened 2026-09-03
vc_render_tifxyz: do not treat torn per-slice TIFFs as finished output (refs #1404)

Bullo27:
**In one sentence:** `vc_render_tifxyz` no longer treats the unreadable per-slice TIFFs left behind by a killed render as finished output, so rerunning the same command after an OOM kill or `--timeout` repairs the output directory instead of printing `[tif] all slices exist, skipping.` forever.

**One real example:** Rendering PHerc1447 segment `auto_grown_20250703034159599` against the level-3 volume with `vc_render_tifxyz -v <volume> -s <segment> --scale 16 -g 3 -n 8 --tif-output out` (8 slices of 10920×7400 uint8, 12.9 s uninterrupted), I killed the process with SIGKILL at 6.0 s (band 25/58, 43%) as a stand-in for the Docker OOM kill in #1404. That left `out/00.tif … 07.tif` at 4.0–4.4 MB each: a TIFF header whose first-IFD offset is 0, plus the tiles written before the kill. `tifffile` reports "contains no pages", `cv2.imread` returns `None`, `tiffinfo` prints nothing. Rerunning the identical command with `--resume` printed `[tif] all slices exist, skipping.` and exited 0 after 0.1 s. (This was first measured on an older `main` where the skip fired with or without `--resume`; #1528 has since gated it behind `--resume`, so the rebased fix now sits inside that `if (resumeFlag && numParts <= 1)` guard and the case it fixes engages on `--resume`.) After this PR the same rerun prints `[tif] 8 of 8 existing slices are not complete, readable TIFFs (interrupted render or damaged file?), re-rendering all slices.`, renders 58/58 bands and exits 0 after 13.0 s; `cmp` against the uninterrupted render reports all 8 slices byte-identical.

**Before:** the "skip if all slices exist" check in the `--tif-output` path tested `std::filesystem::exists()` only. `TiffWriter` opens the file (header written) in its constructor, writes tiles as they are produced, and writes the directory only in `close()`, so a process killed mid-render leaves every slice existing but without a directory. Every later run skipped them, so the only way out was deleting the files by hand — exactly as #1404 describes.

**After this PR:** each existing slice must also open as a TIFF and every tile of it must decode to its full size (`isReadableTiff()`: `TIFFOpen` for read, then `TIFFReadEncodedTile` over all tiles, with libtiff's error/warning handlers muted for the probe). If any slice is missing, torn or damaged, all slices are rendered again — the same all-or-nothing behaviour the check already had for a missing slice — and the bad slices are reported. Finished output is still skipped with the files untouched; because the probe now reads the whole file, that skip costs 2.2 s instead of 0.1 s for the 8 slices above (73 MB of LZW TIFF). A fresh render is unchanged.

**Proof:** terminal transcripts before/after on the same torn directory, the mixed case (5 finished + 3 torn slices → `3 of 8 …`), the finished case, and the recovered slice:

vc_render_tifxyz torn-slice resume before/after

| output directory state | before (main @ 908aa7f06) | after this PR |
|---|---|---|
| 8 slices torn by SIGKILL at 43% | `[tif] all slices exist, skipping.` — exit 0, 0.1 s, files stay unreadable | `8 of 8 existing slices are not complete, readable TIFFs …, re-rendering all slices.` — exit 0, 13.0 s, 8/8 slices `cmp`-identical to the uninterrupted render |
| 5 finished slices + 3 torn copies | `[tif] all slices exist, skipping.` | `3 of 8 …` — 12.9 s, 8/8 identical |
| 8 finished slices | `[tif] all slices exist, skipping.` — 0.1 s | same message, 2.2 s (every tile decoded), files untouched |
| 8 finished slices, 4 KiB of one slice's tile data overwritten with 0xFF (directory intact, `TIFFOpen` succeeds) | `[tif] all slices exist, skipping.` — 0.1 s, the damaged slice stays | `1 of 8 …` — 14.7 s, 8/8 identical |
| kill at 43% with the patched binary, then rerun | — | `8 of 8 …` — 13.5 s, 8/8 identical |

**Why / where this is useful:** a render is interrupted precisely in the situations where people rely on rerunning it — container OOM kills (the #1404 case), `--timeout`, a lost SSH session. Without this, the rerun looks successful (exit 0, "all slices exist") while the directory holds files no reader can open, and the failure only surfaces later, when something tries to read the layers.

## Details

Refs #1404 (report: @DarthCeltic). This addresses the second half of that issue — "checking that each existing slice actually opens would make resume safe after a crash". #1415 by @NicolasHuberty took the same approach for this half (together with a cache clamp) and was auto-closed for inactivity on 2026-08-27 with only bot comments; this PR is limited to the resume check, is verified on real scroll data, comes with a test, and leaves the `--cache-gb` default alone. In my reproduction the torn files are 4 MB (header plus the tiles written before the kill) rather than the zero-byte files reported in the issue; the check treats both the same, and the tests cover an empty file, a torn file, a finished one and a finished one whose tile data was damaged after the directory was written.

**What changed** (4 files, +192/−5):
- `core/util/Tiff.hpp` / `Tiff.cpp`: new `bool isReadableTiff(const std::filesystem::path&)` next to `TiffWriter` — `TIFFOpen(path, "r")` succeeds only if the first directory is readable, then every tile (or strip, for untiled files) is decoded with `TIFFReadEncodedTile`/`Strip` and must come back at its full size, so a truncated, missing or undecodable payload behind an intact directory is rejected too; libtiff's global error/warning handlers are set to null around the probe and restored, so a bad file does not spam stderr (the caller decides what to report). The header comment states what is and is not checked: structure and decodability, not pixel values.
- `apps/src/vc_render_tifxyz.cpp`: the single-part `--tif-output` skip check counts missing and torn slices instead of testing existence only; skips only when both are zero; logs `N of M existing slices are not complete, readable TIFFs (interrupted render or damaged file?), re-rendering all slices.` when such slices are found. The multi-part `.partN.tif` path (`numParts > 1`) and zarr output are untouched.
- `core/test/test_tiff.cpp`: new `TEST_CASE("isReadableTiff: finished vs torn, empty and absent files")` — writes a 32×32 tiled TIFF with `TiffWriter`, copies the file after two of four tiles and before `close()` (what a killed writer leaves on disk), and checks finished → true, torn → false, empty file → false, absent → false. Registered through the existing `test_tiff` target (label `vc-core`, CI core shard). With the probe reduced to `exists()` the two torn/empty checks fail and nothing else does. Second `TEST_CASE("isReadableTiff: intact directory over corrupt tile data")` — writes a 128×128 file of random bytes as four 64×64 LZW tiles, overwrites one tile's payload with 0xFF (offsets and byte counts taken from libtiff itself), and checks that the file still opens with `TIFFOpen` but is rejected by `isReadableTiff`.

**Tested:** first round, commit 0e1e0ef, Ubuntu 24.04.4, gcc 13.3.0. ASan+UBSan Debug build: `test_tiff` 14/14, full `vc-core` label 142/142. Real-data runs above used a Release build of the same commit. The skip path on 8 finished slices takes the same 0.1 s as before. Rebased on 2026-09-20 onto main `23c3d759e` as `4373ab9b9` (`git range-diff`: patch unchanged; main has not touched these files since `757f70c01`); on that Release build the torn-directory rerun with `--resume` again prints `8 of 8 …`, exits 0 after 13.9 s with all 8 slices `cmp`-identical to the uninterrupted render, and the finished directory is skipped in 0.15 s with the files untouched.

Review round (2026-09-21, commit `8742eb0d7` on top, Release build of this branch): `test_tiff` 15/15. Same data and command as above: torn directory → `8 of 8 …`, exit 0, 13.0 s, 8/8 `cmp`-identical; finished directory → `all slices exist, skipping.`, 2.2 s, files untouched; finished directory with 4 KiB of `03.tif`'s tile data overwritten → `1 of 8 …`, exit 0, 14.7 s, 8/8 identical (the previous head skipped that directory in 0.14 s and left the damaged slice in place).

**Not changed:** the `--cache-gb` default and the silent stall at the container memory limit (first half of #1404 — a separate decision about defaults/warnings); per-slice partial resume (a torn directory is re-rendered in full, as a directory with a missing slice already was); zarr chunk resume.

**From the author:** I contribute to VC through an AI-assisted workflow (Claude, directed by me). Going through older open issues, and looking for reports that could be reproduced and fixed with a small, tested change, we found the following. #1404 was still open, the earlier fix (#1415) had gone stale without a review, and the failure is easy to reproduce on real data: kill a render midway, rerun it, and the tool declares the output complete. We picked it because a resume path that keeps unreadable output does not work in the one situation it exists for, namely renders interrupted by memory limits or timeouts, which is the case reported in #1404 (a Docker OOM kill). The PR only touches the resume check and leaves the cache-size default alone. I reviewed the investigation, the issue history and the evidence; every number in this PR was produced by running the stated commands on my machine.

**AI disclosure:** investigation, patch, test and this write-up were produced with Claude (Anthropic) under my direction; every number above comes from running the stated commands on the stated data on my machine.

### DIFF
--- a/volume-cartographer/apps/src/vc_render_tifxyz.cpp
+++ b/volume-cartographer/apps/src/vc_render_tifxyz.cpp
@@ -1680,15 +1680,26 @@ int main(int argc, char *argv[])
                 return p;
             };
 
-            // Skip if all exist
+            // Skip if all exist and open as TIFFs. A render killed mid-way
+            // (OOM, --timeout) leaves every slice with a header but no
+            // directory -- TiffWriter writes the IFD in close() -- and a bare
+            // exists() check would treat those torn files as done and skip
+            // them on every rerun (#1404).
             bool tifSkip = false;
             if (numParts <= 1) {
-                bool all = true;
-                for (int z = 0; z < tifSlices; z++) if (!std::filesystem::exists(makePartPath(z))) { all = false; break; }
-                if (all) {
+                int missing = 0, torn = 0;
+                for (int z = 0; z < tifSlices; z++) {
+                    const auto p = makePartPath(z);
+                    if (!std::filesystem::exists(p)) missing++;
+                    else if (!isReadableTiff(p)) torn++;
+                }
+                if (missing == 0 && torn == 0) {
                     if (!wantZarr) { logPrintf(stdout, "[tif] all slices exist, skipping.\n"); return true; }
                     logPrintf(stdout, "[tif] all slices exist, skipping tif output.\n");
                     tifSkip = true;
+                } else if (torn > 0) {
+                    logPrintf(stdout, "[tif] %d of %d existing slices are not readable TIFFs (interrupted render?), re-rendering all slices.\n",
+                              torn, tifSlices - missing);
                 }
             }
             if (!tifSkip) {
--- a/volume-cartographer/core/include/vc/core/util/Tiff.hpp
+++ b/volume-cartographer/core/include/vc/core/util/Tiff.hpp
@@ -90,6 +90,14 @@ private:
     std::filesystem::path _path;     // For error messages
 };
 
+// True if `path` opens as a TIFF with a readable first directory.
+// TiffWriter (like libtiff in general) writes the directory in close(), so
+// a file left behind by a writer that was killed mid-way has a header and
+// tile data but no directory: exists() says "done", this says "torn".
+// libtiff diagnostics are suppressed while probing; the caller decides
+// what to report.
+bool isReadableTiff(const std::filesystem::path& path);
+
 // Merge partial .partN.tif files into final TIFFs.
 // Scans outputPath (or its parent directory) for .partN.tif files,
 // merges them tile-by-tile, and removes the part files.
--- a/volume-cartographer/core/src/Tiff.cpp
+++ b/volume-cartographer/core/src/Tiff.cpp
@@ -312,6 +312,26 @@ void TiffWriter::close() {
 }
 
 // ============================================================================
+// isReadableTiff
+// ============================================================================
+
+bool isReadableTiff(const std::filesystem::path& path)
+{
+    // TIFFOpen("r") reads the first directory and fails without one; a
+    // torn file makes libtiff complain through its global handlers, so
+    // mute them for the probe and hand back a plain yes/no.
+    TIFFErrorHandler prevError = TIFFSetErrorHandler(nullptr);
+    TIFFErrorHandler prevWarning = TIFFSetWarningHandler(nullptr);
+    TIFF* tf = TIFFOpen(path.string().c_str(), "r");
+    TIFFSetErrorHandler(prevError);
+    TIFFSetWarningHandler(prevWarning);
+    if (!tf)
+        return false;
+    TIFFClose(tf);
+    return true;
+}
+
+// ============================================================================
 // mergeTiffParts implementation
 // ============================================================================
 
--- a/volume-cartographer/core/test/test_tiff.cpp
+++ b/volume-cartographer/core/test/test_tiff.cpp
@@ -1,6 +1,7 @@
 // Coverage for core/src/Tiff.cpp. atomicImwriteMulti has its own test elsewhere;
 // here we exercise voxelSizeToDpi, writeTiff (typed & untiled), TiffWriter
-// tiled writes, type-conversion paths, and mergeTiffParts no-input.
+// tiled writes, type-conversion paths, isReadableTiff, and mergeTiffParts
+// no-input.
 
 #define DOCTEST_CONFIG_IMPLEMENT_WITH_MAIN
 #include <doctest/doctest.h>
@@ -11,6 +12,7 @@
 #include <opencv2/imgcodecs.hpp>
 
 #include <filesystem>
+#include <fstream>
 #include <random>
 #include <stdexcept>
 #include <string>
@@ -199,6 +201,31 @@ TEST_CASE("TiffWriter: move construction transfers file handle")
     w2.writeTile(0, 0, tile);
     w2.close();
     CHECK(fs::exists(p));
+    fs::remove_all(d);
+}
+
+TEST_CASE("isReadableTiff: finished vs torn, empty and absent files")
+{
+    auto d = tmpDir("readable");
+    auto p = d / "out.tif";
+    auto torn = d / "torn.tif";
+    {
+        TiffWriter w(p, 32, 32, CV_8UC1, 16, 16);
+        cv::Mat tile = ramp(16, 16, CV_8UC1);
+        w.writeTile(0, 0, tile);
+        w.writeTile(16, 0, tile);
+        // What a writer killed at this point leaves on disk: header and
+        // tiles, no directory (that is written by close()).
+        fs::copy_file(p, torn);
+        w.writeTile(0, 16, tile);
+        w.writeTile(16, 16, tile);
+    }
+    CHECK(isReadableTiff(p));
+    CHECK(fs::file_size(torn) > 0);
+    CHECK_FALSE(isReadableTiff(torn));
+    { std::ofstream empty(d / "empty.tif"); }
+    CHECK_FALSE(isReadableTiff(d / "empty.tif"));
+    CHECK_FALSE(isReadableTiff(d / "absent.tif"));
     fs::remove_all(d);
 }
 


nerln:
I reproduced this independently on the failure #1404 describes (I had patched it locally too; never opened, superseded by this PR).

Two platforms, both starting from a zero-byte `01.tif` standing in for a killed render:

- macOS arm64, `main` @ 76370e1a6 (our usual LLVM/OpenBLAS/lld toolchain, PCH off, OpenCV 5): rendering 3 slices of a PHerc0125 segment, emptying `01.tif`, then rerunning with `--resume` prints `[tif] all slices exist, skipping.` and exits 0 with `01.tif` still 0 bytes.
- The official Windows build of the same series, `VC3D-be09a85-2026-09-11-win64.zip` (be09a85), on a second machine: identical result on different data (PHerc1218) — same log line, `01.tif` stays 0 bytes, exit 0, no source changes, no rebuild.

Then your binary (`fix/render-tifxyz-tif-resume-torn-slices`, built here at 0e1e0ef77): same 3-slice render, same empty `01.tif`. Without `--resume`: `[tif] 1 of 3 existing slices are not readable TIFFs (interrupted render?), re-rendering all slices.`, exit 0, `01.tif` back to 1925266 bytes. With `--resume`: the identical message and result — on this branch the skip check isn't gated on `resumeFlag` at all (`if (numParts <= 1)`), so it fires the same way either way, which matches what your PR description says. Both runs fix exactly the case we reproduced on macOS and on the Windows build.

One thing to flag before this merges: the branch is based on an old `main` (merge-base 908aa7f06, 2026-09-01) and no longer applies cleanly — `main` has since picked up 50 commits this branch doesn't have. `git merge-tree --write-tree origin/main <this branch>` reports exactly one conflict, in `vc_render_tifxyz.cpp`, in the block this PR touches. In those 50 commits, #1528 (merged 2026-09-10T07:46:03Z) rewrote the same skip check: it's now gated behind `--resume`,

```
volume-cartographer/apps/src/vc_render_tifxyz.cpp:1624   // Skip existing TIFFs only when explicitly resuming.
volume-cartographer/apps/src/vc_render_tifxyz.cpp:1626   if (resumeFlag && numParts <= 1) {
```

so the line in the PR description — "with or without `--resume` (`--resume` only concerns zarr chunks)" — was accurate against the `main` it was written against, but isn't anymore: on today's `main` the skip (and therefore the bug this PR fixes) only engages when `--resume` is passed. That's where to rebase from; the fix itself doesn't need to change, it just needs to sit back inside `if (resumeFlag && numParts <= 1)` instead of replacing it with `if (numParts <= 1)`.

On the code itself: `isReadableTiff()` (`Tiff.hpp:99`, `Tiff.cpp:318`) is a cleaner version of the same idea my local patch used inline. Both wrap `TIFFOpen`/`TIFFClose`, so the detection is the same: libtiff itself rejects a header whose first-IFD offset is still 0 or a file truncated after the header. What's better in yours: the error/warning handlers are muted around the probe and restored after (`Tiff.cpp:323-327`), and it is a named function with its own unit test (`test_tiff.cpp:207`). Once this lands I'll drop my inline variant.

Not asking to widen this PR: rebasing onto current `main`, inside the `resumeFlag` block, is all it needs.

Bullo27:
Thanks for reproducing this on both macOS and the official Windows build, and for building the branch directly — that's the strongest confirmation I could have asked for.

You're right about #1528: it moved the skip behind `--resume`. I've rebased onto current `main` with the readable-TIFF check now sitting inside `if (resumeFlag && numParts <= 1)` rather than replacing it, so it only engages on resume, the way that block works now. The rest of the diff is unchanged and `isReadableTiff` keeps its unit test. Branch is force-updated; `git merge-tree` onto `main` is clean.

I also corrected the PR description's "with or without `--resume`" line, which was accurate against the older `main` this was first written on but isn't after #1528.

Appreciate you pinning down the exact merge-base.

*AI-assisted (Claude), human-directed account; the rebase and merge check above were run on my machine.*

Bullo27:
Status, for whoever picks this up: rebased onto today's main (`23c3d759e` → head `4373ab9b9`, `git range-diff` shows the patch unchanged), CI green on that head, review pending.

Re-checked on the rebased Release build: rerunning a SIGKILL-torn output directory with `--resume` prints `8 of 8 existing slices are not readable TIFFs (interrupted render?), re-rendering all slices.`, exits 0, and all 8 slices are `cmp`-identical to an uninterrupted render; a finished directory is still skipped in 0.15 s with the files untouched. nerln's comment above reproduces the bug and confirms the fix independently on macOS and on the official Windows build.

The change is 4 files, +71/−5: the probe sits in the existing `if (resumeFlag && numParts <= 1)` block, with a unit test in `test_tiff`.

*AI-assisted (Claude), human-directed account; the checks above were run on my machine.*

hendrikschilling:
```
1. P2: isReadableTiff() validates only the TIFF header/first directory
     volume-cartographer/core/src/Tiff.cpp:318 uses TIFFOpen("r"), which does not verify that all tile/strip payloads are readable. A file with an intact directory but truncated or corrupt
     tile data can still pass the probe and be skipped. This fixes the reported kill-before-close() case, but it is not a complete integrity check. Either document that limitation or      
     validate the required tiles before skipping.

```

Bullo27:
Agreed, and fixed in 8742eb0d7 by validating rather than documenting: `isReadableTiff()` now decodes every tile (or strip) of the first directory with `TIFFReadEncodedTile`/`Strip` and requires the full size back, so a truncated, missing or undecodable payload behind an intact directory is rejected as well. The header comment states what is checked (structure and decodability) and what is not (pixel values).

Measured on the same 8-slice render as in the description: a finished directory with 4 KiB of `03.tif`'s tile data overwritten by 0xFF (`TIFFOpen` still succeeds) was skipped by the previous head in 0.14 s; it is now reported as `1 of 8 existing slices are not complete, readable TIFFs …`, re-rendered, 8/8 `cmp`-identical to the uninterrupted render. The cost is that skipping a finished directory reads it: 2.2 s instead of 0.15 s for 73 MB of LZW TIFF. Torn-directory case unchanged (8 of 8, 13.0 s). New test case builds exactly that file; `test_tiff` 15/15. PR description updated.

PR #1696, opened 2026-09-04
spiral: make the spiral tutorial work on native Windows

7jycwjmbfn-eng:
**In one sentence:** You can skip WSL and run the whole spiral pipeline natively on Windows: `uv sync`, `fit_spiral.py`, `flatten_spiral_checkpoint.py`, and rendering with the released VC3D binaries.

**Motivation:** I wanted this work to run on native Windows. WSL is too much hassle for me. I went through the tutorial on Windows and hit a few things that make the native path unusable. Today it only really works under WSL. This PR fixes them. The same workflow now runs directly on Windows.

**One real example:** On a laptop with an RTX 4080, starting from the published `spiral-input` PHercParis4 dataset, I followed the four commands in the [spiral fitting tutorial](https://scrollprize.org/tutorial_spiral). `uv sync` failed while building libigl. After working around that by hand, flatten failed twice: once on `symlink_to`, once on `gpu_pause` importing `fcntl`. With these four commits the same commands run through to a flattened surface, and rendering it with the released `vc_render_tifxyz.exe` needs no changes at all.

**Before:**

- `uv sync` exits 1. The lockfile pins `libigl==2.6.2`, the one release with no `win_amd64` wheel, so every Windows install is pushed into a CGAL/Boost source build.
- `flatten_spiral_checkpoint.py` raises `OSError [WinError 1]` from `symlink_to`, and cannot work on exFAT at all. `lasagna/fit_service.py` imports `gpu_pause` before checking the flag, so `--no-gpu-pause` does not stop it reaching a bare `import fcntl`.
- The fit runs, but PyPI marks torch's `triton` dependency `sys_platform == 'linux'`, so `_HAS_TRITON` is False and the fused RK4 and gap kernels fall back to the eager path. Nothing is printed, so the run looks the same while using different kernels from the ones Linux uses.

**After this PR:**

- The lockfile takes `libigl==2.6.3`, released 2026-09-01, which ships the Windows wheels 2.6.2 dropped. Nothing is gated off; SLIM reparameterisation still works.
- flatten copies where it cannot symlink, and the `gpu_pause` import follows the flag.
- The fallback prints one line: the fused kernels are unavailable, the eager path is slower, and its results differ slightly.

**Proof:**

1. The lockfile pins libigl 2.6.2. On the same machine, `--no-build` shows that 2.6.2 has no usable wheel, and that 2.6.3 installs and imports.

libigl before and after

2. I ran flatten on a checkpoint from my own fit (PHercParis4, z 5000-5400). Clean main stops with `No module named 'fcntl'`. This branch writes `flattened.tifxyz`, 163.5 MB, and exits 0.

flatten before and after

3. I ran two 20-step fits on real data with triton not installed. On clean main, nothing in the output mentions triton. This branch prints the two warnings and reaches the same `satisfied_area`.

fused kernel fallback before and after

Tests: 4 for the fallback, 11 around flatten and `gpu_pause`. Reverting the source changes and keeping the tests fails them.

Speed, for context: with the community `triton-windows` build installed so the fused path is available, the same window and step count runs at 3.8 it/s against 1.7 it/s. The first fused run pays a one-off kernel compile, and `satisfied_area` moves between the two arms by more than the run-to-run spread, so this is not only a speed difference.

**Why / where this is useful:** For people who want to work on native Windows, and for people who would rather not use WSL or are not familiar with it. With this in, they can just run the workflow.

## Details

Four commits, each self-contained:

1. `spiral: take the libigl release that ships a Windows wheel` — a floor in `pyproject.toml` plus `uv lock --upgrade-package libigl`. No other package moves in the lockfile.
2. `spiral: flatten a checkpoint where symlinks and gpu_pause are unavailable` — the object store is private to the run and only ever read, so a copy serves the same reads; the manifest hash is over the segment's bytes and does not move.
3. `spiral: say when the fused kernels fall back to the eager path` — warn-once, only when triton is missing. `FIT_SPIRAL_TRITON=0` stays silent, since that is the user's own choice. Two places in this repo already do this: `spiral-fitting/tracks.py:1450` and `vesuvius/src/vesuvius/ink_detection/inference/inference_runtime.py:158-171`.
4. `docs(spiral): what the install line does on Windows` — the CUDA index, and that `triton-windows` is what puts the fused path back.

The screenshots were taken at `1ee7f94d3`; the branch is rebased onto `23adee047`, which only changes a default in `spiral-fitting/config.py`.

Tested on Windows 11, Python 3.14.7, torch 2.11.0+cu128, RTX 4080 Laptop.

### DIFF
--- a/lasagna/fit_service.py
+++ b/lasagna/fit_service.py
@@ -66,6 +66,20 @@ _VC3D_SOURCE_HEADER = "X-VC3D-Source"
 os.environ.setdefault("LASAGNA_CHECK_SPARSE_CACHE", "1")
 
 
+def _gpu_pause_or_null():
+    """Return the GPU-pause context, or a no-op where pausing is disabled.
+
+    gpu_pause is imported here rather than at module scope because it is
+    built on fcntl and unix sockets. A host started with --no-gpu-pause has
+    nothing to coordinate with and may not have either.
+    """
+    if not _gpu_pause_enabled:
+        from contextlib import nullcontext
+        return nullcontext()
+    from gpu_pause import gpu_pause_context
+    return gpu_pause_context()
+
+
 def _mib(n_bytes: int) -> float:
     return float(n_bytes) / (1024.0 * 1024.0)
 
@@ -1901,11 +1915,8 @@ def _run_optimization(job: _JobState, body: dict[str, Any]) -> None:
             kwargs["cancel_fn"] = _check_cancel
             return _orig_optimize(**kwargs)
 
-        from contextlib import nullcontext
-        from gpu_pause import gpu_pause_context
-
         opt_mod.optimize = _patched_optimize
-        with (gpu_pause_context() if _gpu_pause_enabled else nullcontext()):
+        with _gpu_pause_or_null():
             try:
                 import fit as fit_mod
                 job.set_running("loading", 0, 0, 0.0)
@@ -2348,9 +2359,7 @@ class _Handler(BaseHTTPRequestHandler):
                         data_input = str(candidate)
 
             import lasagna_analyze
-            from contextlib import nullcontext
-            from gpu_pause import gpu_pause_context
-            with (gpu_pause_context() if _gpu_pause_enabled else nullcontext()):
+            with _gpu_pause_or_null():
                 lasagna_analyze.export_vis_obj(
                     model_path=str(model_input),
                     data_path=str(data_input),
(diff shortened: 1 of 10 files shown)


7jycwjmbfn-eng:
Just following up on this one since it’s been open for a couple of weeks. The native Windows path is still working on my end, and I’m happy to make any changes or split this up if that would make it easier to review.

Would appreciate a review when someone has a chance. Thanks!

pmh47:
Having windows support in spiral is worthwhile. If this were rebased, and triton made to use triton-windows in windows (if possible), and the tests shortened, we will consider merging.

7jycwjmbfn-eng:
Thanks. Rebased onto current main. On Windows, `uv sync` now installs `triton-windows` 3.6 alongside torch 2.11, so the fused kernels run there. I dropped the missing-triton warning, since Windows has triton now. The tests are down to one for the flatten copy fallback and one for the `gpu_pause` import. I've updated the description and screenshots.

PR #1697, opened 2026-09-04
Fix VC3D downloading legacy level_N files on cold start with no ~/.VC…

spelufo:
**In one sentence:** [see discord](https://discord.com/channels/1079907749569237093/1243576621722767412/1545371691813703730)

**One real example:** Move ~/.VC3D out of the way and run VC3D. It no longer creates deprecated level_N files.

### However

I hope this doesn't break the read path. How should it work? It reads from either, but only writes new chunks to N (not level_N) ? 

Please review if this is the right thing, didn't take the time to look at it in depth, sorry.


hendrikschilling:
working on a more comprehensive fix right now

PR #1698, opened 2026-09-04
ink_detection: retry transient remote reads in flat inference and bbox reader (fixes #1666)

### LINKED ISSUE #1666: ink_detection inference: one truncated remote chunk read aborts the whole run (exit 1, no output)
Streaming a published surface volume into `vesuvius.ink_detection.inference.infer` aborts the entire run when a single remote chunk read is truncated. The process exits 1 with no output after minutes of work.

I hit this twice on the same segment while running the [18 August First Letters workflow](https://scrollprize.substack.com/p/from-ct-scan-to-ancient-text-a-first) against `PHerc1447/segments/20250703025628-auto_grown_20250703025628283/surface-volumes/8.64um-1.2m-116keV-volume-20250521151220.zarr`:

```
aiohttp.client_exceptions.ClientPayloadError: Response payload is not completed:
<ContentLengthError: 400, message='Not enough data to satisfy content length header
(received 173631 of 507920 bytes).'>. SSLError(1, '[SSL: RECORD_LAYER_FAILURE] record layer failure (_ssl.c:2713)')
```

**The object is not corrupt.** The second attempt truncated at a *different* offset — 299,585 of the same 507,920 bytes — and:

- fetching real chunks directly over plain sequential HTTPS gave **12/12 successes, 4,663,415 bytes, zero failures**;
- after mirroring the store locally with per-object retries, the identical segment inferred successfully in about a minute with the same checkpoint on the same machine.

So the read path is the only variable. This looks like a transient truncation under the async reader's concurrency (default `--num-workers 4` plus prefetch), not a data or general connectivity problem.

**Why it matters:** these are exactly the workloads the streaming path exists for. A single transient chunk failure discarding an entire multi-minute inference makes unattended runs over the published volumes impractical, and the failure is silent about its cause until you read the traceback.

**Suggested fix:** the same treatment already merged for the prediction path in #1244 ("Retry transient remote reads instead of aborting the whole run") — bounded retries with backoff around the chunk read, so a truncated response is retried rather than propagated.

Environment: villa `main` @ `5479453a7`, torch 2.13.0+cu126, `scrollprize/ink_9um` seed42 `step-075000.pth`, Windows, RTX 3090.

Full reproduction, the local-mirror workaround (`scripts/fetch_zarr.py`), and all measurements: https://github.com/TAUIL-Abd-Elilah/pherc1447-ink-survey

Happy to open a PR for the retry wrapper, though I am currently at three open PRs (#1572, #1578, #1647) and would rather clear one first than open a fourth.

<sub>Investigation done with AI assistance (claude), directed and reviewed by me.</sub>

modaribrahim:
**In one sentence:** Flat ink-detection inference (`vesuvius.ink_detection.inference.infer`) now retries transient remote chunk reads instead of aborting the whole run when a single network hiccup truncates one chunk download (issue #1666).

**One real example:** Reading the published First Letters surface volume `PHerc1447/segments/20250703025628-auto_grown_20250703025628283/surface-volumes/8.64um-1.2m-116keV-volume-20250521151220.zarr` (level 0, 31×4100×4260 uint8) via `FlatPatchReader` — the exact workflow from #1666 — a single truncated chunk response previously aborted the read with `ClientPayloadError` and no output. With this PR the same read retries and completes.

**Before:**
```
READ FAILED: ClientPayloadError: Response payload is not completed: <ContentLengthError: 400,
message='Not enough data to satisfy content length header (received 232 of 423 bytes).'>
...
RESULT: ABORT
```

**After this PR:**
```
transient remote read error (ClientPayloadError), retry 1/3 in 0.5s: Response payload is not
completed: <ContentLengthError: 400, message='Not enough data to satisfy content length header
(received 232 of 423 bytes).'>
READ OK: patch shape=(128, 128, 31), dtype=uint8
RESULT: SUCCESS
```

**Proof:** Same real volume, same real code path (`FlatPatchReader._read_raw` → zarr → fsspec → aiohttp), same one-shot truncation of chunk `0/0.0.0`, same error type/message as #1666. Pre-fix aborts (exit 1, no output — the reported symptom); post-fix logs one warning line and completes. The truncation is transient, not data corruption: the retried attempt reads the same chunk successfully from the real store.

- `tests/ink_detection/test_volume_io_retry.py` — 15 passed (bbox + flat-reader retry, cap exhaustion, deterministic fail-fast, verbatim #1666 message classification)
- `tests/data/test_volume_retry.py` — 14 passed (refactor is behavior-identical)
- `tests/ink_detection/test_volume_io.py` — 12 passed (no regression)

**Why / where this is useful:** The 18 Aug First Letters tutorial command hits this verbatim (per #1666). One truncated chunk currently discards everything a multi-minute/overnight streaming run already computed. The fix also protects native 3D inference (`infer_full3d_tifxyz`) and the training patch loaders, which read through the same shared `read_bbox_with_padding`.

## Details

**Root cause:** `ink_detection/volume_io.py` and `inference/infer.py` read remote zarr chunks directly (`array[selection]`) with no retry. The prediction path already got this treatment in #1244 (`data/volume.py`), but ink-detection uses a separate reader that never received it.

**Fix:**
1. Extracted the #1244 transient-error classifier into a shared leaf module `vesuvius/data/_transient_reads.py` (`_TRANSIENT_READ_MARKERS`, `_is_transient_read_error`) plus a new `_read_array_with_retry(volume, selection, ...)`: 4 attempts, 0.5 s doubling to 8 s, `max(1, retries)` guard; deterministic errors (bad coords, missing node, auth) re-raise on the first attempt.
2. `data/volume.py` now imports the helpers from the shared module — no behavior change (14/14 existing retry tests still pass).
3. `inference/infer.py`: the flat-read path `FlatPatchReader._read_raw` (the #1666 path) and the startup occupancy scan `compute_nonempty_mask_from_lowres_array` read through the retry wrapper; one warning line logged per retry.
4. `volume_io.py`: `read_bbox_with_padding` — used by full_3d inference and all training patch loaders — reads through the retry wrapper.

**How the error was reproduced:** a local proxy (`proxy_truncate.py`) forwards every request to the real public S3 object and serves the real bytes, but for exactly one chunk — `0/0.0.0`, 423 bytes — sends the full `Content-Length` header, writes only 232 bytes (~55%), then drops the connection. The client therefore sees the identical `ClientPayloadError` that #1666 reports. This is a deterministic reproduction of a real transient network truncation on the actual published volume; the volume is untouched and re-reads cleanly on the retry.

**Verified:** Python 3.12 / zarr 3.3.0 / CPU torch on a Linux VM, against the real public PHerc1447 surface volume. Test-matrix note: this demo ran on zarr 3; repo CI covers zarr 2.18.7 and 3.2.1. The retry helper is transport-agnostic (classifies on message text across the whole `__cause__`/`__context__` chain), so it does not depend on the zarr/fsspec version.

**Determinism:** retries only trigger on exceptions matching the transient markers; the happy path performs exactly one read. Retry re-reads the whole bbox selection (coarser than #1244's per-read retry) — acceptable at inference patch granularity.

**Reproduction (reviewer can rerun):**
```
python proxy_truncate.py 8899
# open http://localhost:8899/PHerc1447/segments/20250703025628-auto_grown_20250703025628283/
#     surface-volumes/8.64um-1.2m-116keV-volume-20250521151220.zarr
# then read one patch via FlatPatchReader.read(0, 0, 128, 128):
#   pre-fix code -> ClientPayloadError ... RESULT: ABORT
#   this PR      -> retry 1/3 in 0.5s ... READ OK: (128,128,31) ... RESULT: SUCCESS
```
(Proxy + demo scripts and raw before/after logs available on request.)

---

*Written with AI assistance as a coding assistant; directed, reviewed, and verified by a human (community norm — see #1244/#1667).*

### DIFF
--- a/vesuvius/src/vesuvius/data/_transient_reads.py
+++ b/vesuvius/src/vesuvius/data/_transient_reads.py
@@ -1,0 +1,93 @@
+"""Classification and bounded retry for transient remote reads.
+
+zarr and fsspec wrap the underlying transport (aiohttp, botocore, urllib3,
+ssl) differently across versions and backends, so transient failures are
+classified by message substring rather than by exception type.  Retrying
+happens only for failures that look like a network hiccup; deterministic
+errors (missing array, bad coordinates, auth) re-raise immediately.
+
+This module is a leaf: it imports nothing from vesuvius, so both
+``data.volume`` and ``ink_detection.volume_io`` can import it without any
+circular-import risk.
+"""
+
+from __future__ import annotations
+
+import time
+from typing import Any, Callable, Optional
+
+
+_TRANSIENT_READ_MARKERS = (
+    'payload is not completed',
+    'not enough data to satisfy content length',
+    'contentlengtherror',
+    'connection reset',
+    'connection aborted',
+    'connection closed',
+    'server disconnected',
+    'record layer failure',
+    'ssl',
+    'timed out',
+    'timeout',
+    'temporarily unavailable',
+    'slowdown',
+    'throttl',
+    'too many requests',
+    'internal error',
+    'service unavailable',
+    'bad gateway',
+    'gateway timeout',
+    ' 429',
+    ' 500',
+    ' 502',
+    ' 503',
+    ' 504',
+)
+
+
+def _is_transient_read_error(exc: BaseException) -> bool:
+    """True when a read failure looks like a network hiccup rather than a bug."""
+    seen = set()
+    while exc is not None and id(exc) not in seen:
+        seen.add(id(exc))
+        text = f"{type(exc).__name__}: {exc}".lower()
+        if any(marker in text for marker in _TRANSIENT_READ_MARKERS):
+            return True
+        exc = exc.__cause__ or exc.__context__
+    return False
+
+
+def _read_array_with_retry(
+    volume: Any,
+    selection: tuple[Any, ...],
+    *,
+    retries: int = 4,
+    base_delay: float = 0.5,
+    max_delay: float = 8.0,
+    warn: Optional[Callable[[str], None]] = None,
+) -> Any:
+    """Read ``volume[selection]``, retrying transient remote failures.
+
+    Remote object stores drop connections routinely — truncated payloads,
+    SSL record errors, 5xx, throttling.  A caller streaming a whole volume
+    issues one read per patch, so without this a single hiccup propagates
+    out and aborts the job, discarding everything computed so far.
+
+    Deterministic failures (bad coordinates, missing array, auth) are not
+    retried; they re-raise on the first attempt.
+    """
+    retries = max(1, int(retries))
+    delay = base_delay
+    for attempt in range(retries):
+        try:
+            return volume[selection]
+        except Exception as e:
+            if attempt == retries - 1 or not _is_transient_read_error(e):
+                raise
+            if warn is not None:
+                warn(
+                    f"transient remote read error ({type(e).__name__}), retry "
+                    f"{attempt + 1}/{retries - 1} in {delay:.1f}s: {e}"
+                )
+            time.sleep(delay)
+            delay = min(delay * 2, max_delay)
--- a/vesuvius/src/vesuvius/data/_transient_reads.py
+++ b/vesuvius/src/vesuvius/data/_transient_reads.py
@@ -13,6 +13,7 @@ circular-import risk.
 
 from __future__ import annotations
 
+import re
 import time
 from typing import Any, Callable, Optional
 
@@ -37,22 +38,53 @@ _TRANSIENT_READ_MARKERS = (
     'service unavailable',
     'bad gateway',
     'gateway timeout',
-    ' 429',
-    ' 500',
-    ' 502',
-    ' 503',
-    ' 504',
+)
+
+# HTTP statuses worth retrying, but only when the number appears as a status:
+# "ClientResponseError: 503, message=...", "An error occurred (503) when
+# calling GetObject", "HTTP 429", "status: 502". A bare " 504" also matched
+# "index out of bounds for dimension with length 504" and "size 5040", which
+# turned a coding error into four attempts and 3.5 s of backoff.
+_TRANSIENT_HTTP_STATUS = re.compile(
+    r'(?:(?:error|status|http|response|code)\W{0,4}|\(\s*)(?:429|500|502|503|504)(?!\d)'
+)
+
+# Exceptions that describe a bug or a bad request rather than a network
+# hiccup. Their messages are not scanned for markers (zarr's BoundsCheckError
+# is an IndexError whose text names the dimension length), but the cause
+# chain is still walked in case one wraps a genuine transport error.
+_DETERMINISTIC_ERROR_TYPES = (
+    IndexError,
+    KeyError,
+    TypeError,
+    ValueError,
+    AttributeError,
+    NotImplementedError,
 )
 
 
+_NON_TRANSIENT_TYPES = _DETERMINISTIC_ERROR_TYPES
+
+
 def _is_transient_read_error(exc: BaseException) -> bool:
-    """True when a read failure looks like a network hiccup rather than a bug."""
+    """True when a read failure looks like a network hiccup rather than a bug.
+
+    Coding errors are never transient, regardless of their message: zarr's
+    ``BoundsCheckError`` subclasses ``IndexError`` and its message contains a
+    dimension length (e.g. ``... length 504``) that a bare numeric marker
+    (``' 503'``, ``' 504'``) would otherwise match.  Skip the marker scan for
+    those types, but keep walking the cause chain so a genuine transport
+    error wrapped behind a deterministic wrapper is still detected.
+    """
     seen = set()
     while exc is not None and id(exc) not in seen:
         seen.add(id(exc))
-        text = f"{type(exc).__name__}: {exc}".lower()
-        if any(marker in text for marker in _TRANSIENT_READ_MARKERS):
-            return True
+        if not isinstance(exc, _NON_TRANSIENT_TYPES):
+            text = f"{type(exc).__name__}: {exc}".lower()
+            if any(marker in text for marker in _TRANSIENT_READ_MARKERS):
+                return True
+            if _TRANSIENT_HTTP_STATUS.search(text):
+                return True
         exc = exc.__cause__ or exc.__context__
     return False
 
--- a/vesuvius/src/vesuvius/data/_transient_reads.py
+++ b/vesuvius/src/vesuvius/data/_transient_reads.py
@@ -42,17 +42,23 @@ _TRANSIENT_READ_MARKERS = (
 
 # HTTP statuses worth retrying, but only when the number appears as a status:
 # "ClientResponseError: 503, message=...", "An error occurred (503) when
-# calling GetObject", "HTTP 429", "status: 502". A bare " 504" also matched
-# "index out of bounds for dimension with length 504" and "size 5040", which
-# turned a coding error into four attempts and 3.5 s of backoff.
+# calling GetObject", "HTTP 429", "status: 502", urllib3's "too many 503
+# error responses". A bare " 504" also matched "index out of bounds for
+# dimension with length 504" and "size 5040", which turned a coding error
+# into four attempts and 3.5 s of backoff.
 _TRANSIENT_HTTP_STATUS = re.compile(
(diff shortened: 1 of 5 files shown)


Bullo27:
**Verdict: works as described here; one measured edge, not a blocker.**

Ran `test_volume_io_retry.py` plus `test_volume_io.py` and `test_flat_inference.py` on 8828f19 (Linux, zarr 3.2.1, torch 2.13.0): 47 passed. The marker list and classifier are moved from `volume.py` unchanged.

The edge: the bare numeric markers (`' 429'`, `' 500'`, `' 502'`, `' 503'`, `' 504'`) also match zarr's bounds error. `FlatPatchReader` with `layer_indices=[0, 2, 600]` (fancy path) on a 504-layer array: `BoundsCheckError: index out of bounds for dimension with length 504` is classified transient — 4 reads, three "transient remote read error" warnings and 3.5 s of backoff before it raises. The same array with 300 layers raises on the first read (lengths 5000 and 5048 match as well). Inherited from `volume.py`'s loop, so pre-existing; skipping the marker scan for links that are `IndexError`/`KeyError`/`TypeError`/`ValueError` (`BoundsCheckError` subclasses `IndexError`) makes that repro fail fast and keeps all 15 new tests green here.

ge-al:
**Verified end to end on the real segment (measured):** I streamed the exact volume from #1666 (`PHerc1447/segments/20250703025628-auto_grown_20250703025628283/surface-volumes/8.64um-1.2m-116keV-volume-20250521151220.zarr`) with `ink_9um/hybrid_3d2d-seed42/step-075000.pth` and a 512×512 mask (80 patches) through a local HTTP proxy that forwards to the open-data bucket but stops 3% of chunk responses short of `Content-Length` (what aiohttp reports as `ContentLengthError`, the failure in the issue), on this PR's head `8828f19` vs `main` `23adee0`, same proxy seed both times.

| run | exit | retried reads | output |
|---|---|---|---|
| `main` 23adee0 | 1 at the first truncated chunk (`_read_raw`, `ClientPayloadError … received 30841 of 92524 bytes`) | — | none |
| this PR 8828f19 | **0** | 7 (`retry 1/3` ×6, one `retry 2/3`), 7 truncations injected | 360,504-byte TIFF, **byte-identical** to the same run on a local copy of the volume |

The occupancy-scan read (`compute_nonempty_mask_from_lowres_array`, `array[:]`) is also covered here, which matters: at a harsher 15% truncation rate my first attempt at this fix died there, before the patch loop, not in `_read_raw`.

Logs and the ~60-line proxy script: [infer_proxy_pr1698.log](https://raw.githubusercontent.com/ge-al/villa/proof-assets/infer_proxy_pr1698.log), [proxy4.log](https://raw.githubusercontent.com/ge-al/villa/proof-assets/proxy4.log), [truncating_proxy.py](https://raw.githubusercontent.com/ge-al/villa/proof-assets/truncating_proxy.py) (`python truncating_proxy.py PORT FRACTION SEED`, then point `infer` at `http://127.0.0.1:PORT/<bucket key>`).

I had an equivalent branch in flight ([`ge-al:fix/ink-inference-read-retries`](https://github.com/ge-al/villa/tree/fix/ink-inference-read-retries)); yours was first and passes the same test, so I am not opening it. Two things from it you may want to fold in, or I can send as a follow-up once this lands: a `--read-retries` flag on `infer` and `infer_full3d_tifxyz` (mirrors `vesuvius.predict --read-retries` from #1244; `1` restores fail-fast), and `NativePatchDataset(read_retries=...)` so the native path can be tuned the same way.

*AI-assisted (Claude Code), human-directed; the runs above were executed on my machine on the published data.*

ge-al:
Follow-up on the classifier edge above, since it lives in already-merged code too (`vesuvius/data/volume.py` from #1244, which `vesuvius.predict` streams through):

**Measured on a published volume.** PHerc0500P2 `segments/20250611132918-z_dbg_gen_00302_inp_hr/surface-volumes/9.362um-1.2m-113keV-volume-20250820143440.zarr` is 28×**5040**×3260. Reading `vol[0:4, [0, 2, 6000], 0:4]` through `Volume(type="zarr", path="s3://…", anon=True)` on `main` 23adee0:

```
  transient read error (BoundsCheckError), retry 1/3 in 0.5s: index out of bounds for dimension with length 5040
  transient read error (BoundsCheckError), retry 2/3 in 1.0s: ...
  transient read error (BoundsCheckError), retry 3/3 in 2.0s: ...
BoundsCheckError: index out of bounds for dimension with length 5040
classified transient=True  attempts+backoff elapsed=4.25s
```

Two of the 368 published level-0 arrays I checked have such an axis (that one and `…00505_inp_hr`, 5020), and numpy-style `index 500 is out of bounds` messages hit it on any shape.

Fix on [`ge-al:fix/transient-read-classifier-fail-fast`](https://github.com/ge-al/villa/tree/fix/transient-read-classifier-fail-fast) (`86f0de2`): skip the marker scan for `IndexError`/`KeyError`/`TypeError`/`ValueError`/`AttributeError`/`NotImplementedError` links while still walking their cause chain, and only count a status number when it appears as one (`ClientResponseError: 503, …`, `An error occurred (503) when calling …`, `HTTP 502`, `status: 504`). Same read afterwards: `classified transient=False  elapsed=0.59s`, no retries; the #1244 transient fixtures still pass; `tests/data/test_volume_retry.py` 14 → 28 cases. GitHub is currently refusing PR creation from my account on this repo (works on my fork), so I cannot open it as a PR right now; it applies verbatim to `_transient_reads.py` here if you would rather fold it in.

*AI-assisted (Claude Code), human-directed; measured on my machine on the published data.*

modaribrahim:
@ge-al thanks for the careful check and the catch on the classifier. I folded your fix from `86f0de2` into `99a272d`: skip the scan for deterministic error types (still walking the cause chain) and only match status numbers as real tokens. Tests plus attribution included.

@Bullo27 same edge should be fixed now.

ge-al:
Thanks for folding it in, and for the attribution. Two small follow-ups landed on my branch after `86f0de2` (both in `fcbfdf6`), from checking the regex against real exception objects from the installed aiohttp / botocore / s3fs / fsspec / urllib3 / requests:

1. urllib3's retry-exhaustion text, `too many 503 error responses` (what `requests` surfaces inside `MaxRetryError`), has no `error|status|http|response|code` before the number, so `99a272d` classifies it non-transient where `main` retried it. Widening the alternation to `(?:error|status|http|response|code|returned|got|many)` keeps it (and `server returned 502`); test strings are in `tests/data/test_volume_retry.py` on my branch if useful. Low reach today (the chunk path goes through aiohttp/botocore, not requests), but cheap.
2. `ssl.SSLCertVerificationError` (and aiohttp's `ClientConnectorCertificateError`) inherit `ValueError`, so the deterministic-type skip turns certificate failures from retry into fail-fast. That is the right outcome — a bad certificate does not fix itself — but it is a behaviour change worth a comment on `_DETERMINISTIC_ERROR_TYPES` and a pin (`assert not _is_transient_read_error(ssl.SSLCertVerificationError(...))`, while a plain `SSLError` record-layer failure stays transient).

Everything real the six stacks produce for 5xx/429 (`ClientResponseError: 503, message=…`, `An error occurred (503) when calling …`, s3fs's `OSError: [Errno 16] Slow Down` with a `ClientError` cause, `HTTPError: 503 Server Error …`) matched on both versions; 403/404 and zarr bounds errors do not.

Since `volume.py` imports the classifier from `_transient_reads.py` here, I'll drop my separate branch once this lands rather than open a competing PR.

*AI-assisted (Claude), human-directed; measured on my machine.*

Bullo27:
Confirmed on 99a272d (measured): the repro from my first comment (`FlatPatchReader`, `layer_indices=[0, 2, 600]`, 504-layer array) now raises `BoundsCheckError` on the first read, 0 warnings, 0.00 s, against 4 reads / 3 warnings / 3.5 s on 8828f19. `test_volume_io_retry.py` + `test_volume_io.py` + `tests/data/test_volume_retry.py` + `test_flat_inference.py`: 72 passed; the 3 `test_flat_inference.py` failures are a cuDNN version mismatch on my box and fail identically on 8828f19.

Ran ge-al's two points through the classifier here as well: `OSError('too many 503 error responses')` is transient on 8828f19 and not on 99a272d; `ssl.SSLCertVerificationError` likewise flips to fail-fast, while a record-layer `ssl.SSLError` stays transient. So both observations reproduce.

modaribrahim:
@jrudolph @bruniss friendly ping for review when you get a chance. Both community reported classifier edges are confirmed fixed (Bullo27 measured 0 wasted retries and 72 tests passing), and ge-al's follow-up from fcbfdf6 is now folded in as 5732003. Nothing else pending on my side. Happy to address anything that comes up.

PR #1699, opened 2026-09-04
tifxyz: compute the full-resolution canvas the way QuadSurface::size() does (fixes #1694)

### LINKED ISSUE #1694: tifxyz: Tifxyz.shape / full_resolution_shape truncate the full-resolution canvas by 1 px for float32-rounded scales
**In one sentence:** `Tifxyz.shape` / `Tifxyz.full_resolution_shape` compute the full-resolution canvas with `int(h / scale)`, which is one pixel short for every tracer mesh whose `meta.json` stores the scale as float32-rounded 0.05 (`0.05000000074505806`).

**I was trying to:** render surface volumes from `vc_grow_seg_from_seed` meshes (PHerc0358, PHerc0813, PHerc0826 patches grown on the released m7 surface predictions, and the published PHerc0139 w035 mesh) with a Python renderer built on `vesuvius.tifxyz`, then run ink inference on them and compare against the published surface volumes / maps.

**Using:** `vesuvius/src/vesuvius/tifxyz/types.py` at villa `37e300d3` (also present on main as of 2026-09-01); meshes written by `vc_grow_seg_from_seed` (release `vc-tracer-de3c2494`); published mesh `PHerc0139/segments/20260317000000-w035_2026031718`; `read_tifxyz(dir, load_mask=False, validate=False).use_full_resolution().shape`.

**What happened:** every full-resolution canvas came out one pixel smaller than the C++ pipeline's. For a 152×152 stored grid with `"scale": [0.05000000074505806, 0.05000000074505806]`, `use_full_resolution().shape` is `(3039, 3039)`; `vc_render_tifxyz` and the published surface volumes use 3040×3040. For w035 the Python path gives 5819×5239 against the published 5820×5240 volume. Rendered maps, masks and labels produced through the two paths are misaligned by a pixel at the far edge and fail exact-shape checks; nothing warns.

**What I expected or needed:** the same canvas the C++ renderer uses, i.e. the rounded value (`152 / 0.05000000074505806 = 3039.99995…` → 3040).

**Evidence / reproduction:**

Real meshes (any tracer output; these are from PHerc0813 grown on 2026-09-02):

```
$ python - <<'EOF'
import json, tifffile
m = json.load(open("auto_grown_20260902135207525/meta.json"))
h, w = tifffile.imread("auto_grown_20260902135207525/x.tif").shape
print(m["scale"], (h, w), "int:", (int(h/m["scale"][0]), int(w/m["scale"][1])), "round:", (round(h/m["scale"][0]), round(w/m["scale"][1])))
EOF
[0.05000000074505806, 0.05000000074505806] (152, 152) int: (3039, 3039) round: (3040, 3040)
```

Shortest standalone reproduction (no download):

```python
import json, os, tempfile, numpy as np, tifffile
from vesuvius.tifxyz import read_tifxyz
d = tempfile.mkdtemp(); h = w = 152
for n in "xyz":
    tifffile.imwrite(os.path.join(d, f"{n}.tif"), (np.random.rand(h, w) * 1000).astype(np.float32))
json.dump({"format": "tifxyz", "type": "seg", "uuid": "t",
           "scale": [float(np.float32(0.05)), float(np.float32(0.05))]}, open(os.path.join(d, "meta.json"), "w"))
s = read_tifxyz(d, load_mask=False, validate=False)
print(s.shape, s.use_full_resolution().shape)   # (152, 152) (3039, 3039) -- expected (3040, 3040)
```

## Details

Cause: both properties do `(int(h / scale_y), int(w / scale_x))`. The tracer serializes the float32 value of 0.05 as a double, so `h / scale` lands a hair below the integer and `int()` truncates. Suggested fix: `int(round(h / scale_y))` in both properties (matching `vc_render_tifxyz`), plus a test using the float32-rounded scale above.

Our workaround in the meantime: compute the canvas as `round(stored / scale)`, allocate at that size, and render the region villa exposes.

ge-al:
**In one sentence:** `vesuvius.tifxyz` now reports the same full-resolution canvas that `vc_render_tifxyz` renders — for tracer meshes whose `meta.json` stores the scale as float32-rounded 0.05 and for the non-round scales `vc_obj2tifxyz` writes — so Python renderers and label producers stop coming out one pixel short.

**One real example:** Starting with the published PHerc1447 segment `20250703025628-auto_grown_20250703025628283` (`mesh/intermediate/tifxyz_original`, 202×215 stored grid, `"scale": [0.05000000074505806, 0.05000000074505806]`), I read it with `read_tifxyz(...).use_full_resolution().shape`, and it produced `(4039, 4299)` before and `(4040, 4300)` after — the latter is the canvas `vc_render_tifxyz` renders for that grid.

**Before:** `Tifxyz.shape` / `full_resolution_shape` did `int(h / scale)` in float64. `202 / 0.05000000074505806 = 4039.99994…`, truncated to 4039. Every mesh written by `vc_grow_seg_from_seed` (and the published w-series meshes with the same float32 scale, e.g. PHerc0139 w035: 5819×5239 vs the published 5820×5240 surface volume) rendered a canvas one pixel smaller than the C++ pipeline; maps/masks/labels produced through the two paths then fail exact-shape checks, and nothing warns (#1694).

**After this PR:** the extent is computed the way the renderer sizes its canvas (`vc_render_tifxyz.cpp`, "Compute render scale": `double sx = render_scale / surf->_scale[0]; lround(full_size.width * sx)` with `_scale` a `cv::Vec2f`), i.e. `max(1, lround(stored * (1.0 / float32(scale))))` in double, rounding half away from zero. Meshes with an exact 0.05 scale are unchanged (`205×213 → 4100×4260`, matching the published `canvas_size` of the same segment's surface volume), and the seven `vc_obj2tifxyz` meshes with non-round scales that @Bullo27 measured against the binary (see the review below) now match on all seven.

**Proof:**

```
$ uv run python check_shape.py   # reads the two tifxyz dirs of the segment above

# main (23adee0):
mesh/tifxyz                         scale= 0.05                stored= (205, 213) full= (4100, 4260)
mesh/intermediate/tifxyz_original   scale= 0.05000000074505806 stored= (202, 215) full= (4039, 4299)   <-- short by 1

# this PR:
mesh/tifxyz                         scale= 0.05                stored= (205, 213) full= (4100, 4260)
mesh/intermediate/tifxyz_original   scale= 0.05000000074505806 stored= (202, 215) full= (4040, 4300)
```

(`check_shape.py` is `read_tifxyz(dir, load_mask=False, validate=False).use_full_resolution().shape` for each directory, printed next to `meta.json["scale"]`.) The surface volume published next to that mesh declares `"canvas_size": [4260, 4100]` in its `.zattrs`, which the exact-scale case reproduces unchanged.

Tests: `tests/tifxyz/test_full_resolution_shape.py` (the 152→3040 patch from the issue, the w035 291/262→5820/5240 canvas, the PHerc1447 grid above, the seven measured `vc_render_tifxyz` cases from the review, the exact-half case 7×5 @ 2.0 → 4×3, exact-scale unchanged, minimum extent 1, unusable/underflowing scales fall back, and `tifxyz_label_transfer.core.Surface` agreeing with `Tifxyz`). `tests/tifxyz` + `tests/tifxyz_label_transfer` + `tests/test_surface_preflight.py`: 168 passed, 5 skipped; `tests/test_surface_preflight.py`, `tests/ink_detection/test_native_full3d_inference.py`, `tests/ink_detection/test_data_foundation.py` also pass.

**Why / where this is useful:** anything that builds a canvas from a tifxyz in Python — the flat renderer path in `ink_detection`, `tifxyz_label_transfer`, custom renderers comparing against published surface volumes — now agrees pixel-for-pixel with `vc_render_tifxyz` output, so shape assertions between the two pipelines pass instead of failing at the far edge.

## Details

Fixes #1694. Tested at `23adee0` on macOS / Python 3.14 / zarr 3.2.1.

Behaviour notes (all consistent with the renderer, but broader than the float32 case): any quotient with fraction ≥ .5 now rounds up instead of truncating (152 at scale 20.0: 7 → 8), a negative or non-finite scale no longer yields a negative/zero extent, the result is at least 1, and a scale that underflows float32 to 0 (e.g. 1e-46) falls back to the stored extent instead of raising. `QuadSurface::size()` in VC3D still truncates in float32; its only in-tree consumer is `surface_metrics.cpp`'s sampling bounds, so the renderer, not `size()`, is treated as the canvas of record here.

One shared definition: the helper lives in a dependency-free module, `vesuvius/tifxyz_canvas.py`, and `tifxyz_label_transfer/core.py`'s `Surface.full_resolution_shape` (and the native-canvas candidate in `infer_output_shape`) now call it too. `core.py` previously rounded the float64 quotient, which differs from the renderer whenever `stored / scale` sits within ~1e-4 of .5 (3681 at scale 0.3313977: 11108 vs the rendered 11107), so the two packages could report canvases one pixel apart on the same mesh; a test pins their agreement. `neural_tracing/datasets/common.py` (`int(round(h_s / scale))`) and `tifxyz/upsampling.py` size their own resampling grids with yet another rounding; I left those alone since they are not the canvas the renderer writes, but they are the next candidates for the shared helper.

Revision history: the first version mirrored `QuadSurface::size()` (`static_cast<int>(cols / _scale[0])`, float32 truncation). @Bullo27 measured `vc_render_tifxyz` on seven `vc_obj2tifxyz` meshes and showed the renderer is what defines the canvas people compare against, and it rounds in double; `446dfaf` switches to that formula and adds the seven cases as tests. It also agrees with `tifxyz_label_transfer/core.py`'s `_positive_lround`-based `full_resolution_shape`, so the package carries one canvas definition. Why not plain `round()`: Python rounds half to even, `std::lround` half away from zero, and exact halves occur with integer scales. Whether `QuadSurface::size()` in VC3D should adopt the renderer's rounding is a separate question for the maintainers.

Written with Claude Code as a coding assistant, directed and reviewed by me; the runs above are from my machine on the published data.

### DIFF
--- a/vesuvius/src/vesuvius/tifxyz/types.py
+++ b/vesuvius/src/vesuvius/tifxyz/types.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import math
 import os
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -18,29 +17,13 @@ import cv2
 import numpy as np
 from numpy.typing import NDArray
 
+from vesuvius.tifxyz_canvas import full_resolution_extent
+
 InterpolationMethod = Literal["linear", "bspline", "catmull_rom"]
 
 
-def _full_resolution_extent(stored: int, scale: float) -> int:
-    """Stored grid extent -> full-resolution canvas extent, as vc_render_tifxyz sizes it.
-
-    The renderer (``vc_render_tifxyz.cpp``, "Compute render scale") does::
-
-        double sx = render_scale / surf->_scale[0];      // _scale is cv::Vec2f
-        full_size.width = std::max(1, int(std::lround(full_size.width * sx)));
-
-    i.e. the stored scale is a float32, the reciprocal is taken in double,
-    the product is rounded half away from zero, and the result is at least 1.
-    Tracer meshes store ``scale`` as float32 0.05 (``0.05000000074505806``);
-    ``stored / scale`` in float64 then lands just below the integer
-    (``152 / 0.05000000074505806 == 3039.99995...``), so truncating it is one
-    pixel short of the canvas the renderer and the published surface volumes
-    use. Plain ``round()`` is not equivalent either: Python rounds half to
-    even, ``std::lround`` half away from zero, and exact halves occur with
-    integer scales (7 x 5 at scale 2.0 renders as 4 x 3).
-    """
-    quotient = float(stored) * (1.0 / float(np.float32(scale)))
-    return max(1, int(math.floor(quotient + 0.5)))
+# Shared with tifxyz_label_transfer so the package has one canvas definition.
+_full_resolution_extent = full_resolution_extent
 
 
 @dataclass
--- a/vesuvius/src/vesuvius/tifxyz/types.py
+++ b/vesuvius/src/vesuvius/tifxyz/types.py
@@ -20,6 +20,24 @@ from numpy.typing import NDArray
 InterpolationMethod = Literal["linear", "bspline", "catmull_rom"]
 
 
+def _full_resolution_extent(stored: int, scale: float) -> int:
+    """Stored grid extent -> full-resolution canvas extent, as VC3D computes it.
+
+    ``QuadSurface::size()`` in volume-cartographer evaluates
+    ``static_cast<int>(_points->cols / _scale[0])`` with ``_scale`` held as
+    ``cv::Vec2f``, i.e. a *float32* division followed by truncation. Tracer
+    meshes store ``scale`` as the float32 value 0.05 (``0.05000000074505806``),
+    and ``stored / scale`` in float64 then lands just below the integer
+    (``152 / 0.05000000074505806 == 3039.99995...``), so ``int()`` on the
+    float64 quotient is one pixel short of the canvas ``vc_render_tifxyz``
+    and the published surface volumes use. Doing the division in float32
+    reproduces the C++ result exactly rather than approximating it with
+    ``round``.
+    """
+    quotient = np.float32(stored) / np.float32(scale)
+    return int(quotient)
+
+
 @dataclass
 class Tifxyz:
     """A 3D surface represented as a 2D grid of (x, y, z) coordinates.
@@ -177,11 +195,7 @@ class Tifxyz:
         if self.resolution == "stored":
             return self._x.shape  # type: ignore[return-value]
         # Full resolution
-        h, w = self._x.shape
-        scale_y, scale_x = self._scale
-        if scale_y == 0 or scale_x == 0:
-            return (h, w)
-        return (int(h / scale_y), int(w / scale_x))
+        return self.full_resolution_shape
 
     @property
     def full_resolution_shape(self) -> Tuple[int, int]:
@@ -194,7 +208,10 @@ class Tifxyz:
         scale_y, scale_x = self._scale
         if scale_y == 0 or scale_x == 0:
             return (h, w)
-        return (int(h / scale_y), int(w / scale_x))
+        return (
+            _full_resolution_extent(h, scale_y),
+            _full_resolution_extent(w, scale_x),
+        )
 
     @property
     def _stored_shape(self) -> Tuple[int, int]:
--- a/vesuvius/src/vesuvius/tifxyz/types.py
+++ b/vesuvius/src/vesuvius/tifxyz/types.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import math
 import os
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -21,21 +22,25 @@ InterpolationMethod = Literal["linear", "bspline", "catmull_rom"]
 
 
 def _full_resolution_extent(stored: int, scale: float) -> int:
-    """Stored grid extent -> full-resolution canvas extent, as VC3D computes it.
+    """Stored grid extent -> full-resolution canvas extent, as vc_render_tifxyz sizes it.
 
-    ``QuadSurface::size()`` in volume-cartographer evaluates
-    ``static_cast<int>(_points->cols / _scale[0])`` with ``_scale`` held as
-    ``cv::Vec2f``, i.e. a *float32* division followed by truncation. Tracer
-    meshes store ``scale`` as the float32 value 0.05 (``0.05000000074505806``),
-    and ``stored / scale`` in float64 then lands just below the integer
-    (``152 / 0.05000000074505806 == 3039.99995...``), so ``int()`` on the
-    float64 quotient is one pixel short of the canvas ``vc_render_tifxyz``
-    and the published surface volumes use. Doing the division in float32
-    reproduces the C++ result exactly rather than approximating it with
-    ``round``.
+    The renderer (``vc_render_tifxyz.cpp``, "Compute render scale") does::
+
+        double sx = render_scale / surf->_scale[0];      // _scale is cv::Vec2f
+        full_size.width = std::max(1, int(std::lround(full_size.width * sx)));
+
+    i.e. the stored scale is a float32, the reciprocal is taken in double,
+    the product is rounded half away from zero, and the result is at least 1.
+    Tracer meshes store ``scale`` as float32 0.05 (``0.05000000074505806``);
+    ``stored / scale`` in float64 then lands just below the integer
+    (``152 / 0.05000000074505806 == 3039.99995...``), so truncating it is one
+    pixel short of the canvas the renderer and the published surface volumes
+    use. Plain ``round()`` is not equivalent either: Python rounds half to
+    even, ``std::lround`` half away from zero, and exact halves occur with
+    integer scales (7 x 5 at scale 2.0 renders as 4 x 3).
     """
-    quotient = np.float32(stored) / np.float32(scale)
-    return int(quotient)
+    quotient = float(stored) * (1.0 / float(np.float32(scale)))
+    return max(1, int(math.floor(quotient + 0.5)))
 
 
 @dataclass
--- a/vesuvius/src/vesuvius/tifxyz_canvas.py
+++ b/vesuvius/src/vesuvius/tifxyz_canvas.py
@@ -1,0 +1,41 @@
+"""Full-resolution canvas extent of a tifxyz grid, as vc_render_tifxyz sizes it.
+
+Dependency-free on purpose: ``vesuvius.tifxyz`` (OpenCV) and
+``vesuvius.tifxyz_label_transfer`` (SciPy) both need this and neither should
+import the other for it.
+"""
+
+from __future__ import annotations
+
+import math
+
+import numpy as np
+
+
+def full_resolution_extent(stored: int, scale: float) -> int:
+    """Stored grid extent -> full-resolution canvas extent.
+
+    ``vc_render_tifxyz.cpp`` ("Compute render scale") does::
+
+        double sx = render_scale / surf->_scale[0];      // _scale is cv::Vec2f
+        full_size.width = std::max(1, int(std::lround(full_size.width * sx)));
+
+    i.e. the stored scale is a float32, the reciprocal is taken in double, the
+    product is rounded half away from zero, and the result is at least 1.
+    Tracer meshes store ``scale`` as float32 0.05 (``0.05000000074505806``);
+    ``stored / scale`` in float64 then lands just below the integer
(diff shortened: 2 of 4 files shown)


Bullo27:
**Verdict (measured): right for the 0.05 tracer meshes, but `QuadSurface::size()` is not the formula that defines the rendered canvas. For the non-round `vc_obj2tifxyz` scales the Details section cites as the reason for float32, `_full_resolution_extent` is one pixel short of what `vc_render_tifxyz` actually renders.**

`vc_render_tifxyz.cpp:1569-1571` sizes the canvas as `double sx = render_scale / surf->_scale[0]; lround(full_size.width * sx)`: double precision, rounded, not `static_cast<int>(cols / _scale[0])`. The two only coincide when `stored/scale` lands within float32 rounding of an integer, which is the 1/20 case, so every test in the PR passes either way.

I ran `vc_render_tifxyz` (main build of 2 Sep; those lines last changed in #1554 and are identical in `23adee0`) with `--scale 1 -g 0` on seven `vc_obj2tifxyz` outputs from my #1683 runs and compared its `rendering WxH` line with the PR helper (all W×H):

| stored | scale (x, y) | `vc_render_tifxyz` | PR #1699 |
|---|---|---|---|
| 272×185 | 0.0499802, 0.0498272 | 5442×3713 | 5442×**3712** |
| 14×10 | 0.0023985, 0.0024457 | 5837×4089 | **5836×4088** |
| 1085×737 | 0.1998486, 0.1992154 | 5429×3700 | 5429×**3699** |
| 10841×7361 | 1.9987041, 1.9921989 | 5424×3695 | 5424×**3694** |
| 5421×3681 | 0.9992994, 0.9960730 | 5425×3696 | **5424×3695** |
| 1994×2001 | 0.3675560, 0.5414373 | 5425×3696 | 5425×**3695** |
| 272×185 | 0.05f, 0.05f | 5440×3700 | 5440×3700 |

Six of seven are off by one, the same class of mismatch as #1694. `tifxyz_label_transfer/core.py:158-163` (`_positive_lround(height / scale_y)`) matches the binary on all seven, so after this PR the package would carry two full-resolution canvases that disagree on obj2tifxyz meshes.

Plain `round()` is not the answer either: Python rounds half to even, `std::lround` half away from zero, and exact halves do occur with integer scales (crafted 7×5 grid at scale 2.0: binary 4×3, PR 3×2, `round()` 4×2). This matched the binary in all ten runs and keeps your 8 tests plus 339 neighbouring tests passing unchanged (the 3 `test_flat_inference.py` failures in my env are identical on `23adee0`):

```python
def _full_resolution_extent(stored: int, scale: float) -> int:
    # vc_render_tifxyz.cpp:1569-1571: double sx = render_scale / float(scale); lround(stored * sx)
    quotient = float(stored) * (1.0 / float(np.float32(scale)))
    return max(1, int(math.floor(quotient + 0.5)))
```

Whether VC3D's `QuadSurface::size()` should move to the renderer's formula is a separate question for the maintainers; the surface volumes people compare against come from the renderer.

ge-al:
Thank you — measured and right. I had taken `QuadSurface::size()` as the definition, but the canvas people compare against is the renderer's, and `vc_render_tifxyz.cpp` sizes it as `lround(stored * (render_scale / float32(scale)))` in double, not `int(cols / _scale)` in float32.

Pushed `446dfaf`:

- `_full_resolution_extent` is now `max(1, floor(stored * (1.0 / float(np.float32(scale))) + 0.5))` — the renderer's formula (reciprocal of the float32 scale in double, `std::lround` half-away-from-zero, at least 1). It agrees with `tifxyz_label_transfer/core.py`'s `_positive_lround` path, so the package no longer carries two canvases that disagree on obj2tifxyz meshes.
- Your seven measured `vc_render_tifxyz` cases are in `tests/tifxyz/test_full_resolution_shape.py` as a parametrized test (credited), plus the exact-half case (7×5 at scale 2.0 → 4×3; `round()` would give 4×2).
- The 0.05 tracer cases (152→3040, w035 291/262→5820/5240, the PHerc1447 202×215→4040×4300 mesh) are unchanged; `tests/tifxyz` + `tests/tifxyz_label_transfer`: 146 passed, 5 skipped.

Whether `QuadSurface::size()` should adopt the renderer's rounding is left to the maintainers, as you say; this PR only makes the Python side agree with what gets rendered.

ge-al:
Pushed `7343b42` after a self-review pass:

- **One canvas definition for the package.** The helper moved to a dependency-free module (`vesuvius/tifxyz_canvas.py`) and `tifxyz_label_transfer/core.py`'s `Surface.full_resolution_shape` (plus the native-canvas candidate in `infer_output_shape`) now use it. My earlier claim that core.py's `_positive_lround(height / scale_y)` agreed with this helper was not exactly true: core.py rounds the float64 quotient, so the two differ whenever `stored / scale` sits within ~1e-4 of .5 (e.g. 3681 at scale 0.3313977 → core.py 11108, renderer 11107). A test now pins the two packages to the same answer.
- **Guard:** a scale that underflows float32 to 0 (1e-46) slipped past the float64 zero check into `ZeroDivisionError`; non-positive/non-finite float32 scales now fall back to the stored extent.
- Behaviour notes added to the description (round-half-up on any fraction ≥ .5, minimum extent 1, `QuadSurface::size()` still truncates and is not the canvas of record).

`tests/tifxyz` + `tests/tifxyz_label_transfer` + `tests/test_surface_preflight.py`: 168 passed, 5 skipped.

Bullo27:
Measured against the renderer for the near-.5 claim in 7343b42 (crafted square grids, same `vc_render_tifxyz` build and `--scale 1 -g 0` recipe as the table above): in all six cases the binary matches `full_resolution_extent` and not the float64 `lround` that `tifxyz_label_transfer/core.py` used before, so moving core.py onto the shared helper is the right direction.

| stored | scale | `vc_render_tifxyz` | helper (7343b42) | old core.py |
|---|---|---|---|---|
| 4 | 1.6 | 2 | 2 | 3 |
| 6 | 0.9230769 | 6 | 6 | 7 |
| 10 | 0.2247191 | 44 | 44 | 45 |
| 12 | 0.5106383 | 24 | 24 | 23 |
| 16 | 0.256 | 62 | 62 | 63 |
| 17 | 2.2666667 | 8 | 8 | 7 |

`tests/tifxyz` + `tests/tifxyz_label_transfer` + `tests/test_surface_preflight.py`: 168 passed, 5 skipped here as well.

ge-al:
Thank you for measuring the near-.5 cases against the binary — that is exactly the evidence the core.py change needed. Added all six as `test_near_half_quotients_match_vc_render_tifxyz` in `acc7862` (credited), so the shared helper is now pinned to the renderer on 7 obj2tifxyz meshes, 6 crafted near-.5 grids, the exact-half case and the 0.05 tracer cases. Same three suites: 174 passed, 5 skipped.

ge-al:
Measured how much of the bucket this touches. Reading the `meta.json` + `x.tif` header of every published mesh directory (593 mesh variants across all 315 segments; metadata only), **207 / 593 (35 %)** store the scale as the float32 value `0.05000000074505806`, and on every one of them the pre-PR `int(h / scale)` canvas is one pixel short of what `vc_render_tifxyz` renders. Cross-checked against the published surface volumes: 262 / 628 reproduce their `canvas_size` exactly from a published mesh variant with this PR's formula and one isotropic render scale (143 at render scale 1, 103 at exactly 4.0, 16 at other clean ratios) — 0 of the float32-rounded ones with the old formula. The other 366 (cross-scan renders with non-integer implied scales) do not reproduce to the pixel with either formula, so that is a render-scale/affine question rather than this rounding one (#1727); per-volume numbers in https://github.com/ge-al/vesuvius-cpu-repro (audit section).

*AI-assisted (Claude Code), human-directed; measured on my machine from bucket metadata. Edited: the first version of this comment said all 262 were at render scale 1; 119 are cross-scan renders at clean ratios.*

ge-al:
@jrudolph @bruniss a friendly review request when you have a moment, for this and the two small PRs opened alongside it (#1700 zarr 3 for `compute_st`, #1702 YAML training schedule). All three are rebased onto today's `main` (`2dcfaf6`) with their tests passing, and @Bullo27's renderer and zarr 3 measurements are folded into this one and #1700.

I'm asking rather than waiting because they reach the 28-day auto-close on 2 Oct, and with the three-open-PR limit (#1743) four more measured fixes are parked as drafts behind them (#1824–#1827, including the `NameError` that stops `vesuvius.finalize_outputs` from starting at all). If any of these isn't useful, closing it is a fine answer too, since that frees a slot.

*AI-assisted (Claude Code), human-directed.*

PR #1700, opened 2026-09-04
structure_tensor: write zarr through helpers that exist under zarr 2 and 3 (fixes #1448)

### LINKED ISSUE #1448: zarr 3 installs: compute_st and several image_proc modules call APIs zarr 3 removed

ge-al:
**In one sentence:** `vesuvius.compute_st` (structure tensor → eigenanalysis → OME `first_component/second_component/normal/confidence` outputs) runs again on a zarr 3 install, and it no longer reports "Completed Successfully" after the child process failed to create its output store.

**One real example:** Starting with a 31×512×512 crop of the published PHerc1447 surface volume `segments/20250703025628-auto_grown_20250703025628283/surface-volumes/8.64um-1.2m-116keV-volume-20250521151220.zarr` (level 0), I ran `vesuvius.compute_st --input_dir crop.zarr --output_dir st --patch_size 31,128,128 --num-workers 0 --keep-eigen` on a `uv sync --extra models` environment (zarr 3.2.1), and it produced the full `structure_tensor / eigenvectors / eigenvalues / first_component / second_component / normal / confidence` store in 19 s on CPU.

**Before:** the child died at store creation with

```
  File ".../vesuvius/structure_tensor/create_st.py", line 206, in _create_output_stores
    self.output_store = root_store.create_dataset(
AttributeError: 'Group' object has no attribute 'create_dataset'
An error occurred during inference: 'Group' object has no attribute 'create_dataset'
...
--- Structure Tensor Processing Completed Successfully ---      <-- wrong: the store holds only a v3 zarr.json
Output saved to: .../st_before2.zarr

--- Running Eigenanalysis ---
FileNotFoundError: structure_tensor
```

(full log: [compute_st_before.log](https://raw.githubusercontent.com/ge-al/villa/proof-assets/compute_st_before.log)). With `--structure-tensor-only` the runner exits **0** on that empty store, so a batch script has no way to notice. The same removed APIs sit behind the eigenanalysis writers, `create_vf.py`, `vf_format.py`, `vesuvius.voxelize_obj` and `generate_mutex_graph.py` (#1448 lists every line).

**After this PR:**

```
--- Structure Tensor Computation Finished ---
Structure tensor saved to: .../st_after.zarr/structure_tensor
--- Structure Tensor Processing Completed Successfully ---
--- Running Eigenanalysis ---
[Eigen] Chunks: 100%|██████████| 16/16 [00:07<00:00,  2.17it/s]
Eigenanalysis completed successfully
--- All computations completed successfully ---
  - Structure tensor: .../st_after.zarr/structure_tensor
  - first_component/, second_component/, normal/, confidence/ (scale '0', ds=1)
  - Eigenvectors: .../st_after.zarr/eigenvectors
  - Eigenvalues:  .../st_after.zarr/eigenvalues
EXIT: 0
```

On disk it is the same Zarr v2 layout the C++ tools and every existing reader expect:

```
group format: 2
structure_tensor       shape=(6, 31, 512, 512) chunks=(6, 31, 128, 128) dtype=float32 .zarray=True
eigenvectors           shape=(9, 31, 512, 512) chunks=(1, 31, 128, 128) dtype=float32 .zarray=True
eigenvalues            shape=(3, 31, 512, 512) chunks=(1, 31, 128, 128) dtype=float32 .zarray=True
normal/z/0             shape=(31, 512, 512)    chunks=(31, 128, 128)    dtype=uint8   .zarray=True
first_component/x/0    shape=(31, 512, 512)    chunks=(31, 128, 128)    dtype=uint8   .zarray=True
confidence/0           shape=(31, 512, 512)    chunks=(31, 128, 128)    dtype=uint8   .zarray=True
compressor: {'id': 'blosc', 'cname': 'zstd', 'clevel': 3, 'shuffle': 1, 'blocksize': 0}
```

**Proof:** layer 15 of the input crop, the leading eigenvalue, and the written `confidence/0` (the masked holes in the surface volume come out as zero confidence, papyrus as high):

compute_st output on the PHerc1447 crop

Tests: new `tests/structure_tensor/test_zarr3_compat.py` (11 tests) drives the real writers — `_finalize_structure_tensor_torch` with `ome_out` + `keep_eigen`, `OMEU8VectorWriter` create-then-reopen, and the helpers' v2 layout / nested separator / `write_empty_chunks` / `require` semantics — and runs under whichever zarr major is installed. `tests/structure_tensor` + `tests/data/test_utils.py` + `tests/data/test_open_zarr_format.py`: 39 passed.

**Why / where this is useful:** structure-tensor / fiber-direction volumes feed the fiber tools and VC3D's normal grids; anyone who installs with the lockfile (zarr 3) currently cannot produce them. The fix is three small helpers in `vesuvius.data.utils` (`open_zarr_group`, `create_zarr_array`, `require_zarr_array`) so the same pattern can be applied to the remaining `create_dataset` sites elsewhere.

## Details

Fixes #1448. Tested at `23adee0` on macOS (CPU), Python 3.14, zarr 3.2.1, numcodecs 0.16; @Bullo27 ran the test file under zarr 2.18.7, 3.2.1 and 3.3.0 (see thread).

- The helpers always create groups as **Zarr v2** (`zarr_format=2` under zarr 3, no-op under zarr 2): a v3 group rejects the numcodecs `Blosc` compressor these writers pass and would write `zarr.json` instead of `.zgroup`/`.zarray`. `create_zarr_array` maps `write_empty_chunks` → `config={...}` and `dimension_separator` → `chunk_key_encoding` under zarr 3, and uses the non-deprecated `compressors=` keyword. If a caller appends to a store somebody created as v3, a `Blosc` compressor is translated to `zarr.codecs.BloscCodec`; anything else raises a clear error instead of zarr's internal `TypeError`.
- `require_zarr_array` keeps zarr 2's `require_dataset` contract: an existing array is returned after a shape check and a safe dtype-cast check, with its chunks and compressor untouched (`overwrite` is not consulted), so a resumed `compute_vf` run keeps what it already wrote. `create_zarr_array` pins `write_empty_chunks=True` under zarr 3 unless the caller passes `False`, matching zarr 2's default so chunk files on disk agree across majors.
- `open_zarr_group` in append modes (`'a'`/`'r+'`) keeps the format of a store that already holds arrays, recreates an *empty* leftover v3 group (what a failed run on zarr 3 leaves behind) as v2, and creates v2 when nothing is there; `'w'`/`'w-'` always create v2.
- `vesuvius.voxelize_obj --chunk_size N` ran its dask rechunk through `da.to_zarr(store, component=…)`, which under zarr 3 writes a v3 array and a root `zarr.json` into the v2 label store and hides the level arrays. The temporary array is now created through the v2 helper and handed to dask as an object, and the final rename no longer relies on `Group.move` (unimplemented in zarr 3).
- `create_st.main` now returns 1 when `infer()` returns `None` (it swallows the exception and prints the traceback itself), so `run_create_st` no longer announces success on an empty store.
- Sites changed: `run_create_st.py`, `create_st.py` (store creation, eigenanalysis OME/eigen arrays), `create_vf.py` (`require_dataset`), `vf_format.py`, `image_proc/run/voxelize_objs.py` (`zarr.DirectoryStore` + `create_dataset`), `image_proc/run/generate_mutex_graph.py`. `stack_composite_tifs.py` also uses `DirectoryStore`, but it imports a module that does not exist on `main` (noted in #1448), so it is left alone.
- `tests/structure_tensor/test_structure_tensor.py` still has three commented-out tests that use `root.create_dataset`; the new test file covers the same eigenanalysis path through the helpers.

Written with Claude Code as a coding assistant, directed and reviewed by me; the runs above are from my machine on the published data.

### DIFF
--- a/vesuvius/src/vesuvius/data/utils.py
+++ b/vesuvius/src/vesuvius/data/utils.py
@@ -78,7 +78,9 @@ def create_zarr_array(group, name: str, *, shape=None, data=None, chunks=None, d
     zarr 3 removed ``Group.create_dataset``/``require_dataset`` and moved
     ``write_empty_chunks`` into ``config`` and ``dimension_separator`` into
     ``chunk_key_encoding``. This wraps both APIs so callers keep one code path
-    and the array is written exactly as before (v2 metadata, same chunk keys).
+    and the array is written as zarr 2 wrote it (v2 metadata, same chunk keys,
+    and empty chunks materialised unless ``write_empty_chunks=False`` is
+    passed, which is zarr 2's default but not zarr 3's).
     """
     if data is not None:
         data = np.asarray(data)
@@ -95,8 +97,11 @@ def create_zarr_array(group, name: str, *, shape=None, data=None, chunks=None, d
         create_kwargs['compressors'] = _v3_compressor(group, compressor)
         if dimension_separator is not None:
             create_kwargs['chunk_key_encoding'] = {'name': 'v2', 'separator': dimension_separator}
-        if write_empty_chunks is not None:
-            create_kwargs['config'] = {'write_empty_chunks': bool(write_empty_chunks)}
+        # zarr 2 defaults write_empty_chunks=True, zarr 3 defaults False; pin
+        # zarr 2's behaviour so the chunk files on disk match across majors.
+        create_kwargs['config'] = {
+            'write_empty_chunks': True if write_empty_chunks is None else bool(write_empty_chunks)
+        }
         array = group.create_array(name, **create_kwargs)
         if data is not None:
             array[...] = data
@@ -117,19 +122,27 @@ def create_zarr_array(group, name: str, *, shape=None, data=None, chunks=None, d
 def require_zarr_array(group, name: str, *, shape, **kwargs) -> zarr.Array:
     """``Group.require_dataset`` for both zarr majors.
 
-    Returns the existing array when one of that name and shape is present,
-    otherwise creates it through :func:`create_zarr_array`. ``overwrite=True``
-    always (re)creates, matching zarr 2's ``require_dataset(..., overwrite=True)``.
+    Same contract as zarr 2's ``require_dataset``: an existing array of that
+    name is returned after a shape check and a dtype cast check (its chunks
+    and compressor are left as they are, and ``overwrite`` is not consulted);
+    otherwise the array is created through :func:`create_zarr_array`. A
+    resumed ``compute_vf`` run therefore keeps the chunks it already wrote.
     """
-    overwrite = bool(kwargs.pop('overwrite', False))
-    if not overwrite and name in group:
+    if name in group:
         existing = group[name]
-        if tuple(existing.shape) != tuple(shape):
+        if not hasattr(existing, 'shape'):
+            raise TypeError(f"{name!r} exists in {group.path!r} but is a group, not an array")
+        if tuple(int(v) for v in existing.shape) != tuple(int(v) for v in shape):
             raise TypeError(
                 f"array {name!r} exists with shape {tuple(existing.shape)}, requested {tuple(shape)}"
             )
+        dtype = kwargs.get('dtype')
+        if dtype is not None and not np.can_cast(existing.dtype, np.dtype(dtype), casting='safe'):
+            raise TypeError(
+                f"array {name!r} exists with dtype {existing.dtype}, requested {np.dtype(dtype)}"
+            )
         return existing
-    return create_zarr_array(group, name, shape=shape, overwrite=overwrite, **kwargs)
+    return create_zarr_array(group, name, shape=shape, **kwargs)
 
 # Function to get the maximum value of a dtype
 def get_max_value(dtype: np.dtype) -> Union[float, int]:
--- a/vesuvius/src/vesuvius/data/utils.py
+++ b/vesuvius/src/vesuvius/data/utils.py
@@ -16,11 +16,24 @@ def open_zarr_group(path, mode: str = 'r', storage_options: Optional[Dict[str, A
     different on-disk layout (``zarr.json`` instead of ``.zgroup``/``.zarray``).
     Every writer in this package predates zarr 3 and its consumers (VC3D,
     ``vesuvius.data.utils.open_zarr``, the eigenanalysis readers) expect v2,
-    so ask for ``zarr_format=2`` whenever the group may be created. zarr 2 has
-    no ``zarr_format`` argument, and an existing store keeps its own format.
+    so ask for ``zarr_format=2`` whenever the group is created. Append modes
+    (``'a'``, ``'r+'``) keep the format of an existing store and only fall
+    back to v2 when nothing is there yet. zarr 2 has no ``zarr_format``
+    argument.
     """
-    if _ZARR_V3 and mode in ('w', 'w-', 'a', 'r+') and 'zarr_format' not in kwargs:
-        kwargs['zarr_format'] = 2
+    if _ZARR_V3 and 'zarr_format' not in kwargs:
+        if mode in ('w', 'w-'):
+            kwargs['zarr_format'] = 2
+        elif mode in ('a', 'r+'):
+            # Appending: keep whatever format the store already has. Forcing
+            # v2 here would write a second .zgroup next to an existing
+            # zarr.json and leave a store that reads as v3 and empty.
+            try:
+                kwargs['zarr_format'] = zarr.open_group(
+                    path, mode='r', storage_options=storage_options
+                ).metadata.zarr_format
+            except FileNotFoundError:  # GroupNotFoundError: nothing there yet
+                kwargs['zarr_format'] = 2
     if storage_options is not None:
         kwargs['storage_options'] = storage_options
     return zarr.open_group(path, mode=mode, **kwargs)
--- a/vesuvius/src/vesuvius/data/utils.py
+++ b/vesuvius/src/vesuvius/data/utils.py
@@ -25,15 +25,24 @@ def open_zarr_group(path, mode: str = 'r', storage_options: Optional[Dict[str, A
         if mode in ('w', 'w-'):
             kwargs['zarr_format'] = 2
         elif mode in ('a', 'r+'):
-            # Appending: keep whatever format the store already has. Forcing
-            # v2 here would write a second .zgroup next to an existing
-            # zarr.json and leave a store that reads as v3 and empty.
+            # Appending: keep the format of a store that already holds data.
+            # Forcing v2 would write a second .zgroup next to an existing
+            # zarr.json and leave a store that reads as v3 and empty. An
+            # *empty* v3 group, though, is what a failed run on zarr 3 leaves
+            # behind (zarr.json only); there is nothing to keep, so recreate
+            # it as v2 rather than write a v3 store no consumer reads.
             try:
-                kwargs['zarr_format'] = zarr.open_group(
+                existing = zarr.open_group(
                     path, mode='r', storage_options=storage_options
-                ).metadata.zarr_format
+                )
             except FileNotFoundError:  # GroupNotFoundError: nothing there yet
                 kwargs['zarr_format'] = 2
+            else:
+                fmt = int(existing.metadata.zarr_format)
+                if fmt != 2 and len(existing) == 0:
+                    mode, kwargs['zarr_format'] = 'w', 2
+                else:
+                    kwargs['zarr_format'] = fmt
     if storage_options is not None:
         kwargs['storage_options'] = storage_options
     return zarr.open_group(path, mode=mode, **kwargs)
--- a/vesuvius/src/vesuvius/data/utils.py
+++ b/vesuvius/src/vesuvius/data/utils.py
@@ -6,6 +6,109 @@ from typing import Union, Dict, Any, Optional, Tuple
 
 _ZARR_V3 = int(zarr.__version__.split('.', 1)[0]) >= 3
 
+
+def open_zarr_group(path, mode: str = 'r', storage_options: Optional[Dict[str, Any]] = None,
+                    **kwargs) -> "zarr.Group":
+    """``zarr.open_group`` that yields a Zarr v2 group under both zarr majors.
+
+    zarr 3 creates *v3* groups by default, and arrays created inside a v3
+    group reject the numcodecs compressors this package uses and produce a
+    different on-disk layout (``zarr.json`` instead of ``.zgroup``/``.zarray``).
+    Every writer in this package predates zarr 3 and its consumers (VC3D,
+    ``vesuvius.data.utils.open_zarr``, the eigenanalysis readers) expect v2,
+    so ask for ``zarr_format=2`` whenever the group may be created. zarr 2 has
+    no ``zarr_format`` argument, and an existing store keeps its own format.
+    """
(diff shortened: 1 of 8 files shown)


Bullo27:
**Measured on 2978dab under zarr 3.2.1 and 3.3.0: `open_zarr_group(path, mode='a')` on an existing zarr v3 store does not "keep its own format"; it writes a second, v2 group next to `zarr.json`, so the "appends to a store somebody created as v3" branch of `_v3_compressor` is unreachable through the helper.**

Repro (v3 store from `zarr.open_group(p, mode='w', zarr_format=3)` with one array `existing`):

```
open_zarr_group(p, mode='a')      -> group format 2; files: .zattrs .zgroup existing zarr.json
create_zarr_array(g, 'new', ...)  -> new/.zarray
zarr.open_group(p, mode='r')      -> format 3, members ['existing']  (ZarrUserWarning "Both zarr.json ... and .zgroup ... exist. Zarr format 3 will be used")
open_zarr(p, mode='r')            -> members []
```

That is the state a #1448 user is left in: your "Before" log shows the failed run leaving a `zarr.json`-only store. Same for `create_vf.py` on `main` (measured: `zarr.open_group(out, mode='a')`, then `require_dataset` → `AttributeError`, `zarr.json` left behind); re-running this branch on the same `--output` yields `.zgroup` + `zarr.json`, `U` readable via `open_zarr(out/U)` but `zarr.open_group(out)` reads as empty. Only the two `mode='a'` sites (`create_vf.py`, `OMEU8VectorWriter`) are affected; the `mode='w'` sites wipe the store first (measured).

Variant that measured OK under 2.18.7 / 3.2.1 / 3.3.0 (missing store → v2; existing store keeps its format, which makes the BloscCodec branch reachable; `GroupNotFoundError` subclasses `FileNotFoundError`):

```python
if _ZARR_V3 and 'zarr_format' not in kwargs:
    if mode in ('a', 'r+'):
        try:
            kwargs['zarr_format'] = zarr.open_group(path, mode='r', storage_options=storage_options).metadata.zarr_format
        except FileNotFoundError:
            kwargs['zarr_format'] = 2
    elif mode in ('w', 'w-'):
        kwargs['zarr_format'] = 2
```

Refusing v3 stores outright would be defensible too, since every consumer wants v2. Everything else checked out: all 6 new tests pass here with torch (zarr 3.2.1), and the 5 helper tests plus a fill_value / empty-chunk / `.zarray` parity probe pass identically under zarr 2.18.7 and 3.3.0.

ge-al:
Measured and right, thank you — forcing `zarr_format=2` in append mode on a store that already has a `zarr.json` produces exactly the split you describe, and it is the state a #1448 user's first failed run leaves behind.

Pushed `7961d67` with your variant: `'w'`/`'w-'` still create v2; `'a'`/`'r+'` read the existing group's `metadata.zarr_format` and only fall back to v2 on `FileNotFoundError` (`GroupNotFoundError`). That also makes the `BloscCodec` translation in `_v3_compressor` reachable, so `create_vf.py` / `OMEU8VectorWriter` appending to a v3 store now write a v3 array with a v3 codec instead of a stray `.zgroup`.

Tests added in `tests/structure_tensor/test_zarr3_compat.py`: append to a missing path → `.zgroup`, no `zarr.json`, v2 array; append to an existing v3 store (skipped under zarr 2) → format stays 3, no `.zgroup` written, `zarr.open_group(path)` lists both arrays and the new one round-trips. `tests/structure_tensor` + `tests/data/test_utils.py` + `tests/data/test_open_zarr_format.py`: 36 passed under zarr 3.2.1.

I went with keeping the format rather than refusing v3 stores, since a user who already has the stray-`zarr.json` store from a failed run on `main` should be able to re-run `--output` in place; happy to switch to a hard refusal if the maintainers prefer one on-disk format.

Bullo27:
One measured consequence of 7961d67 for the in-place re-run you describe: the stray store from a failed `main` run is an *empty* v3 group, so keeping its format makes `create_vf.py` re-run on that `--output` write a v3 store (files: `U/c`, `U/zarr.json`; no `.zarray`/`.zgroup` anywhere), which by the PR's own description the C++ tools and the v2 readers will not open. Same result under zarr 3.2.1 and 3.3.0.

Since an empty leftover group has nothing to keep, recreating it as v2 gets that user the v2 store they wanted; the format-keeping path then only applies to v3 stores that actually hold data (a warning there would also be defensible). Measured OK under 3.2.1 / 3.3.0 (v2 and missing paths unchanged):

```python
try:
    existing = zarr.open_group(path, mode='r', storage_options=storage_options)
except FileNotFoundError:
    kwargs['zarr_format'] = 2
else:
    fmt = existing.metadata.zarr_format
    if fmt != 2 and len(existing) == 0:
        mode, kwargs['zarr_format'] = 'w', 2   # empty leftover group: nothing to keep
    else:
        kwargs['zarr_format'] = fmt
```

7961d67 otherwise checks out here: 8/8 in the new test file under 3.2.1 with torch, 7/7 under 3.3.0 and 6 passed + 1 skipped under 2.18.7.

ge-al:
Right again — I had only thought about a v3 store with data in it, and the leftover from a failed `main` run is exactly the empty case. Pushed `43a626e` with your variant: an empty non-v2 group on `'a'`/`'r+'` is recreated as v2 (`mode='w'`, `zarr_format=2`); a v3 store that holds arrays keeps its format as before. Added `test_open_zarr_group_append_recreates_empty_v3_leftover_as_v2` (zarr.json-only group → after append: `.zgroup`, no `zarr.json`, `U/.zarray`, and `zarr.open_group(path)` lists `['U']`). `tests/structure_tensor` + `tests/data/test_utils.py` + `tests/data/test_open_zarr_format.py`: 37 passed under 3.2.1. Thanks for running it under 3.3.0 and 2.18.7 as well.

ge-al:
Pushed `13ad9df` after a self-review pass against zarr 2.18's semantics; three things changed:

- `require_zarr_array` now behaves like `Group.require_dataset` did: an existing array is returned after a shape check and a safe dtype-cast check, chunks/compressor untouched, and `overwrite` is not consulted. My first version deleted and recreated it when `overwrite=True` (which `create_vf.py` passes), which on zarr 2 would have wiped the chunks of a resumed `compute_vf` run — a behaviour change the old code never had. A group at the requested name raises a clear `TypeError` instead of an `AttributeError`.
- `create_zarr_array` pins `write_empty_chunks=True` under zarr 3 unless the caller passes `False` (zarr 2's default is True, zarr 3's is False), so the chunk files on disk match across majors, not just the values.
- `vesuvius.voxelize_obj --chunk_size N` (non-default) still corrupted the store under zarr 3: the dask rechunk used `da.to_zarr(store, component=…)`, which writes a v3 array plus a root `zarr.json` into the v2 label store and hides the level arrays. The temporary array is now created through the v2 helper and handed to dask as an object, and the rename no longer relies on `Group.move` (unimplemented in zarr 3). Test `test_voxelize_rechunk_keeps_store_v2` covers it end to end.

`tests/structure_tensor` + `tests/data/test_utils.py` + `tests/data/test_open_zarr_format.py`: 39 passed under 3.2.1.

hendrikschilling:
```
[P2] Recovering an “empty” v3 group deletes its attributes.
  At vesuvius/src/vesuvius/data/utils.py:42 (https://github.com/ScrollPrize/villa/blob/912ab0c673d270a6662c980fac519792885f3955/vesuvius/src/vesuvius/data/utils.py#L42), len(existing) == 0
  means no child arrays/groups, not no metadata. Switching to mode='w' also deletes existing attributes.

  I reproduced this with a metadata-only group containing multiscales: after opening in append mode, its attributes were {}. Preserve those attributes during conversion, or restrict       
  automatic recovery to groups without attributes. This does not affect ordinary fresh outputs or populated stores.
```

ge-al:
Right, and thank you — `len(existing) == 0` was the wrong test for "nothing to keep". A group can hold attributes before its first array: OME `multiscales` in your reproduction, and this package's own `patch_size`/`sigma`/`original_volume_shape` that `run_create_st.py` writes on the root group.

Pushed `565908c`, taking the first of your two options so the leftover recovery still works on a group that has metadata: the attributes are read before the conversion and written back onto the v2 group.

```python
if fmt != 2 and len(existing) == 0:
    preserved_attrs = dict(existing.attrs)
    mode, kwargs['zarr_format'] = 'w', 2
...
group = zarr.open_group(path, mode=mode, **kwargs)
if preserved_attrs:
    group.attrs.update(preserved_attrs)
```

Test `test_open_zarr_group_append_keeps_attrs_of_empty_v3_group` seeds an empty v3 group with `multiscales` and `patch_size`, appends through the helper, and asserts the group is v2 with both attributes intact and readable from a fresh open. It fails on the previous commit with `KeyError` from `zarr/core/attributes.py`, which is your reproduction. `tests/structure_tensor`: 32 passed.

*AI-assisted (Claude Code), human-directed.*

PR #1701, opened 2026-09-04
patch_cache: notice a label that changed under the same path

khj1222:
**In one sentence:** The patch cache now notices when the labels underneath it changed, instead
of handing back a split that describes labels that no longer exist.

**One real example:** Starting with `phercparis4-w00` from `ink_9um`, I ran patch discovery and
it found **1,266** patches. I halved that segment's supervision mask **in place**, at the same
path, and re-ran against the same `out_dir`: **1,266 again, with identical bounding boxes**. A
fresh directory gives **1,162**.

**Before:** 104 patches kept training on supervision that had been deleted, and nothing in the
run looked wrong — the count matched, so the natural reading was that the mask edit had not
taken effect. That is what a held-out split looks like while you are building one, and it cost
a day in July before the cache was the suspect.

**After this PR:** the second run rejects the cache and rediscovers, returning 1,162 — the same
answer as the fresh directory.

**Proof:** the reproduction above, on that segment, before and after the change. The cache is
keyed on the label *paths*, so pointing a run at a different tree was always caught; a mask
regenerated at the same path was not.

patch cache, before and after

*Console output of the script linked below, run against a copy of this segment's real ink labels: the cache returning the stale split before, and rejecting it after.*

Cost of the check, on this repository's real w00 label assets — 12,619 files across the ink
labels and the supervision mask:

| | |
|---|---|
| the fingerprint, five runs | **22.5 – 32.4 ms** |
| the same thing written on `Path.stat()` | **187.4 – 270.5 ms** |

No chunk data is read; `os.scandir` hands back the size as it walks, and that is the entire
difference.

Script and raw output: https://github.com/khj1222/vesuvius-challenge/tree/main/runs/f4_main (`f4_shot.py`, `f4_console.txt`, `real_fingerprint_cost.json`)

Anyone building a validation split gets caught by this the same way. It cost me a day in July.

## Details

Each label asset is fingerprinted by the relative names, sizes and modification times of the
files under it (the mtime was added after review, see the update below), and
that fingerprint goes into the cached record and into the lookup key — so the rejection path
that already exists does the work, and there is no new invalidation logic. Names and sizes
rather than contents, because that is what moves for the changes that matter: rewriting a zarr
changes chunk sizes, and dropping annotation deletes chunks outright when the store does not
write empty ones.

The whole change fits in `ink_detection/data/patch_cache.py`; the `Segment` type is untouched.

Three tests are added beside the existing round-trip test, which passes unmodified: a mask
regenerated in place is rejected when a chunk disappears, when one is rewritten to a
different size under the same name, and when one is rewritten to the same size; an untouched tree keeps hitting the cache; and the
fingerprint is stable, order-independent, sensitive to each asset, and defined for a missing
path.

Limits, stated rather than found later: a copy of a tree fingerprints the same as its source
only when the copy preserved modification times, and a label rewritten to the same bytes at a
new time misses once — both err towards rebuilding the split, which is the cheap side; and a
cache written before this change carries no fingerprint, so it misses once and is rebuilt —
no migration.

**Update after review (2026-09-18):** on hendrikschilling's suggestion the fingerprint now
includes `st_mtime_ns` next to `st_size`, so a mask edited to the same size under the same
name is caught too. `stat()` was already called per file, and on the same 12,619 real label
files the walk costs 24.2 – 27.3 ms against 22.9 – 25.7 ms for the size-only digest, inside
the run-to-run spread. The one property this trades away is the size-only version's
"byte-identical copy hits the cache" — it now hits only if the copy kept mtimes
(`cp -p`, `robocopy`, `shutil.copytree` do; a plain `cp` does not), and the paragraph above
is rewritten accordingly. Commit `0c2ce57`.

This supersedes [#1661](https://github.com/ScrollPrize/villa/pull/1661), which does the same
thing across three files against `merge-ink-pipelines`, where this identity is not yet in one
place. I will close that one. Same branch question as on
[#1608](https://github.com/ScrollPrize/villa/pull/1608).

**Disclosure:** most of the work in this project, including this reproduction, the change and
the tests, is done with an AI coding assistant. The problem, the data and the decisions are
mine; the failure above is one my project hit while building held-out splits over the
`ink_9um` corpus.

Tested at `5479453` (`main`), Python 3.12, Windows 11.


hendrikschilling:
While you are at it I suggest including mtime to catch more changes - stat is already used so thats basically for free.

```
[P2] Same-size mask edits still return stale patches.
  vesuvius/src/vesuvius/ink_detection/data/patch_cache.py:57
  (https://github.com/ScrollPrize/villa/blob/ec9b5cb6f9841194a054124bac4976f0e2dcf776/vesuvius/src/vesuvius/ink_detection/data/patch_cache.py#L57)

  The fingerprint includes file names and sizes, but neither contents nor modification times. This is explicitly acknowledged in the PR, but it occurs with ordinary masks, not just        
  unlikely compression collisions.
```

khj1222:
Done in 0c2ce57: the fingerprint now hashes `st_mtime_ns` next to `st_size`, from the same `stat()` call the walk already makes. Agreed that the same-size case is ordinary rather than a compression corner — a mask painted over in place lands there whenever the chunk stays the same length, which for uncompressed chunks is always.

Cost on the same 12,619 real label files (five runs each): 24.2 – 27.3 ms with mtime against 22.9 – 25.7 ms without, so inside the run-to-run spread.

The one property it trades away is the size-only version's "a byte-identical copy of the tree hits the cache": that now holds only when the copy preserved modification times (`cp -p`, `robocopy`, `shutil.copytree` do; a plain `cp` does not), and a label rewritten to the same bytes at a new time misses once. Both err towards rebuilding the split, which is the cheap side, so I think that is the right trade — but if you would rather keep copies cache-stable I can drop the mtime and leave the same-size hole documented instead.

Tests: the in-place test now covers the same-size edit (content changed, mtime bumped), and the fingerprint test covers an mtime-preserving copy hitting and a touched copy missing; the repository's round-trip test still passes unmodified. The PR body's fingerprint description and limits paragraph are updated to match.

PR #1704, opened 2026-09-04
3D DCT based compression backend for VC3D

SuperOptimizer:
**In one sentence:** This adds [volume-compressor](https://github.com/SuperOptimizer/volume-compressor) support to VC3D as a 3D DCT based compression backend. 

**One real example:** I exported all 39 scrolls and ~65 volumes from the AWS bucket to [my community uploads]( https://dl.ash2txt.org/community-uploads/forrest/volcomp/ ) that users can stream and download. The layout is the same as the aws open data access bucket so users should be able to replace e.g. `s3://vesuvius-challenge-open-dataPHercParis4/volumes/20260411134726-2.400um-0.2m-78keV-masked.zarr/` with `https://dl.ash2txt.org/community-uploads/forrest/volcomp/PHercParis4/volumes/20260411134726-2.400um-0.2m-78keV-masked.zarr/`

**Before:** The user would have to download ~300TB of uncompressed data to have all available data

**After this PR:** I compressed all the data at Q=8/4/2/1 and hit a ~44x compression ratio so all the data is now ~6TB or so. Level 0 is compressed at Q=8, level 1 at Q=4, level 2 at Q=2, levels 3-5 at Q=1

**Proof:** [https://dl.ash2txt.org/community-uploads/forrest/volcomp/](https://dl.ash2txt.org/community-uploads/forrest/volcomp/) Is the link to the data. Some benchmark data is [here](https://github.com/SuperOptimizer/volume-compressor/blob/main/docs/BENCHMARKS.md) compared to a different codec I worked on. Some sample numbers on my 275HX from Paris 4 are

| q | codec | ratio | PSNR dB | SSIM | MAE | P90 | P95 | P99 | max | enc MB/s | dec MB/s |
|--:|---|--:|--:|--:|--:|--:|--:|--:|--:|--:|--:|
| 2 | volcomp | **17.56** | 41.33 | .9840 | 1.573 | 3 | 4 | 7 | 49 | 471 | 702 |
| 4 | volcomp | **30.05** | 38.04 | .9700 | 2.282 | 5 | 6 | 11 | 77 | 574 | 906 |
| 8 | volcomp | **52.92** | 35.00 | .9463 | 3.217 | 7 | 9 | 15 | 121 | 784 | 1375 |
| 16 | volcomp | **95.58** | 32.22 | .9088 | 4.416 | 10 | 13 | 21 | 143 | 1033 | 2135 |
| 32 | volcomp | **176.5** | 29.67 | .8532 | 5.911 | 13 | 17 | 28 | 155 | 1113 | 2484 |

**Why / where this is useful:**  It's a nearly 50x reduction in bandwidth and storage size. You can start to see artifacting and banding at Q=8 at > 2x zoom at native resolution so I'd argue it's fine for visual inspection. 

Models that haven't been finetuned on compressed data are sensitive to it so I wouldn't use the released models on comrpessed data. I am currently working on finetuning the upstream models on this data. Based on some older experiments I think most of the discrepancy due to compression can be recovered but I've not finished that work yet.

- [✅] I personally verified that the example and proof above were produced by this PR on the stated data.

## Details

Volume-Compressor is a 3D DCT based compression codec for volumetric data. It does a 3d DCT 16^3 on each of the 512 blocks in a chunk. These are entropy coded together into a single 128^3 chunk. These chunks are sharded together into 1024^3 shards. The shards are placed into a zarr v3 compatible container.


hendrikschilling:
Looks good on its own, however this PR also adds trailing index support for zarr and does so incompletely, see below.
So either remove that part and use index at start format, or fix it.
I'd prefer index start for cleaner header access patterns, but not a requirement.
FYI this exposed a bug in vc3d remote shard handling code which I'm fixing right now.

```
[P1] Reject or correctly implement writes to trailing-index shards

  This PR now preserves index_location="end" in Zarr metadata and honours it when reading, but the shard writers still assume the index is at the beginning.

  write_shard() therefore writes a front-index shard whose metadata advertises a trailing index. More seriously, write_inner_chunk_to_shard() updates the index entry at linear * 16, which 
  can overwrite compressed payload bytes in an existing trailing-index shard while leaving its actual index stale. I reproduced a failed write/read round trip with the PR code.

  Please either implement trailing-index writes consistently or explicitly reject them before modifying any files, including other shard-mutating operations. Add a regression test for     
  writing and updating an index_location="end" array. Read-only access and byte-for-byte remote cache mirroring can remain supported independently.
```

Issue #1660, opened 2026-08-31
render_ink.py: a mesh/volume frame mismatch renders a black strip and exits 0, and the published container has no working --flatten-keep


jonmarrs:
Following on from #1655. Trying to get from a fit to `total_fg_pixels` using only published artifacts, I hit two things. The first is a silent wrong answer rather than a failure, which is why I am reporting it even though the underlying cause may be my own configuration.

## 1. A mesh/volume frame mismatch renders a black strip and exits 0

`render_ink.py` assumes the fitted mesh coordinates are already in the ink volume's level-0 frame. It exposes `--scale` and `--group-idx`, which choose the pyramid level and pixels-per-voxel, but nothing that says the mesh is in a different frame.

When the assumption does not hold, the run completes normally. It writes a strip, exits 0, and the only trace is `p95=0.0` in one line of output:

```
rendering 1915x712 at scale 0.25 crop [1915x712 from (0,0)]
[1/1] wrote /work/meshes/ink/w010-019.jpg (1915px wide, p95=0.0)
Done. Strips in /work/meshes/ink
```

The strip is entirely black. Scored, that is `total_fg_pixels = 0` reported as a successful render.

In my case the mesh frame is 4x coarser than the volume. Sampling the published ink volume directly at the mesh's own vertices, 200 valid vertices from one `_spliced` winding:

| coordinate scale | in bounds | nonzero | mean | max |
|---|---:|---:|---:|---:|
| 1 (as rendered) | 100.0% | 0.0% | 0.00 | 0 |
| 2 | 100.0% | 66.5% | 12.04 | 250 |
| 4 | 100.0% | 90.0% | 19.64 | 250 |

Note that every scale is "in bounds", so a bounds check cannot catch this. Note also that scale 2 reads 66.5% nonzero: the scroll is dense enough that a wrong-but-close scale still lands on papyrus, so a nonzero-fraction heuristic would not be reliable either.

[...]

PR #1717, opened 2026-09-07
vc_render_tifxyz: report surface points that fall outside the volume

kadenpool:
**In one sentence:** `vc_render_tifxyz` now says how much of the surface fell outside the volume, instead of writing a partly blank image and reporting success.

**One real example:** I rendered a published surface against a published volume, both straight from the open-data catalogue. Segment `20250919184429` (PHerc0500P2) is shipped against four volumes. The catalogue's own `overlap_ratio` says the surface fits three of them completely and only a third of it fits the 0.55um scan. So the surface and the command stay the same and only the volume changes:

```
$ vc_render_tifxyz --volume .../20250528085330-4.317um-...zarr -s 20250919184429-on-20250528085330-4.317um.tifxyz \
    --scale 1 -g 3 --tif-output out
rendering 150x110 at scale 1 crop [150x110 from (0,0)]
$ echo $?
0
```

```
$ vc_render_tifxyz --volume .../20250821110041-0.550um-...zarr -s 20250919184429-on-20250821110041-0.55um.tifxyz \
    --scale 1 -g 3 --tif-output out
rendering 1177x867 at scale 1 crop [1177x867 from (0,0)]
$ echo $?
0
```

Neither run says anything is wrong, and both exit 0. The second image is 70% blank.

Same published surface, same command, two published volumes. Left: 4.317um volume, the surface fits and nothing is printed. Right: 0.55um volume, 70% of the image is fill value; before this PR the run exits 0 with no message, with it the run prints the warning below.

**Before:** samples that land outside the volume read nothing and their pixels keep the fill value. Nothing counts them, so the run cannot tell the difference between a surface that sits inside the volume and one that mostly does not. Measured from the surface's own coordinates against the volume shape: 57.3% of its grid points are valid, and of those only 52.8% address a voxel that exists in the 0.55um volume. 0.573 x 0.528 = 30.3% predicted content; the image has 29.9%. The blank area is exactly the part of the surface that is not there.

**After this PR:**

```
WARNING: 273835 of 580867 surface points (47.1%) fall outside the volume and were not rendered; those pixels are blank.
```

The 47.1% agrees with the independent coordinate measurement above (47.2% outside) to a tenth of a point. The control render against the 4.317um volume prints nothing, as it should.

**Proof:** the rendered pixels are byte-identical before and after, checked with `numpy.array_equal` on the two TIFFs. This changes what the run says, not what it produces.

**Why / where this is useful:** this is not a rare surface. On 8 Sep I checked the grid points of 833 published surfaces (every fourth point on each axis) against the shape of the volume each one names: **243 of the 833 (29%) have more than 0.1% of their points outside that volume**, across six scrolls, with a median of 77% of points inside and the worst at 29%. That check reads each surface's own x/y/z grids from the bucket and compares them with the volume's `.zarray` shape, so a point outside is a published coordinate that addresses no published data. The script and the run's output are in [scroll-lineup/surface_in_volume](https://github.com/kadenpool/scroll-lineup/tree/main/surface_in_volume), where `--count` re-counts the 243 from them.

The catalogue also carries its own figure, `overlap_ratio`, for 882 of its 903 transformed entries. On 21 Sep it flags 327 of those 882 (37%) as not wholly inside the volume they name. It is a different measure and the catalogue does not publish its definition (#1734 asks): on the example above it gives 0.34 where the surface's own points put 53% inside. So the two figures should not be compared, but both say the same thing about scale, and what was missing is the tool saying so at the moment someone renders one. The survey behind the catalogue figure is attached below.

The same surface through `--zarr-output`, which uses the tile renderer (the OMP-parallel path, hooked separately from the band renderer above): the identical figure, 273835 of 580867 (47.1%), a populated 3-level zarr, exit 0. The control stays silent there too. Verified 2026-09-08 on a from-source build; peak resident 112 MB.

## Details

- Counted in `renderBands` and `renderTiles`, where the sample positions are already computed in volume coordinates, so it costs one comparison per point and no extra reads. The prefetch planner walks the same points a second time and is deliberately not counted.
- `base` is `(x, y, z)` and `ds->shape(level)` is `(z, y, x)`; the comparison is written against that, and the control case is what shows it is not mirrored.
- Non-finite points are the surface's own holes and are excluded from both the numerator and the denominator, so the figure is "of the points the surface actually has, how many the volume lacks".
- **Reported once, at the end**, so it cannot be lost in the progress output.
- **Not changed:** the rendered pixels (byte-identical, above), the exit code of any render that has at least one point inside the volume, and the prefetch, caching and slice paths.
- **Related work:** #1463 (merged 1 Sep) adds a Python-side preflight that checks a tifxyz against its volume before rendering, and #1660 / #1728 (jonmarrs) report and fix the same kind of silent black output in spiral-fitting's `render_ink.py`. This PR is the C++ renderer's own count, so it needs no extra step and also catches a surface that partly leaves the volume after a preflight has passed.
- **Exit code changes only when nothing could be rendered at all.** A partly outside surface is still a useful render and still exits 0, with the warning. A surface with no point inside the volume produces an image with no data in it, and that now exits non-zero, in line with the other fail-loudly changes in this repo.
- **Six other open PRs touch this file** (checked 2026-09-20 with `git merge-tree` against each branch head; a seventh, #1695, has since merged). #1797, #1807, #1818 and #1831 merge cleanly with this one. #1761 conflicts with `main` itself, independently of this change. #1817 (fixing #1809) wraps the tile loop in a `try` block, and the only conflict is this PR's one added call, `countSamplesOutside(base, ds, level);` after `prepareBaseAndDirs`, so whichever lands second re-adds that line; I will do it here if #1817 lands first. Of the PRs listed here before, #1656 and #1657 were merged on 14 Sep and this branch includes both; #1681 was closed.

Verification:

```
$ cmake --build vc-build-apps --target vc_render_tifxyz     # Homebrew LLVM 23, macOS arm64
$ ./vc-build-apps/bin/vc_render_tifxyz ... 0.55um volume ...
WARNING: 273835 of 580867 surface points (47.1%) fall outside the volume and were not rendered; those pixels are blank.
$ ./vc-build-apps/bin/vc_render_tifxyz ... 4.317um volume ...
(no warning)
```

Built with Homebrew LLVM rather than Apple clang for the same `std::format` reason as `.github/workflows/vc3d-macos.yml`.

Re-checked 2026-09-11: branch brought up to date with `main`, the counter moved below the log-flusher globals it had split (same lines, no behaviour change), compiled again with the build's own flags, and CI re-run on the new head. Brought up to date with `main` again on 19 Sep (`b494a5ce6`); its build and test checks pass on that head.

I am happy to make the threshold for the non-zero exit configurable, or to drop the exit-code change entirely and keep only the warning, if you would rather.

---

The patch was written with Claude Code under my direction. The renders and measurements above were done with it on my own machine, and I reviewed the results before opening this PR.

*Edited 21 Sep 2026: the survey figure now carries its date and method, the catalogue's own figure is added beside it as a different measure, the sentence under the two example runs is corrected (they print different sizes), and the 19 Sep update is noted.*

<details>
<summary><code>survey_coverage.py</code>, the survey behind the catalogue's 327 of 882, run 21 Sep (standard library plus fsspec, anonymous read of the public catalogue)</summary>

```python
"""The catalogue's own figure for PR #1717's question: how many transformed surfaces its `overlap_ratio`
marks as not wholly inside the volume they name.

The PR's own figure, 243 of 833 (29 %), comes from a different and more direct measure: each surface's
grid points checked against the shape of the volume it names, run on 8 Sep with surface_in_volume.py (scroll-lineup, surface_in_volume/). This script is the second measure, not a re-run of that one.

MEASURE. The catalogue publishes, for most segments (882 of the 903 transformed entries here have one for
their volume), `properties.volume_coverage[<volume>]`, and
inside it `overlap_ratio`: the fraction of the segment that lies inside that volume. A ratio below
1 is the catalogue's own flag that the surface is not wholly inside. It is NOT a point fraction: on
#1717's worked example it is 0.34 where the surface's own points put 0.53 inside, and the catalogue
does not publish its definition (#1734 asks). Report it as the catalogue's flag, never as points.
`exp_1717/survey_outside.py` answers a neighbouring question, whether the declared bounding BOX
leaves the volume, by fetching each mesh's meta.json. The two do not have to agree and this script
prints both so the difference is visible rather than smoothed over.

PAIRING. Each `tifxyz-transformed` entry names one target volume. The coverage lives on the
enclosing segment node, so the walk carries the segment down with it instead of parsing ids out of
path strings.

WHAT IT IS PINNED TO. metadata.json is remote and mutable, so the output records its length and
sha256 along with the clock time of the run. A later reader can tell whether they are looking at
the same catalogue this figure was taken from.

  .venv/bin/python exp_1717/survey_coverage.py
"""
import gzip, hashlib, json, os, subprocess, sys, time

import fsspec

ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
BUCKET = "vesuvius-challenge-open-data"
OUT_DIR = os.path.join(ROOT, "exp_1717/coverage_runs")

def clock():
    """The run's timestamp, from the OS, with its source named."""
    stamp = subprocess.run(["date", "+%Y-%m-%d %H:%M:%S %Z"],
                           capture_output=True, text=True).stdout.strip()
    return stamp, "`date` on the machine that ran this"

def catalogue(fs):
    raw = fs.cat(f"{BUCKET}/metadata.json")
    if raw[:2] == b"\x1f\x8b":
        raw = gzip.decompress(raw)
    return json.loads(raw), len(raw), hashlib.sha256(raw).hexdigest()

def pairs(node, segment, out):
    """Every tifxyz-transformed entry, carried together with the segment node enclosing it.

    A node is treated as a segment when it has volume_coverage in its properties; that is the
    thing the coverage hangs off, so it is found by what it carries, not by where it sits.
    """
    if isinstance(node, dict):
        props = node.get("properties")
        if isinstance(props, dict) and isinstance(props.get("volume_coverage"), dict):
            segment = node
        if (node.get("format") or node.get("type")) == "tifxyz-transformed":
            out.append((node, segment))
        for v in node.values():
            pairs(v, segment, out)
    elif isinstance(node, list):
        for v in node:
            pairs(v, segment, out)

def pct(n, d):
    return 100.0 * n / d if d else 0.0

def median(xs):
    s = sorted(xs)
    if not s:
        return None
    m = len(s) // 2
    return s[m] if len(s) % 2 else 0.5 * (s[m - 1] + s[m])

def box_survey_rows():
    """The box test's finished run, if one is on disk, for the side-by-side."""
    p = os.path.join(ROOT, "exp_1717/survey_outside.json")
    try:
        d = json.load(open(p))
    except Exception:
        return None
    if "n_measured" not in d:          # a checkpoint from an interrupted run, not a result
        return None
    return d

def main():
    stamp, stamp_src = clock()
    t0 = time.time()
    fs = fsspec.filesystem("s3", anon=True)
    cat, nbytes, digest = catalogue(fs)
    print("catalogue: %d bytes, sha256 %s" % (nbytes, digest[:16]))

    found = []
    pairs(cat, None, found)
    print("tifxyz-transformed entries: %d" % len(found))

    rows, orphan = [], 0
    for entry, segment in found:
        origins = entry.get("origins") or []
        path = origins[0].get("path") if origins else None
        vid = (entry.get("parameters") or {}).get("target_volume")
        if not path or not vid:
            continue
        cov = None
        if segment is not None:
            cov = (segment["properties"]["volume_coverage"] or {}).get(vid)
        if not isinstance(cov, dict) or "overlap_ratio" not in cov:
            orphan += 1
            rows.append({"path": path, "volume": vid, "overlap_ratio": None})
            continue
        r = cov["overlap_ratio"]
        rows.append({"path": path, "volume": vid, "overlap_ratio": r,
                     "object": path.split("/")[0]})

    measured = [r for r in rows if r["overlap_ratio"] is not None]
    outside = [r for r in measured if r["overlap_ratio"] < 1.0]
    ratios = [r["overlap_ratio"] for r in outside]
    objects = sorted({r["object"] for r in outside})

    print()
    print("  entries whose segment publishes a coverage for the volume they name: %d" % len(measured))
    print("  of those, some of the surface is outside that volume: %d (%.1f %%)"
          % (len(outside), pct(len(outside), len(measured))))
    print("  no coverage published for that pair: %d" % orphan)
    if ratios:
        print("  among those that poke out: median %.1f %% inside, worst %.1f %% inside"
              % (100.0 * median(ratios), 100.0 * min(ratios)))
    print("  objects affected: %d (%s)" % (len(objects), ", ".join(objects)))

    box = box_survey_rows()
    if box:
        print()
        print("  cross-check, the box test in survey_outside.py: %d of %d (%.1f %%) declare a box "
              "that leaves the volume" % (box["n_outside"], box["n_measured"], box["pct"]))
        print("  the two measure different things; a box can fit while points do not, and a box "
              "test cannot see a surface that pokes out inside its own box")

    os.makedirs(OUT_DIR, exist_ok=True)
    # one output name per run, fixed when it starts, so a later run cannot overwrite this one
    tag = stamp.split()[0].replace("-", "") + "_" + stamp.split()[1].replace(":", "")[:4]
    out_path = os.path.join(OUT_DIR, "coverage_%s.json" % tag)
    result = {
        "surveyed": stamp,
        "timestamp_source": stamp_src,
        "elapsed_s": round(time.time() - t0, 1),
        "catalogue_bytes": nbytes,
        "catalogue_sha256": digest,
        "measure": ("properties.volume_coverage[target_volume].overlap_ratio < 1 for every "
                    "tifxyz-transformed entry: the catalogue's own flag, not a point count (#1734)."),
        "n_entries": len(rows),
        "n_measured": len(measured),
        "n_outside": len(outside),
        "pct": round(pct(len(outside), len(measured)), 1),
        "n_no_coverage": orphan,
        "median_inside_pct": round(100.0 * median(ratios), 1) if ratios else None,
        "worst_inside_pct": round(100.0 * min(ratios), 1) if ratios else None,
        "objects": objects,
        "rows": rows,
    }
    json.dump(result, open(out_path, "w"), indent=1)
    latest = os.path.join(ROOT, "exp_1717/coverage_latest.json")
    slim = {k: v for k, v in result.items() if k != "rows"}
    slim["run_file"] = os.path.relpath(out_path, ROOT)
    json.dump(slim, open(latest, "w"), indent=1)
    print()
    print("  wrote %s" % os.path.relpath(out_path, ROOT))
    print("  wrote exp_1717/coverage_latest.json (summary, points at the run above)")
    return 0

if __name__ == "__main__":
    sys.exit(main())
```

</details>

*Rebuilt 21 Sep 2026 as one commit on current `main`, with no merges. The `b494a5ce6` above was a merge of `main`, and it carried #1829's commit into this branch's history. The change itself is byte-identical.*


kadenpool:
Moving this to draft myself, for one reason and not as an abandonment.

The number that motivates it, "243 of 833 (29 %) have points outside the volume they name", was surveyed when this PR was opened and it is the one figure here I never dated. The catalogue has since grown to 903 such entries, so the proportion needs re-measuring before I ask anyone to weigh it. I have dated the sentence in the body rather than quietly leaving it.

The change itself stands on its own: it is +54 and adds no behaviour beyond a warning when a surface's points fall outside the volume being rendered, plus a non-zero exit when no point of the surface is inside it. I will re-run the survey, put a dated figure back, and mark it ready.

This also frees one of my three PR slots for #1846's patch, which is measured on two platforms and which @Bullo27 has asked me to open.

Issue #1716, opened 2026-09-07
First Letters spiral recipe has fitted with no winding-placing input since #1651


ttendoscopie-creator:
**In one sentence:** the `FIT_SPIRAL_CONFIG_OVERRIDES` block published in *From CT Scan to Ancient Text: A First Letters Prize Workflow* (18 August) does not set `input_use_tracks`, and #1651 flipped that default to `False` on 30 August, so following the post on a Grand-Prize-eligible scroll now fits with nothing that can place a point on a winding.

**I was trying to:** fit a spiral on PHerc0826 by following the published First Letters workflow, to see what the pipeline gives on an eligible scroll.

**Using:** `main` at `23adee047`, rebased and re-checked on `b218ace8`. PHerc0826, volume `20250821151701`, the published `surface-m7-L0-th0.2` tracks from `dl.ash2txt.org/datasets/spiral_datasets/`, and the umbilicus published under `representations/umbilicus/` on 3 September. Band z ∈ [11000, 12000), 30,000 iterations, seed 1, one RTX 3090, about 30 minutes per run.

```
FIT_SPIRAL_CONFIG_OVERRIDES='{"z_begin":11000,"z_end":12000,"input_disable_patches":true,
 "loss_weight_shell_outer":0,"loss_weight_shell_patch_radius":0,
 "dense_spacing_mode":"grad_mag","loss_weight_dense_spacing":0}' \
  python fit_spiral.py --dataset <root>
```

**What happened:** the tracks the dataset declares were never loaded and never mentioned, and the fit ran to completion anyway.

```
skipping all verified/unverified patch loading
pcls: 0 cross-patch, 0 unattached
theta topology ready (0.0s, peak RSS 1.02 GiB)
step 0: loss = 2.2, umbilicus = 1.6, sym_dirichlet = 0.6, min_spacing = 0.0
```

There is no `track_radius` term. The post says the fit's geometric evidence is "umbilicus coordinates, extracted tracks, and X/Y surface-normal volumes"; two of those three are gone. Tracks are off by the new default, and normals give orientation rather than winding — two adjacent windings have parallel normals — so nothing is left that can say which winding a point is on. None of the 13 eligible scrolls ships verified patches.

**What I expected or needed:** either the recipe sets `input_use_tracks`, or the post says that tracks are no longer part of the evidence. Right now it names them and does not use them.

**Evidence / reproduction:** the same scroll, same band, same seed, same published umbilicus, one config key apart.

[...]

PR #1719, opened 2026-09-07
docs(spiral tutorial): note that input_use_tracks defaults to false

ttendoscopie-creator:
**In one sentence:** the tutorial's Configure section lists the per-input paths but not the `input_use_*` toggles, so a reader who has a tracks DBM and points the fitter at it has no way to learn that `input_use_tracks` has defaulted to `false` since #1651.

**One real example:** following the First Letters workflow post on PHerc0826 — published tracks, published umbilicus, the post's overrides verbatim — the tracks are never loaded and never mentioned, and the fit runs its 30,000 iterations anyway. The tutorial is where a reader would go to find out why, and it does not say.

**Before:** the section says the per-input paths "default to locations inside it. Set any of them to `None` to fit without that input." Nothing about the toggles that decide whether a path that *is* set gets used.

**After this PR:** one bullet, before the `z_begin, z_end` one, naming the toggles, giving the override, and saying why it matters on scrolls whose only published supervision is tracks:

> `input_use_tracks` has defaulted to `false` since #1651, so a dataset that ships a tracks DBM will not use it unless you say so […] with tracks off and no verified patches, nothing left in the fit places a point on a winding — normals give orientation, and two adjacent windings have parallel normals. The fit still converges and writes a normal-looking checkpoint.

**Proof:** measured on PHerc0826, volume `20250821151701`, z ∈ [11000, 12000), 30,000 iterations, seed 1, the umbilicus published on 3 September, one config key apart:

| | recipe as published | + `"input_use_tracks": true` |
|---|---:|---:|
| tracks loaded | 0 | 480,117 |
| `dr_per_winding` | 16.8663 vx | 14.8851 vx |
| satisfied track points | 20.9% | 41.6% |

Full write-up and logs in #1716. #1713 makes `fit_spiral` say this at runtime; this PR is the documentation half, and is the cheaper of the two fixes that issue proposes.

**Why / where this is useful:** anyone reading the tutorial to set up a fit on one of the 13 eligible scrolls, nine of which publish tracks and nothing else, and none of which publishes verified patches.

### DIFF
--- a/scrollprize.org/docs/38_tutorial_spiral.md
+++ b/scrollprize.org/docs/38_tutorial_spiral.md
@@ -121,6 +121,14 @@ re-running rclone resumes interrupted downloads. The dataset contains verified a
 Configuration is straightforward: the input paths and fitting region are plain variables at the top of `fit_spiral.py`. Edit them to point at your download:
 
 - `dataset_path` — the root of the dataset; the per-input paths below it (`verified_patches_path`, `unverified_patches_path`, `pcl_json_paths`, `fibers_path`, `shell_path`, `tracks_dbm_path`, the normals/grad-mag zarr paths, …) default to locations inside it. Set any of them to `None` to fit without that input.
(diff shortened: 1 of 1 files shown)


pmh47:
If this just added the text about enable inputs flags then it'd be fine to merge -- without the issue reference etc which is not appropriate for a tutorial, and without being overly specific to the prize scorlls

PR #1725, opened 2026-09-07
vesuvius: make the documented segment call work without scroll_id

mithilp:
I thought it would be smart to take a peek at Scroll 1, and specifically the ink on it. I copied the segment example from the `Volume` docstring and it wasn't running on my computer. The error said it couldn't determine the energy or resolution for "scroll None", and that didn't really help me much. I took a look at the init code and figured out that segments need a scroll id, even though the documented call doesn't pass one. The config already knows which scroll a segment belongs to, so I made it look that up. I didn't want everyone else to run into the same thing.

---

**In one sentence:** `Volume(type="segment", segment_id=...)` now works without also passing `scroll_id`, which is how it is written in the docstring.

**One real example:** Starting with the segment named in the `Volume` docstring, I ran `Volume(type="segment", segment_id=20230827161847)` to look at ink on Scroll 1, and it failed before reaching the data.

**Before:**
```
>>> Volume(type="segment", segment_id=20230827161847)
ValueError: Could not determine energy/resolution for scroll None.
            Please provide them explicitly.
```
Energy and resolution are looked up by scroll id, but `type="segment"` leaves `scroll_id` as whatever was passed, i.e. `None`. The undocumented requirement was to also pass `scroll_id=1`. The same call appears in the class docstring and in the library's own error message, so the first thing a new user copies does not run.

**After this PR:**
```
>>> seg = Volume(type="segment", segment_id=20230827161847)
>>> seg.scroll_id, seg.energy, seg.resolution
('1', 54, 7.91)
>>> seg.shape(0)
(65, 9163, 5048)
```

**Proof:** the shipped `scrolls.yaml` already records which scroll each segment belongs to, so nothing new is fetched. Explicit values still win:

```
Volume(type="segment", segment_id=20230827161847)                    -> scroll 1, 54, 7.91   (was: ValueError)
Volume(type="segment", segment_id=20230827161847, scroll_id=1)       -> scroll 1             (unchanged)
Volume(type="segment", segment_id=..., energy=54, resolution=7.91)   -> 54, 7.91             (unchanged)
Volume(type="segment", segment_id=99999999999999)                    -> ValueError: Segment 99999999999999 was not
                                                                        found in .../scrolls.yaml. Pass scroll_id
                                                                        explicitly if it lives elsewhere.
```

**Why / where this is useful:** it is the first call in the docs, so it is the first thing that fails for anyone starting with segments. The error it produced named `scroll None`, which does not point at the missing argument.

## Details

Tested against `d8c5f48`.

Adds `Volume._infer_scroll_from_segment`, which scans the already-loaded config for the first entry whose `segments` map contains the id and returns `(scroll_id, energy, resolution)`. Called from `__init__` only when `type == "segment"` and `scroll_id is None`, before energy and resolution are resolved. No new dependency and no network access; the config is the same file `get_url_from_yaml` already reads.

Tests in `vesuvius/tests/data/test_volume_segment_scroll_inference.py` cover inference for two scrolls with different energy and resolution, an unknown segment, a missing config file, and a check that the segment named in the docstring resolves from the shipped config. All five fail on main.

`pytest vesuvius/tests/data/` passes (33 passed, 5 skipped) with a `volume-only` install.

### DIFF
--- a/vesuvius/src/vesuvius/data/volume.py
+++ b/vesuvius/src/vesuvius/data/volume.py
@@ -335,6 +335,13 @@ class Volume:
             # energy/resolution are keyed by scroll, so without this the lookup
             # fails on scroll_id=None.
             if self.type == "segment" and self.scroll_id is None:
+                if not self.configs or not os.path.exists(self.configs):
+                    # Distinguish "no config to look in" from "segment not listed",
+                    # and keep the error type the same as get_url_from_yaml's.
+                    raise FileNotFoundError(
+                        f"Configuration file not found at {self.configs}. It is needed to "
+                        f"determine which scroll segment {self.segment_id} belongs to. "
+                        f"Pass scroll_id explicitly to skip this lookup.")
                 inferred = self._infer_scroll_from_segment()
                 if inferred is None:
                     raise ValueError(
@@ -563,15 +570,15 @@ class Volume:
         """Find which scroll a segment belongs to by scanning the config.
 
         Returns (scroll_id, energy, resolution) for the first entry whose
-        ``segments`` map contains ``self.segment_id``, or None if absent.
+        ``segments`` map contains ``self.segment_id``. Returns None if the
+        config does not exist, or if it exists but does not list the segment.
+        A config that exists but cannot be read or parsed raises, rather than
+        being reported later as a missing segment.
         """
         if not self.configs or not os.path.exists(self.configs):
             return None
-        try:
-            with open(self.configs, 'r') as file:
-                config_data: Dict = yaml.safe_load(file) or {}
-        except Exception:
-            return None
+        with open(self.configs, 'r') as file:
+            config_data: Dict = yaml.safe_load(file) or {}
 
         target = str(self.segment_id)
         for scroll_id, energies in config_data.items():
--- a/vesuvius/src/vesuvius/data/volume.py
+++ b/vesuvius/src/vesuvius/data/volume.py
@@ -340,8 +340,9 @@ class Volume:
                     # and keep the error type the same as get_url_from_yaml's.
                     raise FileNotFoundError(
                         f"Configuration file not found at {self.configs}. It is needed to "
-                        f"determine which scroll segment {self.segment_id} belongs to. "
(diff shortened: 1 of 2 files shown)


mithilp:
Both error-handling points were right, fixed in the latest commit.

The helper no longer catches everything. It returns None only when the config file is absent, and a config that exists but will not parse now raises instead of being reported later as a missing segment. Init also distinguishes the two cases: no config at all raises FileNotFoundError with the same message shape `get_url_from_yaml` already uses, so the error type does not depend on which code path happened to notice first. Added a test that malformed YAML propagates.

On caching the parsed config: leaving it as is. `scrolls.yaml` is 39 lines and it is read once per `Volume` construction, which is not a hot path, so a cache would add invalidation questions for no measurable gain. Happy to add one if you would rather have it.

mithilp:
Both points addressed.

You were right that the missing-config message gave advice that would not have worked. The config is read a second time in `get_url_from_yaml` to resolve the segment URL, so passing `scroll_id` would only have moved the failure a few lines down. It now says to restore `scrolls.yaml` or open the data directly with `Volume(type="zarr", path=...)`.

Added two init-level tests so the documented call is covered directly rather than only through the helper. They stub the network-dependent steps and let everything before them run for real, including reading the shipped config: one asserts `Volume(type="segment", segment_id=20230827161847)` resolves scroll, energy, resolution and URL, the other asserts an explicit `scroll_id` is still not overridden. Against `origin/main` seven of the eight tests in the file fail; the eighth is the explicit-`scroll_id` guard, which passes both before and after by design.

mithilp:
Fixed. `yaml.safe_load(...) or {}` turned an empty or null config into an empty mapping, so the lookup fell through and reported the segment as missing. `get_url_from_yaml` already raises "is empty or invalid YAML" for exactly this case, so the helper now raises the same thing rather than inventing a second description of the same problem. Covered for empty, whitespace-only and explicit-null configs.

That is everything the automated reviewer raised across three passes. I am going to stop pushing on its suggestions here so the diff stops moving. The one I did not take is caching the parsed config, for the reason above. Happy to pick anything back up if a maintainer wants it.

PR #1726, opened 2026-09-07
vesuvius: normalize 2D and 1D reads over the whole slice

mithilp:
I was looking at the ink signal layer by layer, because my computer isn't strong enough to load whole slabs, and I used `instance_zscore` because that's what the docs said to use. When I cross-checked one of those layers against the same voxels read as a 3D slab, the numbers didn't match. It just gave me different values, with no error. If I hadn't compared the two read shapes I would have kept going. Hopefully this helps other people avoid it.

---

**In one sentence:** Reading a single slice with `instance_zscore` or `instance_minmax` now normalizes the whole slice, instead of silently normalizing each row against itself.

**One real example:** Starting with Scroll 1 at `z=4000, y=3000:3384, x=3000:3384` read through `Volume(type="scroll", scroll_id=1, normalization_scheme="instance_zscore")`, I compared the 2D read `vol[z, y0:y1, x0:x1]` against the 3D read `vol[z:z+1, y0:y1, x0:x1]` of the same voxels, and they disagreed by up to 1.2763.

**Before:** The temporary channel axis was only inserted when a read came back with exactly 3 dimensions. A 2D read reached the normalization loop with no channel axis, so `for c in range(data_slice.shape[0])` walked the first spatial axis and normalized each row against its own mean and standard deviation. All 384 rows in the example come back with a mean of ~1e-7. A 1D read was worse: every voxel was normalized against itself and the read returned all zeros. Neither case raised or warned.

**After this PR:** The axis is inserted for any read that has no channel dimension, so 1D, 2D and 3D reads all normalize over the full slice and agree with each other. Reads that do have a channel dimension (ndim > 3) are untouched and keep per-channel statistics.

**Proof:** Segment 20230827161847, surface layer 32, the 512x512 window with the most ink in the segment (y=2560, x=2560). Same window and settings in every panel. Left to right: CT, the ink label, what `instance_zscore` returns today for the 2D read, the same read after this PR, and the absolute difference. The horizontal banding is the row-by-row normalization.

ink window on segment 20230827161847

The same comparison on raw Scroll 1 at z=4000, away from any segment:

normalization comparison on Scroll 1

```
segment 20230827161847, layer 32, 512x512 ink window
                                   before      after
2D read == whole-slice zscore      False       True
2D read == 3D read, same voxels    False       True
max |2D - 3D|                      1.7568      0.0
rows with |mean| < 1e-6  (of 512)  512         0

scroll 1, z=4000, 384x384
max |2D - 3D|                      1.2763      0.0
rows with |mean| < 1e-6  (of 384)  384         0
1D read all zeros                  True        False
```

**Why / where this is useful:** Anyone reading 2D slices with instance normalization has been feeding row-normalized input to models and to visual inspection without any indication. Slice-wise reads are a normal way to look at ink and surface data, and this makes them agree with the 3D path.

## Details

Tested against `d8c5f48`.

`vesuvius/src/vesuvius/data/volume.py`, in `__getitem__`: insert the pseudo-channel axis whenever `has_channel_dim` is false rather than only when `original_ndim == 3`, and key its removal off `data_slice.ndim == original_ndim + 1` rather than a hardcoded 3 to 4 transition.

Also in `load_data`: the `verbose` branch printed `data.shape` unconditionally, but a multiscale scroll opens as a `zarr.Group`, which has no `.shape`. `Volume(type="scroll", scroll_id=1, verbose=True)` raised `AttributeError` for every scroll. It now prints the level keys for a group and the shape for an array.

Tests added in `vesuvius/tests/data/test_volume_normalization_ndim.py`: whole-slice correctness for 2D reads under both instance schemes, 2D/3D agreement, the 1D collapse, and a guard that channel-first reads still normalize per channel. Five of the six fail on main; the sixth is the channel guard, which passes both before and after by design.

`pytest vesuvius/tests/data/` passes (34 passed, 5 skipped) with a `volume-only` install; `test_affine.py`, `test_vc_dataset_bbox.py` and `test_zarr_chunk_index.py` fail to collect for want of scipy/torch/tqdm, which is unrelated to this change.

### DIFF
--- a/vesuvius/src/vesuvius/data/volume.py
+++ b/vesuvius/src/vesuvius/data/volume.py
@@ -983,7 +983,12 @@ class Volume:
         # Add temporary channel dim for 3D data (Z, Y, X) -> (1, Z, Y, X) for consistent logic
         original_ndim = data_slice.ndim
         has_channel_dim = original_ndim > 3  # Heuristic: assume >3D means channels exist at dim 0
-        if not has_channel_dim:  # Add channel dim for any read without one (1D/2D/3D)
+        # Only the instance schemes iterate over channels. Give them a single
+        # pseudo-channel when the read has no channel axis of its own, so a 1D
+        # or 2D read is normalized as one unit rather than per spatial row.
+        needs_channel_axis = not has_channel_dim and self.normalization_scheme in (
+            'instance_zscore', 'instance_minmax')
+        if needs_channel_axis:
             data_slice = data_slice[np.newaxis, ...]
             if self.verbose: print(f"  Added temporary channel dim for normalization: {data_slice.shape}")
 
@@ -1039,7 +1044,7 @@ class Volume:
             raise ValueError(f"Internal Error: Unknown normalization scheme '{self.normalization_scheme}' encountered.")
 
         # Remove temporary channel dimension if it was added
-        if not has_channel_dim and data_slice.ndim == original_ndim + 1:
+        if needs_channel_axis and data_slice.ndim == original_ndim + 1:
             data_slice = data_slice[0, ...]
             if self.verbose: print(f"  Removed temporary channel dim: {data_slice.shape}")
 
--- a/vesuvius/src/vesuvius/data/volume.py
+++ b/vesuvius/src/vesuvius/data/volume.py
@@ -735,7 +735,10 @@ class Volume:
             
             if self.verbose:
                 print(f"Successfully opened zarr store: {data}")
-                print(f"Shape: {data.shape}, Dtype: {data.dtype}")
+                if isinstance(data, zarr.Array):
+                    print(f"Shape: {data.shape}, Dtype: {data.dtype}")
+                else:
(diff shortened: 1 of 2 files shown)


mithilp:
Both points addressed in the latest commit.

The pseudo-channel axis is now only inserted for the schemes that actually index `data_slice[c]`, so `none`, `global_zscore` and `ct` no longer pay for an axis they never use, and the removal is keyed off the same condition. The adjacent comment is rewritten to say what it now does. Added a test asserting the non-iterating schemes are unaffected.

The 1D regression test is parameterized over both instance schemes. Worth noting that `instance_minmax` did not collapse to zeros the way `instance_zscore` did, it collapsed to a constant, so the assertion is now on uniqueness rather than on being non-zero.

Re-verified on segment 20230827161847, surface layer 32, the 512x512 window at y=2560 x=2560: 2D and 3D reads agree exactly, and none of the 512 rows carry the row-normalized signature any more.

Issue #1674, opened 2026-09-01
Local zarr volumes render as fill-value: chunk lookup uses wrong indices (silent failure)


bogdanexpl:
Local zarr volume reads return fill-value; chunk lookups use wrong indices

**Summary.** Rendering from a *local* OME-Zarr copy (`-v /path/vol.zarr`, no
`--remote-url`) silently produces all-zero images: the reader never opens any
chunk file. `strace` shows chunk existence checks against index triples that
match neither (z,y,x) nor (x,y,z) of the volume. The same segment rendered from
the same volume via `--remote-url` streaming reads the correct chunks.

**Setup.** Local sparse copy of
`PHerc1447/volumes/20250521151220-8.640um-1.2m-116keV-masked.zarr` (level 0
only, zarr v2, uint8, 128³ chunks, `dimension_separator: "/"`, shape
(24297, 8343, 8343)); segment bbox chunks present, verified readable in Python
(zarr-python reads correct data).

**Evidence.** Segment bbox spans chunk indices z 172–193, y 20–36, x 34–45.
Rendering it locally:

```
newfstatat(... "/vol.zarr/0/34.41.21") = ENOENT
```

— all probed triples fall in ranges dim1 28–56, dim2 21–43, dim3 15–34 (~4–5×
too small in the z dimension to be the correct chunk grid). No chunk file is
ever opened; output = fill value everywhere. Renaming chunks to `.`-separated
form and flipping `dimension_separator` changes nothing (both layouts present →
still zero). The `--remote-url` path renders the identical segment correctly,
so the segment geometry and volume data are fine.

Related but distinct: #1343 (all-black render at an *absent* pyramid level,
closed) — here level 0 exists and is readable; the probed chunk keys simply
match no level's grid.

Suspect: axis order / scale mapping in the local `FileSystemChunkSource` (or
`utils::ZarrArray`) chunk-key computation, possibly missing the OME multiscale
coordinate transform that the HTTP path applies.

---

*This and the `--zarr-output` segfault (filed separately) bite the exact
"alternative unwrapping pipelines integrated with VC3D" workflow the team encourages: streaming render → zarr surface volume →
ink inference. Workaround used: `--remote-url` + `--tif-output`, then repack
TIFFs to zarr out-of-band.*

Bullo27:
**Measured on current `main` (5479453a7): the local chunk-key computation is correct — the probed indices are where this segment's data actually lives.**

[...]

PR #1729, opened 2026-09-08
Fix file-log cleanup on renderer early returns

jauntinoanderson-lang:
**In one sentence:** Preserve renderer exit codes and file diagnostics on early returns, and wake the log flusher immediately during shutdown.

**Real example / motivation:** During the previously documented Codex-assisted PHerc1447 rendering experiment, missing local pyramid metadata returned a useful error without `--log-path`, but enabling it caused a shutdown abort and an empty log.

**Before / after:** The scope guard preserves the normal failure code and diagnostics, including successful multipart merges. Following the review of the five-second shutdown stall, the flusher now uses a condition-variable wait. Stop changes the predicate under the same mutex, notifies the thread, joins it, and flushes/closes the file. The periodic flush interval remains five seconds.

**Latest validation (Linux, GCC Release):**
- Missing-volume error, three runs per binary with identical flags/dependencies: before min/median/max **5.138 / 5.208 / 5.242 s**; after **0.252 / 0.292 / 0.339 s**. Every completed run returned 1 and retained the diagnostic.
- Missing-segmentation timing is scheduling-dependent: the prior binary exceeded three seconds in the regression run but exited quickly in the separate three-run benchmark. No claim that every old exit stalls.
- The tightened three-second CLI timeout fails two cases on the prior binary; all four cases pass on the update through CTest (1.70 s for the test).
- Re-ran the existing PHerc1447 fixture: valid render and multipart merge succeed; logged errors retain exit 1 and the diagnostic; logged/unlogged render and merge pixels match. No rendering mathematics changed.
- Published head `c8b7ebdb0922dfeb9edbf26696f45588c50eb676` has exactly the tested local Git tree `2d36d4ce4b9877fe1aa0c5ae874d8f41374b6e30`.

**Reproduce after configuring the existing build:** 
```sh
cmake --build build --target vc_render_tifxyz --parallel 2
python3 core/test/test_render_tifxyz_logging.py build/bin/vc_render_tifxyz
```
Run from `volume-cartographer`, substituting your configured build directory. The CLI CI job already invokes this script; the GitHub workflows currently require maintainer approval.

**Why useful:** Rendering and multipart jobs need reliable exit codes and logs, without an unnecessary five-second shutdown wait that delays diagnosis or retries.

**Original real-data evidence:** [renderer-log-cleanup-public-evidence.zip](https://github.com/user-attachments/files/31934093/renderer-log-cleanup-public-evidence.zip). This archive documents the original scope-guard validation; it predates the wakeup follow-up. Fixture: unchanged 5×5 XYZ crop from PHerc1447 segment `20250702235910-auto_grown_20250702235910292`, matching `20250521151220-8.640um-1.2m-116keV-masked.zarr`, original chunk `0/89/24/33`. Provenance is in the archive.

**Review and assistance:** Juantino Anderson reviewed the original published change/results and authorized submission, then authorized addressing the shutdown-delay feedback. Codex implemented and tested this follow-up; no claim is made that the contributor personally reran these tests or separately reviewed the new code. The original relevance explanation was adopted by the contributor.

**Limits:** Latest validation is Linux only; no rebuilt Windows/macOS binary, full-suite pass, or newly recovered text is claimed. Watchdog `_exit` and uncaught termination remain outside scope. This does not address indexing report #1674.


Bullo27:
**Verdict: reproduces on current main on Linux, and the guard fixes it here.**

Measured with `vc_render_tifxyz -v absent.zarr -s absent.tifxyz --scale 1 -g 0 --tif-output out` (gcc build; renderer source identical to main tip):

| | rc | stderr | log file |
|---|---:|---|---|
| main, no `--log-path` | 1 | `Error opening local zarr: …` | – |
| main, `--log-path` | 134 | `terminate called without an active exception` | 0 bytes |
| this PR's guard, `--log-path` | 1 | – | 232 bytes, the diagnostic |

Root cause matches the description: after the log setup, `main` has 28 `return` statements and only the last one calls `stopLogFlusher()`, so a joinable `g_logFlushThread` is destroyed at exit. The merge path is the ugly case: `return mergeTiffParts(...) ? EXIT_SUCCESS : EXIT_FAILURE` (L1227) turns a *successful* logged merge into an abort. `stopLogFlusher()` is idempotent (joinable check, null file), so the explicit call #1717 adds before its new early return stays harmless with this guard. `test_render_tifxyz_logging.py`: 4/4 OK against the patched binary.

jauntinoanderson-lang:
Addressed the CI coverage gap in d4ba027: the existing CLI build job now runs `python3 core/test/test_render_tifxyz_logging.py build/ci-fast-cli/bin/vc_render_tifxyz` after compiling the CLI targets, with a two-minute step timeout. The CLI path filter already includes `core/**`, `apps/src/**`, and this workflow. Re-ran the four cases against the patched Linux renderer locally: 4/4 passed (6.927s). GitHub CI still needs maintainer approval to run. Thanks to Bullo27 for independently reproducing and verifying the fix.

hendrikschilling:
```
  - [P2] Logged exits can stall for five seconds at volume-cartographer/apps/src/vc_render_tifxyz.cpp:73 and :82. The guard joins the flush thread, but that thread sleeps for five seconds 
    and cannot be awakened. Locally, logged missing-segmentation and missing-volume failures each took 5.01 s to exit. Use a condition variable or interruptible wait so stopLogFlusher()   
    wakes it immediately.
```

Issue #1618, opened 2026-08-26
PHercParis4: 28 published w-series segments ship bboxes whose lower corner is the tifxyz -1 missing-point marker (true minima ~[2200+, 2200+, 3000])


bkalita-git:
**In one sentence:** 28 published PHercParis4 winding segments (the `w###` series dated 2026-07-01) ship `meta.json` bounding boxes whose lower corner is the tifxyz **missing-point marker `-1`** instead of the true data minimum (which sits near `[2200+, 2200+, ~3000]`), so anything windowing by bbox reads ~3,000 phantom z-slices per segment.

**I was trying to:** Extend a corpus-wide integrity audit of the open-data catalog to segments — validating each segment's `bbox`, `scale`, grid dimensions, and volume references — as groundwork for surface/ink work on the eligible scrolls.

**Using:** `metadata.json` fetched from `s3://vesuvius-challenge-open-data` on 2026-08-27, and the segments' own published `tifxyz` grids (`x.tif`/`y.tif`/`z.tif` + `meta.json`) under `PHercParis4/segments/<id>-w*/mesh/intermediate/tifxyz_original/`. The defect is in the published `meta.json` files themselves, not only the catalog copy.

**What happened:** For 28 of the 29 `w###` segments, the stored bbox lower corner contains `-1` on at least one axis (most are exactly `[-1, -1, -1]`). tifxyz marks missing grid points with `-1`, so a bbox computed over the raw grid inherits the marker as its minimum. Recomputing the bbox **over valid points only** (`!= -1`) from the published grids gives the true extents:

<details><summary>All 28 segments: stored vs true lower corner (recomputed from the published grids)</summary>

| segment | stored bbox lo | true bbox lo (valid points) |
|---|---|---|
| PHercParis4/20260701183124 | [-1.0, -1.0, -1.0] | [2824.15, 2809.43, 2980.76] |
| PHercParis4/20260701183125 | [-1.0, -1.0, -1.0] | [2669.52, 2623.16, 2982.44] |
| PHercParis4/20260701183126 | [-1.0, -1.0, -1.0] | [2537.74, 2439.66, 2992.54] |
| PHercParis4/20260701183127 | [-1.0, -1.0, -1.0] | [2376.99, 2306.72, 2993.93] |
| PHercParis4/20260701183128 | [-1.0, -1.0, -1.0] | [2213.3, 2195.4, 2980.4] |
[...]

PR #1731, opened 2026-09-08
tifxyz: recompute the bbox from valid points when the stored one carries the -1 marker (refs #1618)

kadenpool:
**In one sentence:** a surface whose published `meta.json` bbox carries the `-1` missing-point marker as a minimum no longer reports that marker as its extent, so `list_tifxyz(z_range=...)` and `neural_tracing`'s z filter stop returning it for ranges its coordinates never touch.

**One real example:** macOS arm64, Python 3.14, villa at `main`. I pointed `list_tifxyz` at a folder holding one published surface, `PHercParis4/segments/20260701183124-w010-027/mesh/intermediate/tifxyz_original`, plus a control with an ordinary bbox. That surface's `meta.json` says its bbox z runs from **-1** to 18021. Its `z.tif`, over the points that are not the marker, runs from **2981** to 18021. So:

```
>>> [s.path.name for s in list_tifxyz(folder, z_range=(0, 100))]
['bad_bbox']
>>> [s.path.name for s in list_tifxyz(folder, z_range=(500, 1000))]
['bad_bbox']
>>> [s.path.name for s in list_tifxyz(folder, z_range=(2000, 2900))]
['bad_bbox']
>>> next(s for s in list_tifxyz(folder) if s.path.name == "bad_bbox").z_min
-1.0
```

Three ranges the surface does not reach, and it is returned for all of them. The filter is `bbox_z_min > range_z_max or bbox_z_max < range_z_min`, and with `bbox_z_min = -1` the first clause can never be true.

**Before:** `list_tifxyz` builds `TifxyzInfo.bbox` straight from `meta.json`. #1618 documents 28 published PHercParis4 surfaces whose stored bbox lower corner is the `-1` marker (most exactly `[-1, -1, -1]`), because it was computed over the raw grid rather than the valid points. Every one of them therefore passes every `z_range` whose upper bound is above -1, and reports `z_min = -1.0` to anyone who asks.

**After this PR:** when any bbox component is exactly `-1`, the extent is recomputed over the valid points, in both places the bbox is built: `list_tifxyz` (which reads the three grids to do it) and `TifxyzReader.read` (where the grids are already loaded, so it costs nothing). A warning names the directory:

```
WARNING vesuvius.tifxyz.reader: .../bad_bbox: meta.json bbox contains the -1 missing-point marker; using the extent of the valid coordinates instead
>>> [s.path.name for s in list_tifxyz(folder, z_range=(0, 100))]
[]
>>> [s.path.name for s in list_tifxyz(folder, z_range=(3000, 4000))]
['bad_bbox']
>>> next(s for s in list_tifxyz(folder) if s.path.name == "bad_bbox").z_min
2980.8
```

The range it genuinely overlaps still returns it. The control is untouched: `read_tifxyz` on it returns the same six bbox values before and after, to three decimals, because an honest stored bbox is trusted exactly as before.

The same file through `read_tifxyz`, then `neural_tracing.datasets.common._segment_z_bounds`, which prefers `seg.bbox` whenever it is not `None`:

```
before: (-1.0, 18020.88)
after:  (2980.76, 18020.88)
```

That function already has the right fallback in its `else` branch, over the valid coordinates; a `-1` bbox is simply not `None`, so it never got there.

**Proof:** three new tests. A 4x6 grid whose valid z runs 3000..4000 with a stored bbox floor of -1 is excluded from (0, 100) and (500, 1000), included in (3000, 4000), and reports `z_min` 3000; the `Tifxyz` object from `read_tifxyz` carries bbox z 3000..4000 and `_segment_z_bounds` agrees; and an all-marker grid yields no bbox and is not filtered, as before. Alongside the existing tifxyz and tifxyz_label_transfer suites: 141 passed, 5 skipped.

**Why / where this is useful:** `list_tifxyz` is the documented way to pick segments by z before loading them, and 28 published surfaces answer that question wrongly today. A pipeline selecting "everything between z 0 and 1000" on PHercParis4 gets those 28 back with nothing in that range.

## Details

- **The trigger is the marker, not "negative".** A `tifxyz-transformed` surface can legitimately have negative coordinates: it is expressed in another volume's frame and may extend outside it (the same thing #1717 reports at render time). So the check is `component == -1` exactly, matching how the grids mark a missing point.
- **Cost of the fallback.** The published coordinate grids are uncompressed, untiled float32, so `read_coordinate` is `tifffile.memmap` in copy-on-write mode: opening the 5 MB `z.tif` above took 0.01 s and the min/max over valid points 0.06 s, and nothing was written to disk. The fallback reads x, y and z so the recomputed bbox is a complete six-tuple, not just z.
- **All-marker grids** return `bbox=None`, which `list_tifxyz` already treats as "cannot filter, include", the existing behaviour for a missing bbox.
- **A read-then-write of a bad file now emits a correct bbox.** `read_tifxyz` on the surface above then `write_tifxyz` produced a `meta.json` bbox z of 2980.8..18020.9, where the input said -1. The writer uses `surface.bbox` when no override is given, so this quietly covers the re-save case that #1606 (closed unmerged) addressed in the writer itself.
- **`neural_tracing` is deliberately not edited.** Its z-bounds code is correct given a correct bbox; this fixes the bbox at the source.
- **In `TifxyzReader.read` the valid set is the reader's own mask** (`z > 0` and finite when no `mask.tif` ships), so the recomputed extent is over exactly the points the returned object treats as valid.
- **Relationship to #1606** (closed unmerged): that PR fixed the *writer* so new files get a correct bbox. This is the *reader*, for the 28 files already published.
- **No open PR touches `tifxyz/reader.py`.** Checked 2026-09-08 by listing every open PR's files.

Verification:

```
$ python -m pytest -q src/vesuvius/tifxyz/ tests/tifxyz_label_transfer
141 passed, 5 skipped
```

Tested on villa `main`, Python 3.14.0 (the version `pyproject.toml` requires), macOS arm64.

Happy to make it recompute only z if you would rather keep the fallback to a single grid read.

---

The patch was written with Claude Code under my direction. I ran `list_tifxyz` on the published surface myself, measured its real extent against the stored bbox myself, and wrote the example above myself.


hendrikschilling:
1.  list_tifxyz() ignores mask.tif while recomputing the bbox. The fallback in vesuvius/src/vesuvius/tifxyz/reader.py:82 defines validity solely as all coordinates being unequal to  -1. However, the full reader uses mask.tif when present. Consequently, listing and loading the same surface can produce different bboxes.
2. bbox = None is ignored and not handled as missing bbox
3. if we only want to list and not compare bbox values then this still incurs the full scan cost - make this on-demand when accessing/requesting bbox info

kadenpool:
Thanks, all three are fair. Addressed on the branch:

1. The recompute in `_bbox_from_valid_points` now uses the same validity as `read()`: `mask.tif` when present and shape-matching, otherwise `z > 0` and finite. Listing and loading the same surface give the same bbox now.

2. `bbox` resolves to `None` when there are no valid points, and `z_min`/`z_max` return `None`, so a segment with no usable bbox is left unfiltered rather than being treated as `z_min = -1`.

3. `TifxyzInfo` resolves the bbox lazily. `list_tifxyz` reads only `meta.json`; the coordinate scan runs on first access to `bbox`/`z_min`/`z_max`, and only when the stored bbox carries the `-1` marker. An ordinary listing reads no grids.

141 tests pass across `src/vesuvius/tifxyz/` and `tests/tifxyz_label_transfer`.

hendrikschilling:
@codex review

hendrikschilling:
Fix that codex raised issue then we are good to go!

kadenpool:
Fixed in 86fef65.

Codex was right, and the positional case was the worse half: making `bbox` lazy had renamed the field and moved it from third argument to fourth, so `TifxyzInfo(path, scale, bbox, uuid)` silently assigned the bbox tuple to `uuid` and the uuid string to the stored bbox, then failed somewhere unrelated later. The keyword form raised TypeError outright.

An explicit `__init__` restores the original `(path, scale, bbox, uuid)` signature while keeping the lazy resolution. `stored_bbox` is still accepted by name, and `bbox` wins if both are passed.

Added a test covering all four construction forms so a future rename fails the suite rather than a caller. 142 passed, 5 skipped across `src/vesuvius/tifxyz/` and `tests/tifxyz_label_transfer`.

hendrikschilling:
Getting there!

```
  1. [P2] Unresolved TifxyzInfo objects are corrupted by pickle round-trips.
     vesuvius/src/vesuvius/tifxyz/reader.py:28 uses object() as a sentinel, stores it in each instance at line 70, and tests it by identity at line 77. Pickling recreates that object with 
     a different identity. The restored object therefore appears “resolved,” and bbox returns the anonymous object instead of a tuple or None.

     This breaks multiprocessing queues, ProcessPoolExecutor, or metadata caches:

     restored = pickle.loads(pickle.dumps(info))
     restored.bbox  # <object ...>, not the bbox
     restored.z_min # TypeError: 'object' object is not subscriptable

     Use a pickle-stable sentinel, a separate _bbox_resolved boolean, or explicit pickle state handling. Add an unresolved pickle round-trip test.
```

kadenpool:
Fixed in e3df5c6.

You were right that this reaches further than it looks: an unresolved `TifxyzInfo` that crosses a process boundary comes back believing it is resolved, and `bbox` hands over the sentinel object itself, so `z_min` raises `TypeError` on something that is not subscriptable. Any process pool or metadata cache would hit it.

Replaced the `object()` sentinel with a `_bbox_resolved` boolean, which pickles intact.

The test covers three cases: a round trip while still unresolved, a round trip after resolving, and a segment with no bbox staying `None` rather than becoming a sentinel. I checked it fails against the previous implementation before trusting it, on exactly the symptom you described:

```
restored.bbox   # <object object at 0x...>, not the bbox
restored.z_min  # TypeError: 'object' object is not subscriptable
```

143 passed, 5 skipped across `src/vesuvius/tifxyz/` and `tests/tifxyz_label_transfer`.

PR #1733, opened 2026-09-08
VC3D: consistently use the configured remote cache directory

hendrikschilling:
**In one sentence:** VC3D now always uses the remote cache directory selected in global settings when opening or attaching remote volumes.

**One real example:** On a newly installed Windows system, I selected a cache directory on a different drive and then opened a scroll from the Open Data Catalog.

**Before:** Downloads were written to the default cache location because some volume-loading paths ignored the configured directory or used project-specific cache state.

**After this PR:** All remote volume loading paths resolve the cache directory from the same global VC3D setting, with the existing default used only when no directory has been configured.

**Why / where this is useful:** VC3D users can place remote-volume data on a drive with sufficient space and rely on that location across projects, Open Data Catalog volumes, Lasagna datasets, and direct remote attachments.

## Details

- Moves remote cache path resolution into a shared core utility.
- Removes project-specific remote cache roots and the separate attachment-time cache prompt.
- Migrates VC3D, Open Data, Lasagna, segmentation, and agent-bridge callers to the global setting.
- Resolves relative configured paths consistently and creates the selected directory when needed.
- Adds tests covering global setting changes and remote-volume loading behavior.


hendrikschilling:
@codex review

hendrikschilling:
@codex review

bruniss (approved):

PR #1736, opened 2026-09-08
spiral-fitting: README recipe for tracks-only scrolls, umbilicus and winding-count estimators

gmDevi:
**In one sentence:** A README recipe, plus two small scripts, for fitting the 2025-2026 scrolls whose spiral datasets ship only `tracks/`: how to assemble the missing inputs (umbilicus, lasagna slab, scroll spec), which overrides to set, and what winding count to use instead of Scroll 1's default of 130.

**One real example:** Starting from the public tracks of PHerc0358 (no published umbilicus), I estimated the umbilicus with `estimate_umbilicus.py` from the organizers' surface prediction, set the winding count to 100 from `winding_counts.py`, fitted slices 8000-9500 for 30,000 steps with the documented overrides, and got a mesh with 50% of track points satisfied and 90 exported windings; the same recipe produced the meshes used for the PHerc0125, PHerc0826 and PHerc0211 ink sweeps in https://github.com/gmDevi/vc-windows-tools.

**Before:** Nothing in the README says how to fit a tracks-only dataset. Following the existing docs on PHerc0125 I hit, in order: tracks silently off (`input_use_tracks` defaults to false), the outer-shell crash (#1732), the winding count left at 130 exporting dozens of windings outside the papyrus, and no way to fit six of the nine track-ready scrolls at all because they have no umbilicus anywhere.

**After this PR:** The README has a "Fitting a scroll that ships only tracks" section with the dataset layout, where each missing input comes from, the exact `FIT_SPIRAL_CONFIG_OVERRIDES` block, a winding-count table for nine scrolls, and what a successful run looks like. `estimate_umbilicus.py` writes an `umbilicus.json` for scrolls without one; `winding_counts.py` prints the crossing counts behind the table.

**Proof:** PHerc0358 fit log, estimated umbilicus, 30,000 steps:

```
overrides: {"z_begin": 8000, "z_end": 9500, "optimizer_num_training_steps": 30000, ..., "input_use_tracks": true, "input_use_outer_shell": false, "shell_outer_winding_idx": 100, "model_gap_expander_num_windings": 100}
step 29800: loss = 294.9, umbilicus = 0.0, sym_dirichlet = 35.4, dense_normals = 10.2, min_spacing = 0.0, track_radius = 139.2, track_dt = 110.2
satisfied_track_points = 37566918/74547820 (50.4%)
save_mesh fitted: winding range [10, 100)
```

The estimated PHerc0125 umbilicus is 0.2-2.6 mm from the published one through the middle of the scroll (worse near the ends); the estimated umbilici for six scrolls and the per-scroll crossing counts are in `results/umbilici` of the repository above.

**Why / where this is useful:** Anyone starting on a 2025-2026 scroll from the tracks route, which is the only route those datasets support. It turns four separate dead ends into one documented path and makes six more scrolls fittable.

## Details

Motivation: (owner to fill in)

Change: `spiral-fitting/README.md` gains one section (inserted before "Sweep runner output"). `spiral-fitting/estimate_umbilicus.py` (36 lines, numpy/scipy/numcodecs, already dependencies) reads pyramid level 3 of a surface-prediction zarr over HTTP and writes `{"control_points": [{x, y, z, score}]}`. `spiral-fitting/winding_counts.py` (40 lines) counts sheet crossings along radial rays at three heights. Both read straight from the public bucket; no local volume needed.

Limitations: the umbilicus estimate degrades within a few millimetres of the scroll ends; the crossing count is a lower bound where sheets are pressed together, hence the advice to sit above the range. The recipe was run on one RTX 3090 under WSL2.

Written with an AI coding assistant (Claude); the runs, the data and the failures are mine.


evilaliv3:
I think this tool by @gmDevi is very very useful and needed.

I'm already using it with satisfaction an i'm building some contributions directly on your branch @gmDevi; thank you if you could validate them and eventually include them and we could continue to test it together.

The contributions i would like to propose are:
- https://github.com/gmDevi/villa/pull/1
- https://github.com/gmDevi/villa/pull/3
- https://github.com/gmDevi/villa/pull/4

pmh47:
The methods for finding umbilicus and sheet-counts here are not reliable. You can raise a PR on our "community projects" webpage if you want us to link them, however they are not strong enough to go in villa. In particular, umbilici can be far from the foreground / all-papyrus centroid, and sheets are often not well separated physically in compressed regions.

pmh47:
The README updates might be useful if removed from these umbilicus-estimating and sheet-counting tools (i.e. if it tells the user to estimate these by hand or some other method). Raise another PR if so. Note for umbilicus there's already a tool in VC3D to estimate this from normal-grid representation, though it is not guaranteed to give super accurate results.

PR #1737, opened 2026-09-08
vc_grow_seg_from_seed: deal more reasonably with voxelsizes

### LINKED ISSUE #1603: vc_grow_seg_from_seed silently discards completed traces against released surface predictions (metadata 'scan' nested under source.metadata)
**In one sentence:** `vc_grow_seg_from_seed` traces a surface successfully against a released *surface prediction* volume and then silently discards it, because prediction `metadata.json` nests the scan metadata one level deeper than volume `metadata.json`, so voxel size resolves to 0 and the `min_area_cm` check drops every result.

**I was trying to:** Trace a patch on PHerc. Paris 4 from a seed, to check whether text-line phase can flag a tracer wandering onto a neighbouring sheet in a compressed region.

**Using:** `ghcr.io/scrollprize/villa/volume-cartographer:edge` (meta reports `vc_gsfs_version: "dev"`), 4-core x86_64, real data from the public bucket.

```bash
docker run --rm -v ~/vc-work:/work \
  ghcr.io/scrollprize/villa/volume-cartographer:edge \
  vc_grow_seg_from_seed \
    -v https://vesuvius-challenge-open-data.s3.amazonaws.com/PHercParis4/representations/predictions/surfaces/20260411134726-surface-20260413141734-surface-recto-2um-ps256-L0-th0.45.zarr \
    -t /work/out -p /work/params.json -s 17963 14015 27830
```

**What happened:** The trace ran to completion — 99 generations, seed correctly read as 255 — then produced no output and no error explaining why. The output directory stayed empty.

```
[WARN ] Failed to load remote volume metadata for '...th0.45.zarr':
        metadata.json missing 'scan' key
...
seed location [17963, 14015, 27830] value is 255
gen 99 processing 788 fringe cands (total done 38412 fringe: 780)
generated surface 15627277.310839 vx^2 (0.000000 cm^2)
```

Every `area` readout is `0.00 mm^2` and the final area is `0.000000 cm^2`, despite 15.6M vx². With the default `min_area_cm: 0.3` the surface is dropped at the area check after all the tracing work is done.

**What I expected or needed:** Either the surface is written, or the tool says why it was not. Losing a completed trace to a silent area check is expensive — the user sees a successful-looking run and an empty directory.

**Evidence / reproduction:** Setting `min_area_cm: -1` in the params makes the identical run save correctly:

```
generated surface 15627277.310839 vx^2 (0.000000 cm^2)
saving "/work/out/auto_grown_20260825232050529"
```

The saved `meta.json` records the inconsistency directly:

```json
"area_vx2": 15627277.310839027,
"area_cm2": 0.0
```

The result is a valid 202×202 tifxyz, 39,204 valid points (96.1%), spanning x 16041–19985, y 12107–16020, z 26218–29506.

- [x] I personally encountered and reproduced this using the version and data stated above.

## Details

The two metadata schemas differ. `Volume.cpp:1083` requires a top-level `scan`:

```cpp
auto full = vc::json::load_json_file(altPath);
if (!full.contains("scan")) {
    throw std::runtime_error("metadata.json missing 'scan' key: " + altPath.string());
}
metadata_ = full;
metadata_.update(full["scan"]);
```

But for released prediction volumes the same information is present one level down:

| | path to `samplePixelSize` | value |
|---|---|---|
| `volumes/…-2.400um-….zarr/metadata.json` | `scan.tomo.acquisition.detector.samplePixelSize` | 0.0024 |
| `representations/predictions/surfaces/….zarr/metadata.json` | `source.metadata.scan.tomo.acquisition.detector.samplePixelSize` | **0.0024** |

Prediction metadata top-level keys are `generated_at, kind, model, output, post_processing, schema_version, source, workflow` — no `scan`, but `source.metadata.scan` is the complete scan block.

A fallback to `source.metadata.scan` when top-level `scan` is absent would fix this, and `Volume.cpp:99` already uses a similar multi-key search pattern (`{"scan", "volume", "properties", "metadata"}`) elsewhere.

Two smaller things noticed alongside:

1. The `voxelsize` params override documented in `main` (`vc_grow_seg_from_seed.cpp:330-333, 367-371`) has no effect in the `:edge` image — that binary's startup header prints `mode / step size / min_area_cm / tgt_overlap_count` with no `voxelsize:` line, so it predates that code. Not a bug in `main`, but it means the natural workaround does not work with the published image.
2. It may be worth logging the area check when it discards, e.g. `discarding generated surface because area_cm2 0 < min_area_cm 0.3`, since the current run looks like a success.

I have not built VC3D from source, so I have not compile-tested a patch — this is a diagnosis with a proposed fix rather than a verified one.

jrudolph:
1. Fail early if voxelsize cannot be determined. Before, if no voxelsize for the surface volume could be found, the process discarded results late.
2. Take more metadata into account when figuring out metadata. Most importantly, for some surface predictions, we specify on which level the prediction of the base volume has been run. We can use that value to determine the actual voxelsize. If available we can also determine voxelsize both from zarr and ome-zarr metadata directly if available.
3. some detail changes in the use of the areaCm2

Fixes #1603.

Superseded #1541

### DIFF
--- a/volume-cartographer/apps/VC3D/SegmentationCommandHandler.cpp
+++ b/volume-cartographer/apps/VC3D/SegmentationCommandHandler.cpp
@@ -1027,9 +1027,10 @@ private:
 
     // Compute area_vx2 from the freshly-written tifxyz at 'dir' and write
     // both area_vx2 and area_cm2 into its meta.json. voxelSize is in micrometers
-    // per voxel (same units as Volume::voxelSize()); area_cm2 = vx2 * vs^2 / 1e8
-    // matches SurfaceAreaCalculator's convention. Returns true on success;
-    // failures are non-fatal — caller may continue without updated area.
+    // per voxel (same units as Volume::voxelSize()); area_cm2 is derived by
+    // vc::surface::areaCm2FromVox2, and omitted when the voxel size is unknown,
+    // the same way SurfaceAreaCalculator and the tracers record it. Returns true
+    // on success; failures are non-fatal — caller may continue without updated area.
     static bool updateAreaInMeta_(const QString& dir, double voxelSize) {
         std::unique_ptr<QuadSurface> qs;
         try {
@@ -1042,10 +1043,7 @@ private:
         const double area_vx2 = vc::surface::computeSurfaceAreaVox2(*qs);
         if (!std::isfinite(area_vx2) || area_vx2 <= 0.0) return false;
 
-        double area_cm2 = std::numeric_limits<double>::quiet_NaN();
-        if (std::isfinite(voxelSize) && voxelSize > 0.0) {
-            area_cm2 = area_vx2 * voxelSize * voxelSize / 1e8;
-        }
+        const auto area_cm2 = vc::surface::areaCm2FromVox2(area_vx2, voxelSize);
 
         const QString metaPath = QDir(dir).filePath(QStringLiteral("meta.json"));
         QJsonObject root;
@@ -1058,8 +1056,14 @@ private:
             }
         }
         root.insert(QStringLiteral("area_vx2"), area_vx2);
-        if (std::isfinite(area_cm2)) {
-            root.insert(QStringLiteral("area_cm2"), area_cm2);
+        if (area_cm2) {
+            root.insert(QStringLiteral("area_cm2"), *area_cm2);
+        } else {
+            // The geometry just changed. Leaving the previous area_cm2 beside
+            // the new area_vx2 would describe neither: consumers that recover
+            // a voxel size from the pair would read it scaled by the ratio of
+            // the two geometries.
+            root.remove(QStringLiteral("area_cm2"));
         }
 
         QFile out(metaPath);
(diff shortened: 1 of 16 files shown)


hendrikschilling:
```
  1. [P2] Local copies of the affected prediction stores still cannot open.
     volume-cartographer/core/src/Volume.cpp:1037 still requires metadata.json to contain a top-level scan before calling the new resolver at line 1043. The prediction schema this PR      
     targets has only source.metadata.scan, so remote access works but a local copy throws metadata.json missing 'scan' key. The tests exercise the resolver directly, not Volume::New()    
     with this schema. The local loading path should accept such metadata and derive the standard volume fields from the Zarr as the metadata-free path does.

  2. [P2] Explicit GUI area recalculation still fabricates a 1 µm voxel size.
     volume-cartographer/apps/VC3D/SurfaceAreaCalculator.cpp:161-167 converts every unknown voxel size to 1.0, then persists the resulting area_cm2 at lines 234-248. This conflicts with   
     the PR’s new rule that unknown physical scale must remain unavailable. A user selecting “calculate area” can therefore reintroduce incorrect physical metadata. This path should retain
     area_vx2 but omit/fail area_cm2 when Volume::voxelSize() returns zero.
```

jrudolph:
Yes, I had these fixed before as well but that extended the scope even more, so I kept them out for now but I can create follow ups.

PR #1747, opened 2026-09-09
vesuvius: say which extra to install instead of failing as None

mithilp:
When you run `list_files()`, which the README lists as an actual feature and something I wanted to use, you get `TypeError: 'NoneType' object is not callable`. That is not a very helpful error. After my fix it raises an ImportError instead, and it tells you that you need to run `pip install 'vesuvius[all]'`.

---

**In one sentence:** A function that is unavailable because an optional dependency is missing now tells you which extra to install, instead of failing as `None`.

**One real example:** Starting with a `pip install vesuvius` environment (core dependencies only, which is what the README's install line gives you), I called `vesuvius.list_files()`, and it produced `TypeError: 'NoneType' object is not callable`.

**Before:** `vesuvius/__init__.py` binds `list_files`, `list_cubes`, `update_list` and `is_aws_ec2_instance` to `None` when `vesuvius.utils` fails to import, and `vesuvius/data/__init__.py` does the same for `VCDataset`. `utils` needs `aiohttp` and `nest_asyncio`, which neither a bare `pip install vesuvius` nor the `volume-only` extra installs. Calling any of those names raises `TypeError: 'NoneType' object is not callable`, which names neither the missing dependency nor the extra that provides it. The original ImportError is swallowed, so there is nothing to trace back to.

**After this PR:** Each name is bound to a placeholder that raises `ImportError: vesuvius.list_files needs optional dependencies that are not installed. Install them with: pip install 'vesuvius[all]'`, and chains the original ImportError as `__cause__` so the missing module is still visible in the traceback. `vesuvius.utils` itself stays `None`, so existing `if vesuvius.utils is None` checks keep working.

**Proof:** Same script both times, run against `40f9c6b` and then against this branch. It makes `aiohttp` and `nest_asyncio` unimportable via a `sys.meta_path` hook, which reproduces a core-only install without needing a second environment:

```
########## on origin/main (40f9c6b) ##########
vesuvius.utils is None: True

########## with this PR ##########
vesuvius.utils is None: True
ImportError: vesuvius.list_files needs optional dependencies that are not installed. Install them with: pip install 'vesuvius[all]'
```

Test suite, run with `PYTHONPATH=src python -m pytest tests/test_optional_dependency_errors.py -q`:

```
this branch      7 passed
origin/main      2 failed, 5 passed
```

The two that fail on main are `test_placeholder_raises_importerror_naming_the_extra` and `test_placeholder_keeps_the_name_it_stands_in_for`. The other five assert behaviour that must not change: that `Volume` still imports on a core-only install, and that `vesuvius.utils` is still `None` rather than a placeholder.

**Why / where this is useful:** The README tells you to run `pip install vesuvius` and then lists `list_files()` among the things you can do. Following both instructions in order produces a `TypeError` about `NoneType`, which reads like a bug in the library rather than a missing extra. This is most likely to hit someone in their first ten minutes with the package, when they have the least context for diagnosing it. The fix is a message that names the command to run.

## Details

Tested against `40f9c6b`.

- `vesuvius/src/vesuvius/_optional.py` (new): `requires_extra(name, extra, cause)` returns a callable that raises `ImportError` naming the extra, `raise ... from cause`, and carries the original `__name__`, `__qualname__` and a `__doc__` so `help()` and error messages still identify it.
- `vesuvius/src/vesuvius/__init__.py`: the `except` around the `utils` import binds the four public callables to placeholders instead of `None`. `utils` itself is still set to `None`.
- `vesuvius/src/vesuvius/data/__init__.py`: same treatment for `VCDataset`, pointing at the `models` extra since that is what supplies torch.
- `vesuvius/tests/test_optional_dependency_errors.py` (new): 7 tests.

Limitations: the placeholder raises when the name is *called*, not when it is imported, so `from vesuvius import list_files` still succeeds on a core-only install. That is deliberate — raising at import time would break `vesuvius.utils is None` checks and any module that imports the name without calling it. Code that tests truthiness (`if list_files:`) will now see a truthy callable where it previously saw `None`; I found no such call site in this repository.

### DIFF
--- a/vesuvius/src/vesuvius/__init__.py
+++ b/vesuvius/src/vesuvius/__init__.py
@@ -1,13 +1,11 @@
 """Public entry point for the Vesuvius package."""
 
 from . import data, install
+from ._optional import requires_extra as _requires_extra
 
-# Always expose Volume; protect VCDataset because it depends on PyTorch.
-from .data import Volume
-try:
-    from .data import VCDataset  # requires the 'models' extra (torch)
-except Exception:
-    VCDataset = None  # type: ignore
+# Volume needs only the core dependencies. VCDataset needs torch, and becomes
+# a placeholder that names the missing extra when it is absent.
+from .data import Volume, VCDataset
 
 # Guard optional heavy modules.  They will be None unless their extras are installed.
 try:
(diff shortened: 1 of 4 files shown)


hendrikschilling:
```
  [P2] The suggested installation command is not a reliable fix for a clean pip installation.
  _optional.py:17–20 (https://github.com/ScrollPrize/villa/blob/de44ff73cd7749aa0a8acf2cbd6661dad419c3e2/vesuvius/src/vesuvius/_optional.py#L17-L20) recommends pip install 'vesuvius[all]' 
  or [models]. Both extras include volume-cartographer, which the same revision explicitly identifies as a monorepo-only dependency unavailable on PyPI. Pip does not resolve its local     
  [tool.uv.sources] mapping. The new recovery instructions therefore lead the targeted user into another installation failure. Recommend the specific missing dependency or the documented  
  repository setup instead.

```

PR #1753, opened 2026-09-10
vc_render_tifxyz: create ome-zarr pyramid downscales isotropically

jrudolph:
And make it the default.

So far, we rendered all segments with z-anisotropic pyramid creation for the benefit of webknossos users. However, this
now means that users that require isotropic downscales (e.g. to retrieve 4x downscales from 2um segments) need to redo the downscale in z direction leading to extra storage and processing cost.


jrudolph:
Still need to consider if that's even the best idea going forward.

PR #1773, opened 2026-09-13
ink_detection: robust flat normalization must exclude the reader's zero padding

### LINKED ISSUE #1481: ink_detection: robust flat inference includes zero padding in normalization statistics

rodriguescarson:
Fixes #1481.

`FlatBlockDataset.__getitem__` normalizes the whole patch after
`FlatPatchReader.read()` has already inserted zero padding, so for
`tifxyz_robust` the artificial zeros enter the percentiles, median and MAD and
shift every real CT voxel. Training does the opposite — it normalizes only the
voxels read from the source and leaves padding at zero:

```python
# data/dataset.py::_tensor_sample
if image_valid_slices is not None:
    image[image_valid_slices] = normalize_image(
        image[image_valid_slices], self.config.normalization
    )
```

## Verified, and it reproduces the issue's number exactly

On unpatched `main`, with a 3×4×4 source centered into a 5×4×4 model input, the
padded planes come out at **−2.2658674716949463** instead of 0 — the issue
predicted "about −2.266":

```
Mismatched elements: 16 / 16 (100%)
 [0, 0]: -2.2658674716949463 (ACTUAL), 0.0 (DESIRED)
Max absolute difference among violations: 2.2658675
```

Two regressions in `tests/ink_detection/test_flat_inference.py`:

* `test_flat_robust_normalization_excludes_reader_padding` — asserts both padded
  planes stay exactly zero and every real voxel matches
  `normalize_robust(source)` placed into a zero tensor. **Fails on `main`,
  passes here.**
* `test_flat_divide_255_is_unchanged_by_the_padding_rule` — the control.
  `divide_255` needs no special case because zero over 255 is still zero, and it
  **passes on both**, so the pair distinguishes the fix rather than asserting
  current behaviour.

Suites on this branch (Python 3.12, torch 2.13): `test_flat_inference.py`,
`test_native_full3d_inference.py`, `test_inference_boundaries.py`,
`test_model_foundation.py`, `test_data_foundation.py` — **84 passed**.

## Implementation note

Rather than recomputing the padding arithmetic in the dataset, I factored it out
of `FlatPatchReader.read()` into `source_placement()` and used it from both
sides, so the placement and the normalization region cannot drift apart. That
also covers the case the issue's sketch does not: when a block starts before the
array, `read()` places source data at `source_y0 - y0`, not at 0, so a
`slice(0, block.valid_h)` would normalize the wrong rows.

`valid_region_zyx()` returns the same placement as a Z/Y/X slice tuple for the
dataset's post-`moveaxis` view, and `None` when a block is entirely outside the
volume — those patches are all zeros and are already dropped downstream by
`keep = metadata[:, 4] > 0`, so their behaviour is deliberately unchanged.

---

Reopened as a new pull request because #1646 was closed by the staleness bot today (14 days, no review), and neither reopening nor pushing to the closed PR was permitted for this account. The branch, the fix and the regression test are unchanged from that PR.

### DIFF
--- a/vesuvius/src/vesuvius/ink_detection/inference/infer.py
+++ b/vesuvius/src/vesuvius/ink_detection/inference/infer.py
@@ -455,16 +455,44 @@ class FlatPatchReader:
             block = array[y0:y1, x0:x1, self.layer_indices]
         return np.asarray(block)
 
-    def read(self, y0: int, x0: int, out_h: int, out_w: int) -> np.ndarray:
-        """Read one padded H/W/Z patch with centered short-depth placement."""
+    def source_placement(self, y0: int, x0: int, out_h: int, out_w: int):
+        """Where in an output patch `read` puts source voxels, or None.
 
+        Everything outside this region is padding the reader inserted, and
+        training leaves such padding at zero rather than folding it into
+        normalization statistics.  Both `read` and the flat dataset derive
+        their placement from here so the two cannot drift apart.
+        """
         y1, x1 = y0 + out_h, x0 + out_w
         source_y0, source_x0 = max(0, y0), max(0, x0)
         source_y1, source_x1 = min(self.height, y1), min(self.width, x1)
+        if source_y1 <= source_y0 or source_x1 <= source_x0:
+            return None
+        depth_start = self.output_depth_start
+        return {
+            "source": (source_y0, source_y1, source_x0, source_x1),
+            "rows": slice(source_y0 - y0, source_y1 - y0),
+            "cols": slice(source_x0 - x0, source_x1 - x0),
+            "depth": slice(depth_start, depth_start + self.layer_indices.size),
+        }
+
+    def valid_region_zyx(self, y0: int, x0: int, out_h: int, out_w: int):
+        """`source_placement` as a Z/Y/X slice tuple, or None."""
+
+        placement = self.source_placement(y0, x0, out_h, out_w)
+        if placement is None:
+            return None
+        return (placement["depth"], placement["rows"], placement["cols"])
+
+    def read(self, y0: int, x0: int, out_h: int, out_w: int) -> np.ndarray:
+        """Read one padded H/W/Z patch with centered short-depth placement."""
+
         dtype = np.float32 if self.preprocessing == "tifxyz_robust" else np.uint8
         output = np.zeros((out_h, out_w, self.output_depth), dtype=dtype)
-        if source_y1 <= source_y0 or source_x1 <= source_x0:
+        placement = self.source_placement(y0, x0, out_h, out_w)
+        if placement is None:
             return output
+        source_y0, source_y1, source_x0, source_x1 = placement["source"]
         block = self._read_raw(source_y0, source_y1, source_x0, source_x1)
         if self.preprocessing == "tifxyz_robust":
             block = np.asarray(block, dtype=np.float32)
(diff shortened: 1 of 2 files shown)


rodriguescarson:
Closing this. I checked it against real data too. `ink_9um` does train with `robust_mad`, so it's exposed in principle, but the flat reader almost never pads: the last block is boundary-aligned and depth is centre-cropped. Padding only happens when a render has fewer layers than the model input (17 for `ink_9um`) or a canvas is under 128 px.

I ran the numbers on the 340 published meshes of the eight eligible 9 um scrolls. Every render has 31 layers and every canvas is wider than 128 px, so on that set the patch changes no output. Leaving this here in case someone hits the edge case, since #1481 is still open.

PR #1775, opened 2026-09-13
tifxyz label transfer: release temporary mapping arrays before rasterization

Kotoro78:
**In one sentence:** TIFXYZ label transfer releases mapping arrays once they are no longer needed, reducing memory retained during rasterization without changing the labels.

**Motivation:** I've been evaluating label-transfer workflows on public scroll surfaces, with attention to what can run with limited memory. We measured this small improvement on PHerc1667 and PHerc0139 data. The saving is modest, but it comes from shortening the lifetime of existing arrays and keeps the workflow unchanged.

**One real example:** In the recorded PHerc1667 w018 transfer to a 12,960 × 30,040 target, using ink and supervision labels, eight workers and query batches of 2,048, both versions produced identical decoded label outputs.

**Before:** The gathered affine input, unused mapping distances and the private mapper's search arrays remained referenced beyond their last use.

**After this PR:** Those arrays can be released earlier. The historical PHerc1667 comparison saved a paired median **41.65 MiB of sampled peak RSS (2.43%)** across two timed pairs. A PHerc0139 w016 crop saved **22.77 MiB**. Timing varied; I am not claiming a speedup or an avoided out-of-memory failure. These measurements predate the later upstream geometry changes and are not a new benchmark of today's main.

**Proof:** [Measured results, exact inputs/settings and validation records](https://github.com/Kotoro78/villa/tree/5256754d0fbbc7a9dadc03224e68d77c6b97ef5c/p22-evidence). The report separates the historical real-data measurements from the current correctness tests.

**Why / where this is useful:** People transferring labels between scroll surfaces can retain fewer temporary arrays during rasterization. There is no new dependency or configuration, and callers that reuse their own `SurfaceMapper` keep its search state.

## Details

The proposed diff contains only three production edits and a lifetime regression-test module. Search state is released on the private mapper after its workers finish; the transformed coordinates needed by rasterization are retained.

The reviewed base is `be09a85035059fd83471b1632b5898c62f2c65b1`. On Windows with Python 3.14.7 and the compiled native rasterizer, the final component suite passed **157 tests and 32 subtests**, with no failures or skips, when combined locally with the separately authored [PR1724](https://github.com/ScrollPrize/villa/pull/1724). That PR fixes inherited Windows mapped-file cleanup failures and is **not included** here. Without it, the earlier baseline and candidate checks both had the same eight inherited failures. The added test also handles an absent optional native build while continuing to fail on a broken build.

I have personally reviewed the code and the evidence. AI tools assisted with implementation, testing and this description. The measured benefit is small; no external adoption or result on an untested platform is claimed.


Kotoro78:
Hi, just a gentle follow-up on this memory-lifetime improvement. I appreciate that reviewing contributions takes time. Is there anything I could clarify or provide to help with the review?

I also noticed that the GitHub Actions runs show “action_required” without any jobs running. Could you let me know whether anything is needed from my side?

Thank you for your time!

PR #1778, opened 2026-09-14
VC3D fiber map: inspectable markers, delete from the map, floating axes with sheet distance

skyward7187:
**In one sentence:** In the VC3D Fiber Map tab you can zoom into a fiber and still see its line under the markers, delete a fiber from the map or its dock list, and read winding number, sheet distance and height off axes that stay in view at any zoom.

**One real example:** With PHercParis4 fibers loaded and the Fiber Map built, I zoomed into a linked crossing until one winding filled the view; the control-point dots and link dots stayed small, sat on their fibers, and the winding numbers, sheet distance and height stayed readable along the edges. I selected a stray fiber, ctrl+right-clicked it, chose Delete, confirmed, and the map updated without the fiber.

**Before:** Markers were sized for print and only clamped from below, so zoomed in a control dot grew to hundreds of voxels and hid the line, and the single link dot at the midpoint of the two linked control points sat on neither fiber. The winding numbers were pinned text inside the scene above the map's top edge, so any zoom that scrolled that edge off lost them; there was no distance or height axis. Fibers could only be deleted from the Fibers dock in the main tab.

**After this PR:**
- Control dots, link dots and suspect rings stop growing at 6, 8 and 7 px on screen. Each link is a dot on both of its control points joined by a dotted connector; zoomed out the two merge into one, zoomed in each stays on its own fiber. Fibers draw unsmoothed so every marker lands on its line.
- Ctrl+right-click on the selected fiber offers "Go to control point N" (when a dot was hit) and "Delete <fiber>...", with a confirmation naming the fiber. The dock's fiber rows offer the same Delete. The map updates about a second after.
- Three axes are painted over the map: winding numbers above the scroll ceiling, sheet distance below the floor, height left of the map. Each floats at its extent edge while that edge is on screen and clamps to the viewport edge once it scrolls off; bands run only along the data. Distance and height label in mm/cm/m (height capped at cm) from a 1-2-5 tick ladder, or in voxels when the package has no voxel size.
- Sheet distance is measured along the sheet, not at the reference radius: the layout fits the fibers' umbilicus radius as a line in the winding coordinate and integrates it, so outer windings measure longer than inner ones. The bottom axis tooltip states the fitted pitch. With under half a winding of placed fibers it falls back to the reference-radius arclength.

**Proof:** GUI-tested by the author on the local fiber dataset through the iteration: the marker cap and the two-endpoint links at high zoom, the axes floating at the extent when zoomed out and clamping when zoomed in, panning without axis echoes, the height axis in cm. Screenshots to follow. Unit tests below pass.

**Why / where this is useful:** Fiber annotators use the map to find winding errors and stray fibers; they can now inspect a crossing at full zoom without the markers covering it, remove a bad fiber without leaving the tab, and read where they are on the scroll in physical units.

## Details

28 commits on main d82e13edf (rebased; main has since moved without overlap), all under `volume-cartographer/apps/VC3D`:

| Commit | What |
|---|---|
| 5d4164c75 | VC3D fiber map: cap marker size on zoom, delete fibers from the map |
| 593d9d07a | VC3D fiber map layout: fit a sheet model for distance along the sheet |
| a8e55f5a0 | VC3D fiber map: rulers for windings, sheet distance and height |
| 07af5ddc7 | VC3D fiber map: draw fibers unsmoothed so markers sit on their lines |
| f007147dd | VC3D fiber map: axes float at the scroll extent, clamp to the viewport |
| 8871af507 | VC3D fiber map: axes run only along the data, no echoes on pan |
| 8bce330c7 | VC3D fiber map: draw links as a dot on each linked control point |
| 58a6c43fd | VC3D fiber map: thin the link endpoint fill so overlapping dots match the palette |
| 0081b9084 | VC3D fiber map: the height axis stays in centimetres |
| e4eea17c8 | VC3D fiber map: harden the delete path, the sheet inverse and the rulers (review round 1) |
| 5e57a9042 | VC3D fiber map: ruler labels at a band's end move inward rather than being cut |
| af592a5e4 | VC3D fiber map: delete runs after the confirmation's signal; winding marks culled in floating point (review round 2) |
| 508fcf51f | VC3D fiber map: one delete at a time (review round 3) |
| ecde03675 | VC3D line annotation: deleteFibers runs one at a time and deletes loaded fibers only (review round 4) |
| 9159d5c38 | VC3D line annotation: deleteFibers resolves its targets by file name across the save drain |
| f158ca25c | VC3D line annotation: renames wait for a pending delete; sessions matched by file name (review round 1) |
| 6899c9884 | VC3D line annotation: imports wait for a pending delete; link cleanup matches by file name (review round 2) |
| a1ad5bcd2 | VC3D line annotation: import guard on the shared path; snapshot upsert by file name (review round 3) |
| 2243defc0 | VC3D line annotation: delete reports through captured names; scheduler owners by identity (review round 4) |
| 5d6bc9683 | VC3D line annotation: link-cleanup owners scheduled under current ids (review round 5) |
| c091fdc58 | VC3D line annotation: refresh every pane's branch overlay after a delete's link cleanup (review round 6) |
| 19552fe5d | VC3D line annotation: the all-pane branch refresh belongs to the delete cleanup, not the rename sync |
| 5faddb350 | VC3D line annotation: runtime fiber ids stay stable across in-package reloads |
| 2dd5b168b | VC3D line annotation: identity bound on persist, retired on delete, moved on rename; loads stand down when superseded (review round 1) |
| 86e9336ea | VC3D line annotation: identity bound only at lifecycle points; merge/split retire; rename revalidates (review round 2) |
| 018ada21e | VC3D line annotation: rename from the post-drain record by name and id; retirement checks its package; suppressed sessions refused everywhere (review round 3) |
| fd9dd5cd5 | VC3D line annotation: deleted sessions distinguished from suppressed ones; retirement owns its suppression and follows its package (review round 4) |
| 9c5e8c2dc | VC3D line annotation: narrow two comments and two failure messages (review round 5) |

Build and tests:
```
cmake --build volume-cartographer/build --target VC3D test_fiber_global_layout test_fiber_map_ruler_math test_line_annotation_fiber_deletion test_line_annotation_fiber_identity -j
volume-cartographer/build/bin/test_fiber_global_layout      # 21 passed (3 new: sheet model fit, fallback, mirroring)
volume-cartographer/build/bin/test_fiber_map_ruler_math     # 8 passed (new)
volume-cartographer/build/bin/test_line_annotation_fiber_deletion  # 18 passed (new)
volume-cartographer/build/bin/test_line_annotation_fiber_identity  # 20 passed (new)
volume-cartographer/build/bin/test_fiber_network_layout     # 9 passed
volume-cartographer/build/bin/test_fiber_winding_solver     # 28 passed
```

Notes and limitations:
- Delete is deferred past the menu's nested event loop, re-checks the map's dependency set against the one captured when the menu opened, and re-resolves the fiber from its file name after the confirmation dialog, so a reload or edit during the dialog cannot delete the wrong fiber.
- The sheet model is an Archimedean assumption (radius linear in winding). It is a display scale for the ruler only; the solve and the map geometry are unchanged. Both coefficients are in the output digest so Full rebuild's memoization verification covers them.
- Unsmoothed fibers show the placement noise the de-bumping used to hide. If that reads too rough on some packages, the alternative is to keep smoothing and snap markers onto the smoothed curve.
- `FullViewportUpdate` repaints the whole viewport per pan. The scene is a few hundred cosmetic paths; no slowdown noticed, but a very large map could feel heavier.
- The red winding-suspect markers are unchanged apart from being projected onto the drawn curve; making them actionable is planned separately.
- Controller changes from the two review comments are described in the section below; pre-existing hazards the reviewers set aside are listed under follow-ups.

Review: fiber-map work - one full-context Codex (gpt-6-astra) pass (REWORK, ten findings) and four verification rounds to READY. First review comment (reload during the delete's save drain) - plan review, then eight code rounds to READY. Second review comment (synchronizer accepting an id match with a disagreeing name) - two independent plan reviews of a session-rebind design (rejected), two of the stable-id design (revised), then five code rounds with two independent reviewers each, READY at 9c5e8c2dc.

## Controller changes from the review (two logical parts)

Two review comments found the same class of bug at two sites: the controller renumbered stored fibers from 1 on every in-package reload while everything else kept the id it had. The fix is in two parts, both under `apps/VC3D`.

### 1. Deletion across pending saves (`LineAnnotationFiberDeletion.hpp`, `deleteFibers`)

- The delete captures each target's file (source root + file name) and the package identity (generation + fibers directory) before draining saves, aborts if the identity moved, and resolves by file afterwards; ids are never consulted after the wait.
- Targets that vanished or became ambiguous are skipped and logged; removal failures are collected and reported once after cleanup.
- One delete at a time; rename and import requests made while a delete is pending are refused (not queued).
- A deleted fiber's open session is marked deleted (distinct from the transient save suppression); the session save (entry and after finalize), the link/merge/split handlers and the linked-peer snapshot loop all refuse or skip it.
- The agent bridge's `fiber.delete` reports per requested id from the captured files; partial failure is `-32005` with `detail`, `deleted`, `aborted:false`; a package change is `aborted:true`. Already-absent files count as deleted. Validation of unknown ids stays all-or-nothing.

### 2. Stable runtime ids and one identity rule

Main gained its own registry while this PR was open (`FiberRuntimeIds`: ids keyed by source root + file name, never reused for the controller's lifetime, plus multi-source fibers). This PR now builds on that instead of carrying a parallel id space; its earlier package-local id space and `LineAnnotationFiberIdentity.hpp` are gone.

- One identity rule, `sameFiberIdentity` (in the deletion header): when both records carry a runtime id the ids decide, the file name is the fallback only when an id is missing. `branchReferencesFiber`, `branchLinkedEndpointMatches`, the reciprocal and endpoint synchronizers, the removed-reciprocal test and the deletion helpers all use it. Names-first matching (an earlier version of this PR) is wrong on main now that one file name can exist in several sources.
- The reciprocal synchronizer additionally verifies the mirrored branch points back at the target's owner before it rewrites anything.
- Gap in the registry closed: a session not opened from a stored fiber had no source root, so its saved fiber was remembered under the bare file name and re-keyed under the package directory on the next reload (a fresh id, and the open session then failed the source-root checks in the snapshot and upsert paths). New sessions now take the package's fibers directory as recorded by the loader (`primaryFiberSourceRoot`), merged and split fibers the same at creation, and the registry is told about a file when its save is queued, when a merge/split write succeeds and after a rename.
- Split halves are both allocated from the registry (the second was derived as `prefixId + 1`).
- A load superseded by a package switch or a newer load inside its broken-link or repair-error dialog stands down instead of publishing. The loader is not otherwise made reentrant.
- Id-keyed geometry caches are cleared on every reload.
- Rename re-resolves its fiber by id and name after each yield and writes the post-drain record; merge/split retirement abandons itself on a package change (checked after pane close, after its save drain and after the re-fit), checks the originals before removing them, and releases on failure only the suppression it set; the re-fit itself stops on a package change; import treats a name owned by a loaded fiber or open session as taken.

### Observable behaviour changes

- Imports and renames are refused while a delete is pending, with a message to retry.
- `fiber.delete` outcomes are per request and surface aborts; the MCP tool text describes this.
- Structural edits (link, unlink, pending toggle, merge, split) on the session of a deleted fiber are refused with a message; the transient suppression of a headless save or a retirement window does not affect editing.
- Import never takes a name that a loaded fiber or an open session owns, even when no file exists for it yet.
- A fiber first saved from a new session keeps its runtime id across the next reload (previously it changed, on main too).

### Follow-ups (pre-existing, out of scope of this PR)

- Structural edit handlers and close/save flows keep session and pane references across yields (autosave flush, finalization dialogs); they should revalidate lifetime and identity after each yield.
- The load-time re-optimization prompt and the pane-less re-optimization's dataset picker do not preserve the originating package through their yields; a nested package switch there can continue with stale state.
- Retirement failure recovery is not all-or-nothing: the save job can leave recovery copies, which the merge/split failure messages do not surface.
- Overlapping merge/split operations on the same originals do not compose their temporary save suppression.
- `FiberRuntimeIds` has no rekey for a rename: after renaming `a.json` to `b.json` and importing a new `a.json`, both keys resolve to one id on the next reload.
- `saveSessionAsFiber`'s linked-peer commit loop, its no-op probe, and the linked-seed creation still upsert by file name before id, so a same-named fiber in another source can be overwritten.
- A `redirectFiberSource` during a save drain leaves the paths captured by rename and by the merge/split retirements pointing at the old source.
- The re-optimization prompt has no package check before `reoptimizeMergedFibers` is entered (the checks inside it start from the generation at entry).
- The loader's broken-link prompt is not reentrant in general; only its pre-publication continuations stand down.
- Multi-source dedupe can orphan an open session: when the copy a session was opened from is dropped in favour of a survivor in another source, the session keeps its own id, source root and save destination (its edits go to a file the list no longer shows). This PR remaps the stored fibers' and open sessions' branch refs onto survivors but does not move a session's own identity, because that would change where its edits are saved and would have to merge with an already-open survivor. Consequence with id-first matching: such a session's next linked save can fail validation with "missing reciprocal branch" instead of silently rewriting the survivor's reciprocal; reopening the fiber from the survivor recovers. The dedupe should close or rebind such sessions, a decision for the multi-source work.

### Tests and their limits

`test_line_annotation_fiber_deletion` (19) exercises the deletion helpers and the identity rule, including the reload-with-renumbering scenario, the same file name in two sources, and the capture/wait/resolve boundary with an injected wait. They do not drive the real controller, its save queue or dialogs; there is no harness for that, and I am not presenting them as controller-level coverage.

https://claude.ai/code/session_01RgmjRqrVfdAvDMkamkhazk


hendrikschilling:
```
  [P1] A reload during the save drain can still delete the wrong fiber.
  volume-cartographer/apps/VC3D/LineAnnotationController.cpp:3067 waits for saves using a nested event loop. Afterward, lines 3079–3090 resolve the original numeric ID against the now-    
  current fiber list and delete that file.

  Importing fibers or switching projects during this wait can reload the list, assigning IDs again from 1. The requested ID can therefore belong to a different fiber. The new guard        
  prevents another delete, but not imports or project changes. The map’s dependency check happens before this wait, so it cannot catch this.

  Minimal fix: capture the package identity and target filenames before waiting; afterward, abort on a package change and resolve by those filenames, not the old IDs. Add a regression test
  that reloads/reassigns IDs during the wait.

  This underlying controller weakness predates the PR, but the new map deletion feature inherits it, and the claimed identity hardening is incomplete.

```

skyward7187:
Thanks, agreed on both counts: the map's check ran before the drain and the controller re-resolved the numeric id afterwards. Fixed in the controller so every entry point (both Fibers docks, the map, the agent bridge) gets the same protection:

- `deleteFibers` captures each requested fiber's **file name** and the **package identity** (package generation + fibers directory) before `waitForFiberSaves()`. After the drain it aborts with an error if the identity moved, and otherwise re-resolves each captured file name against the reloaded list and deletes by the *current* id. The requested ids are never consulted after the wait. Fibers without a file name, names that vanished during the wait, and duplicated names are skipped with a log line.
- The capture / wait / resolve sequence is a header-only helper (`LineAnnotationFiberDeletion.hpp`) with an injected wait. The controller passes `waitForFiberSaves`; the regression test `test_line_annotation_fiber_deletion` passes a wait that rebuilds the list with ids reassigned from 1 (your scenario: request id 2 = `b.json`, after the reload id 2 is `a.json` and `b.json` is id 3; exactly `{3}` is deleted), plus order reversal, vanished file, package switch, directory change, empty directory, duplicate names, duplicate requested ids. The test drives the same wrapper the controller calls; it does not spin a real nested event loop (no controller harness exists), which I want to be upfront about.
- Closing the holes around the same reload: renames and imports (menu and bridge `fiber.import`) are refused while a delete is pending, since either could hand a captured name to another fiber; removal errors are collected and shown once after cleanup so the loop never yields again; open sessions and branch refs are matched by file name rather than by their (stale after a reload) ids for save suppression, link cleanup and the metadata scheduler's owner selection; the bridge's `fiber.delete` reports through the captured names and surfaces an abort as an error rather than reading "requested id is gone" as success.

Left alone on purpose (adjacent, not this scenario): `renameFiberFile` still rebinds open sessions by id only; since renames are now refused during a delete it does not interact with this fix, and I would rather do it in its own PR.

Reviewed in eight Codex (gpt-6-astra) rounds; all findings addressed. Commits since your comment: 9159d5c38..19552fe5d.

hendrikschilling:
```
[P1] Post-delete metadata saving can still retarget an unrelated link after ID reassignment.

  At volume-cartographer/apps/VC3D/LineAnnotationController.cpp:14510, the scheduler correctly selects the surviving session by filename, but calls storedFiberFromSession(). That also runs
  branch synchronization.

  The synchronizer at LineAnnotationController.cpp:14658 still accepts matching numeric IDs even when filenames disagree, then overwrites the link’s filename.

  Example: session A retains old ID 1; after reload, ID 1 belongs to another fiber. Saving A’s metadata after deleting a linked fiber can change an unrelated link targeting that other     
  fiber into a link targeting A.

  I reproduced that mutation using the unchanged synchronization method in an isolated harness:

  Before: D -> new.json
  After:  D -> a.json (id 1)

  This is a pre-existing synchronizer weakness reached through the cleanup path, not a failure of the new deletion-target resolver. Apply filename-first matching throughout the invoked    
  synchronization, or avoid that broader synchronization when merely saving deletion cleanup.

```

skyward7187:
Thanks, that is the same bug class as your first comment showing up one call deeper, so rather than patch the synchronizer alone I went after the cause: runtime ids being handed out again from 1 on every in-package reload while open sessions, their branch refs, the docks, the fiber map and bridge clients kept the old numbers (67 id comparisons in the controller could hit the wrong fiber after a reload).

**Package-local stable ids.** The controller now owns a per-package id space (file name -> id, plus a high-water mark). A reload of the fiber list (import, repair reloads, vpkg-ready) keeps every known file's id and gives new files fresh ids; ids are never recycled within a package; only a package change restarts the numbering. Every id creation goes through one allocator (new sessions, merge, split - whose second half no longer takes `prefixId + 1` - and the linked seed), and link validation no longer renumbers. Identity is bound when a save is accepted or a merge/split write succeeds, retired when the app deletes the file (delete, merge, split), and moved by a completed rename; a name merely absent from one load keeps its binding and returns with its id. A load superseded by a package switch or a newer load inside its broken-link or repair-error dialog stands down instead of publishing. The id-keyed geometry caches clear on every reload. This is `LineAnnotationFiberIdentity.hpp`, tested in `test_line_annotation_fiber_identity` (18 cases incl. the reload/renumber scenario, retire/rebind, split halves).

**Names-first matching through the synchronization chain**, as you asked: `branchReferencesFiber`, `branchLinkedEndpointMatches`, `syncReciprocalBranchControlPointReferences`, the three lookups in `syncBranchEndpointPositions` and the removed-reciprocal test all use one rule - names decide when both records carry one, the id counts only when a name is missing. The reciprocal synchronizer additionally checks that the mirrored branch points back at the target's owner before it rewrites anything (your D -> new.json case).

**Around the same yields**: rename re-resolves its fiber by name *and* id after the name dialog and after its save drain and writes the record as it stands after the drain; the deferred merge/split retirements abandon themselves if the package changed and check the originals are still the fibers they consumed; import treats a name owned by a loaded fiber or an open session as taken even without a file; a session whose fiber was deleted underneath it (a new per-session `fiberDeleted` flag, distinct from the transient save suppression the headless save and the merge/split retirement use) is refused by the link, merge and split handlers and by the session save (entry and after its finalize), and skipped by the linked-peer snapshot loop; merge/split release only the suppression they set on failure and re-check their package after each yield.

**What this does not claim**: identity is the package-local file name, so a file replaced on disk under an unchanged name outside the app is the same fiber by definition; the loader is not made generally reentrant (the supersession check covers its pre-publication dialogs only); geometry already held by open sessions or running queries is not refreshed by a reload; and the tests exercise the identity and deletion helpers, not the real controller with its event loop and save queue. The earlier owner-id translation and all-pane refresh in the delete's link cleanup compensated for renumbering and are removed.

Pre-existing reentrancy hazards in adjacent flows that the reviews turned up (finalization crossing a package switch inside re-optimization, overlapping merge/split suppression, an inspection opened during a retirement drain) are listed as follow-ups in the PR description rather than fixed here.

Reviewed with Codex (gpt-6-astra): two independent plan reviews rejected a session-rebind design, two more shaped the stable-id design, then five code rounds with two independent reviewers each, both READY at the end. Commits since your second comment: 5faddb350..9c5e8c2dc.

hendrikschilling:
Looks good but currently has conflicts on merge. Resolve those than we can merge it.

skyward7187:
Main moved under this PR: #1768 brought its own `FiberRuntimeIds` registry (ids keyed by source root + file name, never reused for the controller's lifetime) and multi-source fibers, which conflicted with the package-local id space I had added for your second comment. I resolved the merge by dropping mine and building on upstream's - `LineAnnotationFiberIdentity.hpp` and its test are gone.

Two consequences worth calling out:

- **Matching is id-first again, through one rule.** Names-first (what I did in the last round) is wrong once one file name can exist in several sources, so `sameFiberIdentity` now encodes the same rule as main's `branchReferencesFiber` (ids decide when both are known, the name is the fallback for a missing id) and the synchronizers, `branchLinkedEndpointMatches` and the deletion helpers share it. Your D -> new.json case is covered by the ids being stable now, plus the owner check in the reciprocal synchronizer. The delete captures targets by (source root, file name), since a bare name is no longer unique.
- **One gap in the registry, closed here.** A session not opened from a stored fiber has an empty `fiberSourceRoot`, so the fiber it saved was remembered under the bare name and re-keyed under the package directory on the next reload - fresh id, and the open session then failed the source-root equality checks in `makeStoredFiberSessionSnapshot` / the upsert (a duplicate record on its next save). New sessions, merged fibers and split halves now take the package directory as the loader records it (`primaryFiberSourceRoot`, canonicalised the same way), and the registry is told about a file when its save is queued, when a merge/split write succeeds and after a rename. Happy to split that into its own PR if you would rather review it apart from the UI work.

Everything else from the earlier rounds stands (delete capture/outcome, the guards, merge/split package checks, the loader stand-down, rename/import checks). Reviewed again after the merge in seven Codex (gpt-6-astra) rounds, two reviewers each, converging on READY with the orphaned-session case above accepted as a follow-up. PR description updated.

hendrikschilling:
```
[P1] Renaming leaves the old filename bound to the same runtime ID.

  At volume-cartographer/apps/VC3D/LineAnnotationController.cpp:3434, rename calls FiberRuntimeIds::remember() for the new name. But FiberRuntimeIds.hpp:17 only adds that binding; it does 
  not remove the old one.

  Reproduced using the actual registry:

  1. Load a.json, assigned ID 1.
  2. Rename it to b.json, retaining ID 1.
  3. Import a different fiber as a.json.
  4. Both files receive ID 1. sameFiberIdentity() consequently considers them identical despite different filenames.

  This can confuse selection, branch synchronization and deletion, which resolve fibers by ID.

  Origin: this limitation exists in the registry imported from main. The previous PR iteration explicitly retired the old binding during rename; the merge dropped that protection.         

  Minimal fix: add a source-aware rename operation that moves the binding while preserving the ID, leaving the old filename available for a fresh ID. Add a rename → reuse-name regression  
  test.


```

skyward7187:
Thanks, confirmed: the registry only ever added bindings, so a rename left the old name pointing at the same id. Fixed in 0418fe31a, and I went after the same class everywhere a name stops belonging to its fiber:

- `FiberRuntimeIds` gains `rename(source, oldFile, newFile, id)`, which moves the binding with its id and frees the old name, and `forget(source, file)`, which releases a binding. The counter only grows, so a retired id is never handed out again; the freed name gets a fresh id on the next import. Both are source-aware (keyed by `fiberSourceFileKey`), so a Spiral copy of the same name in another source keeps its own binding.
- Controller: the rename path calls `rename()`; `deleteFibers` calls `forget()` after each successful removal; merge and split, which retire their originals through their own transactional removal rather than `deleteFibers`, release the originals' bindings right after the retirement job succeeds and before their records are erased (a failed job leaves bindings untouched).
- One more route to the same bug that a review pass found: `loadFibersForCurrentPackage` and `nextFiberId` re-remember every open session's name and id, including a session whose fiber was deleted while its editor stayed open, which would have re-bound the deleted name. A deleted session now re-remembers under an empty name: its id stays reserved, the name does not.
- Regression: `test_fiber_runtime_ids` (8 cases): your rename -> reuse-name scenario (a.json id 1, rename to b.json keeps 1, a new a.json gets 2), rename surviving the reload's re-remember, delete -> reuse, per-source independence, the deleted-session reservation pattern, and the monotonic counter. As before, this exercises the registry the controller calls, not a controller lifecycle; there is still no controller harness, so the merge/split/delete call sites were checked by review rather than by an automated test.

PR #1800, opened 2026-09-15
normalgridtools: make segment filtering linear instead of quadratic

evilaliv3:
**In one sentence:** `SegmentGrid::remove` does a linear `std::find` before the swap-and-pop that
the comment beside it calls O(1), and `align_and_filter_segments` calls it once per iteration until
the grid is empty, so that pass is quadratic in the number of segments; this keeps a position index
so the lookup is constant time.

**How big is a real grid.** I counted, on your published normal grids, the quantity that actually
enters `all_segments`: `normalgridtools.cpp:157-166` walks every path, skips those with fewer than
two points, and adds one `SegmentInfo` per consecutive pair, so it is the sum over usable paths of
(points minus one). Read straight out of the `.grid` files, **every published slice of both scrolls,
45,450 of them**, not a sample.

| scroll | plane | slices | segments per slice: min / median / p90 / max |
|---|---|---:|---|
| PHerc0826 | xy | 7,751 | 18,111 / **21,792** / 23,456 / 25,846 |
| PHerc0826 | xz | 8,169 | 454 / **22,343** / 92,591 / **173,815** |
| PHerc0826 | yz | 8,169 | 963 / **45,309** / 73,591 / 75,765 |
| PHerc1203 | xy | 9,693 | 22,199 / **25,398** / 31,861 / 33,664 |
| PHerc1203 | xz | 4,971 | 918 / **77,435** / 117,301 / 122,515 |
| PHerc1203 | yz | 6,697 | 145 / **71,694** / 121,956 / 126,307 |

Two things in that picture matter. The grids are large, tens of thousands of segments and up to a
hundred and seventy thousand. And they are **not uniform**: on the xz plane of PHerc0826 the slices
run from 454 to 173,815, so the largest slice costs about sixty times the median one, and the
largest tenth of the slices carries **45.8 per cent** of the whole plane's cost. For a quadratic
cost the tail is the bill.

**What that costs.** Reproducing the access pattern of lines 444 to 463 with `g++ -std=c++20 -O2`
on one pinned core of an otherwise idle Xeon Gold 6526Y, median of three, at the sizes the table
above says are real:

| segments in the grid | current | with a position index | saved | speedup |
|---:|---:|---:|---:|---:|
| 22,000 (median xy slice) | 22.1 ms | 0.97 ms | 21 ms | 23x |
| 45,000 (median yz slice) | 91.3 ms | 2.56 ms | 89 ms | 36x |
| 77,000 (median xz slice) | 269.1 ms | 5.20 ms | 264 ms | 52x |
| 172,000 (largest slice measured) | 1,673.8 ms | 17.6 ms | 1.66 s | 95x |

Summing those costs **slice by slice** over every published slice, rather than multiplying a median
by a slice count, which for a quadratic cost would throw the tail away:

| scroll | plane | current | with the index | ratio |
|---|---|---:|---:|---:|
| PHerc0826 | xy | 2.8 min | 7.5 s | 22x |
| PHerc0826 | xz | 17.6 min | 19.0 s | 56x |
| PHerc0826 | yz | 15.1 min | 20.1 s | 45x |
| PHerc1203 | xy | 5.2 min | 12.2 s | 26x |
| PHerc1203 | xz | 27.3 min | 26.7 s | 61x |
| PHerc1203 | yz | 33.3 min | 32.1 s | 62x |
| **PHerc0826** | **three planes** | **35.6 min** | **46.6 s** | **46x** |
| **PHerc1203** | **three planes** | **65.9 min** | **70.9 s** | **56x** |

5-cost-per-scroll

The per-plane ratio is not a constant, and the reason is worth a sentence. It runs from 22x on the
xy plane, where every slice is roughly the same size, to 62x on the yz plane of PHerc1203, where
they are not. A quadratic cost is dominated by its largest slices; a nearly linear one is not. On
the xz plane of PHerc0826 the largest tenth of the slices carries 45.8 per cent of the current cost
but only 36.2 per cent of the indexed cost, so the fix does not just make each slice cheaper, it
also flattens which slices matter. The more uneven the plane, the more this change is worth.

**One thing you will notice before I say it, so I will say it first.** In the current tree
`align_and_filter_segments`, `align_and_extract_umbilicus` and `SegmentGrid` are not called from
anywhere outside `core/test/`, and the block in `NormalGridVolume.cpp:374-381` that would read the
umbilicus metadata this pass writes is commented out. The published grids carry no metadata at all,
so they did not come through here either. **Nothing in the tree pays this cost today.** That is
also the reason nobody has noticed it. I am raising it as a correctness fix with a measured cost
attached rather than as a performance win you are currently losing, and if you would rather this
file stay as it is until something calls it, that is a reasonable answer and I will not argue with
it.

**Before:** `remove` scans `all_segments` for the pointer it was handed. The scan is what the loop
in `align_and_filter_segments` pays for: it removes exactly one segment per iteration
(`normalgridtools.cpp:269`) and runs until `unused.count() == 0` (`:202`), so the total is
sum(N, N-1, ...) pointer comparisons. The two lines after the scan are already O(1), and the
comment above them says so.

1-time-vs-n

Least squares on the log-log data over N from 5,000 to 1,000,000: the current code grows as
**N^2.18**, the indexed version as **N^1.38**, and the comparison count as **N^2.00**. Every number
on this page and in the three figures comes from one run of `bench_removal.cpp`; the raw CSV is
attached as `measurements.csv`.

The third figure is the part that does not depend on my machine. The measured comparisons per
removal land exactly on N/4, which is half of what is left, averaged over a container that also
shrinks. Seconds are a property of this Xeon; that count is a property of the algorithm.

3-comparisons-per-removal

**After this PR:** `SegmentGrid` keeps `std::unordered_map<const SegmentInfo*, size_t> index_of`
alongside `all_segments`. `add` records the position, `remove` looks it up, erases the entry, does
the same swap-and-pop, and repairs the position of the one element the swap moved. Same container
state, same order semantics, one hash lookup instead of a scan.

2-speedup-vs-n

**Proof:** equivalence was checked before any timing, by driving both containers with the same
random add/remove sequence and comparing the full contents as a multiset and the count after every
single operation:

```
equivalence: 70264 adds, 66598 removals, 131 removals of an absent element, no divergence over 200000 steps
```

The absent-element removals are in there because `remove` must stay a no-op for a segment the grid
does not hold, which is the one behaviour the early `return` guarantees today.

**What the index costs, stated rather than hidden.** The indexed version is not O(N) overall, it is
N^1.38, and the reason is visible in the data. Its cost per removal goes from 32 ns at N = 5,000 to
251 ns at N = 1,000,000: a hash lookup is a random access, and as the table grows it stops fitting
in cache. Fitting the two halves of the range separately: while the index is smaller than this
machine's 16 MiB L3 the exponent is **N^1.32**, and above it **N^1.52**, which is the shaded region
in the first figure. The current code has the opposite property, a sequential scan over a contiguous
vector, which is slow but cache-friendly. At the sizes a real grid actually has, the index stays
well inside L3, which is why the speedups in the table above are the ones that matter rather than
the 460x at a million.

## Details

Change: one member added to `SegmentGrid` in
`volume-cartographer/core/include/vc/core/util/normalgridtools.hpp`, and its maintenance in `add`
and `remove` in `volume-cartographer/core/src/normalgridtools.cpp`. Sixteen lines added, four
removed.

One behavioural note, because it is the only difference I could find. If the same segment were ever
added twice to one `SegmentGrid`, the old code would remove the first copy and the new code removes
the last one added. The current caller never does that (`align_and_filter_segments` adds each
segment once to `all` and once to `unused`, which are different grids, and moves it to `assigned`
only after removing it from `unused`), and the old behaviour with duplicates was not specified
anywhere. If you would rather it be impossible, `add` could assert on a duplicate key; I left that
out because an assert in a hot path is your call and not mine.

This was the file's only `std::find`; `std::sort`, `std::remove_if` and `std::max` are still there,
and the file gets all of them transitively rather than through its own `#include <algorithm>`, so
the include list needed only the `<unordered_map>` the new member requires.

One thing I could not do: I did not build volume-cartographer. OpenCV is not installed on the
machine I measured on, so the timings come from standalone programs that reproduce the container and
its access pattern, not from your binary. The segment counts are not in that category: those are
read directly from your published `.grid` files, every slice of both scrolls. The per-plane totals
combine the two: measured sizes, and a measured cost curve interpolated in log-log between its
points. Slices smaller than the smallest measured size are 11.9 per cent of the count and 0.017 per
cent of the quadratic weight, so how they are extrapolated cannot move a total.

Not addressed here, on purpose, so this stays a change nobody has to think about:

- `SegmentGrid::nearest_neighbors` scans a full square per ring instead of the ring
  (`:488-489` against the filter at `:496-506`), so it is cubic rather than quadratic in the radius.
  With `grid_step = 64` and twenty neighbours on a dense grid the radius stays at one or two, so I
  measured it, found it to be tens of cells, and left it alone.
- `align_and_extract_umbilicus` has no seed: line 51 is
  `static std::mt19937 gen(std::random_device{}())` and `SegmentGrid`'s own generator (line 428) is
  seeded the same way, so two runs on the same grid return two different axes.
  `core/test/test_normalgridtools.cpp` already records the consequence: "The full RANSAC happy path
  is too slow / non-deterministic to assert in a unit test." Happy to raise it separately.

## Reproduction

Three programs, all attached:

- `count_segments.py`, with `gridstore.py`, counts what a `SegmentGrid` holds, from published
  `.grid` files; it writes one row per slice, so you can check any of them.
- `per_scroll.py` turns those sizes and the timing CSV into the per-plane totals and the last two
  figures. No number in this page is typed by hand.
- `bench_removal.cpp` produces the timing table and the figures' data. It runs both strategies over
  the same N range in one process, medians three repetitions, and prints one CSV line per N.
- `equivalence.cpp` produces the equivalence block above, and asserts the contents after every
  operation rather than only at the end.

```sh
g++ -std=c++20 -O2 -o bench_removal bench_removal.cpp
taskset -c 0 ./bench_removal 5000 10000 20000 22000 45000 50000 77000 100000 172000 200000 500000 1000000

g++ -std=c++20 -O2 -o equivalence equivalence.cpp && ./equivalence
```

Written with an AI coding assistant; the measurements, the reproduction and the review are mine.

[gridstore.py](https://github.com/user-attachments/files/32245254/gridstore.py)
[equivalence.cpp](https://github.com/user-attachments/files/32245256/equivalence.cpp)
[count_segments.py](https://github.com/user-attachments/files/32245257/count_segments.py)
[bench_removal.cpp](https://github.com/user-attachments/files/32245258/bench_removal.cpp)


evilaliv3:
@hendrikschilling @pmh47: do you mind giving this a look?

hendrikschilling:
That function is currently not in active use anywhere. And the supposedly improved functionality was not even run end to end outside a microbenchmark. So without any active uses this speedup of unused code does not in my opinion warrant the extra code complexity even if minor. Reopen if you need this for some reason.

PR #1801, opened 2026-09-15
VC3D: keep grown Open Data patches selected after a restart

### LINKED ISSUE #1466: Open Data projects: patches grown into the project vanish from it on restart — the reopen selects another volume's catalog folder (data intact on disk, entry persisted)
> **Update 2026-09-24:** the root cause below is wrong. As CVasilopoulos showed ([comment](https://github.com/ScrollPrize/villa/issues/1466#issuecomment-5807954858)), the grow does persist a `growpatch` segments entry (since #1121); the patches vanish because the reopen moves the selection to another volume's catalog folder. Reproduced on `33b0b91b7` ([comment](https://github.com/ScrollPrize/villa/issues/1466#issuecomment-5808161370)). Fix proposed in #1801.

## Motivation / how this arose

Grew seed patches on the streamed 45.532 µm PHercParis4 volume inside an Open Data project (the zero-download seeding workflow). In the next session the project opened with the catalog segments only — the patches I had grown were nowhere in the app.

## Measured timeline (agent bridge, main @ aab644c9c; each step from actual runs)

**Session A** — `catalog.open_sample PHercParis4` → `segmentation.grow_patch_from_seed {seed …, volumeId "20260310170716-45.532um-…"}` → job succeeds, `outputDir: …/remote_cache/open_data/segments/PHercParis4/patches`, and `segments.list` shows the new `auto_grown_20260815213206017` (loaded, activatable, editable — the whole editing loop works on it).

**Session B (fresh)** — same `catalog.open_sample`, same volume selected:
```
segments.list → 81 entries, all catalog-published; auto_grown_* absent
```
The patch is fully intact on disk (`x/y/z.tif`, `meta.json`, `generations.tif`, plus a save-backup), and the catalog window still counts it (the "manual" segments column reads the disk via `manualOpenDataSegmentCount`) — but the opened project offers no way to reach it: it is not in the project's segment folders, so it can neither be listed nor activated.

**Session B, after growing a *new* patch** — the completion re-attaches the patches folder, which resurrects the old one too, with a twist:
```
segments.list (before grow) → 81 total, auto_grown = []
segmentation.grow_patch_from_seed …  → succeeded
segments.list (after grow)  → 2 total: [auto_grown_20260815213206017, auto_grown_20260815214031521]
```
i.e. the visible segment folder flips to `patches/` (now the 81 catalog segments are the ones missing from the list). So previously grown patches only reappear as a side effect of growing another patch — or via a manual `segments.attach` with the hand-built cache path.

## Root cause

- When a grow-patch job completes, the patches directory is attached to the session's volume package, but **no persistent entry is written to the project**. The persisted Open Data project (`…/open_data/projects/PHercParis4.volpkg.json`) contains only the catalog aggregate roots in its `segments` array — nothing for `…/PHercParis4/patches` (verified by reading the file after the sessions above).
- `VolumePkg::availableSegmentPaths()` (`core/src/VolumePkg.cpp:2346`) is exactly those persisted entries, so on a fresh open neither the folder list (GUI) nor `segments.list` (bridge) can reach the patches root. (Bridge measured; the GUI shares the same project builder and the same `SegmentationCommandHandler` grow flow whose default output is the patches root — `openDataPatchesRootForVolume`, `SegmentationCommandHandler.cpp:3271` — so the restart behavior should be the same there. Not clicked through interactively.)
- The design elsewhere already solves exactly this: the editable-copy flow persists its folder with `pkg->addSegmentsEntry(editableRoot, {"open-data-editable"})` (`apps/VC3D/segmentation/SegmentationModule.cpp:~168`). The grow-patch completion path has no equivalent.

## Suggested fix

Either (a) on grow-patch completion into an Open Data project, persist the attach the same way the editable-copy flow does (`addSegmentsEntry(openDataPatchesRoot(...), {"open-data-patches"})` or similar), or (b) have the project builder re-attach the sample's `patches/` root whenever it exists and is non-empty — which would also make the catalog window's "manual" count clickable-through instead of a dead-end display. (a) looks like the two-line version with an in-repo precedent.

---
*Found while driving the zero-download seeding workflow on real PHercParis4 data through the agent bridge. Analysis and write-up done with AI assistance (Claude); all listings, job results and file states above are from actual runs, with the one GUI inference explicitly flagged.*

CVasilopoulos:
**In one sentence:** Patches grown from a seed in an Open Data sample are still selected and listed after VC3D restarts, instead of looking deleted.

**One real example:** Starting with the PHercParis4 sample and its streamed volume `20260310170716-45.532um-11.0m-74keV-masked`, I grew a patch from seed `1507.3 1234.3 2640.6` (a vertex of the published segment `20260623144957-w053-058`), restarted VC3D and opened the sample again, and `segments.list` showed the grown `auto_grown_20260915131749849` patch with the `patches` folder still selected.

**Before:** After a restart, the catalog open lands on the sample's preferred volume `20260411134726`, and the segment selection moves to that volume's catalog folder. Selecting the grow volume then picks its catalog folder too, so `segments.list` shows only the 81 catalog segments and no grown patch. The patch is still on disk, and the project file still lists the `patches` folder, which is what #1466 reports.

**After this PR:** The restart opens on the volume the patch was grown on, with the `patches` folder still selected, so the grown patches are listed and editable right away. The catalog folders stay in the segment folder list.

**Proof:** Upstream main `4b3c72882` against this branch on the same base, with the same sample, volume, seed and generation count, and a fresh remote cache for each build. VC3D ran offscreen and was driven through the agent bridge. The attached images show the evidence script's terminal output. The lines starting with `[A]` or `[B]` are printed by the bridge driver. Look at the `volume right after open`, `output_segments` and `segments.list after open` lines in session B.

1466-before-main

1466-after-fix

1466-doctests

1. Session A, same on both builds: open the sample, select the 45.532um volume, grow

Before (main):
```
[A] segments.list after open: 81 segments, auto_grown: []
[A] grow job: state=succeeded message=/src/build/ci-release-gcc-e2e/bin/vc_grow_seg_from_seed completed successfully
[A] segments.list after grow: 1 segments, auto_grown: ['auto_grown_20260915131307506']
[A]   entry location=<remote_cache>/open_data/segments/PHercParis4/patches tags=['growpatch']
[A]   output_segments=<remote_cache>/open_data/segments/PHercParis4/patches
```
After (this PR):
```
[A] segments.list after open: 81 segments, auto_grown: []
[A] grow job: state=succeeded message=/src/build/ci-release-gcc-e2e/bin/vc_grow_seg_from_seed completed successfully
[A] segments.list after grow: 1 segments, auto_grown: ['auto_grown_20260915131749849']
[A]   entry location=<remote_cache>/open_data/segments/PHercParis4/patches tags=['growpatch', 'open-data-patches', 'vc-open-data-target-volume-id:20260310170716', 'vc-open-data-source-coordinate-level:0', 'vc-open-data-coordinate-space:PHercParis4/20260310170716@L0']
[A]   output_segments=<remote_cache>/open_data/segments/PHercParis4/patches
```

2. Session B: restart VC3D, open the same sample and select the same volume

Before (main):
```
[B]   entry location=<remote_cache>/open_data/segments/PHercParis4/patches tags=['growpatch']
[B]   output_segments=<remote_cache>/open_data/segments/PHercParis4/patches
[B] volume right after open: 20260411134726-2.400um-0.2m-78keV-masked.zarr-6ff766d971e6d4b6
[B]   output_segments=<remote_cache>/open_data/segments/PHercParis4/20260411134726
[B] current volume: 20260310170716-45.532um-11.0m-74keV-masked.zarr-3b3dfe0ca0476050
[B] segments.list after open: 81 segments, auto_grown: []
[B]   output_segments=<remote_cache>/open_data/segments/PHercParis4/20260310170716
[B] on disk at end: patches dir exists, entries: ['auto_grown_20260915131307506']
```
After (this PR):
```
[B]   output_segments=<remote_cache>/open_data/segments/PHercParis4/patches
[B] volume right after open: 20260310170716-45.532um-11.0m-74keV-masked.zarr-3b3dfe0ca0476050
[B]   output_segments=<remote_cache>/open_data/segments/PHercParis4/patches
[B] current volume: 20260310170716-45.532um-11.0m-74keV-masked.zarr-3b3dfe0ca0476050
[B] segments.list after open: 1 segments, auto_grown: ['auto_grown_20260915131749849']
[B]   output_segments=<remote_cache>/open_data/segments/PHercParis4/patches
[B] on disk at end: patches dir exists, entries: ['auto_grown_20260915131749849']
```

3. Doctests, `test_open_data_manifest`
```
-- BEFORE
patches root test cases in the source: 0
34 test case(s) passed
-- AFTER
patches root test cases in the source: 2
36 test case(s) passed
```

I picked this up because growing patches from seeds is the zero-download way to start segments on Open Data, and losing them on every restart makes that workflow hard to trust. I reproduced it on the streamed PHercParis4 volume before changing anything, saw that the patch was still saved but lost its selection on reopen, and checked that after the fix a restart keeps the patch selected and listed.

Growing patches from seeds is the zero-download way to start new segments on Open Data samples. Today a restart makes that work look lost, and the only way to get it back is to grow another patch or to find the `patches` folder in the segment folder list by hand. With this change, people and agents that use the agent bridge continue where they stopped. The same reopen rule applies to editable copies of catalog segments, which lost their selection in the same way.

### Root cause

The issue suggests that the grow never writes a project entry. That part is not the cause: `startGrowPatchFromSeedImpl` already calls `addSegmentsEntry(..., {"growpatch"})` and `setOutputSegments`, and both runs above show the entry and the selection in `PHercParis4.volpkg.json` before the restart. The file only lost the selection during the reopen:

1. `createOpenDataSampleProject` reports the volume that sources the most catalog segments as the preferred volume. `refreshVolumeSelectionUi` gives that explicit preference priority over the remembered volume, so the reopen lands on `20260411134726`.
2. `CWindow::updateOpenDataSegmentTransformState` switches the selected segments folder to the catalog folder that matches the loaded volume. A selected editable copy is kept, but the `patches` entry has no Open Data routing tags, so it is replaced.
3. When the 45.532um volume is selected, the selection is already a catalog folder, so its catalog folder wins. `segments.list` only reports the selected folder.

I confirmed this order with temporary logging in `updateOpenDataSegmentTransformState` (not part of this PR).

### What changed

- `OpenDataSegmentCache.cpp/.hpp`: `attachOpenDataPatchesRoot` adds or updates the patches entry with `growpatch`, `open-data-patches` and the routing tags (`vc-open-data-target-volume-id`, `-source-coordinate-level`, `-coordinate-space`, `-source-volume-id`) of the immutable catalog folder that shares the grow volume's coordinate space. The routing tags are replaced on every grow, the same way `attachEditableOpenDataSegmentRoot` reconciles them. Both helpers now share one list of routing tag prefixes.
- `SegmentationCommandHandler.cpp`: when the output folder is the sample's patches root, the grow attaches it with `attachOpenDataPatchesRoot`. Other output folders keep the plain `growpatch` entry.
- `CWindow.cpp`: `isSelectedEditableOpenDataSegmentsEntry` is renamed `isSelectedWritableOpenDataSegmentsEntry` and also accepts `open-data-patches`, so a selected patches root is kept for its own volume.
- `OpenDataSampleProject.cpp`: when the cached project's selected folder is a local `open-data-patches` or `open-data-editable` entry whose target volume is in the sample and attached, that volume becomes the preferred volume of the open.
- `core/test/test_open_data_manifest.cpp`: two doctests.
  - Patches tags: a plain `growpatch` entry gets the tags of the vol1 catalog folder, they persist through save and load together with the selection, a grow on vol2 replaces them, and a volume without a coordinate space leaves only `growpatch` and `open-data-patches`.
  - Reopen volume: without a selection the open prefers the volume with the most segments, with a selected patches root routed to the other volume the reopen prefers that volume and keeps the selection, and a target volume that is not in the sample falls back to the default.

### Tested

- Base: upstream main `4b3c72882`, the same commit for both builds.
- Linux x86_64 only, in `ghcr.io/scrollprize/villa/volume-cartographer:builder-ubuntu-26.04`, cmake preset `ci-release-gcc` with `-DVC_TESTING=ON`. Not tested on macOS or Windows.
- VC3D ran with `QT_QPA_PLATFORM=offscreen` and `--agent-bridge-name`, driven by a small script built on `agent_bridge/test/bridge_client.py` and `vc3d_process.py`. It calls `catalog.list_samples` with `refresh`, `catalog.open_sample`, `volume.select`, `segmentation.grow_patch_from_seed` with `iterations` 20 and `minAreaCm` 0, and `segments.list`, and reads the cached `open_data/projects/PHercParis4.volpkg.json`. Each session ends with SIGTERM because the bridge has no quit method; the grow's project entry is written when the grow starts, so this does not affect the result.
- Sample data streamed from the Open Data bucket: about 1.4 GB per fresh cache.
- Doctests: `test_open_data_manifest`.

### Limitations

- Patches for all volumes of a sample share one `patches` folder, so the routing tags follow the most recent grow. Patches grown earlier on another volume are in that volume's coordinates but reopen with the latest grow volume.
- Projects saved before this change still have a plain `growpatch` entry. The next grow into the patches folder updates it; until then the folder can be chosen in the segment folder list.
- The editable-copy part of the reopen rule is covered by the same code path but was not run on real data here.
- A seed grow into a folder other than the sample's patches root keeps today's behaviour.

### Prior work and credit

- @Bullo27 reported #1466 with a measured timeline on PHercParis4 through the agent bridge, and this reproduction follows it. The suggested `open-data-patches` tag comes from that report. The measurements above correct its root cause: the entry was persisted, and the selection was lost during the reopen.
- #1647 added editable copies of catalog segments with routing tags and the selected-editable preference that this PR extends to patches.
- #1121 added the grow-time `growpatch` entry.
- No other PR or branch addresses this issue.

Fixes #1466


CVasilopoulos:
#1466 was closed as fixed on main yesterday. The half it was filed for — the grow never persisting a segments entry — is indeed on main, and has been since #1121 on 8 July. The half this PR is about is not: after a restart the entry is still in the project file and the patch is still on disk, but the selection moves to a catalog folder and `segments.list` does not show the patch. The closing re-check used PHerc0490B, which has no catalog segments, so it could not reach that path.

Rebased onto `78877eaf3` (no conflicts, `903a5a0fc` → `351ec80af`, nothing else changed) and re-ran the same reproduction on that base, both builds from the project's builder container, VC3D offscreen through the agent bridge, a fresh remote cache per session.

main `78877eaf3`, session B after the restart:

```
[B]   entry location=<remote_cache>/open_data/segments/PHercParis4/patches tags=['growpatch']
[B] volume right after open: 20260411134726-2.400um-0.2m-78keV-masked
[B] current volume: 20260310170716-45.532um-11.0m-74keV-masked
[B] segments.list after open: 81 segments, auto_grown: []
```

this branch, same base, same seed:

```
[B]   entry location=<remote_cache>/open_data/segments/PHercParis4/patches tags=['growpatch', 'open-data-patches', 'vc-open-data-target-volume-id:20260310170716', ...]
[B] volume right after open: 20260310170716-45.532um-11.0m-74keV-masked
[B] segments.list after open: 1 segments, auto_grown: ['auto_grown_20260924044248078']
```

`test_open_data_manifest`: 34 cases on main, 36 on the branch, all passing. The proof section in the description is the original run on `4b3c72882` and is unchanged; the numbers here are from the re-run on current main.

PR #1807, opened 2026-09-16
vc_render_tifxyz: warn when --cache-gb is not below the memory available to the process

nerln:
Refs #1404 (the `--cache-gb` half). Prior work: #1415 addressed both halves of the issue (a clamp plus a torn-TIFF check) and was closed by the repository time limit without a human review; the torn-TIFF half is now in #1695. This PR only adds the warning the issue asks for on the `--cache-gb` side. Nothing is clamped and no default changes.

`--cache-gb` (default 16) at or above the memory the process can use made the render sit at the memory ceiling with no output and no log line, indistinguishable from a slow render (the report in #1404 is a Docker host with a 15 GB limit). The tool now prints one line to stderr when the requested cache is not below the usable memory: the cgroup limit when one is set (`memory.max` or `memory.limit_in_bytes`), otherwise physical RAM. On Windows there is no check.

Reproduced on main (`1f544fa71`) with a 3-slice render of a PHerc0125 segment, on a 16 GB machine:

```
$ vc_render_tifxyz ... --cache-gb 999          # before: renders, no warning
$ vc_render_tifxyz ... --cache-gb 999          # after
Warning: --cache-gb 999 is not below the memory available to this process (16.0 GB): the render can stall without any output. Lower --cache-gb.
```

@hendrikschilling, you merged #1528 on this file: could you take a look when you have a moment?

The patch was prepared with Claude Code under my direction; I built it and ran the reproduction above locally.


nerln:
Status: still applies cleanly on current `main` (`c49028494`, checked with `git merge-tree` after #1695 landed). With #1695 merged, the torn-TIFF half of #1404 is in; this is the remaining `--cache-gb` half, 24 lines, no clamp. Happy to change the wording of the warning or drop the cgroup lookup if either is unwanted.

hendrikschilling:
I would accept this as is given the noted config below is a pretty special case - but if that could be fixed as well it would be appreciated!
```
  [P2] The cgroup lookup can miss the process’s actual memory limit.
  vc_render_tifxyz.cpp:395 (https://github.com/ScrollPrize/villa/blob/17b4411ef3a4a02449f36b690833d78d894aae59/volume-cartographer/apps/src/vc_render_tifxyz.cpp#L395) checks only two fixed
  root paths. This works for common container layouts, but not processes in nested cgroups, such as systemd-limited jobs or containers exposing the host cgroup hierarchy.

  For example, on a 128-GiB host with a 15-GiB process limit, the default 16-GiB cache could still produce no warning. Resolve the process’s cgroup through /proc/self/cgroup and its mount,
  then consider applicable ancestor limits.
```

nerln:
Fixed in 9c7f4bf39: the limit now comes from the process's own cgroup, resolved through `/proc/self/cgroup` and `/proc/self/mountinfo` (v2, and the v1 memory controller), walking up to the top of the mount and keeping the lowest limit. The two fixed paths stay as a fallback and the warning text is unchanged.

In a Linux container limited to 1 GiB and run with `--cgroupns=host`, the previous code read 7.65 GiB, all the RAM the container could see, and the new one reads 1.00 GiB. A job with no limit of its own under a parent limited to 512 MiB now reads 0.50 GiB instead of the container's 2.00 GiB. A plain container reads 1.00 GiB with both, and on macOS nothing changes.

PR #1810, opened 2026-09-16
VC3D line annotation: kollesis termination tag on fiber end control points

skyward7187:
**In one sentence:** Mark the control point where a fiber ends at a kollesis, so the end is recorded in the fiber file, visible in the line annotation views and the Fiber Map, and protected from being extended past by accident.

**One real example:** Starting with the PHerc0139 fibers open in VC3D, Ctrl+right-click the last control point of a fiber that stops at a sheet join and choose "Kollesis termination": the point becomes a hollow yellow ring, the fiber JSON gains `"tags": ["kollesis_termination"]` on that control point, the Fiber Map shows the ring at that point after Update, and a click past the end in the current cut is refused (red marker). _(GUI run and screenshots pending from David; the automated verification is listed under Details.)_

**Before:** There was no way to record that a fiber genuinely ends at a kollesis rather than simply not having been traced further; nothing distinguished the two on screen or in the data, and nothing stopped a stray click in the extrapolated tail from extending a finished fiber.

**After this PR:** The control point context menu has a checkable "Kollesis termination" item, enabled on the first or last control point. A tagged point draws as a hollow ring in the control-point yellow, one step larger than a filled point; a tagged point that is also linked keeps the link colour inside the yellow ring. The overview bar and the Fiber Map (every fiber, not only the selected one) show the same ring. While an end is tagged, no control point can be placed beyond it (click, `/` and `0` keys, and the intersection-inspection panes), and a merge whose join side is tagged is refused with a message. Untag to extend or merge; the toggle is refused while a solve is running or pending. The extrapolated tail still draws past a tagged end.

**Proof:** _(screenshots to be added by David: a tagged unlinked end, a tagged linked end, the Fiber Map ring, and the red blocked marker past a tagged end)_

**Why / where this is useful:** Annotators tracing fibers on PHerc0139 and PHerc1451 need to say "this fiber ends here at a kollesis" so the fiber network and later unrolling can treat the end as real rather than as a gap to fill, and so nobody extends it by mistake.

## Details

**Data format.** Version-3 `vc3d_fiber` control point objects gain an optional `tags` array of strings, written only when non-empty, so untagged fibers are byte-identical to before. Tags belong to the point, not its span: they survive a click that collapses or replaces the point (union of the collapsed points' tags), split, reverse, merge of two fibers (union), optimizer round trips, the stored/session conversions, the single-point reopen path, and a toggle made from another pane while a solve runs. Deleting the point deletes the tag.

**Readers updated** (all rejected unknown control-point fields before): the shared C++ parser `core/include/vc/fiber_tracer/FiberJson.hpp` (atlas, fiber-trace CLI, inspect tools), the VC3D loader in `LineAnnotationFiberSegments.cpp`, python `fiber_merge.is_fiber_doc` (gates vc_sync's three-way merge and hfsync publishing), and the shared python reader `vesuvius/src/vc3d_fiber_format` (neural tracing, lasagna atlas, spiral-fitting). The `vc_lasagna_line_probe --reopt` writer carries tags over. The python v3 span merge previously compared a chunk's terminal control point by position only, which would have silently dropped a tag on a fiber's last point; it now compares terminal tags, so a tag survives a separated remote span change and conflicts with a concurrent refit of the same span.

**Rules.** The menu item is enabled only on the first or last control point by line position (a single point counts), except that an interior tag found in an edited file can still be removed. Placement beyond a tagged end is refused both in the dialog's placement gate and in the controller's `handleGeneratedControlPoint`, which the intersection-inspection clicks reach directly. Transient split/link candidate colours outrank the ring.

**Fiber Map.** Per-point flags travel snapshot -> `InputFiber` -> `PlacedFiber`; they are part of the inputs and result digests (so the memoization check stays exact) but not of the geometry cache key (no geometry reads them). Unlinked tagged points draw as yellow rings on every fiber; linked ones reuse the link endpoint dot with a yellow ring.

**Known limitations.** Older VC3D builds and older vc_sync checkouts refuse to load a tagged fiber. Two panes editing the same fiber concurrently remain last-save-wins for tags as for every other field. Cross-source deduplication fingerprints geometry only, as it already does for fiber-level tags.

**Verification.**
- Full `cmake --build build` (RelWithDebInfo) succeeds.
- `ctest` for the line annotation, fiber map, fiber trace and chunked viewer suites: 18 of 18 pass, including new doctest cases (per-point tag codec, normalisation, rejection of malformed shapes, collapse union, optimizer round trip, reverse, endpoint detection, placement gate) and a new layout test (flag pass-through, cache reuse, digests, mismatched-vector handling).
- `pytest scripts/tests/test_fiber_merge.py`: 87 of 87, with four new tests (schema, final-point tag survives a separated remote change, interior tag survives, same-span refit conflicts).
- The shared python reader was checked on a real PHerc0139 fiber with a tag added, and with malformed tags.
- Six rounds of three independent Codex (gpt-6-astra) reviews (diff-only, full-context, data-lifecycle specialist) until they converged; the fixes they drove are folded into the commits.

https://claude.ai/code/session_01RgmjRqrVfdAvDMkamkhazk


hendrikschilling:
```
[P2] Seed-only reopening reads a potentially invalidated iterator.
  volume-cartographer/apps/VC3D/LineAnnotationController.cpp:2987 reads it->controlPoints after ensureDatasetForSession() and launchSession(). Those calls can process dialog events and    
  finalize/save existing editors, potentially reloading or reallocating _fibers.

  The new tag read can therefore access freed storage, causing a crash or incorrect tags. Previously this seed-only path copied its required data before those calls.

  Minimal fix: copy the seed’s tags alongside seedPoint, before either call. This finding is from code inspection, not a reproduced GUI crash.
```

skyward7187:
Fixed in 9654de047: the seed-only reopen path now copies the seed control point's tags right after `seedPoint`, before `ensureDatasetForSession` and `launchSession`, and passes the copy to `handleLineSeed`. No read of `it` remains after those calls. Build and the line annotation / fiber map test suites pass.

Issue #1764, opened 2026-09-11
`koine_machines.inference`: no MPS branch, so on Apple Silicon the published `ink_9um` checkpoints run on the CPU with the GPU idle and nothing is printed


SurgeFok:
**In one sentence:** `koine_machines.inference` has no MPS branch, so Apple Silicon silently runs the published `ink_9um` checkpoints on the CPU while the GPU sits idle.

**I was trying to:** Run the published `ink_9um` checkpoint through the documented First Letters inference workflow on an Apple Silicon Mac, then shorten the turnaround between repeat inference runs.

**Using:** Current `merge-ink-pipelines`, torch 2.14.0, an Apple M5 Pro with 24 GB unified memory, a 512x512x20 uint8 zarr input, and `python -m koine_machines.inference.infer input.zarr step-075000.pth out.tif --batch-size 1 --no-compile`.

**What happened:** Inference completed with correct output, but selected the CPU without printing a warning; the MPS GPU stayed idle.

**What I expected or needed:** Automatic MPS selection on Apple Silicon, an explicit device log, and a way to request a device without changing the CUDA default for other users.

**Evidence / reproduction:** The source path and measured CPU/MPS comparison are below, together with a tested patch.

- [x] I personally encountered or reproduced this using the version and data stated above.

## Details

Both ink-detection inference entrypoints resolve the device with CUDA or nothing.
`koine_machines/inference/infer.py`, in `prepare_model_for_inference`:

```python
device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```

`infer_full3d_tifxyz.py` has the same line in `prepare_model`. Autocast is gated the same
way, in both files:

```python
autocast_enabled = bool(amp and device.type == "cuda")
...
torch.autocast(device_type="cuda", enabled=True, dtype=amp_dtype)
```

So on any Mac both entrypoints take the CPU path and disable autocast. The run succeeds and
the output is correct, so there is nothing to notice.

## Measured on hardware

Apple M5 Pro, 24 GB unified memory, macOS 26.5.1, torch 2.14.0, checkpoint
`hybrid_3d2d-seed42/step-075000.pth` from `scrollprize/ink_9um`, through the documented CLI:

Five timed runs per device after a discarded warm-up, 49 blocks per run, block/s:

| device | min | median | max | mean | stdev | log line |
| --- | --- | --- | --- | --- | --- | --- |
[...]

PR #1812, opened 2026-09-16
ink-detection: run on Apple Silicon (MPS) instead of falling back to CPU

AndreasHad04:
**In one sentence:** Ink detection inference runs on the Apple Silicon GPU instead of silently falling back to CPU, which makes it 7.63x faster and drops the memory one tile needs from 9.5 GB to 2.6 GB.

**One real example:** Starting with PHerc. 1667 segment `20240304141531-w013_20240304141531_flatboi`, the 2.399 um surface volume read straight from `s3://vesuvius-challenge-open-data/`, I ran `ink_canonical_2um` at `TILE_SIZE=256 STRIDE=128` over layers 25 to 87 on an M1 Max, and it produced a 4096x4096 prediction, 961 tiles, in **404 s** at a sustained 2.38 tiles/s, matching villa's published prediction for those pixels at Pearson **0.9685** (best negative control 0.056).

**Before:** `entrypoint.py` picks `cuda if torch.cuda.is_available() else cpu`, so an Apple Silicon machine takes the CPU path: **3.26 s/tile** and about **9.5 GB per tile** (9.52 GB at batch 1, 19.89 GB at batch 2). On a 16 GB Mac one tile barely fits and two do not. That same 961-tile crop would take over 50 minutes. Separately, `amp_device = "cuda" if device.type == "cuda" else "cpu"` means that when the tensors are on MPS the autocast block does nothing at all, with no error and no warning.

**After this PR:** MPS is selected when present and the autocast device type follows the device the tensors are on. **0.427 s/tile, 7.63x faster**, 2.6 GB per tile.

**Proof:** Nine timed repetitions per MPS configuration and three for CPU, after an untimed warmup, reported min/median/max per `AGENTS.md` 1.4. Same checkpoint, same fixed tiles, batch 1, seconds per tile.

| path | min | median | max |
|---|---|---|---|
| CPU fp32, what Apple Silicon gets today | 3.259 | **3.261** | 3.278 |
| MPS, autocast off | 0.524 | **0.525** | 0.526 |
| MPS, `amp_device` as it is today | 0.525 | **0.526** | 0.527 |
| MPS, `amp_device` corrected | 0.426 | **0.427** | 0.428 |

Rows two and three are the point: today's autocast line times the same as autocast switched off, because on MPS it is a no-op. Confirmed numerically on the same input as well: with `device_type="cpu"` the output is bit-identical to fp32, max absolute difference exactly `0.0`; with `device_type="mps"` it differs by `1.24e-03`.

prediction vs reference

Left is villa's published prediction for those pixels, middle is the same pixels through this patch on MPS, right is the absolute difference.

**Why / where this is useful:** `AGENTS.md` 1.5 lists macOS on arm64 as a target platform, and today that platform gets the slowest path in the repo by a factor of 7.63 at a memory cost most Macs cannot meet. Anyone with an Apple laptop can run ink inference on real segments locally instead of only in a container on a rented GPU.

## Details

**The change.** A new `device_utils.py` plus six changed lines. `select_device()` (cuda, then mps, then cpu, with an `INK_DEVICE` override), `amp_device_type(device)` (the device's own type, falling back to `cpu` for anything unrecognised), and `sync`/`empty_cache` so timing and cache clearing are not cuda-only.

**The cuda path is unchanged.** `amp_device_type(torch.device("cuda"))` is still `"cuda"` and `select_device()` still prefers cuda whenever it is available. Stated plainly: I have no CUDA hardware, so that is verified by unit test and by construction, never by execution. 12 checks, 0 failures (https://github.com/AndreasHad04/villa-apple-silicon, `bench/verify_port.py`), one of which is a positive control that watches the current expression returning `"cpu"` for an MPS device, so if this is fixed elsewhere the control fails loudly rather than passing vacuously.

**Does the patch change what the model computes? No.** Same input, CPU against MPS, single forward: max absolute difference **9.54e-07**. On a different input the same comparison gives **0.165**, so the agreement test discriminates by five orders of magnitude rather than being trivially satisfied.

**On numerical guarantees (`AGENTS.md` 1.3).** Correcting the autocast device type does change MPS numerics, because autocast starts working where it previously did nothing. Measured end to end on a real 1024x1024 crop, 49 tiles, the same pipeline run twice with only that expression differing:

| | |
|---|---|
| wall clock, current vs corrected | 25.37 s vs 20.84 s (1.217x) |
| Pearson between the two outputs | **0.9999995** |
| mean / p99 / max absolute difference | 1.37e-04 / 1.78e-03 / 3.17e-03 |
| pixels whose ink decision at 0.5 flips | **0.014%** |

If you would rather take the device fix without the precision change, the two are separable and I am happy to split the PR: `select_device()` alone changes no numerics and is still 6.21x.

**Against your published prediction: r = 0.99, once START_LAYER is right, and that turned out to be a finding of its own.** Comparing to `...new_canon_autoresearch_recipe-tile256-stride128.tif` for the same pixels, the agreement depends almost entirely on the depth window, not on the device:

| scroll, segment | r at START_LAYER **1** (what the inference README suggests) | best START_LAYER | r there |
|---|---|---|---|
| PHerc. 1667, 20240304141531 | 0.8521 | 25 | **0.9851** |
| PHerc. 0139, 20250108000000 | 0.9054 | 23 | **0.9900** |
| PHercParis4, 20230702185753 | 0.6240 | 24 | **0.9698** |
| PHerc. 0814, 20250925161630 | 0.8263 | 21 | **0.9005** |

Four scrolls, each with its own published canonical prediction, ink-rich window chosen automatically from the reference rather than by hand, every offset also scored against a shuffled copy of that reference (floor never above 0.0017). All four volumes are 109 layers, so the centred window is 23. **START_LAYER=1 is the worst setting on all four.** I have not touched the README in this PR; that belongs in its own change and I am happy to open it.

Your own `vesuvius/docs/ink_detection.md` explains it: for a 2.399 um OME-Zarr it says to "select the centered 84 Z planes", and it records that "Labels occupy Z slice 32 of a 65-plane volume", which is the exact centre. So the writing surface is in the middle of these volumes and a 62-plane window has to be centred on it, `(109-62)//2 = 23`. With `START_LAYER=1` the surface lands about 85% of the way through the window instead. I have opened that as its own documentation PR (#1813) rather than mixing it into this one.

**Your newer code already does this correctly, which is part of why I think this is worth taking.** `scrollprize/hecate`, released 2026-09-15 and built from the same `ink_canonical_2um`, writes its autocast as `torch.autocast(device_type=device.type, ...)`, which is exactly what this PR changes `optimized_inference` to do. I ran hecate at 2.4 um on a real PHerc. 1667 crop on this Mac: `--device mps` works out of the box, **5.76x faster than `--device cpu`** (29.3 s and 29.1 s against 167.7 s), with **13 pixels of 1048576 differing by one quantisation step** and MPS deterministic run to run. Its one remaining gap is the same one this PR fixes: `--device` defaults to `'cuda' if torch.cuda.is_available() else 'cpu'`, so a Mac silently takes the slow path. That is in hecate's own repo rather than villa, so I have not touched it, but the fix is the same line.

**A separate finding, not patched here.** On arm64 CPU, `torch.autocast(device_type="cpu")` routes the 3D convolutions into a bfloat16 reference kernel. One tile: **3.47 s** in fp32 against **714 s and still running when I killed it** under autocast, so at least **205.8x**, a lower bound and not a completed measurement. Attributed by sampling rather than assumed: `BFloat16` symbols sit above `at::native::slow_conv3d_forward_out_cpu` and `cpublas_gemm_impl`. In practice a 49-tile crop that takes 21 s on MPS did not finish in 33 minutes on CPU, which reads like a hang. I left it alone because whether to disable autocast on CPU, gate it on architecture, or keep it is your call, and changing it would affect every existing CPU user.

**Also worth knowing:** `entrypoint.py` assumes `/workspace` and a credentialed `boto3` client, so there is no way to run this outside the container. The linked repo has `run_local.py`, which reads a crop anonymously from the open-data bucket and calls villa's own `run_inference` unmodified. Happy to upstream it separately; I kept it out of here to keep the diff small.

**Environment.** M1 Max, 64 GB, macOS 26.6, torch 2.14.0, python 3.11.15, `scrollprize/ink_canonical_2um` (`r152_3ddec_v2_l5_epoch13.ckpt`).

**Benchmarks, controls and raw result JSON:** https://github.com/AndreasHad04/villa-apple-silicon


AndreasHad04:
Two follow-ups since opening this, one of them a limit of my own patch.

**A minimum torch version, which the PR does not currently state.** On
torch 2.8.0 `torch.backends.mps.is_available()` is True but the model's 3D
max-pool raises

    NotImplementedError: The operator 'aten::max_pool3d_with_indices' is not
    currently implemented for the MPS device

so with this patch applied a 2.8.0 user gets a hard failure where they used to
get a slow CPU run. On torch 2.14.0 the same call returns normally. I have not
bisected which release closed it, so the honest statement is "fails on 2.8.0,
works on 2.14.0". If you would prefer the device selection to fall back to CPU
when the MPS build cannot run the model rather than raising, say so and I will
add that; I did not want to guess which behaviour you want.

**The memory numbers are now regenerated rather than transcribed.** They were
measured once and typed into the README, which meant nothing checked them. They
are now produced by a script that runs each configuration in its own process
(`ru_maxrss` is a process-wide high-water mark, so a single-process sweep
reports the maximum of everything before it) and written to a JSON that the
repo's `verify_claims.py` asserts against every quoted figure:

    CPU  batch 1   9.53 GB      forward RSS the pass adds
    CPU  batch 2  19.89 GB
    MPS  batch 1   2.59 GB      Metal driver allocation
    MPS  batch 4   7.85 GB

Two corrections fall out of that, both against my own earlier text: the ratio at
batch 1 is 3.68x and not the ~6x I first wrote, and the speedup is 7.63x on
medians, not the 7.85x a mean gave. The CPU and MPS columns are different
quantities and the README now says so rather than implying one measurement.

Repo, MIT: https://github.com/AndreasHad04/villa-apple-silicon

AndreasHad04:
Two follow-ups: a gap in my own validation that I have now closed, and a cross reference to work on the sibling inference path that a reviewer should see next to this.

**The gap. Everything measured in this PR used one model type, and it is not the default.** `optimized_inference` dispatches four (`entrypoint.py:717`): `timesformer`, `resnet3d-50`, `resnet3d-152` and `resnet3d-152-3d-decoder`, and `InferenceInputs.model_type` defaults to `timesformer`. Every number in this PR came from `resnet3d-152-3d-decoder`. A device patch can fail in exactly two ways, a missing MPS kernel or a numeric divergence, and neither depends on trained weights, so I ran each architecture with random weights on both devices with the same input. torch 2.14.0, MPS available true, 64 px input, one tile.

| `MODEL_TYPE` | frames | MPS fp32 vs CPU, max abs diff | MPS autocast vs CPU, max abs diff |
|---|---:|---:|---:|
| `timesformer` (the default) | 26 | 1e-06 | 0.000566 |
| `resnet3d-50` | 30 | 0 | 0.000143 |
| `resnet3d-152` | 30 | 1e-06 | 0.000705 |
| `resnet3d-152-3d-decoder` | 62 | 1e-06 | not finite |

**All four run on MPS with no missing kernel, in fp32 and under the `torch.autocast(device_type=device.type)` expression this PR introduces.** In fp32 every model agrees with the CPU to 1e-06 or exactly.

**The one cell that is not a pass, reported as such.** `resnet3d-152-3d-decoder` under autocast with RANDOM weights returns a non finite output, so there is no difference to quote and I have marked it undecided rather than OK. Untrained weights make activations large and float16 overflows. The control that separates "untrained weights overflow" from "this patch is broken" is to repeat it with the real checkpoint, which is the only weights anyone runs: `r152_3ddec_v2_l5_epoch13.ckpt` on MPS gives finite output on both paths with max abs diff 0.001018. So it is a float16 range artifact of the fixture, not a defect. It is also consistent with the end to end run already in this PR, where the two paths agree at Pearson 0.999999511858424 over 49 real tiles with 0.0138% of pixels crossing the 0.5 decision.

**And an honest note about the environment those earlier numbers were taken in.** villa's `optimized_inference/requirements.txt` lists `pytorch-lightning`, and my venv did not have it, because the 3d decoder module is the one model file that does not import it. So at the time I measured this PR, two of the four model modules could not even be imported in my environment and I had not noticed. That is why this check did not happen sooner.

**The cross reference.** @SurgeFok filed #1764 on 2026-09-11 and #1770 on 2026-09-12, which is the same defect class in `koine_machines.inference`: `prepare_model_for_inference` and `prepare_model` both resolve the device as CUDA or CPU and gate autocast the same way, so Apple Silicon silently takes the CPU path there too, with measured 5.97x MPS over CPU on an M5 Pro. That is earlier than this PR and I did not know about it when I opened this one.

The two do not overlap and neither replaces the other: #1770 touches `koine_machines` and zero files under `ink-detection/`, and this PR touches `ink-detection/optimized_inference` only. A maintainer looking at Apple Silicon support probably wants both, and the fix in each is the same shape, so whichever is reviewed first sets the pattern for the other.

Raw numbers and the script are in `results/mps_model_coverage.json` and `bench/mps_model_coverage.py` at https://github.com/AndreasHad04/villa-apple-silicon , MIT.

Issue #1765, opened 2026-09-11
optimized_inference README: the recommended 62-layer window for resnet3d-152-3d-decoder is off-centre on the published 2 um volumes


kadenpool:
**In one sentence:** the README's recommended window for `resnet3d-152-3d-decoder` (`START_LAYER=1`, `END_LAYER=63`) is centred only on a 65-layer surface volume; on the published 109-layer 2 um volumes it does not reproduce the published predictions (pixel match 0.23-0.62 over 9 text windows in 3 segments, against 0.82-0.92 for a window centred on the surface).

**I was trying to:** run the challenge's 2 µm ink model on scans myself, so before relying on it I tested it against a map the team had already published.

**Using:** the canonical checkpoint `scrollprize/ink_canonical_2um` (`r152_3ddec_v2_l5_epoch13.ckpt`), `TILE_SIZE=256`, on the published 109-layer surface volumes of PHerc0139 (2.399 um), compared with the published `new_canon` prediction of the same segment. Inference ran through a single-file copy of `ink-detection/optimized_inference` (same model code, preprocessing and tiling, without the S3 and zarr plumbing; the folder is unchanged on `main` since 25 Aug). On a test crop its output is pixel-identical to the repo's own code (0 of 13,440 pixels differ). Kaggle T4.

**What happened:** with the README's window the output does not match the published prediction; with a window centred on the surface it does, on every window tried.

Segment 20250108000000-w025, one 3120 px window of dense text and one of blank papyrus:

| window | layers | match with published (pixel correlation) | ink area (>0.5) | published ink area |
|---|---|---|---|---|
| dense text | 1-62 (README) | 0.380 | 40.6% | 84.8% |
| dense text | 24-85 (centred) | 0.875 | 85.0% | 84.8% |
| blank papyrus | 1-62 (README) | 0.138 | 3.6% | 0.2% |
| blank papyrus | 24-85 (centred) | 0.628 | 0.1% | 0.2% |

12 more 2048 px windows (random positions, fully inside the segment; text = published ink 10-60%, blank = under 0.5%) over segments w025, w026 and w027, same two windows compared:

| window | published ink | README 1-62: match / ink | centred 24-85: match / ink |
|---|---|---|---|
| text w025 y15501 x14695 | 16.4% | 0.36 / 6.3% | 0.84 / 16.7% |
| text w025 y19820 x23169 | 48.9% | 0.62 / 33.2% | 0.89 / 50.3% |
[...]

PR #1813, opened 2026-09-16
ink-detection: centre the 62-layer window, START_LAYER=1 does not reproduce the published predictions

AndreasHad04:
**In one sentence:** The `START_LAYER=1` this README suggests does not reproduce the ink predictions you publish, and centring the 62-layer window in the surface volume does.

**One real example:** Starting with PHerc. 1667 segment `20240304141531-w013_20240304141531_flatboi`, the 2.399 um surface volume from `s3://vesuvius-challenge-open-data/`, I ran `ink_canonical_2um` at `TILE_SIZE=256 STRIDE=128` and swept `START_LAYER`, and agreement with your own published prediction for those pixels went from **r = 0.8521 at START_LAYER 1** to **r = 0.9851 at START_LAYER 25**.

**Before:** The README says "a 62-layer window such as `START_LAYER=1`, `END_LAYER=63`". Following it, my output disagreed with your published map badly enough that I first assumed I had the wrong checkpoint. I did not: the `ink_canonical_2um` model card gives its recipe as `new_canon_autoresearch_recipe`, which is exactly the recipe in the published prediction filenames.

**After this PR:** The README says to centre the window, `START_LAYER=23`, `END_LAYER=85` for the 109-layer surface volumes in the bucket, and states what that is worth.

**Proof:** 15 measurements across 4 scrolls, each with its own published `new_canon_autoresearch_recipe` prediction. The comparison window on each was chosen automatically as a high-variance region of your own reference, not by hand. **The best window beats `START_LAYER=1` in 15 of 15**, best START_LAYER runs 21 to 25 with 11 of 15 landing exactly on 24, and agreement at the documented setting ranges from **0.1591 to 0.9522** against 0.8525 to 0.99 at the best window. The four rows below are one representative segment per scroll.

| scroll, segment | r at START_LAYER 1 | best START_LAYER | r there |
|---|---|---|---|
| PHerc. 1667, 20240304141531 | 0.8521 | 25 | **0.9851** |
| PHerc. 0139, 20250108000000 | 0.9054 | 23 | **0.9900** |
| PHercParis4, 20230702185753 | 0.6240 | 24 | **0.9698** |
| PHerc. 0814, 20250925161630 | 0.8263 | 21 | **0.9005** |

agreement against START_LAYER

The same 4096 pixels at both settings, so the cost is visible and not only numerical. Top row the full crop, bottom row the same detail from each:

what the documented window costs

Every offset on every scroll was also scored against a **shuffled** copy of that scroll's own reference. That floor never exceeds **0.0022**, so this is depth alignment and not marginal statistics. All four volumes are 109 layers, so the centred window is `(109-62)//2 = 23`, and the measured optima are 21 to 25.

Corroborated by a measure that never sees your reference: on PHerc. 1667 the ink separation of my own output (mean of pixels above 0.5 minus mean of those below) peaks at 0.7210 at START_LAYER 28, against 0.6643 at START_LAYER 1. So the window that matches your production output is also the one that maximises ink contrast.

**Why / where this is useful:** Anyone following this README today gets output that agrees with your published predictions at r 0.62 to 0.91 instead of 0.90 to 0.99, and has no way to tell, because nothing errors. It also costs real ink contrast, so a contributor evaluating a new idea against a weak baseline may reach the wrong conclusion about their own change.

## Details

**Why, and this is your own convention everywhere else.** `START_LAYER=1` in this one document is the outlier:

- `scrollprize/hecate`, released 2026-09-15 and built from `ink_canonical_2um`, describes its own inference script as "The script **selects the central input depth**, slides over XY with half-patch overlap, and blends probabilities using a floored Hann window", and its `--reverse` flag "reverse[s] the **full render depth before selecting the central planes**".
- The same card says why the centre is the reference: a surface volume is "a CT scan **resampled around a mesh** that follows a papyrus sheet", and "The intended sheet can also **wander above and below the centre of the render**".
- `vesuvius/docs/ink_detection.md`, for a 2.399 um OME-Zarr: "**select the centered 84 Z planes**", and "**Labels occupy Z slice 32 of a 65-plane volume**", the exact centre.
- `scrollprize/ink_9um`: "the z window **jitters over 17 of the 21 slices** so the models don't lock onto one exact depth".

For a 109-plane volume the centred window is `(109-62)//2 = 23`. With `START_LAYER=1` the surface at plane 54 lands at position 53 of 62, about 85% of the way through the window. The wandering sheet also explains why the measured optima scatter over 21 to 25 rather than sitting exactly on 23.

**A wrong turn I want to record, because it is an easy mistake.** I first tried to locate the surface from the voxels, using in-plane gradient energy and mean intensity, and their asymmetry led me to think the volume was not centred. That test cannot work: carbon ink on carbonised papyrus has almost no attenuation contrast, so intensity statistics describe bulk papyrus and say nothing about where the writing surface is. The profile is in the linked repo as a measurement of the wrong quantity.

**Stability.** 15 measurements over 4 scrolls and multiple regions per segment, all optima within 21 to 25, median 24.

**Scope.** I searched the repo for every other place this value is set. `START_LAYER` and `END_LAYER` appear only in this README and in `entrypoint.py`, which requires both as environment variables with no default (`os.environ["START_LAYER"]`). So there is no coded default to correct and this is a documentation-only change. If you would rather have a guard in code, I am happy to add one that warns when the requested window is not centred in the surface volume, but that seemed like scope creep for a docs fix.

**Honest limits.** One crop per scroll, 1024x1024, 49 tiles. PHerc. 0814 agrees less well than the others even at its best (0.9005), which I have not explained. The sweep is at stride 1 to 3 in START_LAYER near the optimum and coarser far from it. I did not change the `END_LAYER` in the developer smoke-test snippet lower down, because that block is `STEP=aggregate-profiling` and the values there are placeholders.

**Reproduce it:** `bench/window_generalise.py` and `bench/offset_vs_reference.py` at https://github.com/AndreasHad04/villa-apple-silicon, along with the raw per-offset JSON for all four scrolls and the model-free depth profile.

This is independent of my other PR (#1812, Apple Silicon support) and can be taken on its own.


AndreasHad04:
Credit where it is due, and a correction to this PR.

After filing this I found that the same change was already proposed in #1766 (11 Sep), and that flummoxjr had published the centred window with r = 0.9999997 against the published map on 3 Sep, in `bench/w059_c0/RESULTS.md` of https://github.com/flummoxjr/measure-before-you-hunt

I reached the same conclusion independently and did not know about either. That does not change who was first. #1766 has priority, it is the one that should be considered, and I am happy for this PR to be closed in favour of it. If anything in the evidence here is useful there, please take it freely; the repo it comes from is MIT.

One result here looks additive rather than duplicate, so I will move it to its own issue rather than leave it buried in a redundant PR.

At the documented window the model does not merely lose agreement, it asserts confident ink on input whose depth ordering has been destroyed. Same crop, same checkpoint, same 62 channels, PHerc. 1667 segment `20240304141531`, crop y=20480 x=16384 size 1024:

| window | real input, fraction of pixels above 0.5 | depth-shuffled null, fraction above 0.5 | null r against the real prediction |
|---|---|---|---|
| `START_LAYER=1` | 0.3323 | **0.0732** | **-0.1753** |
| `START_LAYER=25` | 0.4138 | **0.0000** | -0.0691 |

At the centred window the other five nulls also return nothing: `single_layer_repeat` (62 copies of one plane, perfect in-plane texture and no depth information) 0.0000, `per_column_shuffle` 0.0000, `voxel_shuffle` 0.0000, `phase_scramble` 0.0112, `depth_reverse` 0.0131.

So the nulls are a cheap self-check for this failure: if a depth-destroying null still returns confident ink, suspect the window before suspecting the scan or the checkpoint. That needs no reference map, which matters on a segment where none has been published.

Honest limit on that null: the depth permutation is global, the same order for every column, so each plane keeps its in-plane texture. `single_layer_repeat` is the test that closes that gap and it is clean at the centred window, but I did not run it at `START_LAYER=1`.

AndreasHad04:
A correction to my comment above, and the measurement I said I had not run.

**The correction.** I wrote that at the centred window "the other five nulls
also return nothing" and listed five. The battery has six others, and the one I
left out is the only one that does not return nothing: `depth_roll_half` gives
**0.1429** above 0.5 at `START_LAYER=25`. That omission was mine and the full
table is below.

**The measurement.** I had said `single_layer_repeat` was the test that closes
the gap in `depth_shuffle`, whose permutation is global so every plane keeps its
in-plane texture, and that I had not run it at `START_LAYER=1`. I have now run
the whole battery at both windows on the same crop, PHerc. 1667 segment
`20240304141531`, y=20480 x=16384 size 1024, 62 channels:

| input | START_LAYER=1 frac > 0.5 | r vs real | START_LAYER=25 frac > 0.5 | r vs real |
|---|---|---|---|---|
| real | 0.3323 | | 0.4138 | |
| `depth_roll_half` | **0.1512** | +0.7360 | **0.1429** | +0.4234 |
| `depth_shuffle` | **0.0732** | -0.1753 | 0.0000 | -0.0691 |
| `phase_scramble` | 0.0084 | +0.1318 | 0.0112 | +0.2444 |
| `depth_reverse` | 0.0000 | -0.1775 | 0.0131 | +0.3126 |
| `single_layer_repeat` | **0.0000** | +0.0814 | 0.0000 | +0.0078 |
| `per_column_shuffle` | **0.0000** | -0.0240 | 0.0000 | +0.0306 |
| `voxel_shuffle` | 0.0000 | +0.0093 | 0.0000 | +0.0064 |

So the gap is closed in the direction that makes the original claim stronger,
not weaker. 62 copies of one plane carry perfect in-plane texture and no depth
information, and they return **zero** confident ink at both windows, as does
`per_column_shuffle`. The 7.3% at `START_LAYER=1` therefore is not a single
layer texture reader.

The two nulls that do produce ink are exactly the two that leave each plane
intact and only move the stack in depth. Read plainly: this model needs plane
level content and it needs depth variety, and at the documented window it does
not need the depth ORDER to be right. At the centred window shuffled order
returns zero, so correct window inference is the configuration that actually
demands correct depth ordering.

**What I am not claiming.** `depth_roll_half` is not a destroyed input, it is a
differently wrong window, so its surviving 0.1429 is not evidence the model
hallucinates. It is the reason the battery needs to be reported whole: a reader
picking the five clean rows would conclude something the sixth does not support.

Numbers from `results/inkdiag_nulls1.json` and `results/inkdiag_nulls25.json` in
https://github.com/AndreasHad04/villa-apple-silicon , MIT, free to lift into
#1766 if useful there.

kadenpool:
Thanks for this, and for the credit. Your PHerc1667 result is a useful second scroll for the same fix, and the depth-shuffled null is a cheap self-check I had not thought of: it needs no reference map, which is exactly the situation on an unread segment. If #1766 goes in, I would be glad to point to your nulls from it.

Measured with Claude Code under my direction.

kadenpool:
@AndreasHad04, an update from #1766 that I should have posted here too. #1766 has been a draft since 11 Sep, and only because I keep to three ready PRs at a time, not because anything in it is unfinished. Please don't hold this PR for it. If a maintainer prefers this one, which measures the change on four scrolls, I will close #1766; both replace the same README line, so only one can merge.

AndreasHad04:
Two things, one about which PR should carry this and one measurement that answers the half layer question in #1765.

**On #1766 being a draft.** That changes my position, so I am stating the new one plainly rather than leaving the old one standing. My comment above said I was happy for this PR to be closed in favour of #1766. I wrote that believing #1766 was the merge candidate. A draft cannot be merged, so as things stand this PR is the only one of the two that a maintainer can act on today. I am therefore withdrawing the offer to close it and leaving it open and ready. If #1766 comes out of draft I have no objection to it being the one that merges and this one being closed; the point is the README line, not whose PR fixes it.

Priority is unchanged and I want it on the record in this comment rather than only in an earlier one: flummoxjr published the centred window with r = 0.9999997 on 2026-09-03, @kadenpool filed #1765 and #1766 on 2026-09-11, and this PR is from 2026-09-16 and was reached independently without knowledge of either.

**On 23 against 24, and first a correction to my own evidence.** #1765 now records that 23 beats 24 on all 8 text windows there, median match 0.905 against 0.874. I went back to check what my own sweeps actually said about that, and they said nothing: of the 15 sweep grids behind this PR, **14 contained `START_LAYER=24` and not 23, and the one that contained 23 did not contain 24**. So "best `START_LAYER` 21 to 25" in the PR body never separated the two, and any impression that my data preferred 24 was a property of the grid I chose, not a measurement.

So I ran the paired comparison: the same 15 crops, 4 scrolls, each crop scored at 23 and at 24 and nothing else changed. 1024 px crops, `TILE_SIZE=256`, `STRIDE=128`, `resnet3d-152-3d-decoder`, `r152_3ddec_v2_l5_epoch13.ckpt`, scored as Pearson against the published `new_canon_autoresearch_recipe` map of the same segment.

| scroll | crop y,x | r at `START_LAYER=23` | r at `START_LAYER=24` | 23 minus 24 |
|---|---|---:|---:|---:|
| PHerc0139 | 0, 15910 | 0.9900 | 0.9863 | +0.0037 |
| PHerc0139 | 3432, 23954 | 0.9520 | 0.9479 | +0.0041 |
| PHerc0139 | 3492, 3432 | 0.8220 | 0.8525 | -0.0305 |
| PHerc0814 | 6591, 20828 | 0.8095 | 0.8708 | -0.0613 |
| PHerc0814 | 14292, 18224 | 0.9059 | 0.8865 | +0.0194 |
| PHerc0814 | 14903, 92184 | 0.9256 | 0.9239 | +0.0017 |
| PHerc1667 | 5072, 9822 | 0.9725 | 0.9731 | -0.0006 |
| PHerc1667 | 10338, 48536 | 0.9714 | 0.9731 | -0.0017 |
| PHerc1667 | 15312, 38001 | 0.9668 | 0.9714 | -0.0046 |
| PHerc1667 | 20288, 0 | 0.9540 | 0.9601 | -0.0061 |
| PHerc1667 | 20480, 16384 | 0.9836 | 0.9815 | +0.0021 |
| PHerc1667 | 35504, 49110 | 0.9418 | 0.8886 | +0.0532 |
| PHercParis4 | 11646, 18694 | 0.9876 | 0.9831 | +0.0045 |
| PHercParis4 | 16928, 24064 | 0.9876 | 0.9803 | +0.0073 |
| PHercParis4 | 43379, 22110 | 0.9682 | 0.9698 | -0.0016 |

**23 wins 8, 24 wins 7, no ties. Exact two sided sign test p = 1.0000.** Median paired difference +0.0017, mean -0.0007, range -0.0613 to +0.0532. Worst correlation against a shuffled copy of the reference across every one of these runs: 0.0012. Every scroll produces both signs.

**The scale is the part I would act on.** Over the same 15 crops, median r is 0.7474 at the README's `START_LAYER=1` and 0.9698 at the better of 23 and 24. So the window fix is worth +0.2224 in median r and the half layer is worth +0.0017 at p = 1.0000, a factor of about 131. Crop to crop, r ranges 0.8095 to 0.9900, a spread of 0.1805, against a largest single crop half layer effect of 0.0613.

**Where this disagrees with #1765, stated as a disagreement rather than smoothed over.** On PHerc0139, the scroll #1765 measured, my three crops split 2 to 1 for 23 (-0.0305, +0.0037, +0.0041), so I do not reproduce 8 of 8. Two differences could explain it and I have not tested either: crop size, 1024 px here against 2048 px there, and crop selection, the single most ink rich window per segment here against random positions filtered on published ink fraction there. 8 of 8 is itself unlikely under a coin flip, so I am not calling that result noise; I am saying it does not hold when the same comparison is run across four scrolls.

**What I think follows for the README, and it is the cheap conclusion.** Both PRs already write 23, which is what `(109 - 62) // 2` gives. On this evidence there is no reason to change that and no reason for either PR to argue the half layer, because the floor formula is already defensible and the remaining difference is not resolvable at this crop size. If anything is worth adding to the README it is the rule rather than the constant, which #1766 already does in its Tips line.

Raw per crop numbers are in `results/halflayer.json` and the 15 per crop files `results/window_generalise_hl_*.json` in https://github.com/AndreasHad04/villa-apple-silicon , MIT, free to lift into #1766.

One more correction while I am here: one row of my published `results/window_tally.json` carried the string `20240304141531-w013 (offset sweep)` in its `seg` field, which is a label and not a segment id, so that row could not be re-run programmatically. It is now the real id, `20240304141531-w013_20240304141531_flatboi`, with the previous file kept beside it.

kadenpool:
@AndreasHad04, thank you for the paired test and the raw numbers, and for putting the priority on the record. Your test is better evidence on the half layer than mine: 15 crops on four scrolls, 8 to 7, against my 8 windows on one scroll. I have added it to #1765 and #1766 beside my result: 23 is what the floor formula gives, and across four scrolls the half layer makes no measurable difference. On which PR carries the README line, my note above stands: if a maintainer prefers #1813, I will close #1766.

AndreasHad04:
Thanks for adding it to #1765 and #1766. That is the outcome I wanted from posting the raw numbers, and I have nothing further on the half layer: 8 to 7 at p = 1.0000 across four scrolls is not a result either of us should build a README line on, and both PRs already write the 23 that `(109 - 62) // 2` gives. From my side the question is closed and the floor formula stands on its own.

Which PR carries the README line is a maintainer's call and I am not arguing it further in either direction. #1813 is open and ready; if #1766 comes out of draft and a maintainer prefers it, closing this one costs nothing.

One note for anyone arriving at this thread later: #1812 is independent of both. It is the device selection in `ink-detection/optimized_inference`, it touches no README line, and it can merge on its own schedule.

PR #1816, opened 2026-09-17
prepare_9um_isotropic_input: retry transient tile reads so a remote run survives a dropped connection

mayu-meijo:
**In one sentence:** `prepare_9um_isotropic_input` can now finish against a remote surface volume with parallel workers, instead of losing the whole run to one dropped connection.

**One real example:** Starting with the public 2.399 µm volume for PHercParis4/20231016151002 on the open-data bucket, I ran the command with `--workers 6`, and it produced a complete `(21, 7990, 12990)` output in 22m21s.

**Before:** `--workers 6` died after 31 seconds with `aiohttp.client_exceptions.ServerDisconnectedError`. Tile reads have no retry, so `pool.map` propagates the first failure and stops the run. Output is published only after all 416 tiles finish, so a late failure discards everything and leaves a `.partial` that the command then refuses to overwrite. The only way through was `--workers 1`, at 62m38s.

**After this PR:** Tile reads retry up to 5 times with exponential backoff. Only the read is retried; pooling and writing are local and still fail immediately.

**Proof:**

| run | workers | result | time |
|---|---|---|---|
| before | 6 | ServerDisconnectedError | 31s |
| before | 1 | completed | 62m38s |
| after | 6 | completed | 22m21s |

The `--workers 6` output is byte-identical to the `--workers 1` output — compared plane by plane with `np.array_equal`, all 21 planes equal — so the retry does not change results. It only makes the parallel path usable, which is 2.8x faster than the serial workaround.
<details>
<summary>Before: --workers 6 fails after 31s</summary>

```
aiohttp.client_exceptions.ServerDisconnectedError: Server disconnected

real    0m31.409s
```

</details>
<details>
<summary>After: --workers 6 completes</summary>

```
tiles=416/416
wrote .../surface-volume.zarr shape=(21, 7990, 12990)

real    22m21.736s
```

</details>

**Why / where this is useful:** The docs state the input can be a URL, and anyone using the released `ink_9um` checkpoints has to build their input this way. This is a remote-first code path with no retry, so a single dropped connection can discard an hour of work.

### Motivation

I was trying to run the released 9 µm ink model on PHerc. Paris 4. The tutorial segment is only 65 planes deep, so `prepare_9um_isotropic_input` stopped with `Cannot take 84 centered planes from 65`. The docs say the input can be a URL, so I pointed it at the public 2.399 µm volume on the open-data bucket instead. That is where I hit this error.

### Command

```
uv run --extra models python -m \
  vesuvius.ink_detection.preprocessing.prepare_9um_isotropic_input \
  https://vesuvius-challenge-open-data.s3.amazonaws.com/PHercParis4/segments/20231016151002/surface-volumes/2.4um-0.22m-78keV-volume-20260411134726.zarr \
  <out>/surface-volume.zarr --level 2 --workers 6
```

### Environment

Ubuntu 24.04 on WSL2, RTX 3070, uv 0.12.13, CPython 3.14.7.

### Related

#1698 and #1712 add the same kind of retry to inference-side remote reads. This covers the preprocessing path.


hendrikschilling:
Please be a bit more selective in which exception get retries.

`It retries all exceptions inside reading/conversion, including permanent decoding errors and allocation failures, not just dropped connections. Those failures now take longer to  report.`

Issue #1611, opened 2026-08-26
vc_render_tifxyz: remote chunk streaming stalls indefinitely with no error


khj1222:
> **Update, 2026-08-30 — one claim in this report is wrong and is struck.** @Bullo27 identified
> the binary this was run on: `ghcr.io/scrollprize/villa/volume-cartographer:edge` is the image
> built from `1e3f4c021f4e53bea3867772ed05f51a7e586a9c`, created 2026-05-13, digest
> `sha256:bad516f66001abca759454cc43e4fd11e5b19aa55d36bdc2043817291c8083c4` — confirmed against
> my local image. **That revision does retry**: three attempts with a 30 s transfer cap that
> `ZarrChunkFetcher` raises to 60 s. My "no retry" wording came from reading
> `merge-ink-pipelines`, whose cache subsystem this image does not contain, and the title has
> been edited accordingly. What stands is the observation and the mechanism it points at: an
> untimed `cv_.wait` with no attempt cap (`1e3f4c021:render/ChunkCache.cpp:218` and `:749`)
> parks the caller before any retry can be attempted. Also worth stating for anyone reproducing:
> **`:edge` is the only published runtime tag and it has been the same May image since**, so a
> fresh pull reproduces on the same three-and-a-half-month-old binary. Discussion:
> https://github.com/ScrollPrize/villa/issues/1611#issuecomment-5466851805

**In one sentence:** When `--remote-url` streaming stalls mid-render, `vc_render_tifxyz` parks
forever on a chunk instead of failing, so a render that is 20% done simply stops with no error.

**I was trying to:** Render a PHerc1447 auto-grown segment into a ~9 µm surface volume so I
could run the released `ink_9um` ink-detection checkpoints on it. PHerc1447 has 15 segments but
none of them ship a rendered surface volume, so the render is the first step to getting any ink
prediction on that scroll at all.

**Using:** `ghcr.io/scrollprize/villa/volume-cartographer:edge`, digest
`sha256:bad516f66001abca759454cc43e4fd11e5b19aa55d36bdc2043817291c8083c4`, pulled 2026-08-26.
Real data: segment `20250703034159-auto_grown_20250703034159599` (7.40 cm², the largest on that
scroll) against `PHerc1447/volumes/20250521151220-8.640um-1.2m-116keV-masked.zarr` streamed from
the public bucket.

```bash
docker run --rm -v /data/first_letters:/work \
[...]

PR #1817, opened 2026-09-17
vc_render_tifxyz: fail with an error instead of a core dump when a chunk fetch fails (fixes #1809)

### LINKED ISSUE #1809: vc_render_tifxyz: a failed remote chunk fetch aborts with SIGABRT (rc 134) instead of a clean error
### What happens

`vc_render_tifxyz` on a remote volume (`--remote-url`), when a chunk fetch keeps failing after its retries are exhausted, does not exit with an error: it aborts with `SIGABRT` (shell `rc 134`) and a core dump. The last thing printed is a C++ terminate trace, e.g.:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  HTTP 0 fetching 0/41/31/17: Operation timed out after 60002 milliseconds with 60 out of 2097152 bytes received
```

This is the crash that came out of the discussion in #1611, split off here so #1611 (the original "stalls indefinitely with no error" report, which no longer reproduces on current `main`) can be closed.

### Why

`ChunkSampler::updateChunk` rethrows a failed chunk fetch as `std::runtime_error`. The renderer's tile loop in `renderTiles` (and the sampling loops in `core/src/Slicing.cpp`) are OpenMP parallel regions, and an exception cannot leave an OpenMP structured block — so the throw reaches `std::terminate` instead of unwinding to a handler. The result is `SIGABRT` + core dump rather than a clean error and non-zero exit.

### Repro (measured)

Built `vc_render_tifxyz` from `main` at `757f70c0` (Release, gcc 13.3, Ubuntu 24.04). A local HTTP mirror of a real `vol.zarr` serves chunk bodies one byte at a time (metadata full speed) so the connection never idles but no chunk transfer ever completes — the byte-per-second drip is a stand-in for a server that trickles data, which the per-read idle timeout does not catch.

```
vc_render_tifxyz -v cache --remote-url http://127.0.0.1:PORT/vol.zarr \
  -s <segment> --scale 1 -g 0 -n 1 --cache-gb 1 --tif-output out/render
```

Observed on `main` `757f70c0`: the whole-transfer cap fires at ~61 s per attempt (`curl`: `Operation timed out ... with 60 out of 2097152 bytes received`); the fetch is retried, and after the retries are exhausted (four ~61 s attempts, ~4 minutes) the process aborts with `rc 134` and a core dump. In this run it aborted 243 s after start (13:52:49Z start, 13:56:52Z crash) — four ~61 s attempts on the same chunk, then `terminate called after throwing an instance of 'std::runtime_error'` and `the monitored command dumped core`, `rc 134`.

### Suggested fix

Catch the exception at each OpenMP region boundary (keep the first failure, skip the remaining iterations, rethrow once the region has joined), and have `main` print `Error: <what()>` and return `EXIT_FAILURE`. I have this implemented and measured against the drip mirror and a 503 mirror; happy to open it as a PR against this issue.

---
*AI-assisted (Claude), human-directed account; the repro above was run on my machine on the stated build.*

Bullo27:
**From the author:** I contribute to VC through an AI-assisted workflow (Claude, directed by me). This is the fix for #1809, the crash I split off from #1611. On main, a remote render whose chunk fetch fails after its retries ends in a core dump. With this branch it prints which chunk failed and why, and exits with code 1. Renders that do not fail are byte-identical. I ran the drip-server repro, the 503 repro and the control renders on my machine on this rebased branch and checked the table below against the logs.

Fixes #1809.

**In one sentence:** A remote render whose chunk fetch fails after its retries now stops with one line naming the chunk and the transport error, and exit code 1, instead of a core dump.

**One real example:** Starting with the published PHerc1447 segment `auto_grown_20250703034159599` and a local mirror of its volume (`20250521151220-8.640um-1.2m-116keV-masked.zarr`, real level-0 chunks) served by a small HTTP server that answers chunk requests with the right `Content-Length` but delivers the body at 1 byte/s (the connection is never idle, the transfer never completes: the situation in #1611), I ran `vc_render_tifxyz --remote-url … --tif-output …` on main and on this branch. Main dies with `terminate called recursively … core dumped` (rc 134). This branch prints `Error: HTTP 0 fetching 0/41/31/17: Operation timed out after 60002 milliseconds with 60 out of 2097152 bytes received` and exits with rc 1, 0.02 s after the failure.

**Before:** `ChunkSampler::updateChunk` (core/src/Slicing.cpp) rethrows a fetch error as `std::runtime_error` from inside an OpenMP region, and the renderer's tile loop is an OpenMP region too. An exception cannot leave an OpenMP structured block, so the runtime calls `std::terminate`: every failed remote fetch, after its 4 attempts, ends in `terminate called recursively` / `terminate called after throwing an instance of 'std::runtime_error'`, an abort and a core dump, whatever the output mode. On the Clang and MSVC builds the `#pragma omp` lines are no-ops (`CMakeLists.txt` swaps in a shim; the loops run serially), and the process dies the same way for a simpler reason: nothing between the fetch and `main` catches the exception, so it leaves `main` uncaught (measured by @kadenpool on macOS arm64/libc++ in https://github.com/ScrollPrize/villa/issues/1809#issuecomment-5747531092: `libc++abi: terminating due to uncaught exception`, exit 134).

**After this PR:** the failure is caught on the worker thread, the remaining rows/tiles of that region are skipped, and the exception is rethrown by the calling thread once the region has joined. `main` catches it, flips the fetcher's abort flag (`HttpClient::abortAll()`, so transfers still in flight are not waited for), prints `Error: <what()>` and returns `EXIT_FAILURE`. Renders that do not fail are unchanged: byte-identical TIFF output on the local mirror and on the open-data bucket, identical zarr output on the mirror.

**Proof:** same segment, same mirror, same server (60 s transfer timeout × 4 attempts = the existing `ZarrChunkFetcher` settings; nothing about retries or timeouts changes in this PR). Rows without a platform marker were run by me on Linux on 2026-09-17 (the HTML-503 row on 2026-09-20) with this branch rebased on main `2dcfaf6a0`; "main" is `757f70c01` (the only commits after it are VC3D-only). The **macOS** rows are @kadenpool's independent runs, copied from their comments.

| run | binary | rc | wall | last lines |
|---|---|---|---|---|
| slow drip, `--tif-output` | main | **134** | 483.8 s | `terminate called recursively` / `what(): HTTP 0 fetching 0/41/31/17: Operation timed out after 60002 milliseconds with 60 out of 2097152 bytes received` / core dumped |
| slow drip, `--tif-output` (run twice) | this PR | **1** | 483.7 s, 483.7 s | `Error: HTTP 0 fetching 0/41/31/17: Operation timed out after 60002 milliseconds with 60 out of 2097152 bytes received`; process gone 0.02 s later |
| server answers every chunk request with 503 (4 attempts, backoff), `--tif-output` | main | **134** | 2.9 s | `what(): HTTP 503 fetching 0/41/31/17` / core dumped |
| server answers every chunk request with 503, `--tif-output` | this PR | **1** | 2.5 s | `Error: HTTP 503 fetching 0/41/31/17` |
| same, with `--log-path render_logfile.txt` | this PR | **1** | 2.5 s | log file flushed (#1729's guard), its last line is the `Error:` line |
| control, mirror served whole, `--tif-output` | main / this PR | 0 / 0 | 3.1 s / 1.3 s | `00.tif` md5 `80084d370c148a03b78f974877728477` on both |
| control, mirror served whole, `--zarr-output` | main / this PR | 0 / 0 | 0.9 s / 0.8 s | `render.zarr` trees identical (`diff -r`) except the cache path recorded in `.zattrs` `source_zarr` |
| open-data bucket, PHercParis4 `20260310173927-45.532um-11.0m-110keV-masked.zarr`, real segment, `--scale 1 -g 0 --tif-output` | main / this PR | 0 / 0 | 17.6 s / 10.4 s | `00.tif` md5 `b1803be312d3530523c4244c7c355d89` on both |
| server answers 503 with a 32-line nginx-style HTML page, `--tif-output` | main / this PR @ e85724996 / this PR @ dec671593 | **134** / **1** / **1** | 2.9 s / 3.4 s / 3.4 s | main: `what(): HTTP 503 fetching 0/41/31/17: <html>` + the page, core dumped; e85724996: `Error: HTTP 503 fetching 0/41/31/17: <html>` + 32 lines of HTML; dec671593: `Error: HTTP 503 fetching 0/41/31/17: <html> [...]` (one line) |
| @kadenpool's `fail_server.py` ([posted on #1809](https://github.com/ScrollPrize/villa/issues/1809#issuecomment-5747919168), unmodified: real PHerc0139 2.399 µm metadata, every chunk request fails), run by me on Linux on 2026-09-20 with the PHerc1447 segment, `--mode 503` / `--mode drop`, `--tif-output` | main 757f70c01 / this PR @ dec671593 | **134** / **1** (both modes) | 3.5 s / 3.6 s (503), 1.9 s / 3.4 s (drop) | main: `what(): HTTP 503 fetching 0/34/40/20: <!DOCTYPE HTML>` + Python's error page, core dumped; `what(): HTTP 0 fetching 0/34/40/20: Empty reply from server`, core dumped. dec671593: `Error: HTTP 503 fetching 0/34/40/20: <!DOCTYPE HTML> [...]` (one line); `Error: HTTP 0 fetching 0/34/40/20: Empty reply from server` |
| **macOS arm64, Homebrew clang 23.1, system libc++, no OpenMP** (independent run by @kadenpool, [comment](https://github.com/ScrollPrize/villa/issues/1809#issuecomment-5747531092)): stub server fails every chunk (connection closed without reply / 503), real PHerc0139 2.399 µm metadata, 384×384 tile, level 2, 3 slices | main f07d33be6 / this PR e85724996 | **134** / **1** | | main: `libc++abi: terminating due to uncaught exception of type std::runtime_error: HTTP 0 fetching 2/43/28/30: Empty reply from server`, `Abort trap: 6`; this PR: `Error: HTTP 0 fetching 2/43/28/30: Empty reply from server` (and `Error: HTTP 503 fetching 2/43/28/30: ...` for the 503 stub) |
| **macOS**, one real bucket tile, 5 slices, no failure (same run) | main / this PR | 0 / 0 | | five TIFFs, all five MD5s identical |
| **macOS** `test_slicing` (same run) | this PR | 0 | | 15/15, including the new case; main's copy has 14 |
| `test_slicing` (new test case) | main Slicing.cpp | **134** | | `terminate called after throwing … simulated transport failure` (measured on the pre-rebase build) |
| `test_slicing` (new test case) | this PR | 0 | | 15/15 cases |
| @kadenpool's `fail_server.py --mode 503`, same segment, `--zarr-output` (Linux, 2026-09-20) | main 757f70c01 / this PR @ f095ec88b | **134** / **1** | 1.9 s / 5.0 s | main: core dumped; this PR: `Error: HTTP 503 fetching 0/34/41/21: <!DOCTYPE HTML> [...]`. Both leave `render.zarr` with `.zgroup` and six `.zarray` files, no chunk files and no `.zattrs`; `zarr.open` (zarr 3.2.1) returns the right shape and dtype at every level, all zeros, without raising (first seen by @kadenpool on macOS) |
| **macOS**, @kadenpool's stub, all three output modes: `--tif-output`, `--zarr-output`, composite (`--composite-start/-end/--composite-collapse`), clean `origin/main` `f07d33be6` vs this PR @ e85724996 ([comment](https://github.com/ScrollPrize/villa/issues/1809#issuecomment-5749200449)) | main / this PR | **134** / **1** in each of the three modes | | the "whatever the output mode" claim above is measured, not asserted |
| hermetic regression test `core/test/test_render_fetch_failure.py` (added in 7e1ed38cb, hardened in 3875f6af6, rewritten without third-party packages in 46dd13632; no network, synthetic volume, `HOME` pointed at the temp dir; written by @kadenpool), run as `/usr/bin/python3 test_render_fetch_failure.py <binary>` with the system interpreter (3.12, no numpy, no tifffile) (Linux, 2026-09-20) | main 757f70c01 / this PR @ 46dd13632 | test exit 1 / **0** | 14.3 s / 17.6 s | main: `FAILED (failures=5)`: three cases on `-6 != 1 : exit -6: expected a clean exit 1 (-6 is the abort this PR removes, 2 is the --timeout watchdog, which means it hung instead)`, two on the missing `Error:` line (main prints `what():  HTTP 0 fetching 0/0.1.0` and `0/0/1/0` for the two chunk-key styles); the guard case and the recovery case pass on main as they should. This PR: `Ran 6 tests … OK`, also with `-W error`. @kadenpool: 6/6 pass on 7e1ed38cb and 5/6 fail on main 23c3d759e, macOS, clang, non-OpenMP build, with the Homebrew interpreter that made the previous version skip |
| same test through `ctest -R test_render_fetch_failure` | this PR @ 46dd13632 | 0 | 14.9 s | `Passed`, configured with `-DPython3_EXECUTABLE=/usr/bin/python3`. Only where the apps are built with `VC_TESTING=ON`: the three CI jobs that run ctest configure `-DVC_BUILD_APPS=OFF`, so there `if(TARGET vc_render_tifxyz)` is false and neither this test nor `test_render_tifxyz_logging` is registered (pointed out by @kadenpool); in CI the coverage is the workflow step below |
| **VC3D 23c3d759e**, Actions → Render segment on a remote volume (driven offscreen through the agent bridge, `render.tifxyz`; VC3D runs this tool as a subprocess), stub answers every chunk 503 (Linux, 2026-09-20, [#1846](https://github.com/ScrollPrize/villa/issues/1846#issuecomment-5750868717)) | main tool / this PR tool next to the same VC3D | job `failed` / job `failed` | 4.0 s / 7.5 s | main: message `… vc_render_tifxyz: crashed`, console tail `terminate called recursively` / `what(): HTTP 503 fetching 0/20/22/11`; this PR: message `… failed with exit code: 1`, console tail `Error: HTTP 503 fetching 0/20/22/11`. VC3D itself stays up in both (the viewers stream chunks through `ZarrChunkFetcher`, which does not throw) |

About the wall time of the drip runs: the process ends when the chunk the sampler is blocked on has exhausted its four ~61 s attempts. In every run the server log shows two chunk transfers in flight at a time; when the first pair are chunks the sampler is not waiting for (the band prefetch's), their exhausted fetches leave error entries in the cache (existing behaviour, not touched here) and the sampler's own chunk starts its four attempts only after them, so the run takes ~484 s; when the sampler's chunk is in the first pair it takes ~243 s (my earlier runs on the pre-rebase branch, and the run in #1809). In today's runs main crashed at 483.8 s and this branch exited at 483.7 s; only the ending differs. The server log shows the client closing every connection after 61.0 s with 61 bytes received, i.e. `CURLOPT_TIMEOUT` firing (the measurements in #1611).

The partial output is not removed by this PR: a failed `--tif-output` run leaves a small `00.tif` behind on main and on this branch alike (43-45 KB in my drip runs; 8 bytes on main vs 134 bytes on the branch with @kadenpool's stub, on macOS and on Linux alike, where `tifffile` opens main's without raising and refuses the branch's). A failed `--zarr-output` run is less visible still: the pyramid datasets are created before rendering, so on main and on this branch alike the store is left with its six `.zarray` files and no chunks, and `zarr.open` returns arrays of zeros without any error (last proof row; first observed by @kadenpool on macOS). The store is not removed either: with `--resume` it may already hold chunks from an earlier run, so an error path must not delete it. @kadenpool tested that contract on macOS (https://github.com/ScrollPrize/villa/issues/1809#issuecomment-5749625805): a 512×512 render whose server refuses one chunk region exits 1 and leaves one chunk file, byte-identical to the clean render's; rerunning on that store with `--resume` exits 0, fetches 120 chunks instead of the clean run's 128 and produces a store equal to the clean render. So the leftover is useful, not just harmless, and the test pins that a failed run leaves zero chunk files. Scripts should key on the exit code, which is what this PR makes possible.

**Why / where this is useful:** anyone rendering from the open-data bucket over a connection that stalls or drops (#1611's reporter; @kadenpool hit it twice in three days of ordinary bucket renders from macOS, https://github.com/ScrollPrize/villa/issues/1809#issuecomment-5746619200; batch scripts that key on the exit code) now gets the failing chunk and the curl error on stderr and a normal exit code, instead of a core file and a truncated log. The core change also makes every OpenMP sampling entry point (`readInterpolated3D`, `samplePlane`, `readCompositeFast`, `readMultiSlice`, `sampleTileSlices`) propagate a chunk error to its caller instead of terminating the process, which is what the new unit test checks.

- [x] Verified by running the example and proof above on my machine, on the stated data (AI-assisted; see disclosure below).

## Details

Tested commits: `e85724996` (the fix) and `dec671593` (first-line-only error message, 2026-09-20) on top of `2dcfaf6a0` (main as of 2026-09-17); the table says which rows were run on which. On 2026-09-20 the branch was rebased onto `23c3d759e` (main of that day; none of its new commits touch the files of this PR) as `3359147d2` + `f095ec88b`: `git range-diff` shows both patches unchanged, and on the rebased build `test_slicing` passes 15/15 and the `fail_server.py --mode 503` run above gives the same rc 1 / one-line error. `7e1ed38cb`, `3875f6af6` and `46dd13632` (2026-09-20) add the hermetic test and its ctest/CI registration only; no C++ changes. Ubuntu 24.04, GCC 13.3, CMake 3.28, Release build (`-DVC_BUILD_APPS=ON -DVC_BUILD_PYTHON=OFF -DVC_BUILD_UI_TRACER=OFF -DVC_USE_PCH=OFF`), plus `VC_TESTING=ON` for the test binaries.

**Change** (`git diff -w` is the readable view; the rest is the re-indentation of the loop bodies):

- `core/src/Slicing.cpp`: a small `ParallelFailure` collector (mutex + first `exception_ptr` + atomic flag). Each of the four OpenMP regions (`readVolumeImpl`, `samplePlaneImpl`, `readCompositeFastImpl`, `readMultiSliceImpl`) wraps its row body in `try { … } catch (...) { failure.capture(); }`, rows return early once a failure is recorded, and `failure.rethrow()` runs after the region on the calling thread. `ChunkSampler::updateChunk` itself is unchanged.
- `apps/src/vc_render_tifxyz.cpp`: the same idiom around the tile body of `renderTiles` (`#pragma omp parallel for` over `tx`), rethrown before the band/TIF assembly of that tile row; this is the pattern `FiberTrace.cpp` already uses. `main` wraps `process_one` in `try/catch`, calls `utils::HttpClient::abortAll()` (its documented purpose: make shutdown immediate regardless of the S3 timeout, so the unwinding does not wait for transfers still in flight), prints `\nError: <what()>\n` (the newline because the message otherwise lands on the `\r` progress line) and returns `EXIT_FAILURE`; since `dec671593` only the first line of `what()` is printed, followed by `[...]` when there was more, because an HTTP status error carries the server's response body (a whole HTML page for a real 503; @kadenpool's suggestion); the log file is flushed by #1729's `LogFlusherGuard` on the way out (measured with `--log-path`).
- `core/test/test_slicing.cpp`: `FailingChunkArray`, an 8³ `IChunkedArray` whose `tryGetChunk` always returns `ChunkStatus::Error`, and one test case that drives all five sampling entry points with a 64-row grid (enough rows for several threads to fail at the same time) and expects a `std::runtime_error` carrying the fetch message.
- `core/test/test_render_fetch_failure.py` (written, hardened and then rewritten by @kadenpool on #1809; adopted as posted): standard library only. A synthetic 64×128×128 OME-Zarr volume whose chunk bytes are generated on demand and a flat 32×32 tifxyz surface straddling the chunk boundary at x = 64, the three coordinate planes written as baseline float32 TIFFs with `struct`, served by a local `ThreadingTCPServer` (bound to port 0) that refuses chunk requests on demand. The renderer runs with `HOME` (and `USERPROFILE`) set to the temp dir, because `--remote-url` stages chunks under `homeDirectory()/.VC3D/remote_cache` and reads `~/.VC3D/VC3D.ini` (`core/src/RemoteCacheSettings.cpp`): the earlier versions of the test wrote into the real home directory of whoever ran it, which I confirmed from the mtimes in my own `~/.VC3D/remote_cache`. Six cases: every fetch fails with `--tif-output` (exit code exactly 1: `subprocess` reports SIGABRT as `-6`, and `--timeout` starts a watchdog that `_exit(2)`s after a minute, so `> 0` would also have passed on a hung renderer; the leftover TIFF is parsed tag by tag and must declare no `StripOffsets`/`TileOffsets`), the same with `--zarr-output` (the leftover store must hold `.zarray` metadata and zero chunk files, which is stronger than "reads back as zeros" and is what `--resume` relies on), the first two attempts of every chunk refused (must exit 0 with TIFFs byte-identical to a never-failing server), a failure after the render has started (chunk column 0 served, column 1 refused, so the same chunks fail whether the tile loop runs serially under the clang shim or across threads under gcc; the situation of #1809), the message itself for both chunk-key styles (`dimension_separator` `.` and `/`, the latter being what the published volumes use): `Error: HTTP <status> fetching <chunk>`, no `terminating due to`, exit code 1, and a guard: a render against a server that answers everything must come back with more than one pixel value, or the failure cases could pass without a single fetch. A render that hangs 120 s fails with a message saying so, below ctest's 300 s. Registered in `core/test/CMakeLists.txt` next to `test_render_tifxyz_logging` (TIMEOUT 300).
- `.github/workflows/vc3d-ci.yml`: one step in the Linux CLI compile coverage job, after the logging test: `python3 core/test/test_render_fetch_failure.py build/ci-fast-cli/bin/vc_render_tifxyz` (the deps image ships python3; nothing else is needed now). On this PR's own runs the step took 11 s for 7e1ed38cb (three cases, venv and pip install included), 18 s for 3875f6af6 (five cases) and 15 s for 46dd13632 (six cases, nothing to install). All 28 checks were green on the first two; on 46dd13632 every check passed except the Linux base tests shard, where `test_zarr_chunk_fetcher` (a test this branch does not touch) threw `filesystem error: cannot remove: Directory not empty` from the `fs::remove_all` in the teardown of its "native mirror coalesces inner requests" case; the rerun on the empty commit `27534057d` is green on all 28.

**Reproduction.** The drip server is 40 lines of Python (stdlib only):

<details>
<summary>slowdrip.py</summary>

```python
#!/usr/bin/env python3
"""Slow-drip HTTP server: metadata files (.zattrs/.zgroup/.zarray) are served at full
speed; chunk objects get a correct Content-Length but their body is dripped one byte
every DRIP seconds (DRIP == 0: everything served whole, control mode; DRIP < 0: chunks answered 503), i.e. the connection never idles (per-read timeouts never fire)
but the transfer never completes. Logs every request with wall-clock timing."""
import os, sys, time, threading
from http.server import ThreadingHTTPServer, BaseHTTPRequestHandler
PORT=int(sys.argv[1]); ROOT=sys.argv[2]; DRIP=float(sys.argv[3]); LOG=sys.argv[4]
lock=threading.Lock()
def log(msg):
    with lock:
        with open(LOG,'a') as f: f.write(f"{time.time():.3f} {msg}\n")
class H(BaseHTTPRequestHandler):
    protocol_version='HTTP/1.1'
    def log_message(self,*a): pass
    def _path(self):
        p=self.path.split('?',1)[0]
        if not p.startswith('/vol.zarr/'): return None
        return os.path.join(ROOT,p[len('/vol.zarr/'):])
    def do_HEAD(self): self._serve(head=True)
    def do_GET(self): self._serve(head=False)
    def _serve(self,head):
        fp=self._path(); t0=time.time()
        if fp is None or not os.path.isfile(fp):
            self.send_response(404); self.send_header('Content-Length','0'); self.end_headers()
            log(f"404 {self.command} {self.path}"); return
        size=os.path.getsize(fp); meta=os.path.basename(fp).startswith('.')
        if DRIP < 0 and not meta:   # server-error mode: every chunk request fails with 503
            self.send_response(503); self.send_header('Content-Length','0'); self.end_headers()
            log(f"503 {self.command} {self.path}"); return
        self.send_response(200); self.send_header('Content-Type','application/octet-stream')
        self.send_header('Content-Length',str(size)); self.end_headers()
        if head: log(f"200 HEAD {self.path} size={size}"); return
        data=open(fp,'rb').read()
        if meta or DRIP <= 0:
            self.wfile.write(data); self.wfile.flush(); log(f"200 GET-fast {self.path} size={size}"); return
        log(f"200 GET-drip-start {self.path} size={size}")
        sent=0
        try:
            for b in data:
                self.wfile.write(bytes([b])); self.wfile.flush(); sent+=1
                time.sleep(DRIP)
        except (BrokenPipeError, ConnectionResetError, OSError) as e:
            log(f"client-gone {self.path} sent={sent} after={time.time()-t0:.1f}s ({type(e).__name__})"); return
        log(f"drip-complete {self.path} sent={sent} after={time.time()-t0:.1f}s")
srv=ThreadingHTTPServer(('127.0.0.1',PORT),H); srv.daemon_threads=True
log(f"serving {ROOT} on {PORT} drip={DRIP}s"); srv.serve_forever()
```
</details>

```
python3 slowdrip.py 18611 /path/to/vol.zarr 1.0 server.log &     # 1.0 = seconds per byte; 0 = serve whole (control); -1 = answer chunks with a bodiless 503 (add a multi-line HTML body to reproduce the long message)
vc_render_tifxyz -v cache --remote-url http://127.0.0.1:18611/vol.zarr -s segment \
    --scale 1 -g 0 -n 1 --cache-gb 1 --tif-output out/render      # or --zarr-output out/render.zarr
echo rc=$?
```

Any zarr volume with at least one real level-0 chunk under the segment works as the mirror; the six chunks `0/41/31/17..22` of the PHerc1447 volume above were enough here.

**Not in scope:** the retry/timeout policy (60 s transfer deadline × 4 attempts), the prefetch path and what the cache does with a prefetched chunk whose fetch failed are untouched; this PR only changes what happens once a fetch the sampler needs has given up. Follow-up to my measurements in #1611 (https://github.com/ScrollPrize/villa/issues/1611#issuecomment-5593875254). #1611 has since been closed by its reporter.

*AI disclosure: the analysis, the patch, the test, the experiment scripts and this PR text were written by Claude (Anthropic) under my direction; every number above comes from runs on my machine that I reviewed, except the macOS rows, which are @kadenpool's, and the test file, which @kadenpool wrote.*


Bullo27:
On the red "Synthetic rendering regression" check: it failed on `parallel/full_res` at 1.124x reference (tolerance 5%). The commit that triggered it, `dec671593`, only changes the `catch` block in `vc_render_tifxyz.cpp`'s `main`, which `bench_render_synthetic` does not link; `Slicing.cpp` is identical to `e85724996`, whose run passed. Comparing the two job logs: the four serial scenarios agree to the last digit (e.g. `serial/fallback_3` 901346 vs 901346 modeled ns/call, deterministic callgrind counts), while `parallel/full_res` went 0.927x → 1.124x and `parallel/mixed_shuffled` 0.868x → (not reached). So the swing is in the parallel replay model, not in the code. Closing and reopening once to re-run it.

*AI-assisted (Claude), human-directed account; ratios read from the two CI logs.*

Bullo27:
Correction to the above: reopening does not re-run the CI workflow (it triggers on `opened`/`synchronize` only), so I rebased the branch onto today's main instead (`23c3d759e`, no overlap with these files; `git range-diff` shows both patches unchanged; `test_slicing` 15/15 and the 503 stub run give the same result on the rebased build). New head `f095ec88b`.

*AI-assisted (Claude), human-directed account.*

Bullo27:
Seen in the wild today with current main (`33b0b91b7`, Linux): two renders of freshly grown patches on First Letters scrolls (PHerc0343 and PHerc1218, open-data bucket, `--zarr-output`, 28 layers) aborted after 346 s and 424 s with

```
terminate called recursively
terminate called after throwing an instance of 'std::runtime_error'
  what():  HTTP 0 fetching 0/61/44/10: Failed to connect to vesuvius-challenge-open-data.s3.us-east-1.amazonaws.com port 443 after 0 ms: Couldn't connect to server
```

and a core dump (exit 134; the other render failed the same way on `0/66/38/8`). curl could not connect at all (`after 0 ms`), so this was a moment of lost connectivity on my side, not a slow transfer. With this branch that failure ends as the one-line `Error: HTTP 0 fetching …` and exit 1, as in the drop and 503 rows of the table.

PR #1818, opened 2026-09-17
Bicubic interpolation for vc_render_tifxyz.

spelufo:
**In one sentence:** Add bicubic interpolation to vc_render_tifxyz.

**One real example:** A (spiral fit -> sculpted) segment of PHerc1203.

**Before:** 
Screenshot 2026-09-17 at 14 05 10

**After this PR:** with the flag `--surface-interpolation bicubic`
Screenshot 2026-09-17 at 14 05 47

## Details

Implemented on claude code, using opus 5:

[transcript.md](https://github.com/user-attachments/files/32336276/transcript.md)


kadenpool:
I ran your branch through a reading test I had already set up, so here is a number to go with the pictures.

Setup, identical on both arms except for the flag: PHerc0139 segment w016 (public `20250108000004-w029_2025010827`), carried to the 2.399 um scan by the catalogue's own transform, rendered with `vc_render_tifxyz` at level 2, 101 slices, `--flip-normals`, twelve 384 px tiles covering the labelled region. Then the challenge's `ink_9um` checkpoint (seed43, step 60000) on nine 21-layer windows in both depth orders, scored against the published labels over 178,146 held-out pixels. One arm on `main`, one on your branch at `6fe8ac14a` with `--surface-interpolation smooth`.

Both arms peak at the same window. Over the whole region the AUC goes **0.8572 to 0.8756**, and at the model's own threshold the prediction covers more of the labelled ink: **54.4 % to 61.7 %**, with labelled background going 8.2 % to 9.2 %.

The part you should know before quoting that: **it does not hold block by block.** Splitting the same region into blocks and scoring each one separately, the smooth arm is higher in 3 of 9 blocks of 128 label px (median -0.018, sign test p = 0.51), 10 of 22 at 64 px, and 6 of 17 at 96 px. So the region-level gain comes from how the scores rank across blocks rather than from reading any block better, and on one segment it is inside block-to-block noise. I would call it "finds more ink at the same threshold" and leave the AUC out of the headline.

For what it is worth on the other question your PR raises: this says interpolation is not what separates my own render path from the challenge's published input for this segment, which reads 0.912 through their pipeline against 0.857 through mine.

If you want the arm, it is twelve tiles and a nine-window sweep, and I am happy to rerun it on a segment of your choosing, or to hand over the recipe so you can.

hendrikschilling:
```
  [P2] Switching between bicubic and linear interpolation introduces new seams near borders and holes.
  volume-cartographer/core/src/QuadSurface.cpp:885–937
  (https://github.com/ScrollPrize/villa/blob/6fe8ac14a7d9088be58bd7effa5a2ddb752c2c0f/volume-cartographer/core/src/QuadSurface.cpp#L885)

  The interpolation switches abruptly according to whether the full 4×4 neighbourhood exists. Both methods agree at grid vertices, but generally not along the edges between vertices.      
  Consequently, this can introduce a position discontinuity even in the central, zero-offset slice, not just preserve the old normal discontinuities.

  Reproduced through QuadSurface::gen() on a curved, fully valid synthetic mesh:

  - Linear: neighbouring samples differ by 0.004 voxels.
  - Smooth, across the fallback boundary: 0.500 voxels.
  - With a 25-voxel layer offset: 1.036 voxels.
```

PR #1819, opened 2026-09-17
Train and distill Hecate with Hub-compatible EMA checkpoints

giorgioangel:
**In one sentence:** Train or distill Hecate and export EMA checkpoints that load directly in the published Hugging Face inference runtime.

**One real example:** Loaded the released 2.4 µm and 9.6 µm models, compared predictions on PHerc139 and PHerc1447 CT, then trained and exported both architectures using the existing audited data.

**Before:** `distilled_3d_ink` rejects the released architecture configuration with `unexpected keyword argument 'input_depth'` and lacks the resolution-distillation trainer.

**After this PR:** The existing shared architecture supports both released geometries. The trainer accepts released or historical EMA weights, preserves background-masked soft targets, supports registered native/fine-derived inputs, calibrates EMA BatchNorm on training data, and exports only inference weights plus public architecture metadata. Preparation, fresh initialization, full-state resume and export are separate commands.

**Proof:**

| Real-CT check against the Hub runtime | 2.4 µm | 9.6 µm |
|---|---:|---:|
| Maximum 2D/3D output error, forward/reverse, FP32/BF16 | 0 | 0 |
| Newly trained export/reload: maximum 2D/3D output error | 0 | 0 |
| Historical EMA tensors equal released tensors | Yes | Yes |

50 focused tests passed. Two-update training passed for both models, including two-GPU native/fine-derived distillation. Resume restored the schedule and losses; BF16 distributed arithmetic was not bitwise identical (maximum prediction difference 0.00041). No training-quality improvement is claimed from these smoke tests.

**Why / where this is useful — motivation:** Giorgio requested:
> We need to update the code in the branch with a training script that is compatible with the uploaded inference version and also with the version we used for the distillation.

## Details

Tested code: `04b95ff01`. GPU tests used existing H100/PyTorch 2.12.1 environments; no production run was started.

```bash
PYTHONPATH=src HECATE_TEST_ASSETS="$ASSETS" python -m pytest -q \
  tests/ink_detection/test_hecate_training.py \
  tests/ink_detection/test_ema_batchnorm.py \
  tests/ink_detection/test_canonical_projection.py \
  tests/ink_detection/test_multiteacher.py
```

See [the short training/export guide](https://github.com/ScrollPrize/villa/blob/codex/hecate-training/vesuvius/docs/hecate.md) for commands and the external-data contract. Runtime reference: [scrollprize/hecate](https://huggingface.co/scrollprize/hecate). The PR contains no checkpoints, dataset copies, private S3 locations, instance addresses or credentials. The existing multi-teacher workflow remains available; abandoned refinement architectures and orchestration scripts are excluded.


TAUIL-Abd-Elilah:
Since you're training and distilling Hecate: here is the released 9.6 µm checkpoint on the First Letters eligible scrolls, which I don't think anyone had run it on yet. It may be useful as held-out data for the distillation.

Setup: survey renders resampled to 9.6 µm in all three axes, forward and reverse, stride 64, bf16. Support = CT in the 16 central planes, eroded 64 px. Numbers are the fraction of supported pixels above 0.75.

| | fwd >0.75 | rev >0.75 |
|---|---:|---:|
| known ink, PHerc0139 w043 (3 × 3 cm) | 0.0685 | 0.0170 |
| PHerc0813 z6496_w060 (ink_9um-blank) | 0.0048 | 0.0161 |
| PHerc0813 z13696_w060 (ink_9um-blank) | 0.0147 | 0.0202 |
| 99 corpus meshes, each at its densest 3.6 cm window: median / max | 0.0069 / 0.0335 | |

Three things that may matter for your work:

1. **A blank-papyrus baseline on scrolls outside training.** Across the corpus, hecate stays low and separates known ink from ink_9um-blank meshes by 5–14×. Per-mesh scores: `data/hecate/`.
2. **A null on the 3D head.** Over each case's hot pixels, it places the response 0–29 µm from the CT sheet peak, in a layer 48–77 µm thick. That holds for known ink, both blanks, and the region below. So at 9.6 µm, depth placement doesn't separate ink from blank papyrus; only peak strength differs (0.53 vs 0.37–0.44). If that's expected, it may be worth a line in the model card.
3. **One region I can't resolve.** PHerc0813 z12496_w060 reads 0.1107 on the same support rule. That is 3.3× the best window on any other corpus mesh and above known ink, and ink_9um also flags it (3.1× below its control). It is a ~12 mm band that fades 5× below and 35× above within ~7 mm, so it doesn't look like a kollesis seam. There are no letterforms, but known ink at 9 µm shows none either. Is a response this strong on dense non-ink material something you see in training?

Scripts, predictions and a README that re-derives every number: https://github.com/TAUIL-Abd-Elilah/corpus-ink-survey (hecate section; `scripts/verify_readme.py`).

TAUIL-Abd-Elilah:
Correction to point 3 above: "3.3× the best window on any other corpus mesh" counted **forward maps only**. Which face a mesh's forward points at depends on how that mesh was fitted, so reverse counts as much as forward. Best same-width window in either direction, same support rule, 149 meshes so far:

| | direction | >0.75 |
|---|---|---:|
| PHerc0813 z12496_w060 | forward | 0.1107 |
| PHerc0211 z6720_w020 | reverse | 0.0900 |
| PHerc0813 z13088_w040 | forward | 0.0635 |
| every other mesh | | ≤ 0.0405 |

So that region is the highest, but it isn't alone. PHerc0211 z6720_w020 is also flagged by ink_9um in the same reverse direction (5.7× below its control). PHerc0813 z13088_w040 is the mesh directly above the first region, one wrap inward. The README now shows all three with their maps, and `verify_readme.py` fails if the ranking stops holding. The question stands for all three.

TAUIL-Abd-Elilah:
One more fix, to the reason I gave above. I said the face "forward" points at depends on how each mesh was fitted. I measured it, and it doesn't. The surface normal points toward the core on all 320 corpus meshes and on the published PHerc0139 w043 mesh, whose ink reads forward. So forward is the same face everywhere: the one carrying text on the control.

That gives the direction a physical meaning. The two PHerc0813 regions read forward, on the text face. PHerc0211 z6720_w020 reads reverse, on the face a roll usually leaves blank. Its orientation fit is the least certain of the three, since it sits near the core.

One property of the model may be useful for training: on ordinary papyrus, hecate 9.6 µm reads reverse slightly above forward on 62% of 159 meshes (median ratio 1.18). ink_9um shows no such tilt (49%, 1.00). Script: `scripts/mesh_orientation.py`.

PR #1820, opened 2026-09-17
VC3D fiber map: gap heat map, and a tidier Fibers dock

skyward7187:
Branch `fiber-map-gap-heatmap`, eight commits on main.

## What

**Gap heat map.** A toggleable colour layer under the fibers of the global Fiber Map
showing, for every spot on the annotated extent, a surrogate for the shortest distance
from that spot on the sheet to any drawn fiber, counting fibers on neighbouring windings.
Toolbar: `Gaps` checkbox (off by default), the colour scale `0 [ramp] [3.0 cm]` whose
spinbox is the scale's top, and `Fade by [4 windings]` (how many neighbouring windings still
count on either side, 0–7; 0 = only the fiber's own winding).

- The map is periodic in x (one winding = 2π·rRef) and the layout already fits an
  Archimedean sheet model r(W) = radius0 + pitch·W with a monotonic arclength warp u(x).
  The field rasterises every resolved fiber's runs in (u, z), takes one exact Euclidean
  distance transform, then for each cell reads it at the cell's position shifted k windings
  and combines with the modelled radial separation |k|·pitch, minimised over k.
- `Fade by M`: a fiber k windings away is blended toward the saturation by k/(M+1), so the own
  winding counts fully, winding M barely, and winding M+1 and beyond not at all.
- Ramp per theme (steel → periwinkle → violet), chosen so the faint end never fights the
  cyan/green fibers, chips, link dots and rings that by construction sit at that end.
- Built on the rebuild worker after the layout, with the toolbar settings captured at job
  start, published under the existing epoch/package gates, dropped with the layout; takes
  part in no digest. A setting change requests a (memoized) Update; while a build is in
  flight the change queues and the epilogue reconciles.
- ~60 ms for a 100-winding, 30k-voxel-tall synthetic map with 400 fibers (3.2 M cells), plus
  ~0.3 ms to colour the tiles; both on the worker. The GUI thread only wraps the images in pixmaps.

**Fibers dock.** The Full rebuild button goes (Shift+click on Update runs it; the queue's
Full request and cache verification stay). Fibers sort alphabetically by label within each
network, the unlinked list and the unplaceable list. An errors section leads the tree: one
entry per red mark as the status line counts them, numbered left to right; selecting one
brings its mark into view (fitting both rings of a suspect link); its children are the two
fibers. A search box above the tree filters rows by label or annotation name.

## Commits

1. `d1da68ce1` stacked-sheet gap heat map (module `FiberMapGapField.{hpp,cpp}`, test, workspace integration)
2. `2f1edd566` fade neighbouring windings out of the gap heat map
3. `730a3e404` gap heat map toolbar reads as a colour scale
4. `ce1b2cad1` gap heat map ramp follows the theme
5. `99b82d0db` tidy the Fibers dock
6. `e0b2ad871` review follow-ups across the branch (stale tiles hide at once, status suffix follows the checkbox, map click reveals a filtered row, error selection survives a theme change)
7. `926f87d49` defaults 3 cm and 4 neighbouring windings (fade spinbox re-indexed to count neighbours, 0–7); pushed without a review round at the author's request
8. `32b6e7d09` tiles coloured on the worker (256-entry table, parallel rows; GUI thread only wraps pixmaps) — answers the P2 performance finding; colouring 72.7 → 0.3 ms at 3.2 M cells, 546.9 → 3.0 ms at the 24 M cap

## Why not OpenCV's distance transform

`cv::distanceTransform(DIST_MASK_PRECISE)` is not bitwise deterministic across thread counts
(differences of ~1.5e-5), so the module carries its own exact separable
Felzenszwalb–Huttenlocher squared-distance transform in double.

## Verification

- `test_fiber_map_gap_field`: 26 cases, including a brute-force oracle (with and without
  the fade), an exact-distance check to 0.01 voxel against the nearest seed pixel,
  thread-count determinism (cv::setNumThreads 1 vs 4), budget/domain/tile geometry.
- All 16 fiber ctest targets pass; VC3D builds clean (RelWithDebInfo, Ubuntu 24.04).
- Reviews (Codex gpt-6-astra): 1 plan round; per commit: heat map 5 code rounds → READY, fade 2,
  toolbar 1, ramp 1, dock 2 → READY; whole-branch review 4 rounds → READY; Qt lifetime /
  re-entrancy specialist pass: 1 round → READY.
- Proof: GUI-checked by @skyward7187 on PHercParis4 fibers in both themes (Gaps and Fade controls,
  Shift+click Update, the search box, error entries); screenshots to follow in a comment.

## Deferred

Hover readout with predicted 3D position and click-to-jump; local pitch from crossing
radial gaps; H/V family filter for the heat map; hatching beyond the annotated envelope;
removing the Full-rebuild machinery entirely (would be its own PR).

https://claude.ai/code/session_01YKBWyFShke85mC653ujChD


hendrikschilling:
```
Finding
  [P2] Heat-map publication can noticeably stall the GUI.
  volume-cartographer/apps/VC3D/FiberMapWorkspace.cpp:2736

  The distance field is calculated on a worker, but colouring every pixel and creating its pixmaps happens synchronously on the GUI thread. This repeats when publishing a heat map or      
  changing themes.

  Using the PR’s actual colour function, Release build, five measured repetitions:

   Cells                             Colouring time, median
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━
   3.2 million                                        88 ms
  ────────────────────────────────  ────────────────────────
   24 million, the configured cap                    606 ms

  These exclude pixmap creation and other scene work. Move pixel colouring into worker-generated QImage tiles, leaving pixmap publication on the GUI thread. This is a responsiveness issue,
  not an endless update loop.


```

skyward7187:
Addressed in 32b6e7d09: the tiles are now coloured into premultiplied ARGB images on the rebuild worker, for the theme in force at job start, and publication only wraps them in pixmaps. Colouring itself became a 256-entry table read per cell with rows in parallel (buffer pointer and stride taken once before the parallel region, since `scanLine()` bumps QImage's non-atomic detach counter), so the two cases that still colour on the GUI thread, a theme change and re-showing after a scene rebuild without the images, cost milliseconds.

Colouring step alone, synthetic field with 0.1% NaN cells, best of five, 32 cores:

| Cells | Before (interpolate, serial) | After (table, serial) | After (table, parallel rows) |
|---|---|---|---|
| 3.2 M | 72.7 ms | 2.1 ms | 0.3 ms |
| 24 M | 546.9 ms | 16.9 ms | 3.0 ms |

The "before" column matches your measurement within noise. Pixmap wrapping is the remaining GUI-thread cost and is not in these numbers.

PR #1821, opened 2026-09-17
Tutorial 5 on a 6 GB GPU: flat-path chunk selection (2.4 GB instead of 92 GB) + measured VRAM/throughput

mg255:
**In one sentence:** You can now start Tutorial 5 on a 6 GB laptop GPU: download the ~2.4 GB of level-0 chunks your training config actually reads instead of the segment's 92 GB, and use settings that are measured to fit the card.

**One real example:** Starting from an empty directory on a Windows 11 / WSL2 laptop with an RTX 3060 Laptop GPU (6 GB), I synced the labels and `x.tif` (12,625 files, 70 MB), ran `select_flat_training_chunks` against the segment's sync plan, and downloaded what it selected: 15,487 files, **2.42 GB** for `patch_size [64,128,128]`. Training then ran 200 iterations at **0.50 s/iteration** with a **4,385 MiB** peak on the card, no out-of-memory.

**Before:**

- The tutorial says the segment is "about 25 GB". It is **92.02 GB** in 147,785 files, because the surface volume ships all six pyramid levels; level 0 alone is 67.37 GB. A `hf buckets sync` of the whole segment ran 56 minutes on my connection before I stopped it.
- `uv sync --extra models` exits 1 on `Could not find a package configuration file provided by "Ceres"`: the `models` extra pulls in the C++ `volume-cartographer` bindings, which ink detection does not import.
- The tutorial's training config (`patch_size [64,256,256]`, `batch_size 2`) does not fit 6 GB, and on Windows/WSL2 that does **not** surface as an out-of-memory error: the driver moves the overflow into system RAM and the run just gets slower — 44 s/iteration instead of 0.5. The existing tip ("reduce `batch_size` to 1") lands at 10.5 s/iteration, still spilling.

**After this PR:**

- `python -m vesuvius.ink_detection.preprocessing.select_flat_training_chunks <config> full.jsonl subset.jsonl` runs the trainer's own patch discovery on the (small) labels and reduces a `hf buckets sync --plan` to the level-0 chunks those patches touch: 3,534 chunks / 2.95 GB for `[64,256,256]`, 2,862 chunks / 2.42 GB for `[64,128,128]`, out of 101,304 level-0 chunks.
- The tutorial documents the real size per pyramid level, the three-step small download, the install workaround, and a measured table of what fits a 6 GB card — including that memory overflow on WSL2 shows up as a 20–100× slowdown rather than an error.

**Proof:**

proof_download
proof_throughput
proof_training

| File | What to look at |
|---|---|
| `proof_download.png` | 92.02 GB / 147,785 files for the full segment vs 2.42 GB / 15,487 files selected for `[64,128,128]` — both read from the same sync plan |
| `proof_training.png` | 200 iterations, `[64,128,128]` batch 2: GPU memory peaks at 4,385 MiB under the card's 6,144 MiB; the flat first 210 s is patch discovery on the CPU |
| `proof_throughput.png` | seconds per iteration for every configuration measured; only `[64,128,128]` batch 2 stays inside the card |

All three are generated from the raw measurement files by `mess/proof.py` (nvidia-smi samples every 500 ms, tqdm lines, the sync plans) — no hand-cropped screenshots.

**Why / where this is useful:** Anyone with a consumer GPU and a normal internet connection can run the tutorial's training path the same day instead of downloading 92 GB first, and can tell from the table whether their card will train or silently spill.

I'm a hobbyist with a gaming laptop (RTX 3060, 6 GB) and no cloud budget. I wanted to try Tutorial 5 to understand how ink detection works before deciding whether I could contribute anything useful. The tutorial says "about 25 GB" — the segment is actually 92 GB, and that stopped me cold. So I worked out which chunks the trainer actually touches for a given patch configuration, and it turned out to be about 2.4 GB. With that, training runs on my 6 GB card at 0.5 s/iteration with patch 128 and batch 2. I also found that on Windows/WSL2 the driver never reports out-of-memory; it just silently gets 20–100× slower, which is why the existing "reduce batch_size" hint didn't help me. I used Claude Code for most of the implementation and the measurements; the design choices, the test with the mutation check, and every number in the tables were run and read by me on my own machine. I'm submitting this because the next person with a small GPU should be able to start the tutorial in twenty minutes instead of giving up after a 92 GB download.

### What is in the diff

Two commits on top of `5ab585f`:

1. `select_flat_training_chunks` (+140) with a regression test (+93), and the tutorial's new "Small GPU or small disk" section plus the install and out-of-memory notes (+48/−3).
2. The dataset section's per-level table and the `~2.4 GB` statement (+13/−1).

### Prior art

This implements the option suggested by the team on 20 Jul 2026 in the thread *Fetch Data from AWS During Training*: "one option is to download only the zarr data covered by the supervision and validation masks instead of downloading the full segments" (https://discord.com/channels/1079907749569237093/1528861316465430618/1528907566057259048). The 92 GB figure was reported in the same thread the next day (https://discord.com/channels/1079907749569237093/1528861316465430618/1529050657984745493); the docs still say 25 GB, which this PR fixes. In that thread the team also explains that streaming the 2.5D data from Hugging Face is not currently viable because of its rate limits, which is why this takes the local-subset route instead of streaming.

The native 3D path already has `ink_detection.preprocessing.download_required_zarr_chunks`, which selects chunks of the **scroll** volume from `.tifxyz` geometry for `full_3d` training. This tool is its flat-mode counterpart: it selects chunks of the **pre-rendered surface volume** in the Hugging Face bucket, using the flat patch discovery, and edits a `hf buckets sync` plan rather than copying a Zarr. Neither replaces the other.

For the install failure the PR only adds a pointer to villa#1706, which fixes the cause by moving the bindings out of the `models` extra; nothing here duplicates that change.

### Scope / lifetime

The team also said in that thread that the plan is to align the ink labels with the open-data segments so 2D/2.5D data can be streamed from there. If that lands, the selector becomes unnecessary — it is a bridge, not a permanent fixture. The measured table and the WSL2 note stay valid either way, since they are about the card and the driver, not about where the bytes come from.

### Measurements

Host: HP Victus 16, Ryzen 5 5600H, 31 GB RAM (15 GB visible to WSL), **RTX 3060 Laptop GPU, 6144 MiB**, driver 616.92, CUDA 13.4, Windows 11 + WSL2 (Ubuntu 24.04), Python 3.14, `torch 2.12.1+cu130`, `mixed_precision: fp16`, `dataloader_workers: 4`, segment `phercparis4/w00_20231016151002`. GPU memory sampled with `nvidia-smi` every 500 ms (whole-card usage); "s/iteration" is the steady-state rate after ~8 iterations.

Training (`vesuvius.ink_detection.training.train`, tutorial config except for the noted fields):

| `patch_size` | `batch_size` | Iterations | s/iteration | GPU memory | Result |
|---|---|---|---|---|---|
| `[64,128,128]` | 2 | 200 | **0.50** | 4,385 MiB | fits; 5 min 30 s total, ~2.8 h extrapolated for 20,000 |
| `[64,128,128]` | 4 | 20 | 10.25 | full | spills into system RAM |
| `[64,128,128]` | 8 | 20 | 53.75 | full | spills |
| `[64,128,128]` | 16 | 3 | — | full | `torch.OutOfMemoryError` at 18.07 GiB allocated |
| `[64,128,128]` | 32 | 3 | — | full | `torch.OutOfMemoryError` at 17.07 GiB allocated |
| `[64,256,256]` | 1 | 20 | 10.50 | full | spills |
| `[64,256,256]` | 2 (tutorial) | 7 | ~44 | full | spills (run stopped after 7 iterations) |
| `[64,256,256]` | 8 | 3 | — | full | `torch.OutOfMemoryError` at 17.22 GiB allocated |

Inference with the published 9 µm checkpoint (`hybrid_3d2d-seed42/step-075000.pth`) on the published w035 render (28×5820×5240, 6,458 patches), command as in the tutorial:

| `--batch-size` | Wall | GPU memory peak |
|---|---|---|
| 8 | 1:51 | 625 MiB |
| 32 (tutorial) | 1:40 | 1,467 MiB |
| 128 | 1:37 | 4,905 MiB |

Inference is not GPU-bound here: of ~95 s, the GPU is active ~40 s; the rest is process start, `torch.compile` and data loading. That is why the tutorial's "`--batch-size 32` assumes a large GPU" is now stated as a number.

### Reproducing the small download

From the directory that holds `ink-dataset/`:

```bash
# 1. labels, Zarr metadata, x.tif  (12,625 files, 70 MB, 13 min here — many tiny files)
uvx --from huggingface_hub hf buckets sync \
  hf://buckets/scrollprize/datasets/ink/phercparis4/w00_20231016151002 \
  ./ink-dataset/phercparis4/w00_20231016151002 \
  --include "*_inklabels.zarr/*" --include "*_supervision_mask.zarr/*" --include "*_validation_mask.zarr/*" \
  --include "*.zarr/.z*" --include "*.zarr/0/.z*" --include "x.tif" --include "meta.json"

# 2. full plan, nothing downloaded
uvx --from huggingface_hub hf buckets sync \
  hf://buckets/scrollprize/datasets/ink/phercparis4/w00_20231016151002 \
  ./ink-dataset/phercparis4/w00_20231016151002 --plan full.jsonl

# 3. reduce the plan to what the config reads (3.5 min: patch discovery), then fetch it
uv run --extra models python -m vesuvius.ink_detection.preprocessing.select_flat_training_chunks \
  configs/ink_tutorial.json full.jsonl subset.jsonl
uvx --from huggingface_hub hf buckets sync --apply subset.jsonl   # 2.35 GB, 2 min 19 s here
```

Then training with `patch_size [64,128,128]`, `batch_size 2`. In my end-to-end run this produced the same 8,098 patches as a run with the full segment on disk, and the selection is config-specific: change `patch_size`, `patch_overlap`, `patch_min_labeled_coverage` or the labels and step 3 has to run again.

### Limitations

- The 200-iteration run shows that training **runs**; it is not a model and says nothing about ink quality.
- The selector supports `mode: "flat"` with `volume_scale` 0 and needs `x.tif` present, because `gather_segments` requires it for segment discovery (it is not read during flat training).
- Missing chunks read back as the Zarr `fill_value` (0). Chunks outside the selection are never touched by a patch for that config, but a different config against the same partial download would silently read zeros — hence the note to rerun step 3.
- The WSL2 behaviour is reported as measured, not explained: a 6 GB card accepted 12 GiB of allocations and only failed at ~17 GiB, while throughput fell by 3–24× beyond ~5 GB. This matches the NVIDIA driver's "CUDA — Sysmem Fallback Policy" under WDDM, but I did not test that setting, so the name is a pointer, not a verified cause.
- `uv sync` rewrote `vesuvius/uv.lock` here (platform markers only); that change is **not** part of this PR.

### Two related defects, not fixed here

Found while writing this, reported separately rather than mixed into this PR:

1. In `mode: "flat"` without `surface_volume_path(s)`, `gather_segments` skips a segment that has no `x.tif` **silently**, even when it is listed in `segments`; the run then fails with "InkDataset produced no training patches after applying supervision masking", which does not name the cause.
2. That failed run writes an **empty** patch list to `out_dir/flat_ink_patches_*.json`, and the cache is reused afterwards: adding the missing `x.tif` does not help until the cache file is deleted. (Related but, as far as I can tell, not covered by #1701/#1661, which compare label identity — an empty cache has no entries to compare.)

### Tested at

Branch `tutorial5-small-hardware`, commits `6ecc6ae` and `430a050` on `5ab585f`. `pytest tests/ink_detection/test_select_flat_training_chunks.py tests/ink_detection/test_data_foundation.py tests/ink_detection/test_shipped_configs.py` → 23 passed. The new test fails if the chunk-range arithmetic is mutated (checked).


hendrikschilling:
```
  1. [P2] Z-window augmentation can read chunks the selector omitted.
     select_flat_training_chunks.py:71
     (https://github.com/ScrollPrize/villa/blob/430a050117893cba7de351f8afa9f96a1dd21d8d/vesuvius/src/vesuvius/ink_detection/preprocessing/select_flat_training_chunks.py#L71)
     Selection uses the stored patch bbox, but training can shift it through flat_z_window_jitter. Those additional chunks aren’t downloaded and silently read as zeros. I reproduced a     
     valid shifted crop needing four omitted chunks. Include the full possible jitter range, or explicitly reject jitter-enabled configurations.

  2. [P2] The selector crashes with supported Zarr 2 installations.
     select_flat_training_chunks.py:68
     (https://github.com/ScrollPrize/villa/blob/430a050117893cba7de351f8afa9f96a1dd21d8d/vesuvius/src/vesuvius/ink_detection/preprocessing/select_flat_training_chunks.py#L68)
     volume.metadata does not exist in Zarr 2.18.7. The surrounding getattr doesn’t protect that access, so selection raises AttributeError. Handle both Zarr APIs, including their         
     dimension separators. The tutorial’s locked Zarr 3 environment avoids this particular failure.
```

PR #1823, opened 2026-09-17
normalgridtools: score the umbilicus with a weighted sum, not a weighted mean

evilaliv3:
`align_and_extract_umbilicus` in `core/src/normalgridtools.cpp` finds the umbilicus of a normal grid slice in two steps: a random search over candidate points, then a hill climb on a refinement score. The refinement score runs over sampled segment midpoints, weights each by the inverse of its distance from the candidate, adds up the squared cosine between the segment normal and the direction to the segment, and then divides by the sum of the weights:

```cpp
float weight = 1.0f / std::max(100.0f, dist);
score += (cos_angle * cos_angle) * weight;
wsum  += weight;
...
return score/wsum;
```

This pull request removes that division, makes the generator seed an argument, and adds a regression test on sections whose centre is known by construction. It is commit c6bcb2a58 on the branch `normalgrid-umbilicus-score-not-normalised` of the fork, carrying the tree **`b175fb12`**:
the tree is what was measured and what a reviewer should compare, since the commit hash moves whenever the message is reworded and the code does not. Everything below was measured on the upstream code at `23adee047`. The function has not been touched since: at `2dcfaf6a0`, the tip of `main` on 17 September 2026 and 50 commits later, `return score/wsum` is still line 116 of `volume-cartographer/core/src/normalgridtools.cpp`, and none of those 50 commits enters that file
or its header.

## The counter-example

Without the division the score is a weighted sum. Every weight tends to 1/D as the candidate moves a distance D away from the section, so the sum tends to zero, and a positive continuous function that vanishes at infinity has a maximum at finite distance. With the division it is a weighted mean: numerator and denominator vanish together, the limit is the plain mean of cos^2, and that mean is as large far away as it is at the centre, because a mean does not know how many normals are contributing. Every normal keeps contributing to it, however far the candidate goes, and the next section says exactly how large that far field value is.

Measured on a synthetic section, one turn of a spiral of 4000 sampled points, a million units from the centre the sum retains 3.4e-4 of its maximum and the mean 0.52; on a concentric circle, 9.5e-4 and 0.50. The hill climb is therefore not searching for a maximum that is hard to find: on a partial prediction it is searching for one that is not there, every step outwards improves the mean a little, and nothing clamps the walk to the grid, so the routine returns a point outside the
volume as an estimate.

f1-mechanism

*The same section and the same samples, scored both ways: the weighted sum falls away with distance, the weighted mean flattens out, so far from the sheets the climb still finds an improvement.*

## The criterion that says which slices will fail

The limit above can be written exactly, and it is what makes the failure predictable. With `u` the unit direction from the candidate towards the section and `n_i` the sampled normals:

formula-far-field

with the maximum over all directions being the largest eigenvalue of that matrix:

formula-bound

Every normal contributes equally to that limit, however far away the candidate is, because a mean does not know how many normals are contributing or where they are. The maximum of the limit over `u` is the largest eigenvalue of `C`, so the far field value of your score is a property of the normals alone: **1/2 on a complete circle of radial normals, whatever `u`, and rising towards 1 on a partial section whose normals are coherent.** When that bound exceeds the score on the axis, the axis is not the maximum and the climb is right to leave it.

Measured with the 10,000 samples the function itself draws: on the PHerc0826 slice shown below the largest eigenvalue is **0.65** against a weighted mean of **0.58** at the published control point, so the far field is 1.11 times the score on the axis; on a synthetic full circle it is 0.50 against 1.00 at the centre, and on a half circle of radial normals still 0.50, because normals spanning half a turn cover both directions of the plane as evenly as normals spanning all of it.

That is the practical point for a maintainer. **The defect cannot show up on the sections a unit test would build** (a full circle, a half circle, concentric rings: the mean is exact on all of them) **and it strikes the partial predictions the toolchain actually runs on.** Put to every slice of the fifteen scrolls with a manual reference, 360 slices, the condition holds on 192 of them and the published estimate leaves the grid on 134; every slice that leaves the grid satisfies
the condition, and 0 of the 168 slices where it fails do. It is necessary and not sufficient:
sensitivity 1.00, specificity 0.74. The 58 slices that satisfy the condition and stay inside anyway are the ones where the climb does not reach the border from its random start.

Those 360 slices are 24 from each of 15 scrolls and not 360 independent observations, so we do not pool them: slices of one scroll share a geometry, a surface prediction and a scanner. With the scroll as the unit the statement needs no distributional assumption. On 15 of 15 scrolls not one run that left the grid escaped the condition, and on the 14 scrolls where any run left, every one of those runs was flagged; the remaining scroll, PHerc0191, has no run to flag. Shuffling the
predictor among the slices of one scroll, independently in each scroll, which holds that scroll's own two counts fixed, 0 of 200,000 shuffles reached the observed count, p < 5.0e-6.

f2-score-field

*PHerc. 0826, xy slice 8000, both panels divided by the score at the published control point (cross) so the scale is shared. Left, as published: the field is as high at the edge of the grid as on the sheets and the climb stops outside it (circle). Right, division removed: the field collapses away from the sheets and the climb stops on them.*

On the 24 scrolls of the challenge (the 23 in competition plus PHerc0332), 24 slices each, 576 slices, the estimate as published lands inside the grid on **351 of 576** slices; 12 scrolls keep fewer than two thirds of their slices inside, and 3 scrolls keep all of them. With the division removed the count is **576 of 576**, every slice of every one of the 24 scrolls. The median spread between 3 seeds is at most 0.71 per cent of the grid width on any scroll.

This is the part of the result that stands on its own, because it needs no reference to be checked. The accuracy numbers below are worth less than it, and we say so in the section where they are. One more count of the same kind, descriptive and chosen after the fact rather than declared in advance, so not a criterion: at 10 mm from the reference, a distance no disagreement between annotators approaches, the published function leaves 7 of 15 scrolls above that line and the weighted sum 0, its worst row being 7.11 mm.

## Against manual references: it is the objective, not the search

Fifteen scrolls carry a manual umbilicus, five published by the challenge and ten by a third party, 24 slices per scroll, 32 to 154 control points each, 20 seeds per slice.

**Which block is the out of sample test, said before the numbers.** The structure of this measurement and its criteria were written down and dated before the run (2026-09-15 15:33:39Z), including the 0.30 mm margin below. That document says in its own words that it was written before the estimator had run on any of the eleven scrolls whose grids were still to be downloaded, and after the four scrolls already reported in an earlier note. Those four are among the five published by the challenge: **4 of the 5 scrolls of the first block had been measured when the criteria were fixed, and 0 of the ten of the second.** So the second block is the out of sample test and it is the one that carries the accuracy claim, 9 of 10; the first block carries the pre-registered structure and the count of estimates leaving the volume, 4 of 5. A maintainer who found that out after merging would be right to be annoyed, so it is here. A first assumption was
that the search was also at fault, and it was rewritten (double precision, walk clipped to the grid, eight probes from a fixed centre with a step that halves when none improves, one objective
for seed and refinement). The four combinations, median distance from the reference in millimetres, with the 95 per cent bootstrap interval in brackets:

| scroll | (a) weighted mean, as published | (b) weighted sum, the one line | (c) rewritten search, mean | (d) rewritten search, sum |
|---|---:|---:|---:|---:|
| PHerc0125 | 133.02 [107.99, 190.24] | 1.59 [1.15, 2.17] | 39.96 [39.50, 40.60] | 1.64 [1.18, 2.21] |
| PHerc0211 | 63.81 [45.21, 95.60] | 1.47 [0.98, 2.38] | 37.97 [33.43, 40.24] | 1.32 [0.85, 1.97] |
| PHerc0332 | 0.35 [0.28, 0.41] | 0.36 [0.30, 0.42] | 0.35 [0.28, 0.41] | 0.36 [0.30, 0.42] |
| PHerc0826 | 53.21 [40.75, 66.12] | 4.27 [3.03, 5.44] | 39.97 [37.27, 41.35] | 4.30 [2.81, 5.37] |
| PHercParis4 | 4.25 [1.47, 26.57] | 1.13 [0.83, 1.48] | 3.12 [1.37, 11.04] | 1.07 [0.81, 1.41] |
| PHerc0191 | 6.28 [4.39, 8.50] | 4.15 [2.64, 6.03] | 6.22 [4.56, 8.23] | 4.66 [3.21, 6.02] |
| PHerc0257 | 8.37 [2.75, 31.17] | 1.87 [1.26, 2.72] | 5.43 [2.17, 30.40] | 1.84 [1.28, 2.73] |
| PHerc0268 | 8.27 [5.63, 11.05] | 4.32 [2.69, 5.87] | 8.18 [5.24, 10.90] | 4.30 [2.79, 5.95] |
| PHerc0358 | 8.46 [3.92, 97.85] | 2.89 [1.74, 3.86] | 10.38 [3.70, 37.37] | 2.75 [1.63, 3.87] |
| PHerc0800 | 5.29 [3.49, 36.18] | 2.68 [1.73, 3.83] | 4.80 [3.30, 33.00] | 2.66 [1.63, 3.70] |
| PHerc0813 | 10.49 [7.10, 12.45] | 7.11 [4.42, 9.33] | 10.26 [7.62, 12.29] | 6.50 [4.31, 8.69] |
| PHerc1203 | 4.80 [1.46, 34.91] | 1.44 [0.98, 1.77] | 2.85 [1.40, 16.83] | 1.38 [1.01, 1.75] |
| PHerc1218 | 104.09 [82.99, 150.90] | 2.47 [1.66, 3.89] | 33.27 [29.88, 35.68] | 2.24 [1.60, 3.62] |
| PHerc1447 | 38.57 [28.75, 55.24] | 1.76 [1.15, 2.56] | 33.20 [25.73, 38.13] | 1.69 [1.17, 2.51] |
| PHerc1545 | 49.55 [17.68, 68.47] | 1.87 [1.24, 3.38] | 32.66 [9.14, 40.16] | 1.80 [1.25, 3.19] |

The first five rows are the challenge's own references; the last ten are a third party's manual annotations, a weaker reference kept as a separate block and never pooled with the first. The count that separates the objective from the search is the one of gross failures. With the weighted mean, the rewritten search still lands 10 mm or more from the reference on 8 of 15 scrolls, at worst 39.97 mm, much as the published walk does (7 scrolls, at worst 133.02 mm), even though it
keeps 7,200 of 7,200 estimates inside the grid against 4,484 of 7,200 for the published walk, because it is clipped there. With the weighted sum neither search fails on any scroll; their worst rows are 7.11 and 6.50 mm. The rewritten search is better than the one line in the median on 11 of 15 scrolls with the sum, and on 1 of 15 with the mean; the one line beats the centroid of the section by the margin declared before the run on 13 of 15 scrolls, the rewritten search with the sum on the same 13, the rewritten search with the mean on 4. A well posed objective makes even a loose search converge, and the estimates outside the volume disappear without a clamp because the score falls away in every direction. That is the argument for the one line rather than for a rewrite.

**What the margins over the centroid are worth.** The threshold those 13 wins are counted against is 0.30 mm, and two independent annotations of one scroll lie 2.2 mm apart: the threshold is 7 times smaller than the disagreement between two people about where the axis is, and 8 of the 13 margins won fall below that disagreement. Closer than the centroid is a criterion satisfied here, not a difference these references can resolve, and nothing in this pull request rests on it.

## Where the weighted sum puts its maximum, and why the synthetic sections favour your version

That the sum has an interior maximum says nothing about where that maximum is. A sum rewards the side of the section that contributes more segments; a mean normalises the density away. On synthetic sections with a centre known by construction, 8000 by 8000 grid units at 9.362 um per unit, concentric circles of radius 300 to 3500, medians over 20 seeds, in grid units:

| section | sum, maximum: error | sum: dx | mean, maximum: error | mean: dx | estimate, sum | estimate, mean |
|---|---:|---:|---:|---:|---:|---:|
| circle 1:1 | 47 | -2 | 0 | +0 | 46 | 0 |
| ellipse 1.4 | 125 | -36 | 6 | +0 | 126 | 6 |
| half circle | 261 | -12 | 0 | +0 | 259 | 0 |
| density 1:2 | 202 | +200 | 0 | +0 | 202 | 0 |
| density 1:5 | 229 | +226 | 0 | +0 | 230 | 0 |
| density 1:10 | 244 | +242 | 0 | +0 | 244 | 0 |

Error is the distance from the true centre; dx is the offset along x, positive towards the side that holds more segments. The "maximum" columns are the maximum of each objective on a lattice of step 4 around the true centre; the "estimate" columns are the full estimate with your search. The density rows keep every segment with x above the centre and one in 2, 5 and 10 of those below it;
the half circle keeps the half with y above the centre, so its offset is along y and its error column carries it.

Read plainly: **on these sections the published version is the more accurate one.** The mean is exact wherever it has a maximum to find (0.004 mm on concentric circles against 0.430 mm for the
sum, both on 20 seeds), and the sum pays for its guaranteed maximum with a bias towards density:
0.44 mm on a uniform circle, 2.29 mm at a 1:10 thinning, 2.44 mm on a half circle. The full estimate follows the maximum in every row. On real predictions the picture is partly confirmed and partly not, and both halves are reported. The segments of each of the 15 scrolls were counted in eight angular sectors around the reference axis, which serves only as the origin of the sectors;
the sector ratio of a scroll is its most populated sector over its least populated one, median over its 24 slices. PHerc0826, the one scroll of the first block where the centroid of the section wins, is the most one-sided of the fifteen, 6.38 against a median of 3.59, and over the five scrolls of the first block the ratio orders them exactly as the distance from the reference does (Spearman +1.00 over 5). **Over all fifteen that ordering does not survive**: Spearman +0.33 with
p 0.23 against the distance, and -0.35 with p 0.20 against the margin over the centroid. The two scrolls of the second block where the centroid wins or nearly wins, PHerc1218 and PHerc0813, are not among the one-sided ones by this measure: their ratios are 4.08 and 3.59, ranks 10 and 8 of 15 counted from the least one-sided. The density bias explains the one loss in the first block and does not order the fifteen, and no second measure of one-sidedness is offered here to rescue the prediction. The one line trades an unbounded failure for a bounded bias towards the side of the section the prediction happens to have found, and on a very one sided prediction that bias can
exceed the distance of the centroid baseline.

## One section of every scroll of the challenge

The figure below (attached as `umbilicus-gallery-24.png`) is one section per scroll of the 24 the run covered, sorted by scroll id: the segments of the normal grid in grey, the estimate of the published function in orange, the estimate with the division removed in blue, and, on the 14 scrolls where one exists, the reference umbilicus as a black cross, drawn only to be measured against and never used to choose anything. The slice is picked by a rule (the twelfth of the 24 the run sampled, in order of z), not by looking. On 11 of the 24 panels the published estimate lies beyond the window, and an arrow gives its direction and distance from the blue circle: 106, 122, 256, 53, 137 and 53 mm where the voxel size is on record, 0.7 to 261 grid widths where it is not; on PHerc. 0490B the published walk did not stop by itself but was stopped by the cap of the search. Over the whole run 576 of 576 estimates fall inside the volume with the division removed, against 351 of 576 as published. On the 10 scrolls without a reference the figure shows where the estimate falls and that it stays inside; it does not show that it is right, and nothing here claims that. Drawn by `tools/figure_umbilicus_gallery.py` from `evidence/positions-24-refit.csv`,
the cached grids and the reference files; no estimate was recomputed for it.

One section of every scroll, published and repaired estimate

*One slice of each of the 24 scrolls, chosen by a rule and not by eye. Orange: the estimate as published, with an arrow when it lies off the panel. Blue: with the division removed. Cross: the manual reference where one exists.*

## What the patch contains

1. **The one line**: `return score;` in place of `return score/wsum;`, with `wsum` gone and a comment that states the trade above, including the bias, rather than presenting the change as a repair without cost.
2. **A seed argument**, `std::optional<std::uint32_t> seed = std::nullopt`. The generator that draws the 10,000 sampled segments and the 1,000 candidate points was a static `std::mt19937` seeded from `std::random_device`, so no run could be repeated. With no seed the behaviour is exactly as before; with a seed a fresh generator is seeded with that value.
3. **A regression test** in `core/test/test_normalgridtools.cpp`: concentric circles, a half circle and circles thinned 1:10 on one side, built as `GridStore` sections, estimated with a fixed seed, and asserted within 1.0, 5.0 and 4.5 mm of the known centre and inside the grid. The tolerances are at least 1.56 times the p90 over 20 seeds of the table above, so no seed fails them by chance. The test says in its own comment what it does not do: it cannot tell the sum from the mean, because on these sections the mean is exact; it bounds the sum's known bias so that it cannot grow unnoticed. The defect itself shows only on real predictions.

The signature change is source compatible: the new argument has a default and the function has no other caller in the tree.

**The test was built and run.** It needs only OpenCV core and imgproc, not Qt, which `core/test/CMakeLists.txt` demands for the viewer tests and not for this target: built from this commit's own sources with g++ 15.2.0 and OpenCV 4.10.0 in 5.76 s and run in 5.78 s. All 15 cases of `core/test/test_normalgridtools.cpp` pass, the 4 new ones with the estimate 0.23 mm from the known centre on the concentric circles (tolerance 1.0 mm), 2.13 mm on the half circle (5.0 mm), 2.02 mm on the circles thinned 1:10 (4.5 mm) and on the dense side of the centre at x 4206 as the mechanism predicts, every estimate inside the grid, and the same point to the bit on the repeated seed; a second run in a fresh process gave the same estimates. Two departures from your own build,
since neither touches the arithmetic: the flags are not the CMake flags (no precompiled headers, no sanitizers) and the OpenCV is the distribution's rather than the one in your image.

## What this does not say

It does not say the objective is now the right one. Two alternatives you may reach for first, the mean multiplied by a coverage term that counts the contributing normals, and a weight that falls faster than 1/D so that the mean itself decays, are untested here: each is a new objective, and offering one without measuring it on the same fifteen references is the mistake this whole exercise exists to avoid.

Three things in the same function are real and are not in this patch, each because it is a new objective or a new search that would have to be measured against these references before being offered:

- **The walk is unbounded.** The hill climb adds the step to the current best with no clamp to the grid. With the sum the score falls away in every direction, so on these runs the walk settles inside the grid every time (576 of 576), but that is measured, not guaranteed.
- **There is no consensus set.** The header comment calls the function a RANSAC and there is none: every sampled segment contributes its squared cosine with no threshold and no outlier rejection, so a segment whose normal is wrong is averaged in rather than discarded.
- **The seed score and the refinement score are two different functions.** The random search of the first stage has neither a distance decay nor a weight, so it can hand the refinement a starting point its own objective would never have chosen.
- **The far field bound looks like a free diagnostic, and it is not one yet.** The largest eigenvalue of `C` above costs nothing beyond the samples the function already draws, so the code could warn when it exceeds the score it is about to return. Measured against the weighted mean at the reference control point, which the code does not have, that comparison flags every walk which left the grid, at 58 false alarms out of 192. Measured against the refined score, the only score the function holds when it returns, it fires on 14 slices of 360, 3 of them on a run that stayed inside, and misses 124 of the 135 runs that left on that single seed, against the 134 slices counted by majority of the 20 seeds, because once the walk has reached the far field its own score has risen to the bound. The form worth measuring is therefore neither of those but the comparison made where the search starts: the seed of the refinement, the best of the 1,000 random candidates of lines 78 to 88, is a point the function already holds, and the weighted mean there is one more call.

**A post hoc probe of that last form, since you will ask it.** We ran it after the three rounds of review of the paper, on the same 360 slices, with no criterion fixed beforehand: that is why the number is here, where you are entitled to the best answer we have, and not in the paper, where a number measured after the fact on the slices it is tested against would need its threshold and its criterion declared in advance to mean anything.

| the bound compared with | fires on | false alarms | misses | sensitivity | specificity |
|---|---:|---:|---:|---:|---:|
| the mean at the reference control point (which the code does not have) | 192 of 360 | 58 | 0 of 134 | 1.00 | 0.74 |
| the refined score (what the function holds when it returns) | 14 of 360 | 3 | 124 of 135 | 0.08 | 0.99 |
| the mean at the seed point (what it holds before the climb moves) | 101 of 360 | 13 | 47 of 135 | 0.65 | 0.94 |

The misses column of the first row counts slices by majority of the twenty seeds, 134 of them; the other two rows count single runs of one seed, 135, which is the same 360 slices scored a different way and not a discrepancy. The mechanism is what makes the three rows readable: at the seed point the walk has not yet risen to the bound, so the comparison still carries most of the signal there, and by the time the search returns its own score has caught the bound up and the comparison has lost it. None of this is a result we would stand behind. A diagnostic worth having would have its threshold and its false alarm rate declared before being measured, and this is not that; it is the
shape of the thing, offered so that whoever measures it properly does not start from nothing.

It also does not say how far the reference itself is from the truth: the distances above are from hand annotations, and two independent annotations of one scroll lie about 2.2 mm apart, so rows below that distance are not interpretable as a statement about the axis.

## Reproducing

The mechanism curve, the density table and the 24 scroll run are computed by the tools of the validation paper from the published normal grids and the published umbilici, on CPU. The numbers here are the paper's `rev1-numbers.tex`, `rev1-ablation.tex` and `rev1-density.tex`, generated from its evidence files and not retyped.


evilaliv3:
\cc @pmh47

hendrikschilling:
I like it. However its incomplete (see below). I'd love for this to include a cli tool that actually creates that umbilicus (maybe even including optional normalgrid creation so it can be run just from predictions?). Such a tool apparently exists and was used in the experiments shown above?.

And because this effectively only uses normals: maybe test structure tensor normals directly from ct as well (just a research suggestion - maybe a follow up).

```
  1. The experiments use a search cap absent from the submitted implementation. The description says the PHerc0490B baseline “was stopped by the cap of the search,” but the repository’s   
     hill climb has no iteration limit. Please provide the runner, specify the cap, explain which variants used it, and explain how capped runs enter the error and convergence statistics. 
     Were these measurements made by calling the actual C++ function or a separate implementation?

  2. The referenced validation artifacts are missing. The description cites a validation paper, scripts, CSVs and A10-images/umbilicus-gallery-24.png, but these are not included or linked 
     accessibly. Please provide them, together with the exact normal-grid and manual-umbilicus sources, coordinate/scale conversion, slice selection and reproduction commands.

  3. The new tests do not catch the reported defect. As the description acknowledges, the old weighted-mean objective also passes these geometric accuracy checks. They test reproducibility
     and bound the new bias, but do not prevent reintroducing the runaway behaviour. Please include a reproducible failing case, preferably a small fixture from the measured real data,    
     that distinguishes the old and new objectives.

  4. The safety claim needs qualification. A score tending to zero at infinity does not guarantee that this particular hill climb terminates within a reasonable number of iterations or    
     returns a point inside the grid. The observed 576/576 in-bounds results are empirical evidence, not an implemented bound.
```

evilaliv3:
Thank you @hendrikschilling for reading it this closely, and for saying what would make it complete. Taking your points in the order you raised them.

**The command line tool.** It exists now: `vc_gen_umbilicus` calls `align_and_extract_umbilicus` itself, on the xy grids of a normal grid volume, with the `seed` argument this patch adds, and writes the file `Umbilicus::FromFile` reads. It is built and checked against a saved reference, and not yet on this branch. Normal grid creation is not in it because `vc_gen_normalgrids` already does that from a prediction zarr, so predictions to umbilicus is those two commands in a row, and a second copy of the first would be two places to fix one thing.

**Structure tensor normals from CT.** Good idea, and not this patch. It is reachable on PHerc0826, the one scroll where we hold the CT, the prediction derived grid and a published umbilicus at once; the work is that the function consumes sampled line segments and not a dense field, so which voxels become segments is the experiment. We will run it there first, criterion written before the run, no date.

**1.** We said the measurements were made on the upstream code. They were not: they were made by a Python transcription of the function. They have since been made on the upstream code, every table, the 24 scroll run and the exponent table, 37,152 uncapped calls to align_and_extract_umbilicus`, and not one walk failed to stop by itself. The arm this patch proposes agrees with the transcription on 15 rows of 15 and no pre-registered verdict moved; one count moved in our favour, PHerc1218's margin reading minus 0.32 mm by the transcription and minus 0.24 by the function, both losses, so the patch is no better for it either way. The cap was ours, `MAX_SWEEPS = 2000` in the transcription, disclosed nowhere; the recomputed tables have none, and what it did to the old ones is measured in the detailed answer.

**2.** Almost all of it is already public and simply unlinked: the paper, the tools, the evidence files, the pre-registrations, the manual umbilici with their provenance and the gallery figure are
in [`e89eb7ad-umbilicus-estimator`](https://github.com/evilaliv3/vesuvius-challenge-pipeline/tree/main/results/e89eb7ad-umbilicus-estimator), whose `src/README.md` gives sources, scale conversion, slice selection and every command. Not yet pushed: the fixture and test of point 3, the tool, the recomputed runs and the paper build that prints them. One pointer of ours is broken, the walkthrough's `UMBILICUS_BENCH`, which names an empty repository; it is being fixed. The detailed answer to all four points is in the same folder.

**3.** There is now a case that fails on the old objective and passes on the new: a 5,840 byte crop of a real published slice, PHerc0125 at z 14811, chosen by a rule written before looking. On all
five seeds the weighted sum lands inside, within 0.41 mm of the published umbilicus; the weighted mean leaves the grid by about four grid widths. The verdict is unchanged at five optimisation levels.

**4.** Agreed, and the claim is now qualified where it is made. A score that vanishes at infinity has a maximum at finite distance, which says nothing about whether this hill climb reaches it, in
how many steps, or inside the grid; that it settled inside on 576 of 576 slices at one seed each is observed and not enforced, and the comment above `return score;` now says so.

evilaliv3:
I've reorganized and updated the results: https://github.com/evilaliv3/vesuvius-challenge-pipeline/tree/main/results/e89eb7ad-umbilicus-estimator

PR #1745, opened 2026-09-09
M7 per scroll presets

gmDevi:
In one sentence: Anyone with a villa checkout can fit and render the eight 2025-2026 tracks-only scrolls with the right winding count and a usable umbilicus on the first try, from one JSON override per scroll instead of a day of reverse-engineering Scroll 1's defaults.

One real example: Starting with the PHerc0358 tracks dataset (dl.ash2txt.org/datasets/spiral_datasets/PHerc0358/20250821151737, slices 8000-9500) and the estimated umbilicus in configs/scrolls/umbilici/PHerc0358.json, I ran fit_spiral with the overrides that configs/scrolls/PHerc0358.json encodes (tracks only, 100 windings, 30k steps), and it produced a fitted mesh with 50.4% of track points satisfied whose windings render as clean continuous sheets; a second band (slices 9500-11000) with the same preset gave 47.4% and the same clean renders.

Before: The fitter's defaults are Scroll 1's (130 windings), and none of the 2025-2026 track-ready scrolls ships an umbilicus or a winding count. A first fit of PHerc0358 or PHerc0257 either exported windings outside the papyrus or fitted against nothing, the two silent traps documented in 
ScrollPrize/villa#1736, and each scroll cost a day of trial fits to find its count and a hand-estimated umbilicus.

After this PR: configs/scrolls/<scroll>.json carries the tracks-only switches and the validated winding count for PHerc0125 (90), PHerc0826 (70), PHerc0211 (90), PHerc0257 (90), PHerc0358 (100), PHerc0813 (80) and PHerc0191 (100); configs/scrolls/umbilici/ carries estimated umbilicus control points (level-0 voxel coordinates) for the six scrolls without a published one (PHerc0191, 0257, 0268, 0358, 0800, 0813). Presets load through Config(overrides=...), umbilici through json_umbilicus_z_to_yx, and the README gets the merge one-liner plus a provenance table.

[...]

PR #1830, opened 2026-09-19
spiral-fitting: let spiral-scroll.json carry the scroll's winding count

rodriguescarson:
**In one sentence:** A dataset's `spiral-scroll.json` can now say how many windings the scroll has, and the fitter uses that instead of Scroll 1's 130 unless you override it.

**One real example:** Starting with PHerc0191 (volume 20250821151635, slices 11600-12400), I fitted the same window with the winding count at 90 and again at 45. The two fits produced identical surfaces (45 of 45 windings matching at offset 0, median distance 0.00 voxels), and the export stopped at whichever count I'd set: 415.63 cm2 of surface against 99.79. So this value decides how far out the fit exports, and it has to be right per scroll. That fit ran on villa 757f70c0, before this change, so it shows why the value matters rather than this PR's own output.

**Before:** `shell_outer_winding_idx` and `model_gap_expander_num_windings` default to 130, Scroll 1's count. On a scroll with fewer windings, any fit that forgets to override both exports windings outside the papyrus. The count for each scroll lives in people's shell history.

**After this PR:** Add `"num_windings": 90` to the dataset's `spiral-scroll.json` and both the headless CLI and the Spiral panel's defaults (and presets) pick it up. `FIT_SPIRAL_CONFIG_OVERRIDES` and panel edits still win. It's deliberately not one of the scroll-owned keys, since the count is a hypothesis people tune. Capacity only goes up when the count wouldn't fit.

**Proof:** 12 new tests in `tests/test_scroll_spec_num_windings.py` (parsing and validation, the defaults, override precedence, and the panel catalog, including that the service's own copy isn't mutated). They pass on Python 3.14, and the file fails to import on unpatched main. The existing `test_config.py`, `test_input_toggles.py` and `ScrollSpecTests` still pass (41 tests together).

I want this because the results are good. I've been working on this day and night, and the boundaries look better and I can see clearly. Paul suggested exactly this on #1745, and I'm not proposing a count for any scroll here.

Not ticked: the PHerc0191 example above ran before this change, so it isn't this PR's own output.

## Details

Files: `fit_session.py` (the `ScrollSpec.num_windings` field, validation, and `scroll_spec_config_defaults`), `fit_spiral.py` (CLI: Python defaults, then the scroll, then env overrides), `spiral_service.py` (`configuration_catalog` overlays defaults and presets), `README.md` (documents the key), and the new test file. It adds no winding count for any scroll.

AI-assisted: written with Claude Code, reviewed by me. https://claude.ai/code/session_01WXWShsy16D6vxiaYnQLpJ4


rodriguescarson:
Ran this PR on real data. PHerc0826 (z 6528 to 7328, tracks and normals only, 300 steps), the same fit twice with identical config overrides and no winding override: once on this branch (`de41bfb`) with `"num_windings": 90` in `spiral-scroll.json`, once on its base (`b1ef996e`) with no count.

| | stored `shell_outer_winding_idx` | `flatten_spiral_checkpoint` reconstruction | windings in the written mesh |
|---|---|---|---|
| this branch | 90 | `reconstructing windings 10..90` | 81 |
| base | 130 | `reconstructing windings 10..130` | 121 |

The 121-winding export is the silent default this PR removes when a dataset carries its count. Logs, specs, overrides and the harness: https://github.com/rodriguescarson/eligible-scroll-atlas/tree/main/artifacts/hecate-3d (`fit_*.log`, `export2_*.log`, `export_meta_*.json`).

PR #1833, opened 2026-09-19
flatten_spiral_checkpoint: voxel size from the scroll, not a default

evilaliv3:
`flatten_spiral_checkpoint.py` defaults `--voxel-size-um` to `9.6`. That number is written into the exported TIFXYZ metadata, where it squares into every area the mesh reports, and is passed to the flattener, where it scales every length. Nothing warns when it is used, because it is a default and not a fallback: the invocation the README documents, checkpoint and output and nothing else, takes it silently. That is the invocation the README shows, and it is the one that has no way of being right unless the scroll happens to measure 9.6 um.

`9.6` is not the voxel size of any scroll in this repository's own catalogue. The values that appear in `scrollprize.org/static/data_browser/index.json` are 0.55, 1.129, 2.215, 2.399, 2.4, 2.401, 2.403, 3.24, 4.317, 7.91, 8.64, 9.362, 9.366 and 45.532. For PHerc0826, the scroll that `spiral-fitting/README.md` uses for its own `spiral-scroll.json` example, the catalogue says **9.362**, the volume is `20250821151701-9.362um-1.2m-113keV-masked.zarr`, and the README's example states `"voxel_size_um": 9.362` on line 122.

## What it costs

Exporting a PHerc0826 checkpoint the documented way, with a dataset whose `spiral-scroll.json` states 9.362:

| | |
|---|---|
| stated by the dataset | 9.362 um |
| written by the exporter | 9.6 um |
| error on every length handed to the flattener | **+2.5422 %** |
| error on every square centimetre written into `meta.json` | **+5.1490 %** |

A square centimetre is the unit a reported segment area is published in, and `render_ink.py` recovers the voxel size back out of `area_cm2 / area_vx2`, so the wrong number is self consistent downstream and nothing contradicts it. One million voxels of surface come out as 0.921600 cm2 instead of 0.876470 cm2.

We have not audited anyone else's published areas and we are not claiming any specific number in the wild is wrong. We are saying the default cannot be right for the scrolls this catalogue lists, and that it fails silently.

## The patch

The scroll specification already owns this fact. `fit_session.py` says so in as many words: nothing `spiral-scroll.json` specifies belongs in the run block, not the scroll's physical facts, and `voxel_size_um` is one of the four keys it lists as owned by the specification.

So `_resolve_voxel_size_um` reads it from there, by **the same walk `_resolve_umbilicus` already does**: `SPIRAL_DATASET` first, then the checkpoint's own directory and every parent. This adds no new requirement, and that is the point: a run that cannot find `umbilicus.json` by that walk already refuses to export, so the dataset root is already reachable from the checkpoint in every working invocation. In every real dataset we hold, `spiral-scroll.json` sits in the same directory as `umbilicus.json`.

`--voxel-size-um` keeps working and still wins when given. What goes away is the number that is used when nobody said anything.

We did not read it from the checkpoint, because it is not there. A fitted checkpoint carries `lasagna_group`, `lasagna_scale` and `spiral_outward_sense`, which are also scroll specification facts, but not `voxel_size_um`.

This is also the convention the rest of the repository already follows, which is why we did not invent one:

- `lasagna/fit2tifxyz.py` takes the same `--voxel-size-um`, for the same purpose, area calculation, and its default is already `None`.
- `vesuvius/.../fiber_trace_2d/planning/specs.md` states the rule outright for the fiber tracer: physical units are reported only when the caller provides an explicit positive `--voxel-size-um`, and the runner "must not invent physical units".
- `volume-cartographer/core/src/VoxelSizeMetadata.cpp` reads `voxel_size_um` back out of `meta.json`, and refuses to assume a pyramid level when the store does not state one, for the reason this patch is about: "Assuming 0 would report a voxel size too small by whatever factor the run really used, and nothing downstream could tell."
- the README already tells the reader that this script finds `umbilicus.json` by walking the checkpoint's ancestors and `$SPIRAL_DATASET`. This patch reads the scroll specification from the same two places, so anyone who can run the documented command already has the file.

One commit, one file changed plus its test. Nothing renamed, nothing nearby tidied.

## The test

Five cases in the existing `tests/test_flatten_spiral_checkpoint.py`, written next to the `_resolve_umbilicus` cases they are modelled on: the specification is read, an absent specification is an error rather than a guess, the flag overrides, a non positive value is refused, and the area that the resolved number produces is the area of the scroll the dataset names.

```
PYTHONPATH=spiral-fitting python -m pytest -q spiral-fitting/tests/test_flatten_spiral_checkpoint.py
```

Before: `6 passed`. After: `11 passed`.

**We should be straight about what that means.** These five are not a test that failed before and passes now. Before the patch the exporter has no way at all to learn the voxel size, so there is no behaviour to assert against: the new cases cannot even import. The evidence that this is a defect and not a preference is the measurement above, taken by running the shipped `_parse_args` on a pristine checkout, not by transcribing it.

## Three questions

1. Is refusing to export the behaviour you want when a dataset states no voxel size, or would you rather it kept a default and printed a warning? We chose refusal because the failure is silent and permanent once the areas are published, but it is your call and it is a two line difference.
2. `spiral-fitting/README.md` documents the two argument invocation without mentioning that a voxel size is chosen for you. If you take this patch, that section wants a sentence saying the value comes from `spiral-scroll.json`. Do you want it in this pull request?
3. The same `9.6` is a literal in two other places, `grow_track_graph.py:25` and `find_inconsistent_windings.py:203`. We left them alone because they are different tools and we would rather send one claim per pull request. Do you want them in this one, or separately, or are they deliberate there?


pmh47:
The hopeful recursive search is messy. We should have some way to neatly pass this resolution metadata through, but this PR is not good as-is. Note for future readers -- the 9.6 default is 4x downsampling of our 2.4um high-resolution volumes.

evilaliv3:
Thank you @pmh47 !

evilaliv3:
Thanks for looking at these, Paul, and for the note on the 9.6 default (4x of the 2.4 um volumes: we did not know its origin, and we will say so where we quoted it).

Two questions so we spend your time well:
1. Would an issue proposing a clean pass through of the resolution metadata (the scroll's voxel size from its manifest into flatten_spiral_checkpoint, in place of the recursive search you rightly called messy) be welcome, or is spiral-fitting better left to the maintainers for now? We are happy to drop 1832 and 1833 either way.
2. 1834 (the attachment loss sampling from the generator it is given) is still open: if it is not wanted, say so and we will close it ourselves.

Our other work on volume-cartographer (the tracer's repeatability, two small fixes in GrowPatch and vc_grow_seg_from_seed, with a before and after) is coming as one pull request with measurements; if you would rather see issues first, that is easy.

PR #1840, opened 2026-09-19
surface_preflight: surface-only mode plus metadata scale and bbox consistency gates

ItIsCuthNotCup:
**In one sentence:** Run `vesuvius.surface_preflight --surface <tifxyz>` (no CT volume needed) right after any tifxyz export and get exit code 2 with a one-line reason when the grid is empty/all-sentinel or when `meta.json` `scale`/`bbox` disagree with the grid, instead of finding out during flattening or rendering.

**One real example:** Starting with the public Scroll 1 segment `dl.ash2txt.org/community-uploads/bruniss/scrolls/s1/autogens/02201554/` (196x225, scale 0.05), I ran `vc_tifxyz2obj` then `vc_obj2tifxyz seg.obj out_nonmetric --uv-non-metric` with the upstream VC3D `f07d33b` build. It printed `Valid grid points: 0 / 4 (0%)`, saved a 2x2 all-sentinel tifxyz with bbox `[[-1,-1,-1],[-1,-1,-1]]`, and exited 0 (#1320). `surface_preflight --surface out_nonmetric` now fails it with exit 2.

**Before:** `surface_preflight` required `--volume` (argparse error without it), so a freshly exported tifxyz could not be checked at all without mounting the CT volume; and even with a volume it passed a surface whose `meta.json` scale implied 3465 voxels/cell on a 20 voxels/cell grid (173x, the #1319 ratio), and passed a surface whose bbox excluded every vertex (#1272).

**After this PR:** `--volume` is optional; without it the volume gates are skipped and `report.volume` is `null`. Two new gates run always: `tifxyz_scale_consistency` (median adjacent-vertex 3D spacing per grid axis vs `1/scale`, tolerance `--scale-tolerance`, default 2x) and `tifxyz_bbox_consistency` (valid extents inside `bbox` within `--bbox-tolerance`, default 0.01 vx; only when `bbox` is present; malformed bbox fails). Failed gates are summarised one per line on stderr. `schema_version` 1 -> 2; exit codes unchanged.

**Proof:**

Real segment round-trip at `f07d33b` (`--uv-non-metric`), producer exit 0:
```
Final grid: 2 x 2
Valid grid points: 0 / 4 (0%)
Saving to tifxyz format...
```
Preflight after this PR, exit 2:
```
FAIL: preflight.json
valid_surface_vertices: surface has no valid vertices (all coordinates are sentinel/invalid); the producer emitted an empty grid
valid_surface_quads: surface has no connected valid quads
tifxyz_scale_consistency: grid has no positive columns spacing pairs to compare with meta.json scale
tifxyz_bbox_consistency: no valid vertices to compare against bbox
```
Same segment re-exported with `--tifxyz-source=<seg>` (224x195, scale 0.05): PASS, measured spacing 19.98 vs expected 20.0 vx/cell. Two untouched real segments (`02201554`, `auto_grown_20250409214252294`) PASS with ratio 0.9989 on both axes, so the gates do not false-positive on correct data.

Synthetic 48x64 fixtures (plane at 20 vx spacing), same synthetic all-ones volume for both tools:

| fixture | `f07d33b` tool | this PR |
|---|---|---|
| `wrong_scale` (scale 0.0002886) | PASS, exit 0 | FAIL `tifxyz_scale_consistency`, exit 2 |
| `stale_bbox` | PASS, exit 0 | FAIL `tifxyz_bbox_consistency`, exit 2 |
| `all_sentinel` | needs `--volume`; then FAIL | FAIL surface-only, exit 2 |
| `good` | PASS | PASS |

Tests: `uv run pytest tests/test_surface_preflight.py -q` -> 16 passed at `f07d33b`, 28 passed here.

**Why / where this is useful:** Anyone producing or ingesting tifxyz (VC3D exports, spiral fits, community uploads) can gate the artifact before flattening, rendering, label transfer or publication with one command and no volume mount; the JSON report says which invariant broke. Complements PR #1781 (producer-side fix for #1320/#1319): this catches the same class of artifact from any producer, including segments already published.

## Details

- Tested commit base: `f07d33b`; producer binaries from release asset `VC3D-f07d33b-2026-09-19-macos-arm64.zip`; macOS 26.5 arm64, CPython 3.14.7, numpy 2.4.6, tifffile 2026.6.1, zarr 3.2.1.
- Scale semantics follow `QuadSurface::size()` (`cols / scale[0]` voxels), i.e. expected spacing = `1/scale`. Spacing is accumulated blockwise (`block_rows`) into a fixed 2^(1/64)-binned log histogram, including row pairs that straddle block boundaries, so memory is bounded and the median is deterministic (quantised to ~1.1%).
- Existing `--volume` behaviour is unchanged apart from the two extra gates; existing tests pass with the gate count bumped 8 -> 10.
- Not reproduced here: a HEAD invocation that emits a wrong scale on real data (#1319); metric mode and `--tifxyz-source` produced self-consistent scales. The scale gate is evidenced on synthetic fixtures plus no-false-positive checks on real segments. Stretch `1000` on this segment OOMs in OpenCV before writing output.
- Full `vesuvius` suite could not be collected locally (unrelated missing `torch`/`tqdm`/`cv2`); only `test_surface_preflight.py` imports this module.
- Issues: closes nothing by itself; relates to #1320, #1319, #1272.


hendrikschilling:
```
  [P2] Histogram rounding can reject grids within the configured tolerance.
  vesuvius/src/vesuvius/surface_preflight.py:246 (https://github.com/ScrollPrize/villa/blob/04c002296/vesuvius/src/vesuvius/surface_preflight.py#L246), comparison at line 514
  (https://github.com/ScrollPrize/villa/blob/04c002296/vesuvius/src/vesuvius/surface_preflight.py#L514)

  The measured median is replaced by a histogram-bin midpoint, then compared as though exact. I reproduced:

  - Exact 32-vx spacing, metadata scale=1/32, --scale-tolerance 1: incorrectly fails because estimated spacing is 32.174.
  - True ratio 2.0, default tolerance 2.0: incorrectly fails as 2.01086.

  Account for the histogram’s uncertainty when applying the threshold, or refine the measurement near a boundary. This is a limited numerical issue, not a broad regression.
```

ItIsCuthNotCup:
@hendrikschilling thanks, confirmed and fixed in 2081a69. The median is now reported with its histogram bin bounds (`median_spacing_bounds_voxels`, `ratio_range`), and the gate fails only when the whole bin lies outside `[1/tol, tol]`, so a true spacing anywhere inside the tolerance (including exactly on the boundary) can no longer be rejected by the ~0.55% quantisation. Both repros are added as regression tests (32 vx at scale 1/32 with `--scale-tolerance 1`; ratio exactly 2.0 at default tolerance, plus a negative control at 1.95 that still fails). They fail on 04c0022 and pass now; `tests/test_surface_preflight.py` -> 30 passed.

Issue #1846, opened 2026-09-20
A failed remote chunk fetch at the seed point still aborts vc_grow_seg_from_seed


kadenpool:
**In one sentence:** #1817 gives `vc_render_tifxyz` a clean exit and a one-line error when a remote chunk fetch is exhausted, but that catch is per binary, so the same failure at the seed point still ends `vc_grow_seg_from_seed` in `SIGABRT` with the server's whole error page on the terminal. **Corrected 21 Sep: this issue also claimed VC3D's viewer was affected. It is not: see the comments. Only the tracer is; #1847 fixes it where a failed fetch outside `tracer()` reaches `main`, as in the run below, and VC3D also runs the tracer for some jobs.**

**This is a different throw site from the one #1809 is about, and that is the point.** #1817 does two things: a `ParallelFailure` marshaller in `core/src/Slicing.cpp`, which is in core and which every binary gets, and a top-level catch in the renderer's `main`, which only the renderer gets. The tracer does not go through `ChunkSampler`; in the run below it throws at `core/src/Volume.cpp:380`, already outside any parallel region. So the core half correctly does nothing for it, and only a top-level catch would help. For the tracer, which like the renderer takes a remote volume URL on its command line, an ordinary `std::runtime_error` walking out of `main` is the failure reproduced below, where every chunk fetch fails and so the first read, the seed point, fails on the main thread; to a user it looks exactly like the crash #1809 reported. A fetch that fails later, during the growth, is different: `main` already wraps `tracer()` in its own catch, and on a build with real OpenMP (GCC) a failure inside the growth's parallel loops aborts from inside the loop.

**Reproduced with the stub from #1809**, `fail_server.py --mode 503`, on macOS arm64 (Homebrew clang, the non-OpenMP shim build), against `PHerc0139`:

```
$ vc_grow_seg_from_seed -v http://127.0.0.1:8823/ -t out -p params.json -s 1728 4800 2000
libc++abi: terminating due to uncaught exception of type std::runtime_error: Volume::read failed
  fetching chunk 0/15/37/13: HTTP 503 fetching 0/15/37/13: <!DOCTYPE HTML ... (the whole page)
Abort trap: 6
$ echo $?
134
```

Identical on `origin/main` at `23c3d759e` and on #1817 at `7e1ed38cb`, so this is not something the PR changed, it is something it did not reach:

| | `origin/main` | #1817 |
|---|---|---|
[...]

PR #1847, opened 2026-09-20
vc_grow_seg_from_seed: report a failed chunk fetch instead of aborting (refs #1846)

kadenpool:
**In one sentence:** a remote chunk fetch that fails after its retries kills `vc_grow_seg_from_seed` with `SIGABRT` and the server's whole error page, and this makes it exit 1 with one line instead, the same way #1817 did for `vc_render_tifxyz`.

Reported and measured in #1846. Related: #1809, and #1817, which fixes it for the renderer. I hit the same failure in the renderer on real S3, without fault injection, twice in three days of rendering PHerc0139 (#1809); the tracer takes the same path whenever it is pointed at the bucket.

**Why this is not covered by #1817.** That PR does two things, and they have different reach. The `ParallelFailure` marshaller in `core/src/Slicing.cpp` is in core, so every binary gets it. The top-level catch is per binary, and only the renderer has one. This tool does not go through `ChunkSampler` at all: it throws at `core/src/Volume.cpp:377`, already outside any parallel region, so the core half correctly does nothing for it and only a top-level catch helps.

**It is reachable without any flag.** `is_remote_volume_path()` accepts `http://`, `https://` and `s3://` and routes the volume argument to `Volume::NewFromUrl`, so a run pointed at the open-data bucket is on this path today, even though the docs only show local paths.

**The change.** The body moves to `run_main` and a thin `main` wraps it in the same catch the renderer got, including the first-line truncation from `dec671593` so a 503's HTML page does not land on the terminal. **25 lines added, none removed**, and no re-indentation.

**Before and after**, against the stub server from #1809, same seed and volume:

| | exit | output |
|---|---|---|
| before | 134 | `terminating due to uncaught exception`, the whole HTML page |
| after | **1** | `Error: Volume::read failed fetching chunk 0/15/37/13: HTTP 503 ... [...]` |

**Two platforms.** Mine is macOS arm64, Homebrew clang, the build where villa's CMake swaps in the no-op OpenMP shim, so those loops run serially. @Bullo27 ran the complement on Ubuntu 24.04, GCC 13.3, Release, OpenMP on, and reported in #1846. He has since confirmed the patch he ran is identical to this PR's commit `650b38635` ([comment](https://github.com/ScrollPrize/villa/pull/1847#issuecomment-5750964966)), so these rows apply to this diff:

| mode | before | after |
|---|---|---|
| `--mode 503` | SIGABRT (rc 134), the whole HTML page, 1.9 s | rc 1, one `Error:` line, 1.8 s |
| `--mode drop` | SIGABRT (rc 134), `HTTP 0 ... Empty reply from server` | rc 1, the same message on one line, 1.7 s |

He also noted no `terminate called recursively` on either abort, which is consistent with this throw being on the main thread rather than escaping a parallel region.

**VC3D is not affected.** @Bullo27 built VC3D and drove it into 144 refused chunk fetches offscreen, and nothing throws: the viewer's sampler counts `ChunkStatus::Error` and returns (#1846). When VC3D launches the renderer with `--remote-url`, that path is the renderer's, which #1817 fixes.


Bullo27:
Confirming the Linux rows apply to this exact diff: the hunks of 650b38635 are identical to the patch I built and ran for [#1846](https://github.com/ScrollPrize/villa/issues/1846#issuecomment-5750589282) (Ubuntu 24.04, GCC 13.3, OpenMP on; `--mode 503` and `--mode drop`, rc 134 → 1, one `Error:` line). Nothing further from me on the code.

kadenpool:
Thanks for checking the hunks against 650b38635. I've added that to the description, with a link to your comment, so the Linux rows are tied to this exact diff for whoever reviews it.

PR #1859, opened 2026-09-21
vc_obj2tifxyz: fail when rasterization produces zero valid points

### LINKED ISSUE #1320: vc_obj2tifxyz: 'Successfully converted' and exit 0 after rasterizing 0 points

Kaluzy:
Fixes #1320.

`vc_obj2tifxyz` currently detects an all-invalid raster, prints a warning, then saves sentinel-only tifxyz files, prints `Successfully converted`, and exits 0. Scripts therefore cannot distinguish an empty conversion from a usable surface.

This change:

- treats `valid_count == 0` as a hard conversion failure before saving output
- prints an actionable message for the common normalized-`[0,1]` UV case
- clarifies the `stretch_factor` help text and corrects the default-mode example
- adds a hermetic CLI regression using the real built `vc_obj2tifxyz` binary

The regression constructs normalized UV geometry that deliberately misses every sample on the default 2x2 grid:

**Before current main**
- `Valid grid points: 0 / 4`
- sentinel-only tifxyz is saved
- `Successfully converted to tifxyz format`
- exit 0

**After**
- same default invocation exits non-zero
- no x/y/z tifxyz output is written
- error explains using a larger `stretch_factor` or `--tifxyz-source`
- the same OBJ with `stretch_factor=20` still converts successfully

Keeping this draft until CI confirms the native test on supported platforms.


Kaluzy:
Update: external upstream workflows were approval-gated, so I validated the final head in the fork using the repository's own workflows. The real vc_obj2tifxyz CLI regression now runs in normal Linux CLI CI and passes; Linux/macOS/Windows builds, full core CI, and CodeQL are green on the final #1859 head. PR is ready for review.

Kaluzy:
Final validation update: this branch has been refreshed onto current upstream main (`33b0b91`). I also validated #1859 together with #1861 on the same current converter base: both real-CLI regressions passed against one compiled `vc_obj2tifxyz`, and the combined matrix passed Linux, macOS, Windows, Python, CodeQL, full core CI, and the synthetic rendering regression. The new upstream workflow runs are currently at GitHub's external-contributor `action_required` approval gate.

